PR #108, opened 2021-12-02
Add basic metrics.

AlekSi:
Refs #74.

Response statuses are not done yet. Adding them requires listener/conn/handler refactoring that should be in a separate PR.


OpenSauce (approved):
LGTM 👍 Not on my home pc at the moment though。

PR #115, opened 2021-12-05
Makefile: add ability to use custom go bin

ekalinin:
Now there's no need to switch go bin via `ln -s`.

```
$ go install golang.org/dl/gotip@latest
$ gotip download

$ make init GO=gotip
```


AlekSi:
I'm afraid that's not enough. For example, we also use `go` there: https://github.com/FerretDB/FerretDB/blob/c52ea450072bd3453f27146c49a6869a8a7c25fb/tools/tools.go#L30-L36

I would rather do nothing until 1.18 release in Februrary, when all this complexity will go away.

ekalinin:
Ah, I see. Ok, let's just close it.

PR #117, opened 2021-12-06
Implement `dropDatabase` command.

### LINKED ISSUE #111: Implement `dropDatabase` command
Handle it with `DROP SCHEMA name CASCADE`.

radmirnovii:
Closes #111.

### DIFF
--- a/internal/handlers/handler_test.go
+++ b/internal/handlers/handler_test.go
@@ -19,7 +19,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/jackc/pgx/v4"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap/zaptest"
@@ -125,7 +124,6 @@ func TestListDatabases(t *testing.T) {
 }
 
 func TestDropDatabase(t *testing.T) {
-
 	ctx := testutil.Ctx(t)
 	pool := testutil.Pool(ctx, t)
 	l := zaptest.NewLogger(t)
--- a/internal/handlers/handler_test.go
+++ b/internal/handlers/handler_test.go
@@ -15,6 +15,8 @@
 package handlers
 
 import (
+	"fmt"
+	"github.com/jackc/pgx/v4"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


internal/handlers/shared/msg_dropdatabase.go
```diff
@@ -0,0 +1,54 @@
...
+
+// MsgDropDatabase removes the current database.
+func (h *Handler) MsgDropDatabase(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
+	document, err := msg.Document()
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+
+	m := document.Map()
+	db := m["$db"].(string)
```
OpenSauce, line 36: Will want to check here that a value is returned from the map. [See here.](https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/shared/msg_listcollections.go#L53)

internal/handlers/shared/msg_dropdatabase.go
```diff
@@ -0,0 +1,54 @@
...
+
+	sql := fmt.Sprintf(`DROP SCHEMA %s CASCADE`, pgx.Identifier{db}.Sanitize())
+
+	_, _ = h.pgPool.Exec(ctx, sql)
```
OpenSauce, line 40: Will want to check for errors etc here

OpenSauce:
Good start. I'd be interested in what @AlekSi thinks in terms of testing this, as we don't currently have a create database command?

internal/handlers/shared/msg_dropdatabase.go
AlekSi: That's not strictly required.

The current code will panic if "$db" is somehow absent, and this panic will be handled by logging a stack trace and closing the connection.

"comma ok" form will return a lazy error in that case, and it will be handled the same way.

I think the whole error handling will be reworked later, but I'm not sure how yet, so there is no preference between panics and lazy errors yet.

internal/handlers/shared/msg_dropdatabase.go
AlekSi: Just return a lazy error if something happened

AlekSi (commented):
Yep, tests would be nice. We should test with both existing and non-existing databases.

(Databases are created on the fly, so to create a database, we should insert some dummy document in some dummy collection)

radmirnovii:
Thx, for review, I'll fix this errors

radmirnovii:
> One more thing to note: what about test? Something like this:
> 
>     * https://github.com/FerretDB/FerretDB/pull/62/files#diff-efe5633155496149a2523f825232224f1dfc096d50a197246fe972d464f0e262R33

> Yep, tests would be nice. We should test with both existing and non-existing databases.
> 
> (Databases are created on the fly, so to create a database, we should insert some dummy document in some dummy collection)

fixed

AlekSi:
There is a merge conflict now. May you please resolve it?

internal/handlers/handler_test.go
```diff
@@ -122,6 +124,120 @@ func TestListDatabases(t *testing.T) {
...
+			require.NoError(t, err)
+			assert.Equal(t, count, 0)
+
+			// checking schema
+			query = fmt.Sprintf(`
+				SELECT COUNT(*) 
+				FROM information_schema.schemata 
+				WHERE schema_name = '%s'`,
+				schemaName)
+			err = pool.QueryRow(ctx, query).Scan(&count)
```
AlekSi, line 234:
```suggestion
				WHERE schema_name = $1`,
				schemaName)
			err = pool.QueryRow(ctx, query, schemaName).Scan(&count)
```

To avoid SQL injections.

We have a similar code already; search by `information_schema.tables`.

AlekSi:
There is no need to do a force-push – the PR will be squashed on merge anyway, and force-pushes require me to re-approve CI checks every time.

radmirnovii:
sorry, next time  will be more tidy

AlekSi:
Please tell when it is ready for review :)

radmirnovii:
> Please tell when it is ready for review :)

Ready for review)

AlekSi (approved):

AlekSi:
Thank you!

radmirnovii:
> Thank you!

There is some golangci-lint problem. Do i need to fix them?

AlekSi:
Don't worry, I will fix it

Issue #122, opened 2021-12-07
Use composite GitHub Action for Go setup


AlekSi:
We have some steps that are shared between jobs, workflows, and repositories. For example: https://github.com/FerretDB/FerretDB/blob/ef352172586ee74b69dcfeb15c4a39f201fc270f/.github/workflows/go.yml#L31-L51

It seems that [composite actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) would allow us to DRY our workflows. Let's try to extract those steps into a common "setup Go" action.

klokar:
@AlekSi Can I work on this improvement?

AlekSi:
Sure thing!

PR #126, opened 2021-12-07
Use composite GitHub Action for Go setup. (#122)

klokar:
Refs #122 

Extracted Go setup and caching to separate reusable action. Input and if statement was added until fuzz is treated in the same manner.

### DIFF
--- a/.github/actions/setup-go/action.yml
+++ b/.github/actions/setup-go/action.yml
@@ -1,0 +1,44 @@
+name: 'Setup Go'
+description: 'Install Go and restore cache.'
+inputs:
+  cache-key:
+    description: 'First part of key for restoring cache.'
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Install Go
+      run: |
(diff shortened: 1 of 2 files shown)


AlekSi (approved):
That's just awesome!

AlekSi:
@klokar Do you by any change know if that's possible to have several actions in the same repository and make `action.yml` actually work? See https://github.com/FerretDB/github-actions and https://github.com/FerretDB/dance/runs/4447027485: the `cache-key` argument seems to be ignored.

AlekSi:
🤦 Thank you so much. I changed one job, but was debugging another. 😅

AlekSi:
But I wonder what `required` even means there. I would expect an error when this argument is not passed, not an empty value.

klokar:
It happens when you're doing multiple things at a time 😄 

That's weird. In docs https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputsinput_idrequired it doesn't state that it should fail when it's not present. But how would that be useful?

AlekSi:
https://github.com/actions/runner/issues/1070

PR #137, opened 2021-12-09
CONTRIBTUING.md: fix "/user/.../" -> "/usr/.../"

GinGin3203:
This seems to be a typo. The title is self-explanatory


GinGin3203:
Btw, if the intention of the command is to replace the symlink pointing to the `go` executable to `gotip`, maybe it could better be done with the following:

``$ ln -sf `which gotip` `which go` ``

I think this version is better, because on my system (Ubuntu 21.10) the `go` symlink actually resides in `/usr/bin/go`.

AlekSi (approved):

AlekSi:
The intention is to put the new `go` somewhere closer to the start of the $PATH, not replace the existing Go.

Thank you for the typo fix!

PR #141, opened 2021-12-09
Support `limit` parameter in delete.

### LINKED ISSUE #82: Support limit parameter for delete operations
We should support `limit: 0` for deleting all matching documents and `limit: 1` for deleting only one. Since PostgreSQL does not support `DELETE … LIMIT …`, we should instead do a subquery:
`DELETE FROM xxx WHERE _jsonb->_id IN (SELECT _jsonb->_id FROM xxx WHERE xxx LIMIT 1)`.

OpenSauce:
Closes #82.


OpenSauce:
I've left the tests in both the msg_delete_test and handler_test as I'm not sure which is best... Could probably clean up the test cases a bit

AlekSi:
Let's pick one place to keep the test. We can always move it later.

OpenSauce:
> Let's pick one place to keep the test. We can always move it later.

Just reverted the handler_test... Think keeping it in a separate file is better. Although the SQL handler isn't tested that way.

AlekSi (approved):

AlekSi:
Thanks!

PR #149, opened 2021-12-11
Wait for PostgreSQL on `make env-up`.

### LINKED ISSUE #92: `make env-up` fails on fast computers
PostgreSQL Docker image starts the temporary server, initializes it, then [stops it](https://github.com/docker-library/postgres/blob/2b5355140df4416b0223078f1367da762311aee7/docker-entrypoint.sh#L331). Waiting for port availability is not enough as we can hit the temporary server, and then the test would fail when the temporary server is stopped.

Instead, we should test with `psql` query like `SELECT 1`. Additionally, our `psql` line actually uses a UNIX socket, not a port; that should be fixed too.

agneum:
Closes #92.

### DIFF
--- a/cmd/envtool/main.go
+++ b/cmd/envtool/main.go
@@ -58,15 +58,24 @@ var (
 )
 
 func runCompose(args []string, stdin io.Reader, logger *zap.SugaredLogger) {
+	if err := tryCompose(args, stdin, logger); err != nil {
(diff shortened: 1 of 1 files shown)


AlekSi (approved):

AlekSi:
Thank you!

PR #151, opened 2021-12-12
feat: Add basic support for aggregation pipeline

alvarowolfx:
I started testing with a real-world application here that uses MongoDB as the main database and the first thing that was not working was that most endpoints use `countDocuments` to generate pagination. Nowadays MongoDB drivers use an aggregation pipeline to query the count of collection, especially as you do a count with filters. 

With this PR, I added support for converting some basic aggregate pipeline queries to SQL.

I also made some extra changes to support boolean parameters (7b10341068aa44b27f8e1fb68d0ca29df078b908) and one change to create table even before inserting a doc ( c16f707c94ed68bc93e2d9f253d31bf6013450c2 ) as I was having issues on our app here that run a unique check to see if the name is being used before inserting.

Some example queries that I tested here:
```
test> db.devices.find({})
[
  {
    _id: ObjectId("61b51b3778adccdffe6c799d"),
    projectId: ObjectId("61b4fc61e1058918cb822a63"),
    createdAt: ISODate("2021-12-11T21:42:15.058Z"),
    updatedAt: ISODate("2021-12-11T21:42:15.058Z"),
    name: 'my-device'
  },
  {
    _id: ObjectId("61b64bbf78adccdffe6c799f"),
    name: 'my-second-device',
    projectId: ObjectId("61b4fc61e1058918cb822a63"),
    createdAt: ISODate("2021-12-12T19:21:35.974Z"),
    updatedAt: ISODate("2021-12-12T19:21:35.974Z")
  },
  {
    _id: ObjectId("61b65be378adccdffe6c79a1"),
    createdAt: ISODate("2021-12-12T20:30:27.767Z"),
    updatedAt: ISODate("2021-12-12T20:30:27.767Z"),
    name: 'my-third-device',
    projectId: ObjectId("61b4fc61e1058918cb822a63")
  }
]

test> db.devices.aggregate([{ $group: { _id: "$projectId", n: { $sum: 1 } } }])
[ { _id: ObjectId("61b4fc61e1058918cb822a63"), n: 5 } 

test> db.devices.aggregate([{ $group: { _id: null, count: { $sum: 1 } } }])
[ { count: 5 } ]

test> db.devices.aggregate([{ $group: { _id: 1, count: { $sum: 1 } } }, { $match: { name: "my-device" }}])
[ { count: 1 } ]

test> db.devices.aggregate([{ $group: { _id: "$name", count: { $sum: 1 } } }, { $sort: { _id: 1 }}]) 
[
  { _id: 'my-device', count: 1 },
  { _id: 'my-second-device', count: 1 },
  { _id: 'my-third-device', count: 1 }
]
```


internal/handlers/handler.go
```diff
@@ -236,6 +238,18 @@ func (h *Handler) msgStorage(ctx context.Context, msg *wire.OpMsg) (common.Stora
 		return nil, lazyerrors.Errorf("Handler.msgStorage: %w", err)
 	}
 
+	if !tableExist {
```
AlekSi, line 241: `db.foo.find()`should not create a table as it doesn't create a collection in MongoDB. It should be created on `insert` only.

internal/handlers/jsonb1/msg_aggregate.go
```diff
@@ -0,0 +1,194 @@
...
+	})
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+
+	return &reply, nil
+}
+
+func aggregateToSQL(pipeline types.Array, db, collection string) (string, []interface{}, error) {
+	filterPairs := []interface{}{}
```
AlekSi, line 86: Let's use `any` instead of `interface{}` in new code

internal/handlers/jsonb1/msg_aggregate.go
```diff
@@ -0,0 +1,194 @@
...
+	groupByFields := map[string]string{}
+	sortByFields := map[string]interface{}{}
+	aggregations := map[string]string{}
+	for _, item := range pipeline {
+		if p, ok := item.(types.Document); ok {
+			if p.Map()["$match"] != nil {
```
AlekSi, line 92: Only on `$match`, `$sort`, `$group` can be present, right? If yes, let's replace it with `switch p.Command()`

internal/handlers/jsonb1/msg_aggregate.go
```diff
@@ -0,0 +1,194 @@
...
+								aggFormat = "count(%s)"
+							}
+							if sval, ok := val.(string); ok {
+								aggValue = fmt.Sprintf("_jsonb->'%s')", sval)
+							}
+							aggregations[k] = fmt.Sprintf(aggFormat, aggValue)
+						}
+					}
+					if fieldName, ok := v.(string); ok {
+						groupByFields[k] = fmt.Sprintf("_jsonb->'%s'", strings.TrimPrefix(fieldName, "$"))
```
AlekSi, line 126: I think we should use a placeholder there to avoid SQL injections. The same in other places.
And why trim suffix?

internal/handlers/jsonb1/msg_aggregate.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}
+
+	filter, err := types.MakeDocument(filterPairs...)
+	if err != nil {
+		return "", nil, lazyerrors.Error(err)
+	}
+
+	groupByFieldItems := []string{}
+	groupByClauseItems := []string{}
```
AlekSi, line 139:
```suggestion
	groupByFieldItems := make([]string{}, 0, len(groupByFields))
	groupByClauseItems := make([]string{}, 0, len(groupByFields))
```

Let's allocate the whole slice where we know the capacity. In other places too

internal/types/types.go
```diff
@@ -82,6 +82,8 @@ func validateValue(value any) error {
 		return nil
 	case Regex:
 		return nil
+	case int:
+		return nil
```
AlekSi, line 86: No, the caller should use specific int types (`int32` and `int64`)

AlekSi (commented):
Despite all my comments, that's just awesome!

alvarowolfx:
Hey @AlekSi, sorry for the late reply. All comments make sense, but I'll only be able to work on them later this week. Busy week for me at work and also entering holiday season around here.

AlekSi:
No worries at all!

AlekSi:
@alvarowolfx gentle ping :) Will you be able to continue working on it?

internal/handlers/jsonb1/msg_aggregate.go
alvarowolfx: Yeah, I need to think a bit more about how to add placeholders for those `select`/`group by` clauses. Right now is only handled on the `where` clause.

About the trim suffix, it's because when specifying a field on an aggregation, on mongo it starts with `$`.

Example aggregating devices by field `projectId`:
```
test> db.devices.aggregate([{ $group: { _id: "$projectId", n: { $sum: 1 } } }])
```

internal/handlers/handler.go
alvarowolfx: I removed this change, but we still need to fix this behavior, because on MongoDB, you can query/count documents on a non-existing collection and it handles that gracefully by returning empty results. 

We might need to catch the `relation "non_existing_collection" does not exist (SQLSTATE 42P01)` error and return empty results.

alvarowolfx:
Sorry for the really long delay in coming back to the project. I updated this branch to the latest and worked on most of the suggested changes ( except for some potential SQL injection issues on the `select`/`group by` clauses ).

AlekSi:
Hi @alvarowolfx. We did a pretty big change recently with [0.1.0 release](https://github.com/FerretDB/FerretDB/releases/tag/v0.1.0): all data filtering and aggregation are performed on the Go side for now. Will you be able to update that PR to do that? Or we can close it and implement the basic aggregation later.

AlekSi (commented):
https://github.com/FerretDB/FerretDB/pull/151#issuecomment-1100101142

PR #188, opened 2021-12-19
Support basic `hostInfo` command.

### LINKED ISSUE #157: Support `hostInfo`

ekalinin:
Closes #157 

Without `extra` field yet.

### DIFF
--- a/internal/handlers/handler_test.go
+++ b/internal/handlers/handler_test.go
@@ -418,11 +418,9 @@ func TestReadOnlyHandlers(t *testing.T) {
 		compareFunc func(t testing.TB, actual, expected any)
 	}
 
-	nowStr := "2021-12-19T12:13:05Z"
+	now := time.Now()
 	shared.Now = func() time.Time {
-		dt, err := time.Parse(time.RFC3339, nowStr)
-		require.NoError(t, err)
-		return dt
+		return now
 	}
 	hostname, err := os.Hostname()
 	require.NoError(t, err)
--- a/internal/handlers/handler_test.go
+++ b/internal/handlers/handler_test.go
@@ -418,11 +418,14 @@ func TestReadOnlyHandlers(t *testing.T) {
 		compareFunc func(t testing.TB, actual, expected any)
 	}
 
-	now := time.Now()
+	nowStr := "2021-12-19T12:13:05Z"
 	shared.Now = func() time.Time {
-		return now
+		dt, err := time.Parse(time.RFC3339, nowStr)
+		if err != nil {
+			panic(err)
+		}
+		return dt
 	}
-	nowStr := now.UTC().Format(time.RFC3339)
 	hostname, err := os.Hostname()
 	require.NoError(t, err)
 
--- a/internal/handlers/handler_test.go
+++ b/internal/handlers/handler_test.go
@@ -417,10 +417,6 @@ func TestReadOnlyHandlers(t *testing.T) {
 		compareFunc func(t testing.TB, actual, expected any)
 	}
 
-	now := time.Now().UTC()
-	shared.Now = func() time.Time {
-		return now
-	}
 	hostname, err := os.Hostname()
 	require.NoError(t, err)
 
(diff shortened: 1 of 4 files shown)


internal/handlers/handler_test.go
```diff
@@ -558,6 +571,26 @@ func TestReadOnlyHandlers(t *testing.T) {
 			),
 		},
 
+		"HostInfo": {
+			req: types.MustMakeDocument(
+				"hostInfo", int32(1),
+			),
+			resp: types.MustMakeDocument(
+				"system", types.MustMakeDocument(
+					"currentTime", fmt.Sprintf("ISODate(%s)", nowStr),
```
AlekSi, line 580: I think that's really BSON time; ISODate is how it is represented in Mongo shell.

When you run FerretDB with `make run`, you should see the diff between FerretDB and MongoDB responses. That should make data types visible.

internal/handlers/handler_test.go
```diff
@@ -558,6 +571,26 @@ func TestReadOnlyHandlers(t *testing.T) {
...
+					"hostname", hostname,
+					"cpuAddrSize", fmt.Sprintf("%d", strconv.IntSize),
+					"numCores", fmt.Sprintf("%d", runtime.NumCPU()),
+					"cpuArch", runtime.GOARCH,
+					"numaEnabled", "false",
```
AlekSi, line 585: I think that should be bool, not string. Again, diff output in `make run` should show that

ekalinin: Yep, fixes (as well as other fields). Thanks again.

internal/handlers/handler_test.go
ekalinin: Got it, fixed. Thanks!

internal/handlers/handler_test.go
```diff
@@ -413,6 +417,13 @@ func TestReadOnlyHandlers(t *testing.T) {
 		compareFunc func(t testing.TB, actual, expected any)
 	}
 
+	now := time.Now().UTC()
+	shared.Now = func() time.Time {
```
AlekSi, line 421: And we never set it back.

I think we should be fine with `testutil.CompareAndSetByPath` there, see this test: https://github.com/FerretDB/FerretDB/blob/630a7e9b336abaedb51ff6cd54b9219dbd4bee31/internal/handlers/handler_test.go#L554

internal/handlers/shared/msg_hostinfo.go
```diff
@@ -0,0 +1,65 @@
...
+			"system", types.MustMakeDocument(
+				"currentTime", now,
+				"hostname", hostname,
+				"cpuAddrSize", int32(strconv.IntSize),
+				"numCores", int32(runtime.NumCPU()),
+				"cpuArch", runtime.GOARCH,
+				"numaEnabled", false,
+			),
+			"os", types.MustMakeDocument(
+				"type", cases.Title(language.English).String(runtime.GOOS),
```
AlekSi, line 55: Is `strings.Title` not enough?

ekalinin: It's deprecated:
- https://github.com/golang/go/blob/7c94355b738170cf06484d502af7f2d935831c2b/src/strings/strings.go#L709

So I decided to use suggested replacement - `golang.org/x/text/cases`.

ekalinin: But, of course, I can use `strings.Title` if you don't mind :)

internal/handlers/handler_test.go
ekalinin: > And we never set it back.

Sure, I can reset the default value with `defer`.

> I think we should be fine with `testutil.CompareAndSetByPath` there, see this test

Could you, please, add more context here?
Should I remove this case into a separate test and use ` testutil.CompareAndSetByPath`?

ekalinin: Fixed. Thanks!

AlekSi (approved):

Issue #198, opened 2021-12-21
Add `bits` to `buildInfo` command


AlekSi:
We already have basic `buildInfo` command implementation, but it misses `bits` response field.

Issue #197, opened 2021-12-21
Add `gitVersion` to `buildInfo` command


AlekSi:
We already have basic `buildInfo` command implementation, but it misses `gitVersion` response field.

PR #204, opened 2021-12-21
Improve output of buildInfo command

GinGin3203:
PR for #197 and #198.

Although I'm not completely sure about obtaining the value of architecture bits, but adding another flag to `buildInfo.Settings` seems a bit cleaner than parsing `GOARCH` (especially if you consider that, for example, `amd64p32` is a valid 32-bit compilation target).

### DIFF
--- a/internal/util/version/version.go
+++ b/internal/util/version/version.go
@@ -55,19 +55,6 @@ func init() {
 			info.Commit = s.Value
 		case "vcs.modified":
 			info.Dirty, _ = strconv.ParseBool(s.Value)
-		case "architecture":
-			temp, _ := strconv.ParseInt(s.Value, 10, 32)
-			if temp == 32 {
-				info.Architecture = x86
-			} else if temp == 64 {
-				info.Architecture = x64
-			}
-
 		}
 	}
 }
-
-const (
-	x86 int32 = 32
-	x64 int32 = 64
-)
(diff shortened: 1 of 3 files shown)


AlekSi:
> Although I'm not completely sure about obtaining the value of architecture bits

We currently use `strconv.IntSize` in `MsgHostInfo`.

> but adding another flag to buildInfo.Settings seems a bit cleaner

`buildInfo.Settings` is populated by `go build`, we can't easily change that.

> amd64p32 is a valid 32-bit compilation target

Support for amd64p32 was removed in 1.14 together with NaCl: https://go.dev/doc/go1.14#nacl Currently, there are no plans to support weird architectures like this.

AlekSi:
See also https://github.com/FerretDB/FerretDB/pull/205

internal/util/version/version.go
```diff
@@ -27,9 +27,10 @@ import (
 var version string
 
 type Info struct {
-	Version string
-	Commit  string
-	Dirty   bool
+	Version      string
+	Commit       string
+	Dirty        bool
+	Architecture int32 // either equals to 32 or 64
```
AlekSi, line 33: That's unused now, right?

GinGin3203: Yup, gonna fix it right away

AlekSi (approved):

AlekSi:
Thank you!

PR #212, opened 2021-12-28
Support projections

### LINKED ISSUE #80: Support projections

ekalinin:
Fixes #80.


internal/handlers/jsonb1/msg_find_or_count.go
```diff
@@ -43,13 +43,28 @@ func (h *storage) MsgFindOrCount(ctx context.Context, msg *wire.OpMsg) (*wire.Op
 	db := m["$db"].(string)
 
 	projection, ok := m["projection"].(types.Document)
+	projectionStr := "_jsonb"
 	if ok && len(projection.Map()) != 0 {
-		return nil, common.NewErrorMessage(common.ErrNotImplemented, "MsgFind: projection is not supported")
+		ks := ""
+		for i, k := range projection.Keys() {
+			if i != 0 {
+				ks += ","
```
AlekSi, line 51:
```suggestion
				ks += ", "
```

The same in other places

internal/handlers/jsonb1/msg_find_or_count.go
```diff
@@ -43,13 +43,28 @@ func (h *storage) MsgFindOrCount(ctx context.Context, msg *wire.OpMsg) (*wire.Op
...
+			}
+			ks += " '" + k + "'"
```
AlekSi, line 53: We should use a placeholder there. The same in other places

internal/handlers/handler_test.go
```diff
@@ -498,6 +498,42 @@ func TestReadOnlyHandlers(t *testing.T) {
...
+			resp: types.MustMakeDocument(
+				"cursor", types.MustMakeDocument(
+					"firstBatch", types.MustNewArray(
+						types.MustMakeDocument(
+							"first_name", "WOODY",
+							"last_name", "HOFFMAN",
+						),
+					),
+					"id", int64(0),
+					"ns", "pagila.actor",
```
AlekSi, line 522: I think it might be a bit more clear to set it to empty string there, and patch for both monila and pagila in compareFunc

ekalinin: Good point, fixed. Thanks!

internal/handlers/jsonb1/msg_find_or_count.go
ekalinin: Fixed.

AlekSi:
Thank you!

PR #218, opened 2021-12-28
Add `buildEnvironment` and `debug` to `buildInfo` command

### LINKED ISSUE #199: Add `buildEnvironment` and `debug` to `buildInfo` command
We already have basic `buildInfo` command implementation, but it misses `buildEnvironment ` and `debug` response field.
For `buildEnvironment`, let's use the result of `runtime/debug.BuildInfo.Settings` field.
For `debug`, let's set it to `true` if FerretDB is built with race detector (buildinfo contains `-race=true`) or test coverage enabled (buildinfo contains `-tags=` line with `testcover` tag).

GinGin3203:
Closes #199.


internal/util/version/version.go
```diff
@@ -48,12 +50,20 @@ func init() {
...
+		info.BuildEnvironment[s.Key] = s.Value
 		switch s.Key {
 		case "vcs.revision":
 			info.Commit = s.Value
 		case "vcs.modified":
 			info.Dirty, _ = strconv.ParseBool(s.Value)
+		case "-race":
+			info.IsDebugBuild, _ = strconv.ParseBool(s.Value)
+		case "-tags":
+			if s.Value == "testcover" {
```
AlekSi, line 64: Will it work if there are several build tags set during build?

internal/util/version/version.go
```diff
@@ -48,12 +50,20 @@ func init() {
...
+	info.BuildEnvironment = map[string]string{}
 	for _, s := range buildInfo.Settings {
+		info.BuildEnvironment[s.Key] = s.Value
 		switch s.Key {
 		case "vcs.revision":
 			info.Commit = s.Value
 		case "vcs.modified":
 			info.Dirty, _ = strconv.ParseBool(s.Value)
+		case "-race":
+			info.IsDebugBuild, _ = strconv.ParseBool(s.Value)
```
AlekSi, line 62: Will it work for something like `-tags=testcover -race=false`?

internal/handlers/shared/msg_buildinfo.go
```diff
@@ -47,3 +49,11 @@ func (h *Handler) MsgBuildInfo(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
 
 	return &reply, nil
 }
+
+func buildEnvironment() (buildEnvironment types.Document) {
```
AlekSi, line 53: I think we should not extract that function – we use it only in one place

internal/handlers/handler_test.go
```diff
@@ -422,6 +422,12 @@ func TestReadOnlyHandlers(t *testing.T) {
 	}
 
 	hostname, err := os.Hostname()
+
+	buildEnvironment := types.MustMakeDocument()
+	for k, v := range version.Get().BuildEnvironment {
+		buildEnvironment.Set(k, v)
+	}
```
AlekSi, line 429: Can we test specific values instead of copying a part of implementation?

internal/handlers/shared/msg_buildinfo.go
GinGin3203: Will fix. Also, maybe it would be better to initialize the `buildEnvironment` document during the start, kinda like `version.info` is initialized?
If so, is it ok if I put `buildEnvironment` in the `version` package or should it reside elsewhere?

GinGin3203: Btw, we can initialize the `buildEnvironment` Document during `version.init()`, so no map iteration and copying during runtime. Maybe do that?

AlekSi: Sure, let's do that

internal/handlers/handler_test.go
GinGin3203: `BuildInfo.Settings` is empty when `go test` is run, so `buildInfo.buildEnvironment` field is actually empty when the project is tested. What do we do?

AlekSi: So that code actually does nothing then :) Let's remove it

AlekSi (commented):
There is a merge conflict that should be resolved

AlekSi:
Resolved :)

internal/util/version/version.go
```diff
@@ -55,12 +58,25 @@ func init() {
...
 		case "vcs.revision":
 			info.Commit = s.Value
 		case "vcs.modified":
 			info.Dirty, _ = strconv.ParseBool(s.Value)
+		case "-race":
+			if raceEnabled, _ := strconv.ParseBool(s.Value); raceEnabled {
+				info.IsDebugBuild = true
+			}
+		case "-tags":
+			// TODO: replace for slices.Contains()
```
AlekSi, line 74: We totally could that now :)

AlekSi (approved):

internal/util/version/version.go
GinGin3203: Yeah, just wasn't sure about depending on `golang.org/x/exp` only to use `slices.Contains()` in this place

PR #220, opened 2021-12-29
Support capped collections [WIP].

### LINKED ISSUE #75: Support capped collections and tailable cursors
https://www.mongodb.com/docs/manual/core/capped-collections/
https://www.mongodb.com/docs/manual/core/tailable-cursors/

Blocks #76.

```[tasklist]
### Tasks
- [ ] #3483
- [ ] FerretDB/dance#604
- [ ] #3469
- [ ] #3484
- [ ] #3466
- [ ] #3462
- [ ] #3518
- [ ] #3458
- [ ] #3490
- [ ] #3467
- [ ] FerretDB/dance#605
- [ ] #3498
- [ ] #2815
- [ ] #419
- [ ] #3638
- [ ] #3459
- [ ] #3818
- [ ] #2283
- [ ] #3739
- [ ] #2983
- [ ] #2984
```

ekalinin:
Closes #75.

Implementation details:
- supported `capped` option for `create` command or `db.createCollection` method:
  - for this kind of collections the underlying postgres table will have additional columns (see below)
- supported `max` option for `create` command or `db.createCollection` method:
  - new unique cycled index created (max value is the same as `max` from the command)
  - added 2 new columns into postgres table:
    - `_id` - reference to cycled index to update the oldest record
    - `_dt` - creation date, to support ordered output in `db.find` operation)
  - into `INSERT` statement added `ON CONFLICT` clause to update old record with a new one
- improved `db.find` method for capped collection:
  - the order of the results is the same as the insertion order (see `_dt` column above)

TODO:
- support `size` option (via `pg_relation_size` call before each insert?)
- tests


AlekSi:
Capped collections are much harder to implement than they appear. I would personally don't do them now.

ekalinin:
> Capped collections are much harder to implement than they appear. I would personally don't do them now.

Yeah, especially with the `size` option. Moved PR to `Draft` for now.

AlekSi:
@ekalinin Any news? Maybe we should abandon that PR for now?

PR #244, opened 2022-01-09
Disable optimization and inlining for debug builds

GinGin3203:
Using `-gcflags="all=-N -l"` helps when debugging with Delve

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ bench-short:                           ## Benchmark for 5 seconds
 	bin/benchstat old.txt new.txt
 
 build-testcover: gen-version           ## Build bin/ferretdb-testcover
-	go test -c -o=bin/ferretdb-testcover -trimpath -tags=testcover -race -coverpkg=./... ./cmd/ferretdb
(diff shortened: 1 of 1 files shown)


AlekSi (commented):
Does it actually work and makes the difference? Can Delve handle -trimpath and -race binaries?

GinGin3203:
Ok, so initially I tried `dlv attach` to a running `ferretdb-testcover` binary compiled without `-gcflags` and it says`Warning: debugging optimized function`, but I get the same warning when I attach to the binary compiled _with_ `-gcflags`. Kinda weird, but seems like too much of a rabbit hole to get into.

But in some cases `-gcflags` does seem to help: https://github.com/go-delve/delve/issues/2867. So maybe it's worth having the flag just in case? Also, [Delve docs](https://github.com/go-delve/delve/blob/4c1f8b4b2536beb8b06c6a6ef0b0383a7928e188/Documentation/usage/dlv_exec.md) explicitly recommend compiling with `-gcflags="all=-N -l"` for debug purposes.

AlekSi (commented):
My biggest concern is that we are adding a line _just in case_ without actually checking if it works or does something. In particular, my argument about `-trimpath` and `-race` stands.

GinGin3203:
Ok, I guess you are right...
Guess I'll check it out more in-depth someday, closing for now

PR #286, opened 2022-01-23
Fix some typos and style

ekalinin:


internal/handlers/handler_test.go
```diff
@@ -870,7 +870,7 @@ func TestListDropDatabase(t *testing.T) {
 }
 
 //nolint:paralleltest // we test a global list of collections
-func TestCreateListDropCollection(t *testing.T) {
+func TestCreateDropCollection(t *testing.T) {
```
AlekSi, line 873: There is TODO item in code to test list function once we have it

AlekSi: Also, test fails due to that rename

AlekSi (approved):

PR #292, opened 2022-01-28
Securely build and push Docker images

### LINKED ISSUE #224: Set more Docker image labels
We should set two more [labels](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys): `org.opencontainers.image.version` and `org.opencontainers.image.revision`.

AlekSi:
Refs #70.
Closes #224.


.github/workflows/docker.yml
```diff
@@ -0,0 +1,106 @@
...
+#   Only a few trusted people have permission to do that.
+# * Thanks to the way `pull_request_target` trigger works, PR changes in the workflow itself are not run
+#   until they are merged.
+# * We use a short-living automatic `GITHUB_TOKEN` instead of a long-living personal access token (PAT).
+#   It also has minimal permissions.
+# * We publish Docker images from PRs as a separate package that should not be run by users.
+# * We limit what third-party actions can be used.
+
+# We also tried a different approach: build Docker image in one normal, secure `pull_request` workflow,
+# upload artifact, and the download and publish in another workflow that has access to secrets, but treats
```
AlekSi, line 18:
```suggestion
# upload artifact, and then download and publish it in another workflow that has access to secrets, but treats
```

internal/util/version/version.go
```diff
@@ -64,9 +69,12 @@ func init() {
 	info.BuildEnvironment = types.MustNewDocument()
 	for _, s := range buildInfo.Settings {
 		info.BuildEnvironment.Set(s.Key, s.Value)
+
 		switch s.Key {
 		case "vcs.revision":
-			info.Commit = s.Value
+			if s.Value != info.Commit {
```
pboros, line 75: I like the way that this ensures that the version info which is used for builds is consistent with git.

pboros (approved):
Overall, looks good to me. Too bad importing the image didn't work. With that caveat, this is the best that can be done.

PR #296, opened 2022-01-31
issue-255, $size operator

taaraora:
`$size` query comparison operator for array was implemented.


internal/handlers/handler_test.go
```diff
@@ -80,6 +80,90 @@ func handle(ctx context.Context, t *testing.T, handler *Handler, req *types.Docu
 	return actual
 }
 
+func TestFindArray(t *testing.T) {
+	t.Parallel()
+	ctx, handler, pool := setup(t, nil)
+	schema := testutil.Schema(ctx, t, pool)
```
AlekSi, line 86: This function creates a test-specific PostgreSQL schema / MongoDB database and drops it if the test passes. There is no need to do the manual cleanup with `cleanArrayTestData`.

internal/handlers/handler_test.go
```diff
@@ -80,6 +80,90 @@ func handle(ctx context.Context, t *testing.T, handler *Handler, req *types.Docu
...
+	loadArrayTestData(ctx, t, handler, schema)
```
AlekSi, line 87: Let's start making a test dataset that would be shared between various query tests. I was thinking about documents with two fields: string `_id` for identification and `value` of various types. That would allow us to test edge cases like calling `$size` operators on NaN double value. https://github.com/FerretDB/dance/pull/33/files#diff-ee104391e4d88f69dce464c4f136bdeeb5c3d9a3b494a639416e6e0454afd4c2R86-R157

internal/handlers/handler_test.go
```diff
@@ -80,6 +80,90 @@ func handle(ctx context.Context, t *testing.T, handler *Handler, req *types.Docu
...
+	defer cleanArrayTestData(ctx, t, handler, schema)
+
+	req := must.NotFail(types.NewDocument(
+		"find", "dmts",
```
AlekSi, line 91: `testutil.TableName` instead of dmts

internal/handlers/jsonb1/where.go
```diff
@@ -142,6 +142,13 @@ func fieldExpr(field string, expr *types.Document, p *pg.Placeholder) (sql strin
 			// {field: {$gte: value}}
 			sql += "_jsonb->" + p.Next() + " >="
 			argSql, arg, err = scalar(value, p)
+		case "$size":
+			// { field: { $size: 2 } }
+			sql += "jsonb_array_length(_jsonb::jsonb->" + p.Next() + ") = "
+
+			var sizeValue any
+			sizeValue, err = expr.Get(expr.Command())
+			argSql, arg = fmt.Sprintf("%s", p.Next()), []any{sizeValue}
```
AlekSi, line 151:
```suggestion
			argSql, arg = p.Next(), []any{sizeValue}
```

internal/handlers/jsonb1/where.go
```diff
@@ -142,6 +142,13 @@ func fieldExpr(field string, expr *types.Document, p *pg.Placeholder) (sql strin
 			// {field: {$gte: value}}
 			sql += "_jsonb->" + p.Next() + " >="
 			argSql, arg, err = scalar(value, p)
+		case "$size":
+			// { field: { $size: 2 } }
+			sql += "jsonb_array_length(_jsonb::jsonb->" + p.Next() + ") = "
```
AlekSi, line 147: I _think_ `::jsonb` casting should not be required there – `_jsonb` column already has this type.

internal/handlers/jsonb1/where.go
```diff
@@ -142,6 +142,13 @@ func fieldExpr(field string, expr *types.Document, p *pg.Placeholder) (sql strin
...
+
+			var sizeValue any
+			sizeValue, err = expr.Get(expr.Command())
```
AlekSi, line 150:
```suggestion
			sizeValue, err = expr.Get(op)
```

But we should also check type

internal/handlers/jsonb1/where.go
taaraora: I totally agree.

internal/handlers/handler_test.go
taaraora: Got it, `cleanArrayTestData` is removed, thank you.

internal/handlers/handler_test.go
taaraora: I tried to mimic, that helped to catch new a bug.
As I got from documentation `$size` is used only with arrays.

internal/handlers/handler_test.go
taaraora: Thank you, fixed.

internal/handlers/jsonb1/where.go
taaraora: Thank you, fixed.

internal/handlers/jsonb1/where.go
taaraora: I've updated the SQL query with type check, thank you!

AlekSi: We should check that the `$size` operator is called on the array with integer (btw, can it be float?) argument, and return MongoDB-compatible errors if types are different. We should also have tests for those errors.

I think that invoking `jsonb_array_length` on the argument and checking the error code later might be easier than doing a type checking in SQL.

AlekSi (commented):
https://github.com/FerretDB/FerretDB/pull/296#discussion_r797619467

internal/handlers/jsonb1/where.go
taaraora: 1. I've added tests for `$size`.
2. Regarding invoking `jsonb_typeof()=array` in SQL. When I filled up the collection with mixed types values `jsonb_array_length()` function just started throwing exceptions for non-array values. At the same time, Mongo handles such cases silently skipping non-array values. So I think that type check is justified.

AlekSi (commented):
I added integration tests. Please make this one pass: https://github.com/FerretDB/dance/blob/51b6c9ba578243af1763d457d255161a151909ad/tests/ferret/core_test.go#L193-L200

Please also merge with the `main` branch

internal/handlers/jsonb1/where.go
AlekSi: > silently skipping non-array values

That what I was missing. Thanks.

internal/handlers/common/validate.go
```diff
@@ -0,0 +1,44 @@
...
+// limitations under the License.
+
+package common
+
+import (
+	"fmt"
+	"math"
+)
+
+func IsPositiveInteger(value any) error {
```
AlekSi, line 22: Let's not generalize and move to common that function just yet. Let's keep it separate, but in jsonb1 package, with a name something like `validateSizeArg`. Once we have several similar fucntions, then we would generalize them.

I think that function can also use `common.NewError` instead of error strings

internal/handlers/handler_test.go
```diff
@@ -37,6 +40,12 @@ import (
 	"github.com/FerretDB/FerretDB/internal/wire"
 )
 
+type testCase struct {
```
AlekSi, line 43: That's a common type for query tests, but not for inserts or updates, for example. Let's move it inside TestSize

internal/handlers/handler_test.go
```diff
@@ -69,15 +87,197 @@ func handle(ctx context.Context, t *testing.T, handler *Handler, req *types.Docu
...
+
+		t.Run(tcName, func(t *testing.T) {
+			t.Parallel()
+
+			actual, err := handleWithError(ctx, t, handler, tc.req)
+			expected := must.NotFail(types.NewDocument(
+				"cursor", must.NotFail(types.NewDocument(
+					"firstBatch", tc.resp,
+					"id", int64(0),
+					"ns", schema+"."+testutil.TableName(parentT),
```
AlekSi, line 219: Maybe we can make `ns` before subtest and avoid using `parentT`?

internal/handlers/handler_test.go
```diff
@@ -69,15 +87,197 @@ func handle(ctx context.Context, t *testing.T, handler *Handler, req *types.Docu
...
+				)),
+				"ok", float64(1),
+			))
+
+			if tc.err != nil && assert.ErrorAs(t, tc.err, &err) {
+				assert.Equal(t, tc.err.Error(), err.Error())
+				// TODO: compare error code
```
AlekSi, line 226: That's actually quite important. `assert.Equal(t, tc.err, err)` does not work?

internal/handlers/handler_test.go
```diff
@@ -69,15 +87,197 @@ func handle(ctx context.Context, t *testing.T, handler *Handler, req *types.Docu
...
+			} else {
```
AlekSi, line 227:
```go
if tc.err != nil {
  // requires and asserts
  return
}
```

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

internal/handlers/jsonb1/where.go
```diff
@@ -142,6 +142,17 @@ func fieldExpr(field string, expr *types.Document, p *pg.Placeholder) (sql strin
...
+		case "$size":
+			// { field: { $size: 2 } }
+			var sizeValue any
+			sizeValue, err = expr.Get(op)
+			if err = common.IsPositiveInteger(sizeValue); err != nil {
+				err = common.NewError(common.ErrBadValue, err)
+				return
+			}
+
+			sql += "jsonb_typeof(_jsonb->" + p.Next() + ") = 'array' AND jsonb_array_length(_jsonb->" + p.Next() + ") = "
```
AlekSi, line 154: That should be the same placeholder (number) to avoid sending the same argument (field name) twice

AlekSi (approved):

PR #307, opened 2022-02-05
Remove extra allocation

peakle:


internal/wire/bits.go
```diff
@@ -15,7 +15,6 @@
 package wire
 
 import (
-	"fmt"
 	"strings"
```
AlekSi, line 18: That should be just `import “string”` now

AlekSi (approved):
LGTM with a small tweak required

AlekSi (approved):

Issue #251, opened 2022-01-13
Support all comparison query operators


AlekSi:
Support all query comparison operators for all data types.

- [x] $eq
- [x] $gt
- [x] $gte
- [x] $lt
- [x] $lte
- [ ] $in
- [x] $ne
- [x] $nin

ribaraka:
I would like to work on this one.

AlekSi:
I propose to do everything end-to-end (implementation, tests, integration tests) for one operator (for example, `$eq`), then do the rest.

ribaraka:
got it. I will create seperate PRs for each operator.

AlekSi:
Sure, that's a good idea. But I propose is to do and _merge_ one PR first, then start with the rest. Otherwise, you would have to change several PRs after each review cycle and do more busy work in the end.

ribaraka:
Agreed. I'm going to do one PR after each merge.

PR #309, opened 2022-02-07
Support `$eq` comparision operator

ribaraka:
Refs #251.


.gitignore
```diff
@@ -3,6 +3,7 @@
 
 /bin/
 /vendor/
+.idea
```
AlekSi, line 6: Please read the comment on top of the file

internal/handlers/jsonb1/where.go
```diff
@@ -34,7 +34,11 @@ func scalar(v any, p *pg.Placeholder) (sql string, args []any, err error) {
 	case string:
 		sql = "to_jsonb(" + p.Next() + "::text)"
 		arg = v
-	case types.ObjectID:
+	case types.ObjectID,
+		float64,
+		int64,
+		*types.Array,
+		types.Binary:
```
AlekSi, line 41: Let's put them on the same line

internal/handlers/msg_comparison_test.go
```diff
@@ -0,0 +1,345 @@
+package handlers
+
+import (
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/must"
+	"github.com/FerretDB/FerretDB/internal/util/testutil"
+	"github.com/stretchr/testify/assert"
```
AlekSi, line 7: Imports should be grouped; check other files

internal/handlers/msg_comparison_test.go
```diff
@@ -0,0 +1,345 @@
...
+	Int32MinDataType               = "int32-min"
+	Int64DataType                  = "int64"
+	Int64ZeroDataType              = "int64-zero"
+	Int64MaxDataType               = "int64-max"
+	Int64MinDataType               = "int64-min"
+	TimestampDataType              = "timestamp"
+	SchemaWithAllTypes             = "values"
+	CollectionWithAllTypes         = "values"
+	BinaryDataType                 = "binary"
+)
```
AlekSi, line 40: Why we need those constants?

internal/handlers/msg_comparison_test.go
```diff
@@ -0,0 +1,345 @@
...
+		Int32MinDataType:               int32(-2147483648),
+		Int64DataType:                  int64(42),
+		Int64ZeroDataType:              int64(0),
+		Int64MaxDataType:               int64(9223372036854775807),
+		Int64MinDataType:               int64(-9223372036854775808),
+		TimestampDataType:              types.Timestamp(42),
+		ArrayDataType:                  must.NotFail(types.NewArray("array", int32(42))),
+		EmptyArrayDataType:             must.NotFail(types.NewArray()),
+		BinaryDataType:                 types.Binary{Subtype: types.BinaryUser, B: []byte{42, 0, 13}},
+	}
```
AlekSi, line 68: Why we need that map?

AlekSi (commented):
Please sign the CLA, merge with `main`, fix CI issues

internal/handlers/msg_comparison_test.go
ribaraka: I've been thinking it's convenient to set the data type names to constants and then use them as keys to access the data value.
Regarding to schema and collection constants, if the name of schemes or collections changes, it will also be easier to change the values of constants in one place than to run through all files and change manually

internal/handlers/msg_comparison_test.go
ribaraka: I made a map that corresponds to the values of the table, as a convenience so that in each test case or in another test I don’t write the value manually.

.gitignore
ribaraka: Sorry. The line has been removed

internal/handlers/jsonb1/where.go
ribaraka: Done.

internal/handlers/msg_comparison_test.go
ribaraka: Fixed.

internal/handlers/jsonb1/where.go
```diff
@@ -34,7 +34,10 @@ func scalar(v any, p *pg.Placeholder) (sql string, args []any, err error) {
 	case string:
 		sql = "to_jsonb(" + p.Next() + "::text)"
 		arg = v
-	case types.ObjectID:
+	case bool:
+		sql = "to_jsonb(" + p.Next() + "::bool)"
+		arg = v
+	case types.ObjectID, float64, int64, *types.Array, types.Binary:
```
AlekSi, line 40: types.Array is not a scalar type, it should not be there. https://github.com/FerretDB/FerretDB/blob/7466a8091080de4724e9f6b87d923c87a47d39fb/internal/types/types.go#L28-L44

internal/handlers/msg_comparison_test.go
AlekSi: Let's remove both constants and this map: data is fixed by test_db, not them, and also replacing one constant with another does not achieve much.

internal/handlers/msg_comparison_test.go
```diff
@@ -0,0 +1,430 @@
...
+	EmptyBinaryDataType            = "binary-empty"
+	BoolFalseDataType              = "bool-false"
+	BoolTrueDataType               = "bool-true"
+	DateTimeDataType               = "datetime"
+	DateTimeEpochDataType          = "datetime-epoch"
+	DateTimeMinYearDataType        = "datetime-year-min"
+	DateTimeMaxYearDataType        = "datetime-year-max"
+)
+
+func TestComparison(t *testing.T) {
```
AlekSi, line 62: Let's integrate it into `TestFind`

internal/handlers/msg_comparison_test.go
ribaraka: Agreed

internal/handlers/msg_comparison_test.go
ribaraka: I'll do it.
Should I leave a `findValueByComparisonOperator` func?

internal/handlers/jsonb1/where.go
AlekSi: Let's make a separate function for now. Once we use it in several places, we will see if it has more in common with `scalar` or not. Right now, I'm not sure if we would need to convert the whole array as often as scalar values.

internal/handlers/jsonb1/where.go
```diff
@@ -68,6 +72,18 @@ func scalar(v any, p *pg.Placeholder) (sql string, args []any, err error) {
 	return
 }
 
+func composite(v any, p *pg.Placeholder) (sql string, args []any, err error) {
```
AlekSi, line 75: Please add documentation comment

internal/handlers/jsonb1/where.go
```diff
@@ -146,9 +162,17 @@ func fieldExpr(field string, expr *types.Document, p *pg.Placeholder) (sql strin
 			argSql, arg, err = common.InArray(value.(*types.Array), p, scalar)
 		case "$eq":
 			// {field: {$eq: value}}
-			// TODO special handling for regex
 			sql += "_jsonb->" + p.Next() + " ="
-			argSql, arg, err = scalar(value, p)
+
+			_, isDocument := value.(*types.Document)
+			_, isArray := value.(*types.Array)
+			_, isRegex := value.(types.Regex)
```
AlekSi, line 169: Regex is not a composite type

ribaraka: It just fits the `composite` function so conveniently.

internal/handlers/handler_test.go
```diff
@@ -438,6 +438,605 @@ func TestFind(t *testing.T) {
 				)),
 			)),
 		},
+		"StringEq": {
```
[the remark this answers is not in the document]
ribaraka: Okay. Should this pattern also be applied to other Eq cases?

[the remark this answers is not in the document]
AlekSi: Of course

internal/handlers/jsonb1/where.go
AlekSi: Well, does it makes the reading and understanding easier? I would expect a function called `composite` to do something (not sure what without documentation) with a composite type, not with a composite type _and regex_

internal/handlers/handler_test.go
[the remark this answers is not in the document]
ribaraka: Fixed.

internal/handlers/jsonb1/where.go
ribaraka: The func was refactored. Documentation to the refactored func was added

internal/handlers/jsonb1/where.go
ribaraka: The func was refactored.

AlekSi (changes requested):
Some of our dance tests fail with MongoDB, meaning that FerretDB and MongoDB behave differently. That's a problem that should be fixed.

ribaraka:
I removed the TODO regex from the $eq comparison operator. I have not find any information in the documentation and on the Internet that these two expressions were used.
I also tried to issue query in mongosh `db.values.find( { value:  {$eq :/foo/i } } )` to the values database manually and it returns only match stored regexes. I mean is there possibility who would store regex datatype in a document? Anyway current implementation FerretDB returns nothing. 
What do you think we will do next?

AlekSi:
> I mean is there possibility who would store regex datatype in a document?

Yes.

> Anyway current implementation FerretDB returns nothing.

We should add a dance test for that case and fix that difference in behavior between MongoDB and FerretDB.

internal/handlers/common/filter.go
```diff
@@ -289,17 +291,34 @@ func filterFieldExpr(fieldValue any, expr *types.Document) (bool, error) {
...
-		return false, err
+	case *types.Array:
+		for i := 0; i < fieldValue.Len(); i++ {
+			arrValue := must.NotFail(fieldValue.Get(i)).(any)
+			s, isString := arrValue.(string)
+			if !isString {
+				continue
+			}
+
+			re, err := regex.Compile()
```
seeforschauer, line 310: regex Compile can be done once right at the func beginning

seeforschauer (changes requested):

internal/handlers/common/filter.go
ribaraka: fixed

internal/handlers/common/filter.go
```diff
@@ -157,14 +157,24 @@ func filterFieldExpr(fieldValue any, expr *types.Document) (bool, error) {
...
+			fieldValueRegex, isFieldValueRegex := fieldValue.(types.Regex)
+			switch {
+			case isValueDocument, isValueArray:
+				return compare(fieldValue, exprValue) == equal, nil
+
+			case isExprValueRegex && isFieldValueRegex:
+				return filterFieldRegex(fieldValueRegex, exprValueRegex)
+
+			default:
+				return compareScalars(fieldValue, exprValue) == equal, nil
```
seeforschauer, line 177: in case if one of variables is regex, it will panic what you probably don't want. Isn't it?

internal/handlers/common/filter.go
```diff
@@ -289,17 +299,31 @@ func filterFieldExpr(fieldValue any, expr *types.Document) (bool, error) {
...
 	if err != nil {
 		return false, err
 	}
 
-	return re.MatchString(s), nil
+	switch fieldValue := fieldValue.(type) {
+	case string:
+		return re.MatchString(fieldValue), nil
+
+	case *types.Array:
```
seeforschauer, line 311: what cases are being processed here? It's probably not only `{field: /regex/}` ?

internal/handlers/common/filter.go
```diff
@@ -289,17 +299,31 @@ func filterFieldExpr(fieldValue any, expr *types.Document) (bool, error) {
...
+			arrValue := must.NotFail(fieldValue.Get(i)).(any)
+			s, isString := arrValue.(string)
+			if !isString {
+				continue
+			}
+			if re.MatchString(s) == true {
+				return true, nil
+			}
+		}
+	case types.Regex:
```
seeforschauer, line 322: If `regex == regex` ? Don't we expect that it's not equal since we are expecting the comparison towards scalar?

internal/handlers/common/filter.go
```diff
@@ -289,17 +299,31 @@ func filterFieldExpr(fieldValue any, expr *types.Document) (bool, error) {
...
+		return fieldValue == regex, nil
```
AlekSi, line 323: I think we should use `compareScalars` there incapsulate all comparisons in that function

internal/handlers/common/filter.go
```diff
@@ -157,14 +157,24 @@ func filterFieldExpr(fieldValue any, expr *types.Document) (bool, error) {
 			continue
 		}
 
-		exprValue := must.NotFail(expr.Get(exprKey))
+		exprValue := must.NotFail(expr.Get(exprKey)).(any)
```
AlekSi, line 160: That `.(any)` part doesn't do anything

internal/handlers/common/filter.go
ribaraka: logic has been rewritten, panic should not appear

internal/handlers/common/filter.go
ribaraka: Good idea. The logic has been implemented

internal/handlers/common/filter.go
ribaraka: fixed

internal/handlers/common/filter.go
seeforschauer: yes, thank you, I mean to add this example to comments above.

ribaraka: `// filterFieldRegex handles {field: /regex/} filter. Provides regular expression capabilities for pattern matching strings  in queries, even if the strings are in an array.`
@seeforschauer, is it okay to add such a comment?

seeforschauer: yes, thank you it makes cases clear

seeforschauer (approved):

AlekSi (approved):

AlekSi:
Thank you!

PR #314, opened 2022-02-11
Clean-up `bson` package

### LINKED ISSUE #260: Remove `types`, refactor `bson` packages
Problems with the current design:
1. If type conversion is used (for example, from bson.doubleType to float64), validation can be skipped, and we might want to make NaN values invalid in the future.
2. If type conversion is not used, then too much copying might be required.
3. "convert" functions are confusing.
4. Too many types are exported.
5. Remove `nolint:deadcode`.
6. Reduce memory allocations.

DoD:
- [ ] The issue number shouldn't be mentioned in the code anymore (all the places that mention it should be  refactored).

AlekSi:
Closes #260.


seeforschauer (dismissed):

AlekSi:
Closing for now

PR #328, opened 2022-02-22
Add fuzzing tests for handlers

### LINKED ISSUE #326: Add fuzzing tests for handlers

AlekSi:
Closes #326.

### DIFF
--- a/internal/handlers/handler.go
+++ b/internal/handlers/handler.go
@@ -70,11 +70,14 @@ func New(opts *NewOpts) *Handler {
 
 // Handle handles the message.
 //
-// Message handlers should:
+// Specific message handlers should do one of the following:
 //  * return normal response body;
 //  * return protocol error (*common.Error) - it will be returned to the client;
-//  * return any other error - it will be returned to the client as InternalError before terminating connection;
-//  * panic - that will terminate the connection without a response.
+//  * return any other error - it will be returned to the client as InternalError before terminating connection.
+//
+// They should not panic on bad input, but may do so in "impossible" cases.
+// They also should not use recover(). That allows us to use fuzzing.
+// (Panics terminate the connection without a response on a different level.)
 //
 //nolint:lll // arguments are long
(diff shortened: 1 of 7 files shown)


seeforschauer (approved):
nothing suspicious,  
and need an intro for fuzzing some day.

AlekSi:
Oops, sorry, that one is not ready yet. I'm extracting smaller things from that PR for easier review.

internal/handlers/handler.go
```diff
@@ -133,43 +136,70 @@ func (h *Handler) Handle(ctx context.Context, reqHeader *wire.MsgHeader, reqBody
 		fallthrough
 	default:
 		requests.WithLabelValues(command).Inc()
-		panic(fmt.Sprintf("unexpected OpCode %s", reqHeader.OpCode))
+
+		err = lazyerrors.Errorf("unexpected OpCode %s", reqHeader.OpCode)
```
seeforschauer, line 140: is the OpCode is unexpected, return here?

seeforschauer: no I see now - it's to be processed.
But the error is not handled somehow, it's overwritten by `	b, err := resBody.MarshalBinary()` later on

internal/handlers/handler.go
```diff
@@ -133,43 +136,70 @@ func (h *Handler) Handle(ctx context.Context, reqHeader *wire.MsgHeader, reqBody
...
 	}
 
+	// set body for error
 	if err != nil {
```
AlekSi, line 144: there

internal/handlers/handler.go
AlekSi: It is handled a few lines below

internal/handlers/handler.go
seeforschauer: now I see.

internal/handlers/handler.go
seeforschauer: ok

seeforschauer (approved):

PR #330, opened 2022-03-01
Support `$gt` comparision operator

ribaraka:
Refs https://github.com/FerretDB/FerretDB/issues/251.


AlekSi (commented):
Let's finish #309 first

internal/handlers/common/compare.go
```diff
@@ -122,7 +122,10 @@ func compareScalars(a, b any) compareResult {
 
 	case time.Time:
 		b, ok := b.(time.Time)
-		if ok {
+		switch {
+		case ok && a.Unix() != b.Unix():
+			return compareOrdered(a.Unix(), b.Unix())
+		case ok && a.Unix() == b.Unix():
```
AlekSi, line 128: I think it can be simplified to something like:

```go
if !ok {
  return notEqual
}

return compareOrdered(a.UnixNano(), b.UnixNano())
```

To avoid `==` and `!=` there, and to simplify

internal/handlers/common/filter.go
```diff
@@ -181,8 +181,18 @@ func filterFieldExpr(fieldValue any, expr *types.Document) (bool, error) {
 
 		case "$gt":
 			// {field: {$gt: exprValue}}
-			if c := compareScalars(fieldValue, exprValue); c != greater {
-				return false, nil
+			_, isValueDocument := fieldValue.(*types.Document)
+			_, isValueArray := fieldValue.(*types.Array)
```
AlekSi, line 185: Can it be a types assertion? Something like (not checked):

```go
switch fieldValue.(type) {
	case *types.Document, *types.Array:
		if compare(fieldValue, exprValue) != greater {
			return false, nil
		}
	default:
		if compareScalars(fieldValue, exprValue) != greater {
			return false, nil
		}
```

AlekSi: Actually, can we just™ make `compare` support arrays and documents?

internal/handlers/common/compare.go
ribaraka: The logic has been simplified

internal/handlers/common/filter.go
AlekSi: I think yes, but we may do that in the next PR

AlekSi: Ah, nevermind, that's already done :)

AlekSi (approved):

AlekSi:
Thank you!

PR #337, opened 2022-03-07
Improve `make env-up` for windows users

w84thesun:


AlekSi (changes requested):
Conflicts, CI fails, etc

PR #339, opened 2022-03-08
Add helper for getting required parameters

AlekSi:
Refs #326.

### DIFF
--- a/internal/handlers/common/required.go
+++ b/internal/handlers/common/required.go
@@ -1,0 +1,40 @@
+// Copyright 2021 FerretDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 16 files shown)


internal/handlers/common/required.go
```diff
@@ -0,0 +1,40 @@
...
+package common
+
+import (
+	"fmt"
+
+	"github.com/FerretDB/FerretDB/internal/types"
+)
+
+// GetRequiredParam returns doc's value for key or protocol error for missing or invalid parameter.
+func GetRequiredParam[T types.ScalarType](doc *types.Document, key string) (T, error) {
```
w84thesun, line 24: It seems that `GetRequiredParam` tightly bounded with `types.Document`. Maybe we should place it somewhere near `types.Document` declaration?

AlekSi: It returns handler-level error. Essentially, that's request validator

w84thesun: Ok then.

w84thesun (approved):

Issue #336, opened 2022-03-07
db.collection.find() works differently for FerretDB and for MongoDB


ribaraka:
The ferretDB finds certain numbers according to the data type, while MongoDB finds all numbers, according to `filter`, no matter if it's int32/64 or even if it's located in array.

There is my output in `MongoDB`:
 **db.values.find({"value": 42})**
{ "_id" : ObjectId("612ec2800000040100000401"), "name" : "array", "value" : [ "array", 42 ] }
{ "_id" : ObjectId("612ec2800000040400000404"), "name" : "array-three", "value" : [ 42, "foo", null ] }
{ "_id" : ObjectId("612ec2800000100100001001"), "name" : "int32", "value" : 42 }
{ "_id" : ObjectId("612ec2800000120100001201"), "name" : "int64", "value" : NumberLong(42) }

And `FerretDB` Output:
**db.values.find({"value": 42})**
[ { _id: ObjectId("612ec2800000100100001001"),  name: 'int32', value: 42  }] 

As you can see, there is only one value found for Ferret DB.

ribaraka:
I have already begun to study this moment with the bug and I want to take on solving it myself, since my [task](https://github.com/FerretDB/FerretDB/issues/251) that I have been doing depends on the solution of this problem.

AlekSi:
I'm not sure if the problem is still present (it is possible I accidentally fixed it), but we definitely need handler and dance tests for it.

If something should be changed, it is likely there: https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/common/compare.go

Issue #335, opened 2022-03-07
Fix fluky tests


AlekSi:
https://github.com/FerretDB/FerretDB/runs/5450924795?check_suite_focus=true

PR #356, opened 2022-03-12
db.collection.find() works differently for FerretDB and for MongoDB [WIP]

ribaraka:
Ferretdb's find method has been refactored and now works similarly to MongoDB but the comparison operators fail.

Tests are failed.
 "InLteGte" one of the test cases that is fails on the moment with $lte comparison operator, pgsql query that is send:
```
SELECT _jsonb FROM monila.actor WHERE (_jsonb->'last_name' IN (to_jsonb('HOFFMAN'::text))) AND
                                      (_jsonb->'actor_id' >= to_jsonb('50'::jsonb)) OR
                                      (jsonb_typeof(_jsonb->'actor_id')) = 'object' AND
                                       _jsonb->'actor_id' @> to_jsonb('{"$l":"50"}'::jsonb) AND
                                       _jsonb->'actor_id' <= to_jsonb('100'::jsonb) OR
                                      (jsonb_typeof(_jsonb->'actor_id')) = 'object' AND
                                       _jsonb->'actor_id' @> to_jsonb('{"$l":"100"}'::jsonb);
```
This test returns `2` result instead of `1`.


AlekSi:
Tests fail due to #335

internal/handlers/jsonb1/where.go
```diff
@@ -31,8 +31,8 @@ import (
 func scalar(v any, p *pg.Placeholder) (sql string, args []any, err error) {
 	var arg any
 	switch v := v.(type) {
-	case int32:
-		sql = "to_jsonb(" + p.Next() + "::int4)"
+	case int32, int64:
+		sql = "to_jsonb(" + p.Next() + "::jsonb)"
```
AlekSi, line 35: That's two different cases, and I think we can use `fjson.Marshal` for int64 (maybe for both)

internal/handlers/jsonb1/where.go
```diff
@@ -266,6 +266,8 @@ func wherePair(key string, value any, p *pg.Placeholder) (sql string, args []any
 	default:
 		// {field: value}
 		switch value.(type) {
+		case int32, int64:
+			sql = fmt.Sprintf("jsonb_typeof(_jsonb->%[1]s) = 'object' AND _jsonb->%[1]s @> to_jsonb('{\"$l\":\"%v\"}'::jsonb) OR _jsonb->%[1]s @>", p.Next(), value)
```
AlekSi, line 270: I think that should be two different `case`s at the very least.

And `'{\"$l\":\"%v\"}'` is definitly not secure. We use `scalar` function below, so I think it could work there too

internal/handlers/jsonb1/where.go
ribaraka: The cases are different but both cases behave the same. 
I tried to use it through `fjson.Marshal` (for int64 and for int32) but it doesn't return values like MongoDB does.

internal/handlers/jsonb1/where.go
taaraora: @AlekSi Why it is not secure? the value passed to the Sprintf() is checked to be a number.

ribaraka:
In https://github.com/FerretDB/FerretDB/pull/356/commits/b9c203b97011e56c7f71ce399c4a1eb6a9c412cf commit I have removed cast for int32 from scalar. It's perplexing to write a scalar for int64 and ind32 separately at the moment because it has to be written in two different conditions for a column. 
` WHERE (_jsonb->'value'->>'$l')::numeric = 42 or (_jsonb->'value' @> to_jsonb('42'::jsonb))`
Although there might be possible to write defer func.

So I suggest we use a special sql query for number datatype without using `scalar`.

Also In this PR tests with operator comparison `$gt` `$lt` are failed due to removed cast from scalar. I think the logic with operator comparison should be rewritten for number type as well.

The query for comparison operators which I have come up with so far. Works not yet identical to MongoDB, but that's all I could think of for now.
`SELECT _jsonb FROM values.values WHERE (_jsonb->'value'->>'$l')::numeric > 41 OR
        (_jsonb->'value'->>'$f')::numeric > 41 OR
    (jsonb_typeof(_jsonb->'value') = 'number' AND (_jsonb->'value') > to_jsonb('41'::jsonb));`
I stuck on searching a query for comparison operators. I'm trying to find something like I use to search for all int32 and the presence of an array like this:
`(_jsonb->'value' @> to_jsonb('42'::jsonb))`
Otherwise there is an option to write sql function. But I would like to know opinion or suggestions on how you would have solve this issue.

AlekSi:
@ribaraka @taaraora My current thinking is that we probably could move comparison logic into the stored procedure. That should greatly simplify our queries. I will try to hack a prototype today.

AlekSi:
> I will try to hack a prototype today.

I made a prototype that showed that it can be done, but it will take quite some time to implement properly. So we rewrote everything in Go.

See https://github.com/FerretDB/FerretDB/issues/336#issuecomment-1073716855

ribaraka:
> I made a prototype that showed that it can be done, but it will take quite some time to implement properly. So we rewrote everything in Go.
> 
> 
> See [[#336 (comment)](https://github.com/FerretDB/FerretDB/issues/336#issuecomment-1073716855)]
> 

Thanks for the prototype! I am working on improving it. It still doesn't work as it should, but with the new implementation it seems like it would be easier to achieve the expected result.

internal/handlers/handler_test.go
```diff
@@ -309,6 +309,472 @@ func TestFind(t *testing.T) {
...
+			resp: must.NotFail(types.NewArray(
+				must.NotFail(types.NewDocument(
+					"_id", types.ObjectID{0x61, 0x2e, 0xc2, 0x80, 0x00, 0x00, 0x04, 0x01, 0x00, 0x00, 0x04, 0x01},
+					"name", "array",
+					"value", must.NotFail(types.NewArray("array", int32(42))),
+				)),
+				must.NotFail(types.NewDocument(
+					"_id", types.ObjectID{0x61, 0x2e, 0xc2, 0x80, 0x00, 0x00, 0x04, 0x04, 0x00, 0x00, 0x04, 0x04},
+					"name", "array-three",
+					"value", must.NotFail(types.NewArray(int32(42), "foo", types.NullType{})),
```
AlekSi, line 516: Just `types.Null` there and in other places

internal/handlers/common/compare.go
```diff
@@ -140,9 +164,29 @@ func compareScalars(a, b any) compareResult {
...
 			return compareOrdered(a, b)
 		default:
-			panic(fmt.Sprintf("unexpected type %T", b))
+			return notEqual
 		}
 
+	case *types.Document:
+		return notEqual
+
+	case *types.Array:
```
AlekSi, line 173: array and document should be there – they are not scalars

internal/handlers/handler_test.go
ribaraka: fixed

internal/handlers/common/compare.go
ribaraka: array and document are not scalars so when compared with scalar types they return not equal

ribaraka:
> @ribaraka is it ready for re-review?

yes, review please

internal/handlers/common/compare.go
```diff
@@ -140,9 +164,12 @@ func compareScalars(a, b any) compareResult {
 		case int64:
 			return compareOrdered(a, b)
 		default:
-			panic(fmt.Sprintf("unexpected type %T", b))
+			return notEqual
 		}
 
+	case *types.Document, *types.Array:
+		return notEqual
```
AlekSi, line 171: That function is called `compareScalars` so it should compare scalar values. Document and Array are not scalars: https://github.com/FerretDB/FerretDB/blob/17a490df209cd40816003e095fe3db5753a4f527/internal/types/types.go#L28-L44

We should either add another function like `func compare(a, b any) compareResult` that would do compare for documents, arrays, and scalars, or make the caller check for document/array. We should not compare non-scalar values in a function called `compareScalars`

internal/handlers/common/filter.go
```diff
@@ -73,6 +73,21 @@ func filterDocumentPair(doc *types.Document, filterKey string, filterValue any)
...
 	default:
 		// {field: value}
+		switch docValue := docValue.(type) {
+		case *types.Document:
+			return false, nil
+		case *types.Array:
+			for i := 0; i < docValue.Len(); i++ {
+				arrValue, err := docValue.Get(i)
+				if err != nil {
+					panic(fmt.Sprintf("cannot get value from array, err is %v, array is %v, index is %v", err, arrValue, i))
```
AlekSi, line 83: That's one case where `must.NotFail(docValue.Get(i))` can be used safely

internal/handlers/common/compare.go
```diff
@@ -47,72 +48,92 @@ func compareScalars(a, b any) compareResult {
 	case float64:
 		switch b := b.(type) {
 		case float64:
+			if math.IsNaN(a) && math.IsNaN(b) {
+				return equal
```
AlekSi, line 52: That's how MongoDB works? Interesting…

internal/handlers/common/compare.go
```diff
@@ -47,72 +48,92 @@ func compareScalars(a, b any) compareResult {
...
-		case 0:
-			return equal
-		case -1:
-			return less
-		case 1:
-			return greater
-		default:
-			panic("unreachable")
+		b, ok := b.(types.Binary)
+		if ok {
```
AlekSi, line 72:
```suggestion
		if !ok {
		  return notEqual
		}
```

Same idea as in https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow – unindent logic, return early

The same in other places

internal/handlers/handler_test.go
```diff
@@ -704,6 +1170,25 @@ func TestFind(t *testing.T) {
...
+
+		response := handle(ctx, t, handler, req)
+		firstBatch, err := testutil.GetByPath(t, response, "cursor", "firstBatch").(*types.Array).Get(0)
+		require.NoError(t, err)
+		responseValue, err := firstBatch.(*types.Document).Get("value")
+		require.NoError(t, err)
+		if nan, ok := responseValue.(float64); ok {
+			assert.True(t, math.IsNaN(nan))
+		}
+	})
```
AlekSi, line 1191: I'm working on a thing that should make writing those tests easier, please bear a bit longer :)

internal/handlers/common/compare.go
ribaraka: How about rename `compareScalars` to `compare`?

AlekSi: I think there are cases where we care only about scalars. Put another way, I think there are cases where we should actually compare documents and arrays, and cases where they are not comparable. So let's _add_ a function `compare` that would return `notEqual` for documents and arrays (for now) and call `compareScalars` for other types

internal/handlers/common/compare.go
ribaraka: Yes, if you are looking for NaN then you retrieve a document with a NaN value

internal/handlers/common/compare.go
ribaraka: Agreed.

internal/handlers/common/compare.go
ribaraka: Suggestions have been applied.

AlekSi (approved):

AlekSi:
Thank you!

PR #358, opened 2022-03-14
Replace `Makefile` with `Taskfile`

### LINKED ISSUE #347: Replace Makefile with Taskfile
https://taskfile.dev/

That would remove our only dependency on Unix tools, allowing an easier setup for contributors on Windows (without WSL).

w84thesun:
Closes #347.


AlekSi:
My only concern is that we might use a different version of `task` that is installed globally with `go install` or package manager for some other project. So I think we should say that `go generate` followed by `./bin/task XXX` is the official way, but `task init` / `task XXX` is likely to work too.

.github/workflows/go.yml
```diff
@@ -30,24 +30,27 @@ jobs:
           fetch-depth: 0
 
       - name: Start pulling environment in the background
-        run: make env-pull &
+        run: ./bin/task env-pull &
```
AlekSi, line 33: Not for this PR, but can task run targets in parallel?

.github/workflows/go.yml
```diff
@@ -78,8 +81,19 @@ jobs:
...
 
+      - name: Setup Go
+        uses: FerretDB/github-actions/setup-go@main
+        with:
+          cache-key: fuzz
+
+      - name: Install Task
+        run: |
+          cd tools
+          go generate -x
```
AlekSi, line 92: Why not `cd tools; go generate -x` as in other steps?

CONTRIBUTING.md
```diff
@@ -30,23 +30,24 @@ $ git remote add origin git@github.com:<YOUR_GITHUB_USERNAME>/FerretDB.git
 
 ## Helpful Commands
 
-1. Install tools with `make init`.
-2. Start the development environment with `make env-up`.
+0. Install [task](https://taskfile.dev/) with `go install github.com/go-task/task/v3/cmd/task@latest`
```
AlekSi, line 33: I think we better start counting from 1 :)

CONTRIBUTING.md
```diff
@@ -30,23 +30,24 @@ $ git remote add origin git@github.com:<YOUR_GITHUB_USERNAME>/FerretDB.git
...
 
-You can see all available "make" commands with `make help`.
+You can see all available "task" commands with `task`.
 
 ## Code Overview
 
 Package `cmd` provides commands implementation. `ferretdb` is the main FerretDB binary; others are tools for development.
 
-Package `tools` uses "tools.go" approach to fix tools versions. They are installed into `bin/` by `make init`.
+Package `tools` uses "tools.go" approach to fix tools versions. They are installed into `bin/` by `task init`.
```
AlekSi, line 50: `./bin` for consistency.

(`bin/XXX` should work on any Unix, no need to sepcify `./`. But I'm not sure about Windows)

Taskfile.yml
```diff
@@ -0,0 +1,193 @@
...
+    desc: "Generate code"
+    cmds:
+    - ./bin/gofumpt
+    - go generate -x ./...
+    - task: fmt
+
+  fmt:
+    desc: "Format code"
+    cmds:
+#      - ./bin/gofumpt{{.EXT}}
```
AlekSi, line 67: Do we need that comment?

Taskfile.yml
```diff
@@ -0,0 +1,193 @@
...
+      - go test -race -shuffle=on -coverprofile=cover.txt -coverpkg=./... ./...
+      - go test -race -shuffle=on -bench=. -benchtime=1x ./...
+
+
+  bench-short:
+    desc: "Benchmark for about 20 seconds (with default BENCHTIME)"
+    cmds:
+      - go test -list='Benchmark.*' ./...
+#      - {{.RM}} new.txt
+      # Running four functions for $(BENCHTIME) each..."
```
AlekSi, line 81: That comment was a hack to echo a message with BENCHTIME substituted. I think we can remove it or print a message by some other mean.

Taskfile.yml
```diff
@@ -0,0 +1,193 @@
...
+    desc: "Run FerretDB"
+    deps:
+      - build-testcover
+    cmds:
+      - bin/ferretdb-testcover -test.coverprofile=cover.txt -mode=diff-normal -listen-addr=:27017
+
+  lint:
+    desc: "Run linters"
+    cmds:
+      - bin/go-sumtype bin/golangci-lint ## Run linters
```
AlekSi, line 136: That was a dependency, not a command: `make lint` depends on `bin/go-sumtype` and `bin/golangci-lint` targets.

CONTRIBUTING.md
w84thesun: `bin\task.exe` would work

AlekSi: The I think we should say "`bin/task XXX` on Linux or macoS,  `bin\task.exe XXX` on Windows" once and then use `bin/XXX` everywhere

Taskfile.yml
w84thesun: So it was to check wheter this tool installed, right?

Taskfile.yml
w84thesun: Removed it

Taskfile.yml
w84thesun: I've added echo to show BENCHTIME

.github/workflows/go.yml
w84thesun: Fixed

Taskfile.yml
AlekSi:
```
lint: bin/go-sumtype
  cmd1 ...

bin/go-sumtype:
  cmd2 ...
```

When invoked as `make lint`, GNU Make first checks if `bin/go-sumtype` file exist (and "fresh"). If it doesn't, it invokes  `bin/go-sumtype` Makefile target first, running `cmd2 ...`.

The current Taskfile invokes `bin/go-sumtype` with argument `bin/golangci-lint` which does nothing useful

(please re-request review if review's attention is needed)

w84thesun: After digging into docs and some researching I've ended up with adding preconditions that would check files existance and output error which would look like this:
```sh
task: Please run `task init` first
task: precondition not met
```
Every other options would make tasks more complicated.

.github/workflows/go.yml
w84thesun: We can do something like this, as docs said:
```
You can also make the tasks given by the command line run in parallel by using the --parallel flag (alias -p). Example: task --parallel js css.
```

AlekSi (approved):

PR #363, opened 2022-03-15
use docker image from docker hub for quick start

Xyolyp:
use the docker image from docker hub in the sample docker-compose.yml file as the github hosted one seems to have an invalid manifest


AlekSi:
What error you get with the one from ghcr?

Xyolyp:
@AlekSi sorry for the confusion but I'm unable to reproduce the error. You may close this pullrequest without merging it.
The error was something about the manifest but I don't remember the specific wording.

PR #365, opened 2022-03-17
Fix Taskfile

w84thesun:
Fix variable error;
Fix echo


AlekSi (changes requested):
CI fails.

Please try it locally :)

AlekSi (approved):

PR #377, opened 2022-03-21
Projection: `<field>: <1 or true>` and `<field>: <0 or false>`

seeforschauer:


internal/handlers/common/projection.go
```diff
@@ -19,5 +19,15 @@ import "github.com/FerretDB/FerretDB/internal/types"
 // ProjectDocuments modifies given documents in places according to the given projection.
 func ProjectDocuments(docs []*types.Document, projection *types.Document) error {
 	// TODO
+
+	projectionMap := projection.Map()
```
seeforschauer, line 23: it can be done w/o creation of new variable..

internal/handlers/jsonb1/msg_count.go
```diff
@@ -100,9 +100,6 @@ func (s *storage) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
 		resDocs = append(resDocs, doc)
 	}
 
-	if err = common.SortDocuments(resDocs, sort); err != nil {
-		return nil, err
-	}
```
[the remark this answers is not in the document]
seeforschauer: I'd better won't change it here.

internal/handlers/jsonb1/msg_count.go
```diff
@@ -99,7 +99,6 @@ func (s *storage) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
 
 		resDocs = append(resDocs, doc)
 	}
-
```
AlekSi, line 102: Let's keep a blank line there to separate that block

internal/handlers/jsonb1/msg_count.go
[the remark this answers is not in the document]
AlekSi: No, sorting is really not required. But that's not for that PR

internal/handlers/jsonb1/msg_count.go
```diff
@@ -45,7 +45,7 @@ func (s *storage) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
 		"let",
 	}
 	if err := common.Unimplemented(document, unimplementedFields...); err != nil {
-		return nil, err
```
AlekSi, line 48: No, `Unimplemented` returns protocol error that we should return as-is to the client

internal/handlers/common/projection.go
```diff
@@ -18,6 +18,17 @@ import "github.com/FerretDB/FerretDB/internal/types"
...
 // ProjectDocuments modifies given documents in places according to the given projection.
 func ProjectDocuments(docs []*types.Document, projection *types.Document) error {
-	// TODO
+	if projection.Len() == 0 {
+		return nil
+	}
+	projectionMap := projection.Map()
+	for i := 0; i < len(docs); i++ {
+		for k := range docs[i].Map() {
+			if _, ok := projectionMap[k]; !ok {
```
AlekSi, line 27: We should check the value too. For example, 0 or false should remove the field.

Or the plan is to do that in a different PR?

seeforschauer: Then I'd add new tests for it too.

seeforschauer: I'd rather add tests and add code to comply with https://docs.mongodb.com/manual/reference/method/db.collection.find/#projection

AlekSi: In this PR, or in different one?

seeforschauer: In this PR

internal/handlers/common/projection.go
```diff
@@ -14,10 +14,47 @@
...
+
+			switch v := v.(type) {
+			case bool:
+				if !v {
+					docs[i].Remove(k)
+					continue
+				}
+				continue
+
+			case int32, int64, float32, float64:
```
AlekSi, line 48: float32 is not possible

internal/handlers/common/projection.go
```diff
@@ -14,10 +14,47 @@
...
 // ProjectDocuments modifies given documents in places according to the given projection.
 func ProjectDocuments(docs []*types.Document, projection *types.Document) error {
-	// TODO
+	if projection.Len() == 0 {
+		return nil
+	}
+	projectionMap := projection.Map()
+	for i := 0; i < len(docs); i++ {
+		for k := range docs[i].Map() {
+			if k == "_id" {
```
AlekSi, line 30: _id handing is more complicated than that

internal/handlers/common/projection.go
```diff
@@ -14,10 +14,47 @@
...
+				continue
+			}
+
+			v, ok := projectionMap[k]
+			if !ok {
+				docs[i].Remove(k)
```
AlekSi, line 36: Do we have a dance test for that? It is not clear from documentation how MongoDB behaves there

internal/handlers/common/projection.go
```diff
@@ -14,10 +14,47 @@
...
+				continue
+			}
+
+			switch v := v.(type) {
+			case bool:
+				if !v {
+					docs[i].Remove(k)
+					continue
+				}
+				continue
```
AlekSi, line 46: We don't need both those `continue` statements

internal/handlers/common/projection.go
```diff
@@ -14,10 +14,47 @@
...
+
+			case int32, int64, float32, float64:
+				if compareScalars(v, int32(0)) == equal {
```
AlekSi, line 49: Can we use compareNumbers?

internal/handlers/common/projection.go
```diff
@@ -14,10 +14,47 @@
...
+					docs[i].Remove(k)
+					continue
```
AlekSi, line 51: We don't need that `continue`

internal/handlers/common/projection.go
seeforschauer: then case with int32 , int64 is not handled. `compareScalars` is more convenient here.

internal/handlers/common/projection.go
seeforschauer: removed

internal/handlers/common/projection.go
seeforschauer: no, dance has no tests for it at all,
 can it be merged to not to stop Dmitry?

AlekSi: Sure

AlekSi (approved):

PR #378, opened 2022-03-21
Add SortDocuments

w84thesun:


w84thesun:
I've added 2 tests and looks like there some problems with `Less` method.

internal/handlers/common/sort.go
```diff
@@ -14,10 +14,159 @@
...
+		// p == q; try the next comparison.
+	}
+	// All comparisons to here said "equal", so just return whatever
+	// the final comparison reports.
+	return ds.sorts[k](p, q)
+}
+
+func getSortType(value any) (sortType, error) {
+	switch value.(type) {
+	case int32:
```
AlekSi, line 135: There might be the same problem as #379: it is possible that MongoDB can accept any whole number there, including int64 and float64, similar to https://github.com/FerretDB/FerretDB/blob/b0804b43a0fae38be0b00e503733981af8ad5528/internal/handlers/common/filter.go#L304-L317

But we should check that with dance tests first

w84thesun: Checked with dance tests. Updated getSortType.

internal/handlers/common/sort.go
```diff
@@ -14,10 +14,173 @@
...
+		return nil
+	}
+
+	if sort.Len() > 32 {
+		return lazyerrors.Errorf("maximum sort keys exceeded: %v", sort.Len())
+	}
+
+	sortFuncs := make([]sortFunc, len(sort.Keys()))
+	for i, sortKey := range sort.Keys() {
+		sortField, err := sort.Get(sortKey)
```
AlekSi, line 46: That's one place where we can and even should use `must.NotFail(sort.Get(sortKey))`. We iterate over keys, so that document must have that key.

Using panics in "impossible" situations (that's what must.NotFail does) instead of returning errors will make fuzzing easier: https://github.com/FerretDB/FerretDB/blob/7613293be8c353eee7b9ced7954424800b7411a3/internal/handlers/handler_fuzz_test.go#L62-L63

internal/handlers/common/sort.go
```diff
@@ -14,10 +14,173 @@
...
+			return err
+		}
+		sortType, err := getSortType(sortField)
+		if err != nil {
+			return err
+		}
+
+		sortKey := sortKey
+		sortFuncs[i] = func(a, b *types.Document) bool {
+			sortType := sortType
```
AlekSi, line 57: That's not needed, we don't change it

internal/handlers/common/sort.go
```diff
@@ -14,10 +14,173 @@
...
+		sortField, err := sort.Get(sortKey)
+		if err != nil {
+			return err
+		}
+		sortType, err := getSortType(sortField)
+		if err != nil {
+			return err
+		}
+
+		sortKey := sortKey
```
AlekSi, line 55: We don't run things in parallel there, so that's not needed

internal/handlers/common/sort.go
```diff
@@ -14,10 +14,173 @@
...
+		sortFuncs[i] = func(a, b *types.Document) bool {
```
AlekSi, line 56: Let's call it lessFuncs or something (and add a comment) because sorting function that accepts two documents and returns a boolean raises questions

internal/handlers/common/sort.go
```diff
@@ -14,10 +14,173 @@
...
+	var sortValue int
+
+	switch value := value.(type) {
+	case int32:
+		sortValue = int(value)
+	case int64:
+		sortValue = int(value)
+	case float64:
+		if value != math.Trunc(value) || math.IsNaN(value) || math.IsInf(value, 0) {
+			return 0, NewErrorMsg(ErrBadValue, "$size must be a whole number")
```
AlekSi, line 146: Not $size.

But I think we can make a common function that accepts a value and something for correct error messages and returns a whole number as int64

internal/handlers/common/sort_test.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+func TestSortDocuments(t *testing.T) {
+	t.Parallel()
+
+	type args struct {
+		docs []*types.Document
+		sort *types.Document
+	}
+	tests := []struct {
```
AlekSi, line 32: testCases for consistency

internal/handlers/common/sort_test.go
```diff
@@ -0,0 +1,131 @@
...
+			sorted: []*types.Document{
+				types.MustNewDocument("_id", int32(3), "name", "Empire State Pub", "borough", "Brooklyn"),
+				types.MustNewDocument("_id", int32(5), "name", "Jane's Deli", "borough", "Brooklyn"),
+				types.MustNewDocument("_id", int32(1), "name", "Central Park Cafe", "borough", "Manhattan"),
+				types.MustNewDocument("_id", int32(4), "name", "Stan's Pizzaria", "borough", "Manhattan"),
+				types.MustNewDocument("_id", int32(2), "name", "Rock A Feller Bar and Grill", "borough", "Queens"),
+			},
+		},
+	}
+	for _, tt := range tests {
```
AlekSi, line 119: tc

internal/handlers/common/sort_test.go
```diff
@@ -0,0 +1,131 @@
...
+		tt := tt
+
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			if err := SortDocuments(tt.args.docs, tt.args.sort); (err != nil) != tt.wantErr {
```
AlekSi, line 125: I think we could simplify that condition to 

```
if tc.wantErr != nil {
  assert.Equal(t, tc.wantErr, err)
  return
}
assert.NoError(t, err)
```

internal/handlers/common/sort.go
w84thesun: Thanks, I'll try to keep that in mind in the future.

internal/handlers/common/sort.go
w84thesun: Added it to params.go.

internal/handlers/common/error.go
```diff
@@ -40,6 +40,7 @@ const (
 	ErrProjectionInEx    = ErrorCode(31253) // Location31253
 	ErrProjectionExIn    = ErrorCode(31254) // Location31254
 	ErrRegexOptions      = ErrorCode(51075) // Location51075
+	ErrIllegalKey        = ErrorCode(15974) // Location15974
```
seeforschauer, line 43: 1. suggest format of the errors:  `Err <Entity Name> <error details related to this entity>`
2. use of this error in context when the sort type is wrong, can it be renamed to smth like `err sort type wrong`?

internal/handlers/common/params.go
```diff
@@ -65,3 +67,24 @@ func AssertType[T types.Type](value any) (T, error) {
 
 	return res, nil
 }
+
+// AssertInt64Value asserts value's type returning protocol error for bad float values and unexpected results.
+// valueType parameter used to generate error message.
+func AssertInt64Value(valueType string, value any) (int64, error) {
```
seeforschauer, line 73: What you do here it's not an assertion, it's type casting if it suits. Shall the function be renamed, WDYT?

internal/handlers/common/sort.go
```diff
@@ -14,10 +14,160 @@
...
+		}
+	case greater:
+		switch sort {
+		case ascending:
+			cmp = false
+		case descending:
+			cmp = true
+		}
+	case notEqual, equal:
+		return false // ???
```
seeforschauer, line 169: what meant be `???` here?

seeforschauer (changes requested):

internal/handlers/common/params.go
seeforschauer: `GetInt64Value`?

internal/handlers/common/error.go
w84thesun: I think `ErrSortBadValue` would be ok as it's about bad sort value

seeforschauer: It is.

internal/handlers/common/sort.go
w84thesun: removed this comment. it marked notEqual situation but it seems that it's ok to return false.

seeforschauer:
dance sort is failing https://github.com/FerretDB/dance/runs/5658676496?check_suite_focus=true#step:7:44

w84thesun:
> dance sort is failing https://github.com/FerretDB/dance/runs/5658676496?check_suite_focus=true#step:7:44

Fixed it.

seeforschauer (dismissed):

internal/handlers/common/error.go
```diff
@@ -40,6 +40,7 @@ const (
 	ErrProjectionInEx    = ErrorCode(31253) // Location31253
 	ErrProjectionExIn    = ErrorCode(31254) // Location31254
 	ErrRegexOptions      = ErrorCode(51075) // Location51075
+	ErrSortBadValue      = ErrorCode(15974) // Location15974
```
AlekSi, line 43: Let's keep them sorted by the error code

internal/handlers/common/params.go
```diff
@@ -65,3 +67,24 @@ func AssertType[T types.Type](value any) (T, error) {
 
 	return res, nil
 }
+
+// GetInt64Value matches value's type returning protocol error for bad float values and unexpected results.
+// valueType parameter used to generate error message.
+func GetInt64Value(valueType string, value any) (int64, error) {
```
AlekSi, line 73: MongoDB calls it "number" in several places. Let's use that name somehow. For example:
```
// GetNumberParam checks if the given value is a whole number and returns it as int64.
// key is used in error messages if the value is not right.
func GetNumberParam(value any, key string) (int64, error)
```

Then the implementation should be cleaned from any references to "sort" as it is not only about sorting.

We should use that function at filterFieldExprSize. And we should move a TODO item about negative zero from that function to the new one.

internal/handlers/common/sort.go
```diff
@@ -14,10 +14,149 @@
...
+}
+
+// matchSortResult matching sort type and compare result.
+func matchSortResult(sort sortType, result compareResult) bool {
+	cmp := false
+	switch result {
+	case less:
+		switch sort {
+		case ascending:
+			cmp = true
```
AlekSi, line 146: It seems that all those assignments could be returns.

And, TBH, that function is harder to understand when it is extracted ("there match sort type and compare result?.. why?.."). We use it one place, so maybe don't extract it at all?

internal/handlers/common/params.go
```diff
@@ -65,3 +67,26 @@ func AssertType[T types.Type](value any) (T, error) {
 
 	return res, nil
 }
+
+// GetNumberParam matches value's type returning protocol error for bad float values and unexpected results.
```
AlekSi, line 71: It, however, does not return protocol error.

internal/handlers/common/params.go
```diff
@@ -65,3 +67,26 @@ func AssertType[T types.Type](value any) (T, error) {
...
+// valueType parameter used to generate error message.
+func GetNumberParam(valueType string, value any) (int64, error) {
```
AlekSi, line 73: Why the first parameter is called "valueType"? It is not a type

internal/handlers/common/params.go
w84thesun: Yeah, it is left after some refactoring.

internal/handlers/common/params.go
AlekSi: > We should use that function at filterFieldExprSize.

That wasn't done

AlekSi (commented):
`filterFieldExprSize` should be updated to use `GetNumberParam`

AlekSi (approved):

PR #380, opened 2022-03-21
Fix small issues after rewrite

AlekSi:


internal/handlers/jsonb1/msg_count.go
```diff
@@ -33,29 +32,15 @@ func (s *storage) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
 
 	unimplementedFields := []string{
 		"skip",
-		"returnKey",
```
seeforschauer, line 36: started to support this or it was added by a mistake?

internal/handlers/jsonb1/msg_count.go
```diff
@@ -69,14 +54,11 @@ func (s *storage) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
...
-	var filter, sort *types.Document
+	var filter *types.Document
 	var limit int32
 	if filter, err = common.GetOptionalParam(document, "query", filter); err != nil {
 		return nil, err
 	}
-	if sort, err = common.GetOptionalParam(document, "sort", sort); err != nil {
-		return nil, err
-	}
 	if limit, err = common.GetOptionalParam(document, "limit", limit); err != nil {
```
seeforschauer, line 62: limit is not necessary for count, isn't it?

internal/handlers/jsonb1/msg_count.go
AlekSi: It was a part of "find or count" that was copied to count by mistake

seeforschauer (changes requested):
pls remove limit from count

seeforschauer (approved):

Issue #379, opened 2022-03-21
Support `int64` `limit` values


AlekSi:
Dance tests fail with `(BadValue) parameter "limit" has type int64 (expected int32)`.

It seems that `limit` can be both int64 and int32. We should also check if it can be whole float64 just like `$size`'s argument.

Fix the problem, add a bunch of tests.

PR #381, opened 2022-03-21
issue 379: support limit `int64`

seeforschauer:


AlekSi:
That's not what #379 says

AlekSi:
https://github.com/FerretDB/FerretDB/blob/b0804b43a0fae38be0b00e503733981af8ad5528/internal/handlers/common/filter.go#L304-L317

But we should check that with dance tests first

AlekSi:
#399

PR #382, opened 2022-03-22
Projection: `<field>: <1 or true>` and `<field>: <0 or false>`: error messages formatting 

seeforschauer:
beginning https://github.com/FerretDB/FerretDB/pull/377 
dance https://github.com/FerretDB/dance/pull/83

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,12 +53,6 @@ Package `tools` uses "tools.go" approach to fix tools versions. They are install
   They use `fjson` package for storing data in PostgreSQL in jsonb columns, but they don't use `bson` package – all data is represented as built-in and `types` types.
 
 
-## Useful links
-
-* escaping https://github.com/lib/pq/issues/782
-* jsonb https://www.postgresql.eu/events/pgconfeu2019/sessions/session/2555/slides/221/jsonpath-pgconfeu-2019.pdf
-* jsnpath guide https://github.com/obartunov/sqljsondoc/blob/master/jsonpath.md
-
 ## Agreements
 
 * in where and projection, let functions be functions, not methods
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,3 +51,21 @@ Package `tools` uses "tools.go" approach to fix tools versions. They are install
   Responses are then converted to `wire`/`bson` messages and sent back to the client.
 * `handlers` handle protocol commands.
   They use `fjson` package for storing data in PostgreSQL in jsonb columns, but they don't use `bson` package – all data is represented as built-in and `types` types.
+
+
+## Useful links
+
+* https://www.postgresql.org/docs/current/functions-json.html
+* https://www.postgresql.org/docs/current/datatype-json.html
+* https://doxygen.postgresql.org/jsonb_8c_source.html
+* https://www.postgresql.org/docs/current/indexes-expressional.html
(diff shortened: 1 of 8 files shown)


internal/handlers/common/error.go
```diff
@@ -67,6 +67,10 @@ func NewErrorMsg(code ErrorCode, msg string) error {
 	return NewError(code, errors.New(msg))
 }
 
+func NewErrorMsgf(code ErrorCode, msg string, args []any) error {
```
AlekSi, line 70: I specifically removed a similar function before: formatting a message should be caller's responsibility

seeforschauer: 1. removed this func from
1. added an agreements in the Contributing file

internal/handlers/common/projection.go
```diff
@@ -32,33 +34,33 @@ func isProjectionInclusion(projection *types.Document) (inclusion bool, err erro
 		case bool:
 			if v {
 				if exclusion {
-					err = lazyerrors.New(errMsg)
```
AlekSi, line 35: That's exactly why there is no XXXf function for errors – `errMsg` is not a template for printf-like function, that's a message. Using messages where templates are expected is one of the common sources of security problems. For example, consider what would happen if the message becomes "projection must be 100% inclusions or exclusions".

In short, it should be either `fmt.Errorf("%s", errMsg)` or `errors.New(errMsg)`

CONTRIBUTING.md
```diff
@@ -51,3 +51,21 @@ Package `tools` uses "tools.go" approach to fix tools versions. They are install
   Responses are then converted to `wire`/`bson` messages and sent back to the client.
 * `handlers` handle protocol commands.
   They use `fjson` package for storing data in PostgreSQL in jsonb columns, but they don't use `bson` package – all data is represented as built-in and `types` types.
+
+
+## Useful links
+
+* https://www.postgresql.org/docs/current/functions-json.html
```
AlekSi, line 58: Are they still useful? :)

CONTRIBUTING.md
```diff
@@ -51,3 +51,21 @@ Package `tools` uses "tools.go" approach to fix tools versions. They are install
...
+* https://www.postgresql.org/docs/current/pageinspect.html
+* escaping https://github.com/lib/pq/issues/782
+* jsonb https://www.postgresql.eu/events/pgconfeu2019/sessions/session/2555/slides/221/jsonpath-pgconfeu-2019.pdf
+* jsnpath guide https://github.com/obartunov/sqljsondoc/blob/master/jsonpath.md
+
+## Agreements
+
+* in where and projection, let functions be functions, not methods
+* use placeholders, must not use sanitizers / bare values
+* formatting a message should be caller's responsibility
```
AlekSi, line 71: Tbh, I think this is less than useful without context.

I would add a comment near the NewError definition mentioning that there are no "constructors" that accept template parameters to avoid pitfalls associated with them.

CONTRIBUTING.md
seeforschauer: three ones are. Left them.

internal/handlers/common/projection.go
seeforschauer: now see.

CONTRIBUTING.md
seeforschauer: added a reference to the NewError func here, and some explanations from this PR there

AlekSi (approved):

PR #383, opened 2022-03-22
Support `$elemMatch` projection query operator

### LINKED ISSUE #257: Support `$elemMatch` projection query operator

seeforschauer:
Closes #257.


internal/types/document.go
```diff
@@ -277,6 +278,40 @@ func (d *Document) Remove(key string) {
 	panic(fmt.Sprintf("types.Document.Remove: key not found: %q", key))
 }
 
+// RemoveByPath removes document by path, doing nothing if the key does not exist.
+func RemoveByPath(d *Document, keys ...string) {
```
AlekSi, line 282: That should be a method to match `GetByPath`

internal/types/document.go
```diff
@@ -277,6 +278,40 @@ func (d *Document) Remove(key string) {
...
+
+	if len(keys) == 1 {
+		d.Remove(key)
+		return
+	}
+
+	key2 := keys[1]
+	switch x := d.m[key].(type) {
+	case *Document:
+		delete(x.m, key2)
```
AlekSi, line 300: That's not enough: we should remove from x.keys. Overall, let's use existing methods: Get, Remove, etc. They have all checks

internal/handlers/common/projection.go
```diff
@@ -85,34 +108,114 @@ func ProjectDocuments(docs []*types.Document, projection *types.Document) error
...
+								continue internal
+							}
+							switch cmpVal := cmpVal.(type) {
+							case *types.Document:
+								docVal, err := cmpVal.Get(k2ConditionField)
+								if err != nil {
+									types.RemoveByPath(doc, k1, strconv.Itoa(j))
+									continue
+								}
+								if testutil.EqualScalars(docVal, elemMatchFieldCondition) {
```
AlekSi, line 199: We definitly should not use testutil in non-testing code – that's a general-purpose functions. For example, it compares NaNs. All that is documented

internal/handlers/common/projection.go
```diff
@@ -85,34 +108,114 @@ func ProjectDocuments(docs []*types.Document, projection *types.Document) error
...
+		}
+	}
+
+	return nil
+}
+
+func projectDocument(inclusion bool, doc *types.Document, projection *types.Document) error {
+	projectionMap := projection.Map()
+
+docMapLoopK1:
```
AlekSi, line 124: Let's refactor this function to reduce cyclomatic complexity and to avoid loop labels. I think some functions can be extracted

internal/util/testutil/equal.go
```diff
@@ -166,20 +166,37 @@ func equalArrays(v1, v2 *types.Array) bool {
...
 //  * float64 NaNs are equal to each other;
 //  * time.Time values are compared using Equal method.
-func equalScalars(v1, v2 any) bool {
+func EqualScalars(v1, v2 any) bool {
 	switch s1 := v1.(type) {
 	case float64:
 		s2, ok := v2.(float64)
 		if !ok {
-			return false
+			s3, ok := v2.(float32)
```
AlekSi, line 177: No, that's a function is a version for assert.Equal that can compare NaNs. For testing. We don't want our testing helper to be smart about types.

AlekSi: For the same reason we don't want `assert.Equal(t, 0, 0.0)` to pass. In tests, we want the comparison to be exact and without type conversions. In finding data we want to do automatic type conversions for numbers, for example, just like MongoDB would do.

internal/types/document.go
AlekSi: We already have this method:

```go
func (d *Document) GetByPath(path ...string) (any, error)
```

I think it makes sense for `RemoveByPath` to be a method too:

```go
func (d *Document) RemoveByPath(path ...string)
```

seeforschauer: thank you, understood, made a method.

internal/types/document.go
seeforschauer: fixed.

internal/handlers/common/projection.go
seeforschauer: changed to compareScalars as agreed on the call.

internal/util/testutil/equal.go
seeforschauer: ok, thank you, that was not visible and clear (RFC for NaN comparison)

internal/handlers/common/projection.go
seeforschauer: done, thank you for note.

internal/util/testutil/equal.go
AlekSi: Now let's revert all changes in that function – unexport it, etc.

internal/types/document.go
```diff
@@ -277,6 +278,40 @@ func (d *Document) Remove(key string) {
 	panic(fmt.Sprintf("types.Document.Remove: key not found: %q", key))
 }
 
+// RemoveByPath removes document by path, doing nothing if the key does not exist.
+func (d *Document) RemoveByPath(keys ...string) {
```
AlekSi, line 282: Let's add some tests

AlekSi:
Merged the last breaking PR for today, moved handler_test.go as is

internal/util/testutil/equal.go
seeforschauer: done

internal/types/document.go
seeforschauer: added tests, fixed 1 bug )

AlekSi (approved):

PR #390, opened 2022-03-23
Minor CI and build tweaks

AlekSi:


.github/workflows/go.yml
```diff
@@ -46,6 +46,9 @@ jobs:
       - name: Run init
         run: bin/task init
 
+      - name: Regenerate and reformat
+        run: bin/task gen fmt
```
AlekSi, line 50: To ensure that files are regenerated and formatted. Checked by `git diff --exit-code` below

.github/workflows/go.yml
```diff
@@ -176,16 +179,6 @@ jobs:
...
 
-      - name: Upload all fuzz corpora
-        if: failure()
-        uses: actions/upload-artifact@v2
-        with:
-          name: fuzz-corpus
-          path: |
-            ${{ env.GOCACHE }}/fuzz
-            **/testdata/fuzz
-            fuzz-corpus
```
AlekSi, line 187: We don't need that because fuzztool and https://github.com/FerretDB/fuzz-corpus are reliable now

Taskfile.yml
```diff
@@ -70,7 +70,6 @@ tasks:
     desc: 'Benchmark for about 20 seconds (with default BENCHTIME)'
     cmds:
       - go test -list='Benchmark.*' ./...
-      - task: build:rm-bin
```
AlekSi, line 73: That simply did not work

Taskfile.yml
```diff
@@ -122,14 +121,24 @@ tasks:
 
   lint:
     desc: 'Run linters'
+    deps:
+      - lint-golangci-lint-required
+      - lint-go-sumtype
+      - lint-go-consistent
```
AlekSi, line 127: Deps are run in parallel, so `bin/task lint` is faster now

seeforschauer (approved):

Issue #256, opened 2022-01-13
Support all bitwise query operators


AlekSi:
Support all query comparison operators for all data types.

- [x] $bitsAllClear
- [x] $bitsAllSet
- [x] $bitsAnyClear
- [x] $bitsAnySet

OpenSauce:
I'll pick this up @AlekSi

AlekSi:
I propose to do everything end-to-end (implementation, tests, integration tests) for one operator (for example, `$bitsAllClear`), then do the rest.

OpenSauce:
@AlekSi Sounds like a plan my friend!

w84thesun:
I'll pick it

PR #394, opened 2022-03-24
Implement `$bitsAllClear`

w84thesun:
Refs #256.


AlekSi:
That PR handles only `$bitsAllClear`, right? Why it closes #256?

internal/handlers/common/params.go
```diff
@@ -93,3 +93,89 @@ func GetNumberParam(parameterName string, value any) (int64, error) {
 	}
 	return numberValue, nil
 }
+
+// GetBinaryMaskParam returns *types.Binary value matched and error if match failed.
+// Possible values are: position array ([1,3,5] == 010101), integer value and types.Binary value.
+func GetBinaryMaskParam(value any) (mask *types.Binary, err error) {
```
AlekSi, line 99: Let's return types.Binary by value as we do with all scalar types (even if that's a small structure).

Let's also do not use named results if we don't actually use them

internal/handlers/common/params.go
```diff
@@ -93,3 +93,95 @@ func GetNumberParam(parameterName string, value any) (int64, error) {
...
+		return types.Binary{}, NewErrorMsg(ErrBadValue,
+			fmt.Sprintf(
+				`value takes an Array, a number, or a BinData but received: $bitsAllClear: %#v`, value),
+		)
+	}
+	return mask, nil
+}
+
+// GetBinaryParam matches value type, returning types.Binary and error if match failed.
+func GetBinaryParam(value any) (types.Binary, error) {
```
AlekSi, line 159: Seems unused

internal/types/binary_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+package types
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBinaryFromArray(t *testing.T) {
```
AlekSi, line 23: The linter is correct there

internal/handlers/common/params.go
w84thesun: It's used in `filterFieldExprBitsAllClear`

AlekSi: Please not resolve the conversation if you want your answer to be seen :)

internal/types/binary_test.go
```diff
@@ -0,0 +1,64 @@
...
+		{
+			name: "bits 1,3,5",
+			args: args{values: MustNewArray(int32(1), int32(3), int32(5))},
+			want: Binary{
+				Subtype: BinaryGeneric,
+				B:       []byte{0b101010, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0},
+			},
+		},
+	}
+	for _, tt := range tests {
```
AlekSi, line 55: There is another good point by linter

w84thesun: Fixed

AlekSi (dismissed):

AlekSi:
@w84thesun why you self-assigned that PR?

AlekSi (approved):

PR #405, opened 2022-03-28
Fix panic on `findOne` query

w84thesun:
Problem: right now for queries like `db.values.findOne({"name":"there is no such name"})` that would ends up with empty result find would try to call `common.Limit` which would return slice with 1 nil document that leads to panic on document validation stage.
Solution: return if there no documents after applying filters.


AlekSi:
#409

PR #426, opened 2022-04-01
Improve `envtool` diagnostics

### LINKED ISSUE #342: Improve `envtool` diagnostics
* [ ] Envtool should check that network/ports are reachable and fail with good error messages before running tools that produce confusing errors like mongoimport.
* [ ] Envtool should log all relevant versions (Go, git, Docker, etc) and wail with an error prompting contributor to create an issue with them and all logs.

w84thesun:
Closes #342.


cmd/envtool/main.go
```diff
@@ -219,7 +198,39 @@ func setupMonilaAndValues(ctx context.Context, pgPool *pgdb.Pool) {
...
+`,
+		runtime.Version(),
+		runtime.GOOS,
+		runtime.GOARCH,
+		info.Version,
+		info.Commit,
+		info.Branch,
+		err,
+	)
+	fmt.Println(msg)
```
seeforschauer, line 221: could you please remove debug log

w84thesun: It's not a debug log. It prints all collected data. I think I should replace it with `fmt.Printf`.

seeforschauer: yep, it was just highlighted by linter that's security was not ok. thanks

seeforschauer: maybe change it to smth better (more safe)?

w84thesun: I've added `nolint` as discussed.

seeforschauer (dismissed):

cmd/envtool/main.go
```diff
@@ -92,35 +93,13 @@ func waitForPort(ctx context.Context, port uint16) error {
...
-			return nil
-		}
-
-		sleepCtx, sleepCancel := context.WithTimeout(ctx, time.Second)
-		<-sleepCtx.Done()
-		sleepCancel()
-	}
-
-	return ctx.Err()
+	return fmt.Errorf("failed to connect to port %d", port)
```
AlekSi, line 96: We should include a full address ("127.0.0.1:%d") as the common networking problem is about bad Docker network configuration

cmd/envtool/main.go
```diff
@@ -219,7 +198,39 @@ func setupMonilaAndValues(ctx context.Context, pgPool *pgdb.Pool) {
 	logger.Infof("Done in %s.", time.Since(start))
 }
 
-func main() {
+//nolint:forbidigo // Printf used to make diagnostic data easier to copy.
+func printDiagnosticData(err error) {
+	info := version.Get()
+	fmt.Printf(`Looks like something went wrong..
```
AlekSi, line 204:
```suggestion
	fmt.Printf(`Looks like something went wrong.
```

internal/util/version/version.go
```diff
@@ -73,7 +73,7 @@ func init() {
 		switch s.Key {
 		case "vcs.revision":
 			if s.Value != info.Commit {
-				panic(fmt.Sprintf("commit.txt value %q != vcs.revision value %q", info.Commit, s.Value))
+				panic(fmt.Sprintf("commit.txt value %q != vcs.revision value %q\nPlease run `bin/task gen-version`", info.Commit, s.Value))
```
AlekSi, line 76: That looks good, but I think we should add `gen-version` task as a dependency for tasks where we need it

cmd/envtool/main.go
```diff
@@ -219,7 +198,39 @@ func setupMonilaAndValues(ctx context.Context, pgPool *pgdb.Pool) {
...
+Please file an issue with all that information below:
+	Go version: %s
+	OS: %s
+	Arch: %s
+	Version: %s
+	Commit: %s
+	Branch: %s
+	Error: %v
```
AlekSi, line 212: I think we can move to a separate block – it is significantly different from other lines

AlekSi (commented):
Where is a part with logging `git verison`, `go version`, `docker version`?

cmd/envtool/main.go
w84thesun: Fixed

cmd/envtool/main.go
w84thesun: Fixed.

internal/util/version/version.go
w84thesun: I've added `gen-version` to `run` task but it's also happen when you edit and run right from editor. It confuses from time to time.

w84thesun:
> Where is a part with logging `git verison`, `go version`, `docker version`?

I've added go version in the first line but forgot about others. Current changes fixing it.

Right now diagnostic output would look like:
```
Looks like something went wrong..
Please file an issue with all that information below:   

        OS: windows
        Arch: amd64
        Version: v0.1.0-21-g4b5d84c-dirty
        Commit: 4b5d84c03e805130b8b75f04086d735d57f202bf
        Branch: issue-342-improve-envtool-diagnostics   

        Go version: go1.18
        git version 2.32.0.windows.2
        Docker Compose version v2.3.3
        git version 2.32.0.windows.2

        Error: failed to connect to 127.0.0.1:5432
```

AlekSi (approved):

PR #432, opened 2022-04-04
Issue 24 - on every new git tag, build a Docker image with tag matching git tag without leading v, and also tag it as latest

seeforschauer:
https://github.com/FerretDB/FerretDB/issues/70
https://github.com/FerretDB/github-actions/issues/24

tested on https://github.com/seeforschauer/FerretDB/pull/5


seeforschauer:
not now

Issue #371, opened 2022-03-20
Large numbers are not compared correctly


AlekSi:
float64 and int64 numbers greater than 2^53 (or less than -2^53) can't be compared by converting them to float64: https://go.dev/play/p/tYJkQoHGr4I

PR #440, opened 2022-04-04
Support `$mod` evaluation query operator

### LINKED ISSUE #406: Support `$mod` evaluation query operator
Support `$mod` evaluation query operator for all data types.

fenogentov:
Closes #406.


internal/handlers/common/filter.go
```diff
@@ -458,3 +465,62 @@ func filterFieldExprBitsAnySet(fieldValue, maskValue any) (bool, error) {
...
+
+	return true, nil
+}
+
+func toInt64(val any) (int64, error) {
+
+	switch v := val.(type) {
+	case float64:
+		return int64(v), nil
+	case float32:
```
AlekSi, line 505: Most of those types are not possible; only 3 numerics types are allowed: float64, int32, int64. See there: https://github.com/FerretDB/FerretDB/blob/main/internal/types/types.go

AlekSi (commented):
For the integration test please do something similar to https://github.com/FerretDB/FerretDB/blob/main/integration/query_array_test.go

seeforschauer (changes requested):
could you please also add some tests?

integration/query_evalution_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+func TestEvalMod(t *testing.T) {
+	//	t.Parallel()
```
AlekSi, line 13: It might be easier to run tests with `go test -parallel=1 run=TestEvalMod` than do temporary modifications in code

internal/handlers/common/filter.go
```diff
@@ -452,3 +459,47 @@ func filterFieldExprBitsAnySet(fieldValue, maskValue any) (bool, error) {
...
+
+	if field%div != rem {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+// toInt64 ...
+func toInt64(val any) (int64, error) {
```
AlekSi, line 494: I'm not sure, but you might want to use existing `GetWholeNumberParam` function

fenogentov: There is a condition in the `GetWholeNumberParam` function
```
if value != math.trunc(value) || math.IsNaN(value) || math.IsInf(value, 0) {
   return 0, errNotWholeNumber
}
```
float, for example 234.5, the function does not consider this value as a number

The `$mod` operator behaves a little differently - the float value is rounded up to 234.5 -> 234.

seeforschauer (dismissed):
t.Parallel is left to be left

internal/handlers/common/filter.go
```diff
@@ -452,3 +460,67 @@ func filterFieldExprBitsAnySet(fieldValue, maskValue any) (bool, error) {
...
+	if arr.Len() < 2 {
+		return false, NewErrorMsg(ErrBadValue, `malformed mod, not enough elements`)
+	}
+	if arr.Len() > 2 {
+		return false, NewErrorMsg(ErrBadValue, `malformed mod, too many elements`)
+	}
+
+	div, err := toInt64(must.NotFail(arr.Get(0)))
+	if err != nil {
+		switch err {
```
w84thesun, line 481: We could move this switch to `internal/handlers/common/error.go` with name like `formatFieldModError`. It would make this function a bit smaller.

AlekSi: But that function will not be shared, so why extract it?

internal/handlers/common/filter.go
w84thesun: We could call it `getModParam` then and move to `internal/handlers/common/params.go`

internal/handlers/common/filter.go
w84thesun: I get your point, we could leave it as is.
My point is that this
```go
if err != nil {
	return false, formatFieldModError(err)
}
```
is easier to read rather then
```go
if err != nil {
	switch err {
	case errUnexpectedType:
		return false, NewErrorMsg(ErrBadValue, `malformed mod, divisor not a number`)
	case errNotWholeNumber:
		return false, NewErrorMsg(ErrBadValue, `malformed mod, divisor value is invalid :: caused by :: `+
			`Unable to coerce NaN/Inf to integral type`)
	default:
		return false, err
	}
}
```

AlekSi: Sure. Because that code translates `toInt64` errors to protocol errors. If we move both functions inside, there will be no error translation code at all. The resulting code will be both shorter and easier to read because the reader will not have a need to jump to the `toInt64` definition to see what errors it returns.

AlekSi (dismissed):

integration/query_evalution_test.go
```diff
@@ -0,0 +1,166 @@
...
+		bson.D{{"_id", "float64_2"}, {"value", float64(114.99)}},
+		bson.D{{"_id", "float64_3"}, {"value", float64(115.5)}},
+		bson.D{{"_id", "int64_1"}, {"value", int64(141)}},
+		bson.D{{"_id", "int64_2"}, {"value", int64(151)}},
+		bson.D{{"_id", "int64_3"}, {"value", int64(161)}},
+		bson.D{{"_id", "int32_1"}, {"value", int32(177)}},
+		bson.D{{"_id", "int32_2"}, {"value", int32(178)}},
+		bson.D{{"_id", "int32_3"}, {"value", int32(179)}},
+		bson.D{{"_id", "nil"}, {"value", nil}},
+		bson.D{{"_id", "string"}, {"value", "12"}},
```
AlekSi, line 28: May you please add tests where the value is a weird number: ±Infinity, NaN, `math.MaxFloat64`, `math.SmallestNonzeroFloat64`, `math.MinInt64`? I suspect there might be a problem with large float64 -> int64 conversion (#371)

AlekSi (commented):
Please add tests as mentioned

fenogentov:
> Please add tests as mentioned

Yes. Soon

AlekSi (commented):

integration/query_evalution_test.go
```diff
@@ -0,0 +1,166 @@
...
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			var actual []bson.D
+			cursor, err := collection.Find(ctx, tc.q)
+			if tc.err != nil {
+				require.Nil(t, tc.expectedIDs)
+				require.Equal(t, tc.err, err)
```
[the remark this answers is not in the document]
fenogentov: ok

internal/handlers/common/filter.go
```diff
@@ -486,3 +494,66 @@ func filterFieldExprBitsAnySet(fieldValue, maskValue any) (bool, error) {
...
+	case int64:
+		remainder = r
+	case float64:
+		if math.IsNaN(r) || math.IsInf(r, 0) {
+			return false, NewErrorMsg(ErrBadValue, `malformed mod, remainder value is invalid :: caused by :: `+
+				`Unable to coerce NaN/Inf to integral type`)
+		}
+		remainder = int64(r)
+	default:
+		return false, NewErrorMsg(ErrBadValue, `malformed mod, remainder not a number`)
```
seeforschauer, line 551:
```suggestion
		return false, NewErrorMsg(ErrBadValue, `malformed mod, remainder is not a number`)
```

internal/handlers/common/filter.go
```diff
@@ -486,3 +494,66 @@ func filterFieldExprBitsAnySet(fieldValue, maskValue any) (bool, error) {
...
+	case int64:
+		divisor = d
+	case float64:
+		if math.IsNaN(d) || math.IsInf(d, 0) {
+			return false, NewErrorMsg(ErrBadValue, `malformed mod, divisor value is invalid :: caused by :: `+
+				`Unable to coerce NaN/Inf to integral type`)
+		}
+		divisor = int64(d)
+	default:
+		return false, NewErrorMsg(ErrBadValue, `malformed mod, divisor not a number`)
```
seeforschauer, line 536:
```suggestion
		return false, NewErrorMsg(ErrBadValue, `malformed mod, divisor is not a number`)
```

internal/handlers/common/filter.go
```diff
@@ -486,3 +494,66 @@ func filterFieldExprBitsAnySet(fieldValue, maskValue any) (bool, error) {
...
+	}
+
+	switch d := must.NotFail(arr.Get(0)).(type) {
+	case int32:
+		divisor = int64(d)
+	case int64:
+		divisor = d
+	case float64:
+		if math.IsNaN(d) || math.IsInf(d, 0) {
+			return false, NewErrorMsg(ErrBadValue, `malformed mod, divisor value is invalid :: caused by :: `+
```
seeforschauer, line 531: what are `::`s? I mean, do we do really need them?

seeforschauer (changes requested):
comments + would you be so kind to use `assertEqualError` in tests

internal/handlers/common/filter.go
AlekSi: All those error messages are the same as returned by MongoDB; we want to use the same for compatibility, at least for now

fenogentov: Yes. errors copied from mongodb

seeforschauer (changes requested):
would you be so kind to add weird numbers such as  ±Infinity, NaN-s  in the test data set to see how they are being processed with this new code? Thank you

fenogentov:
> 		bson.D{{"_id", "NaN"}, {"value", math.NaN}},

`-test.shuffle 1650280464648790621
2022-04-18T14:14:24.658+0300	INFO	debug	debug/debug.go:59	Starting debug server on http://127.0.0.1:43049/
--- FAIL: TestEvalutionMod (0.16s)
    query_evalution_test.go:62: 
        	Error Trace:	query_evalution_test.go:62
        	            	cannot transform type primitive.D to a BSON Document: no encoder found for func() float64
        	Test:       	TestEvalutionMod
    setup.go:116: Keeping database "testevalutionmod" and collection "testevalutionmod" for debugging.
FAIL
coverage: 7.7% of statements in ../...
exit status 1
FAIL	github.com/FerretDB/FerretDB/integration	0.779s`

AlekSi:
@fenogentov `math.NaN` is a function, it should be called: `math.NaN()`

integration/query_evalution_test.go
```diff
@@ -0,0 +1,401 @@
...
+		bson.D{{"_id", "MinInt64"}, {"value", math.MinInt64}},
+		bson.D{{"_id", "MinInt64_float"}, {"value", float64(math.MinInt64)}},
+		bson.D{{"_id", "MinInt64_minus"}, {"value", float64(math.MinInt64 - 1)}},
+		bson.D{{"_id", "MinInt64_overflowVerge"}, {"value", -9.223372036854776832e+18}},
+		bson.D{{"_id", "MinInt64_overflow"}, {"value", -9.223372036854776833e+18}},
+	})
+	require.NoError(t, err)
+
+	for name, tc := range map[string]struct {
+		q           bson.D
```
seeforschauer, line 65: would you be so kind to rename `q` to `filter`  to match other places where we use this kind of entity? Thank you.

seeforschauer (approved):

AlekSi (approved):

AlekSi:
Thank you!

PR #441, opened 2022-04-05
Integration tests improvements

### LINKED ISSUE #411: Integration tests improvements
- [x] Improve logging
- [x] Add test data for documents and arrays
- [x] Add custom `assertEqual` with NaN handling

AlekSi:
Add test data for documents and arrays.
Fix NaNs comparision.

Closes #411.

### DIFF
--- a/integration/helpers.go
+++ b/integration/helpers.go
@@ -20,7 +20,89 @@ import (
 
 	"github.com/stretchr/testify/require"
 	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/testutil"
 )
+
+// convert converts given driver value (bson.D, bson.A, etc) to FerretDB types package value.
+//
+// It then can be used with all types helpers such as testutil.AssertEqual.
+func convert(t testing.TB, v any) any {
+	t.Helper()
+
+	switch v := v.(type) {
+	// composite types
+	case primitive.D:
+		doc := types.MustNewDocument()
+		for _, e := range v {
+			doc.Set(e.Key, convert(t, e.Value))
+		}
+		return doc
+	case primitive.A:
+		arr := types.MakeArray(len(v))
+		for _, e := range v {
+			arr.Append(convert(t, e))
+		}
+		return arr
+
+	// scalar types (in the same order as in types package)
+	case float64:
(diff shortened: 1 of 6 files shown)


integration/helpers.go
```diff
@@ -20,8 +20,90 @@ import (
 
 	"github.com/stretchr/testify/require"
 	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/testutil"
 )
 
+// convert converts given driver value (bson.D, bson.A, etc) to FerretDB types package value.
```
seeforschauer, line 29: :+1:

integration/helpers.go
```diff
@@ -20,8 +20,90 @@ import (
...
+		}
+	case int32:
+		return v
+	case primitive.Timestamp:
+		return types.Timestamp(uint64(v.I)<<32 + uint64(v.T))
+	case int64:
+		return v
+	default:
+		t.Fatalf("unexpected type %T", v)
+		panic("not reached")
```
seeforschauer, line 79: test fatal should fail, isn't it? it's not fails?

integration/query_comparision_test.go
```diff
@@ -15,25 +15,43 @@
...
+			t.Parallel()
+
+			var actual []bson.D
+			cursor, err := collection.Find(ctx, bson.D{{"_id", bson.D{{"$eq", id}}}}, opts)
+			require.NoError(t, err)
+
+			err = cursor.All(ctx, &actual)
+			require.NoError(t, err)
+			require.Len(t, actual, 1)
+			assertEqualDocuments(t, expected, actual[0])
```
seeforschauer, line 53: actual[0] - the first only? then should the limit be added to the Find query options? it won't fail if there are will be more than 1 elements in actual.

integration/helpers.go
AlekSi: It does fail, but compile is not smart enough to understand that return is not needed there
```
./helpers.go:81: missing return
```

I wanted to be sure that all cases have returns, so panic acts there as impossible [terminal statement](https://go.dev/ref/spec#Terminating_statements)

integration/query_comparision_test.go
AlekSi: The previous line ensures that there is exactly one element

seeforschauer (approved):

Issue #442, opened 2022-04-05
Bitwise operators behave differently


AlekSi:
While working on #417 I noticed that MongoDB and FerretDB behave differently on the first test I wrote. See #443.

We should port tests from handler_test.go and dance, and fix the difference in behavior.

PR #443, opened 2022-04-05
Add test stub for bitwise operators

AlekSi:
Refs #442.
Refs #417.


AlekSi:
@w84thesun PTAL at this one and #442

PR #445, opened 2022-04-05
Transfer existing comparison tests

ribaraka:
Refs #417.


integration/query_comparision_test.go
```diff
@@ -15,48 +15,235 @@
...
+		"EqDoubleNaN": {
+			q:           bson.D{{"value", bson.D{{"$eq", math.NaN()}}}},
+			expectedIDs: []any{"double-nan"},
+		},
+		"EqNegativeZero": {
+			q:           bson.D{{"value", bson.D{{"$eq", math.Copysign(0, -1)}}}},
+			expectedIDs: []any{"double-negative-zero", "double-zero", "int32-zero", "int64-zero"},
+		},
+
+		"EqString": {
```
seeforschauer, line 74: empty line - another block of what?  here and below.

integration/query_comparision_test.go
```diff
@@ -15,48 +15,235 @@
...
 	"github.com/stretchr/testify/require"
 	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.mongodb.org/mongo-driver/mongo/options"
 
 	"github.com/FerretDB/FerretDB/integration/shareddata"
 )
 
-func TestQueryComparisionEq(t *testing.T) {
+func TestQueryComparisonEq(t *testing.T) {
```
seeforschauer, line 31: shall we have amount of tests to be passed also? 
then it would be clear how many tests were added / deleted. WDYT?

AlekSi: All integration tests must pass. No need to track the count. What we should track is test coverage on codecov.io

integration/query_comparision_test.go
AlekSi: The separation of types, I think.

But then the one above should be `EqDoubleNegativeZero`, with "Double" in the middle

ribaraka: Another block of data type. If it's redundunt, I will delete empty lines.

AlekSi: I don't have a strong opinion there, but maybe try to add them as separate top-level tests? One reason we want to move away from the previous handlers tests is that one huge table is hard to maintain

seeforschauer: well then empty lines might be useful when we'll think of dividing tests into groups and subgroups.

AlekSi (approved):

PR #446, opened 2022-04-05
Add `$exists` operator

w84thesun:


integration/query_element_test.go
```diff
@@ -0,0 +1,105 @@
...
+		},
+		"query-nan-field": {
+			q:           bson.D{{"nan", bson.D{{"$exists", true}}}},
+			expectedIDs: []any{"nan"},
+		},
+		"query-exists-false": {
+			q:           bson.D{{"field", bson.D{{"$exists", false}}}},
+			expectedIDs: []any{"empty-array", "nan", "null", "string"},
+		},
+		"query-non-bool": {
```
w84thesun, line 72: I made some tests and looks like mongo returns all documents when we pass non bool value. It's a strange behavior but I decided to stick with it for now. As I mentioned earlier mongo shell returns error like `exists must be a boolean`.

integration/query_element_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/mongo-driver/bson"
+)
+
+func TestExistsOperator(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setup(t)
+
+	_, err := collection.InsertMany(ctx, []any{
```
seeforschauer, line 30: When running this test multiple times it would add data to the same container and following tests will fail bcz data is duplicated and IDs are repeated. Maybe, remove data on exit?

w84thesun: How do you ran it? Did you get any errors during those runs? Removing tests data is already handled in `integration/setup.go`. It looks like there some bug or something.

seeforschauer (approved):

internal/handlers/common/filter.go
```diff
@@ -150,13 +163,18 @@ func filterOperator(doc *types.Document, operator string, filterValue any) (bool
 }
 
 // filterFieldExpr handles {field: {expr}} filter.
-func filterFieldExpr(fieldValue any, expr *types.Document) (bool, error) {
+func filterFieldExpr(fieldValue any, expr *types.Document, fieldExist bool) (bool, error) {
```
AlekSi, line 166: `expr` and `fieldExist` describe the same thing there.

I think the right fix could be replace `fieldValue` with `field` and get `fieldValue` inside the function if needed. So, for example, if our document is `{"foo": "bar"}` and our query is `{"foo": {"$exists": true}}`, `fieldValue` would be "bar", but we really need "foo", so let's pass field itself instead of its value.

internal/handlers/common/filter.go
```diff
@@ -53,26 +53,39 @@ func filterDocumentPair(doc *types.Document, filterKey string, filterValue any)
...
 	if err != nil {
-		return false, nil // no error - the field is just not present
+		fieldExist = false
 	}
 
 	switch filterValue := filterValue.(type) {
 	case *types.Document:
 		// {field: {expr}}
-		return filterFieldExpr(docValue, filterValue)
+		return filterFieldExpr(docValue, filterValue, fieldExist)
```
AlekSi, line 65: So pass `filterKey` instead of `docValue` as the first argument there and in all other invocations

internal/handlers/common/filter.go
w84thesun: I tried to do so but we actually need "bar" part almost everywhere to made comparison. So we can't just replace `fieldValue` with `filterKey`

AlekSi: Let's pass the document, field and expr then. From document and field we can always get the value for that field in that document

w84thesun: Yeah, I did it for ‘filterFieldExpr’, for the rest two cases I got the document value.

internal/handlers/common/filter.go
```diff
@@ -278,6 +290,13 @@ func filterFieldExpr(fieldValue any, expr *types.Document) (bool, error) {
 				return false, err
 			}
 
+		case "$exists":
+			// {field: {$exists: boolean}}
```
AlekSi, line 294: This comment doesn't match code – the value there is not boolean

internal/handlers/common/filter.go
```diff
@@ -452,3 +471,20 @@ func filterFieldExprBitsAnySet(fieldValue, maskValue any) (bool, error) {
 
 	return true, nil
 }
+
+// filterFieldExprExists handles {field: {$exists: value}} filter.
+func filterFieldExprExists(fieldExist bool, exprValue any) (bool, error) {
+	expr, ok := exprValue.(bool)
+	if !ok {
+		return true, nil
```
AlekSi, line 479: Without comment that looks like a typo

internal/handlers/common/filter.go
```diff
@@ -53,26 +53,32 @@ func filterDocumentPair(doc *types.Document, filterKey string, filterValue any)
...
+			// not an error, just no such field
+			return false, nil
+		}
 		return filterFieldRegex(docValue, filterValue)
 
 	default:
 		// {field: value}
+		docValue, err := doc.Get(filterKey)
+		if err != nil {
+			// not an error, just no such field
```
AlekSi, line 78: May you please check that we have tests for `{"no_such_field": null}`?

internal/handlers/common/filter.go
```diff
@@ -150,13 +156,19 @@ func filterOperator(doc *types.Document, operator string, filterValue any) (bool
...
+func filterFieldExpr(doc *types.Document, filterKey string, expr *types.Document) (bool, error) {
 	for _, exprKey := range expr.Keys() {
 		if exprKey == "$options" {
 			// handled by $regex
 			continue
 		}
 
+		fieldValue, err := doc.Get(filterKey)
+		if err != nil && exprKey != "$exists" {
+			// exit when not $exists filter and no such field
```
AlekSi, line 168: May you please check that we have tests for `{"no_such_field": {"$eq": null}}`?

internal/handlers/common/filter.go
w84thesun: That's just MongoDB behavior. 😄

AlekSi: I understand that. But there should be a comment

w84thesun: Sure, I'll add one.

integration/query_element_test.go
```diff
@@ -0,0 +1,105 @@
...
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			db := collection.Database()
+			cursor, err := db.RunCommandCursor(ctx,
+				bson.D{
+					{"find", collection.Name()},
+					{"filter", tc.q},
+				})
```
AlekSi, line 86:
```suggestion
			cursor, err := db.RunCommandCursor(ctx, bson.D{
				{"find", collection.Name()},
				{"filter", tc.q},
			})
```

Please use this indentation style everywhere

integration/query_element_test.go
```diff
@@ -0,0 +1,105 @@
...
+		bson.D{{"_id", "null"}, {"null", nil}},
+		bson.D{{"_id", "string"}, {"value", "12"}},
+		bson.D{{"_id", "two-fields"}, {"value", "12"}, {"field", 42}},
+	})
+	require.NoError(t, err)
+
+	for name, tc := range map[string]struct {
+		q           bson.D
+		expectedIDs []any
+		err         error
```
AlekSi, line 42: That field is unsed, let's remove it

integration/query_element_test.go
w84thesun: Ok, I will. Fixed

integration/query_element_test.go
w84thesun: Fixed

PR #452, opened 2022-04-06
Add support for `$gte`, `$lt`, `$lte` operators

ribaraka:
Refs #251.

### DIFF
--- a/integration/query_comparision_test.go
+++ b/integration/query_comparision_test.go
@@ -213,7 +213,7 @@ func TestQueryComparisonGt(t *testing.T) {
 		},
 
 		"GtEmptyString": {
-			q:           bson.D{{"value", bson.D{{"$gt", "boo"}}}},
+			q:           bson.D{{"value", bson.D{{"$gt", ""}}}},
 			expectedIDs: []any{"array-three", "string"},
 		},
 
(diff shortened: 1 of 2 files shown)


w84thesun (approved):

integration/query_comparision_test.go
```diff
@@ -265,13 +265,247 @@ func TestQueryComparisonGt(t *testing.T) {
...
+			q:           bson.D{{"value", bson.D{{"$lt", primitive.Timestamp{T: 41, I: 13}}}}},
+			expectedIDs: []any{"timestamp-i"},
+		},
+		"LtNull": {
+			q:           bson.D{{"value", bson.D{{"$lt", nil}}}},
+			expectedIDs: []any{},
+		},
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
```
seeforschauer, line 415: this can be extracted to a function, WDYT?

AlekSi: Yes, it can be extracted to a helper function to DRY things. Then another test might need a slightly tweaked version of that function, so we just one small addition parameter and one small `if`… And then we end up just like we did in `dance` or `handlers_test.go` that tests everything in one giant test with a lot of branching. We want to avoid that.

So here we don't want to DRY things too much. A few lines copied across several tests there are fine.

AlekSi (approved):

PR #454, opened 2022-04-07
fix integration go mod

seeforschauer:


w84thesun (approved):

AlekSi (changes requested):
Overall, I don't think this is right

PR #458, opened 2022-04-08
Add tests for `null` values

### LINKED ISSUE #456: Improve tests for `null` values
Let's say we have two documents in the collection: `{"foo": "bar"}` and `{"bar": null}`. We should check that those queries work in the same way in FerretDB and MongoDB:

- [ ] `{"foo": null}`
- [ ] `{"bar": null}`
- [ ] `{"foo": {"$eq": null}}`
- [x] `{"bar": {"$eq": null}}`

If we already have tests covering that, great. If not, we should add them and fix bugs if they exist.

w84thesun:
Closes #456.

### DIFF
--- a/integration/query_comparision_test.go
+++ b/integration/query_comparision_test.go
@@ -170,7 +170,28 @@ func TestQueryComparisonEq(t *testing.T) {
 			q:           bson.D{{"value", bson.D{{"$eq", int64(math.MinInt64)}}}},
 			expectedIDs: []any{"int64-min"},
 		},
-		"CompareNoSuchField": {
+
+		"EqNoSuchFieldNull": {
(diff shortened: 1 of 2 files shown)


integration/query_comparision_test.go
```diff
@@ -170,6 +170,29 @@ func TestQueryComparisonEq(t *testing.T) {
 			q:           bson.D{{"value", bson.D{{"$eq", int64(math.MinInt64)}}}},
 			expectedIDs: []any{"int64-min"},
 		},
+		"CompareNoSuchField": {
+			q: bson.D{{"no-such-field", nil}},
```
AlekSi, line 174: `{"no-such-field": {"$eq": nil}}` is missing.

And please rename test cases and group them by blank lines just like other cases

AlekSi (approved):

PR #459, opened 2022-04-08
Add support for $nin operator

ribaraka:


ribaraka:
In this PR only one test case but it tests all data types. I was going to test it as I have done in previous PRs but it looks kind of excessive (example below).
```
{
		"NinDouble": {
			q:           bson.D{{"value", bson.D{{"$nin", bson.A{42.11, 0.0, math.Copysign(0, -1), math.MaxFloat64, math.SmallestNonzeroFloat64, math.Inf(+1), math.Inf(-1), math.NaN(), primitive.Binary{Subtype: 0x80, Data: []byte{42, 0, 13}}, primitive.Binary{Data: []byte{}}, primitive.NewDateTimeFromTime(time.Date(2021, 11, 1, 10, 18, 42, 123000000, time.UTC)), primitive.NewDateTimeFromTime(time.Unix(0, 0)), primitive.NewDateTimeFromTime(time.Date(0, 1, 1, 0, 0, 0, 0, time.UTC)), primitive.NewDateTimeFromTime(time.Date(9999, 12, 31, 23, 59, 59, 999000000, time.UTC)), nil, primitive.Regex{Pattern: "foo", Options: "i"}, primitive.Regex{}, int32(42), int32(0), int32(math.MaxInt32), int32(math.MinInt32), primitive.Timestamp{T: 42, I: 13}, primitive.Timestamp{I: 1}, int64(42), int64(0), int64(math.MaxInt64), int64(math.MinInt64), false, true, "foo", ""}}}}},
			expectedIDs: []any{"array-empty", "document", "document-empty",  "double"},
		},
```

integration/query_comparision_test.go
```diff
@@ -545,8 +545,39 @@ func TestQueryComparisonLte(t *testing.T) {
...
+
+	var docValue bson.A
+	for _, expected := range shareddata.Scalars.Docs() {
+		docValue = append(docValue, expected.Map()["value"])
+	}
+
+	for name, tc := range map[string]struct {
+		q           bson.D
+		expectedIDs []any
+	}{
```
w84thesun, line 561: Could you please add tests for bad values like `nil` or `string` and for empty array?

ribaraka: Do you mean like another seperate test cases? 
`docValue` variable has all scalar data types such as `string`, `nil`, etc. The variable doesn't have composite data type though. So yeah, if need to write tests for composite data type, I'll do it.

w84thesun: Yeah, I mean a query like `{"$nin", nil}` or `{"$nin", "some-string"}`

ribaraka: done

w84thesun (dismissed):

seeforschauer (dismissed):

AlekSi (commented):
Let's finish #471 first, then add documents and arrays to that PR

integration/query_comparision_test.go
```diff
@@ -986,8 +986,84 @@ func TestQueryComparisonLte(t *testing.T) {
...
+		scalarDataTypesFilter = append(scalarDataTypesFilter, scalarDataType.Map()["value"])
+	}
+
+	var compositeDataTypesFilter bson.A
+	for _, compositeDataType := range shareddata.Composites.Docs() {
+		compositeDataTypesFilter = append(compositeDataTypesFilter, compositeDataType.Map()["value"])
+	}
+
+	for name, tc := range map[string]struct {
+		q           bson.D
```
seeforschauer, line 1005: could you please rename `q` to `filter`?

AlekSi: Better yet, remove the common part: `filter := bson.D{{"value", bson.D{{"$nin", tc.value}}}}`. In that case it should be `value`, not `filter` (not the full value of the second argument) or `q` (which we use more generically)

ribaraka: will do

seeforschauer (approved):

integration/query_comparision_test.go
```diff
@@ -986,8 +986,84 @@ func TestQueryComparisonLte(t *testing.T) {
...
+				Name:    "BadValue",
+				Message: `$nin needs an array`,
+			},
+		},
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			var actual []bson.D
```
seeforschauer, line 1052: tbd: actual is used 5 lines later -> move and add empty line

integration/query_comparision_test.go
ribaraka: fixed

integration/query_comparision_test.go
ribaraka: fixed

seeforschauer (approved):

AlekSi (approved):

AlekSi:
Thank you!

PR #465, opened 2022-04-11
Support logical operators

### LINKED ISSUE #252: Support all logical query operators
Support all query comparison operators for all data types. It is likely we already support many of them, but we should add integration tests (moving them from handler tests and dance), and maybe add more tests for weird cases and fix them.

- [ ] $and
- [ ] $not
- [ ] $nor
- [ ] $or

w84thesun:
Closes #252.


integration/query_logical_test.go
```diff
@@ -0,0 +1,298 @@
...
+	"go.mongodb.org/mongo-driver/bson/primitive"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+
+	"github.com/FerretDB/FerretDB/integration/shareddata"
+)
+
+func TestQueryLogicalAnd(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setupWithOpts(t, &setupOpts{providers: []shareddata.Provider{shareddata.Scalars}})
```
AlekSi, line 32: It seems that can be a basic `setup` function.

integration/query_logical_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+	for name, tc := range map[string]struct {
+		filter      bson.D
+		expectedIDs []any
+		err         mongo.CommandError
+	}{
+		"And": {
+			filter: bson.D{{
+				"$and",
```
AlekSi, line 41: `bson.D{{"$and", something}}` is shared across test cases, so we can not repeat it: https://github.com/FerretDB/FerretDB/blob/6e8a7773a9867f76e982ed9229594d6c899e8c46/integration/query_comparision_test.go#L308

integration/query_logical_test.go
```diff
@@ -0,0 +1,298 @@
...
+			t.Parallel()
+
+			cursor, err := collection.Find(ctx, tc.filter, options.Find().SetSort(bson.D{{"_id", 1}}))
+			if err != nil {
+				require.Nil(t, tc.expectedIDs)
+				assertEqualError(t, tc.err, err)
+				return
+			}
+
+			var actual []bson.D
```
AlekSi, line 292: Notice how `actual` is defined there, there is blank line before that block, etc.

integration/query_logical_test.go
```diff
@@ -0,0 +1,298 @@
...
+			cursor, err := collection.Find(ctx, tc.filter, options.Find().SetSort(bson.D{{"_id", 1}}))
+			if tc.err.Code != 0 {
+				require.Nil(t, tc.expectedIDs)
+				assertEqualError(t, tc.err, err)
+				return
+			}
+			require.NoError(t, err)
+			err = cursor.All(ctx, &actual)
```
AlekSi, line 218: Notice how `actual` is defined way above, no blank line, additional `require.NoError(t, err)`, etc

It seems like we continue to copy&paste two different styles instead of reunifiying them

integration/query_logical_test.go
seeforschauer: > `bson.D{{"$and", something}}` is shared across test cases, so we can not repeat it:
> 
> https://github.com/FerretDB/FerretDB/blob/6e8a7773a9867f76e982ed9229594d6c899e8c46/integration/query_comparision_test.go#L308

@AlekSi  sorry what did you mean?

seeforschauer (changes requested):
`actual` semi-similar code that was discussed on the call

integration/query_logical_test.go
w84thesun: Fixed

integration/query_logical_test.go
w84thesun: Fixed

integration/query_logical_test.go
w84thesun: Fixed

integration/query_logical_test.go
w84thesun: Fixed

seeforschauer (approved):

AlekSi (commented):
Tests are failing now

w84thesun:
> Tests are failing now

Fixed

seeforschauer (approved):

AlekSi (approved):

PR #466, opened 2022-04-11
Fix large numbers comparision

### LINKED ISSUE #371: Large numbers are not compared correctly

DoodgeMatvey:
Closes #371.


integration/query_comparison_big_numbers_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"go.mongodb.org/mongo-driver/mongo/options"
+	"golang.org/x/exp/constraints"
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
+
+	"github.com/FerretDB/FerretDB/integration/shareddata"
+)
+
+// Values stores shared data documents as {"_id": key, "value": value} documents.
+type values[idType constraints.Ordered] struct {
```
AlekSi, line 18: Why did we copy all that?

I think it should be enough to add a few test cases to https://github.com/FerretDB/FerretDB/blob/main/integration/query_comparision_test.go

AlekSi:
All commits are squashed on merge; there is no need to sqash them yourself and do a force push

integration/shareddata/bignumbers.go
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+
+package shareddata
+
+// BigNumbersData contains big numbers of different types for tests.
+//
+// This shared data set is not frozen yet, but please add to it only if it is really shared.
+var BigNumbersData = &Values[string]{
+	data: map[string]any{
+		"testBigInt64":   int64(2 << 61),
```
AlekSi, line 22: Let's use kebab-case just like in other sets

AlekSi: Also, no "test" prefix nessisary. I think we can use the same prefix as in scalars: "int64-big", "double-big", "int32-big"

integration/query_comparision_test.go
```diff
@@ -281,9 +281,21 @@ func TestQueryComparisonEq(t *testing.T) {
 				"objectid", "objectid-empty",
 				"regex", "regex-empty",
 				"string", "string-double", "string-empty", "string-whole",
-				"timestamp", "timestamp-i",
+				"testBigFloat64", "testBigInt32", "testBigInt64", "timestamp", "timestamp-i",
 			},
 		},
+		"EqFloat64AndInt64": {
```
AlekSi, line 287: We don't use `Eq` prefix anymore as this is a part of the test name `TestQueryComparisonEq` itself.

Please move the to the right groups above based on the parameter type. If we rename them as suggested in the other comment, they would fit nicely with other Double, Int32 and Int64 tests

internal/handlers/common/compare.go
```diff
@@ -248,8 +249,22 @@ func compareOrdered[T constraints.Ordered](a, b T) compareResult {
...
 // compareNumbers compares two numbers.
-//
-// TODO https://github.com/FerretDB/FerretDB/issues/371
 func compareNumbers(a float64, b int64) compareResult {
-	return compareOrdered(a, float64(b))
+	if math.IsNaN(a) {
+		return notEqual
+	}
+
+	bigFloat := new(big.Float).SetFloat64(a).SetPrec(100000)
```
AlekSi, line 257: Is that precision arbitrary? If yes, please add a comment something like:
```go
// TODO figure out correct precision
```

integration/shareddata/bignumbers.go
DoodgeMatvey: Thank you, fixed

integration/query_comparision_test.go
DoodgeMatvey: Fixed

internal/handlers/common/compare.go
DoodgeMatvey: Fixed

integration/shareddata/bignumbers.go
```diff
@@ -0,0 +1,26 @@
...
+
+package shareddata
+
+// BigNumbersData contains big numbers of different types for tests.
+//
+// This shared data set is not frozen yet, but please add to it only if it is really shared.
+var BigNumbersData = &Values[string]{
+	data: map[string]any{
+		"int64-big":   int64(2 << 61),
+		"float64-big": float64(2 << 60),
```
AlekSi, line 23: BSON spec calls float64 "double", so we use the same name.

integration/query_comparision_test.go
```diff
@@ -260,7 +260,18 @@ func TestQueryComparisonEq(t *testing.T) {
...
+			expectedIDs: []any{"int64-big"},
+		},
+		"Int64AndFloat64": {
+			filter:      bson.D{{"value", bson.D{{"$eq", int64(2 << 60)}}}},
+			expectedIDs: []any{"float64-big"},
+		},
+		"Float32AndInt32": {
+			filter:      bson.D{{"value", bson.D{{"$eq", float32(2 << 25)}}}},
+			expectedIDs: []any{"int32-big"},
+		},
```
AlekSi, line 274: Notice how before that change tests cases were grouped by argument type and separated by blank lines. We should keep that organization.

Rename test cases based on argument type and move them to existing "Double...", "Int32..." and "Int64..." groups.

integration/query_comparision_test.go
AlekSi: The second part wasn't done. Described it again in the other comment

integration/shareddata/bignumbers.go
DoodgeMatvey: fixed

integration/query_comparision_test.go
DoodgeMatvey: I agree, fixed

seeforschauer (approved):

seeforschauer (changes requested):
I see failing tests are not directly connected with the PR, 
however need to fix - either increase limit for dbstats or check what is wrong.

seeforschauer (approved):

internal/handlers/common/compare.go
```diff
@@ -248,8 +249,23 @@ func compareOrdered[T constraints.Ordered](a, b T) compareResult {
 }
 
 // compareNumbers compares two numbers.
-//
-// TODO https://github.com/FerretDB/FerretDB/issues/371
 func compareNumbers(a float64, b int64) compareResult {
-	return compareOrdered(a, float64(b))
+	if math.IsNaN(a) {
```
AlekSi, line 253: I replaced that new code with the old code (just `return compareOrdered(a, float64(b))`) and all tests still pass. We should write tests that fail with the old code to ensure that it is fixed

AlekSi (changes requested):

internal/handlers/common/compare.go
DoodgeMatvey: I've pulled the changes from the master branch and all tests pass, do I need to add additional tests?

seeforschauer: > I've pulled the changes from the master branch and all tests pass, do I need to add additional tests?

Hope my comment would help you:

@DoodgeMatvey please see the issue https://github.com/FerretDB/FerretDB/issues/371 it describes the bug and it has also a playground link https://go.dev/play/p/tYJkQoHGr4I so it might be the test which is necessary to add, 

so on the main branch it fails and in this PR it is not and therefore it's obvious that the PR fixed bug.

seeforschauer (changes requested):
Adding personal `.idea` settings into the repository isn't ok since many of us may have different ones.

internal/handlers/common/compare.go
DoodgeMatvey: Thanks I got it. I added 1 additional case for `TestQueryComparisonEq`, `TestQueryComparisonGt`, `TestQueryComparisonLt`. I tested it and on the main branch it fails. but in this PR it is not

.gitignore
```diff
@@ -8,6 +8,7 @@ cover.html
 integration.txt
 old.txt
 new.txt
+.idea
```
w84thesun, line 11: Please check the comment at the top of this file.

DoodgeMatvey: fixed, I missed it, thank you

internal/handlers/common/compare.go
```diff
@@ -252,10 +253,25 @@ func compareOrdered[T constraints.Ordered](a, b T) compareResult {
...
+	switch bigFloat.Cmp(bigFloatFromInt) {
+	case -1:
+		return less
+	case 0:
+		return equal
+	case 1:
+		return greater
+	}
+
+	return notEqual
```
AlekSi, line 274: Let's add `default: panic("not reached")` to the `switch` above instead as this line should not be reachable

integration/query_comparision_test.go
```diff
@@ -285,6 +293,10 @@ func TestQueryComparisonEq(t *testing.T) {
 			filter:      bson.D{{"value", bson.D{{"$eq", int32(math.MinInt32)}}}},
 			expectedIDs: []any{"int32-min"},
 		},
+		"Int32DoubleBig": {
+			filter:      bson.D{{"value", bson.D{{"$eq", float32(2 << 25)}}}},
```
AlekSi, line 297: `float32` is not a valid BSON type; driver converts it under then hood, so there is no need to test it

integration/query_comparision_test.go
```diff
@@ -189,6 +189,14 @@ func TestQueryComparisonEq(t *testing.T) {
 			filter:      bson.D{{"value", bson.D{{"$eq", math.NaN()}}}},
 			expectedIDs: []any{"double-nan"},
 		},
+		"DoubleBigInt64": {
```
AlekSi, line 192: Tests are grouped by argument type, not results. If the argument is `int64`, it should not be in this block

w84thesun (dismissed):

ribaraka (dismissed):

integration/query_comparision_test.go
seeforschauer: added this to contributng guidelines.

internal/handlers/common/compare.go
DoodgeMatvey: Added

integration/query_comparision_test.go
DoodgeMatvey: Removed

integration/query_comparision_test.go
DoodgeMatvey: Fixed

integration/query_comparision_test.go
```diff
@@ -189,6 +189,10 @@ func TestQueryComparisonEq(t *testing.T) {
...
+			filter:      bson.D{{"value", bson.D{{"$eq", float64(2 << 61)}}}},
```
AlekSi, line 193: We should also test with `float64(2 << 61 +1)` to check that we don't get anything, just like we do with Int64DoubleBigPlusOne below

integration/query_comparision_test.go
```diff
@@ -311,6 +315,14 @@ func TestQueryComparisonEq(t *testing.T) {
 			filter:      bson.D{{"value", bson.D{{"$eq", int64(math.MinInt64)}}}},
 			expectedIDs: []any{"int64-min"},
 		},
+		"Int64DoubleBig": {
+			filter:      bson.D{{"value", bson.D{{"$eq", int64(2 << 60)}}}},
+			expectedIDs: []any{"double-big"},
+		},
+		"Int64PlusOneDoubleBig": {
```
AlekSi, line 322:
```suggestion
		"Int64DoubleBigPlusOne": {
```

integration/shareddata/bignumbers.go
```diff
@@ -0,0 +1,26 @@
...
+package shareddata
+
+// BigNumbersData contains big numbers of different types for tests.
+//
+// This shared data set is not frozen yet, but please add to it only if it is really shared.
+var BigNumbersData = &Values[string]{
+	data: map[string]any{
+		"int64-big":  int64(2 << 61),
+		"double-big": float64(2 << 60),
+		"int32-big":  int32(2 << 25),
```
AlekSi, line 24: All int32 values can be represented by float64 values exactly, so that do not anyhting, I think? We probably can remove it

integration/query_comparision_test.go
DoodgeMatvey: I already answered on this question to @ribaraka. 
"It doesn't make sense because a float of this size rounds up a lot and there is no difference between float64(2 << 61) and float64(2 << 61 + 1) and even with adding 100."

AlekSi: Sure, and that's exactly what we want to try with both MongoDB and FerretDB to ensure that they behave exactly the same. MongoDB does many things that do not make sense 😆

DoodgeMatvey: Ok, fixed

integration/query_comparision_test.go
DoodgeMatvey: fixed

integration/shareddata/bignumbers.go
DoodgeMatvey: Removed

AlekSi (approved):

AlekSi:
Thank you!

PR #471, opened 2022-04-13
Composite data type find handling

ribaraka:


integration/shareddata/composites.go
```diff
@@ -21,11 +21,13 @@ import "go.mongodb.org/mongo-driver/bson"
 // This shared data set is not frozen yet, but please add to it only if it is really shared.
 var Composites = &Values[string]{
 	data: map[string]any{
-		"document":       bson.D{{"foo", int32(42)}},
-		"document-empty": bson.D{},
+		"document":            bson.D{{"foo", int32(42)}},
+		"document-resembling": bson.D{{"foo", int32(42)}, {"42", "foo"}, {"array", bson.A{"foo", int32(42)}}},
```
seeforschauer, line 25: what means resembling? what to resemble? can we please specify more details?

seeforschauer (changes requested):
check dirty and some tests you've added are not ok, can you please fix it?

integration/shareddata/composites.go
ribaraka: "document-resembling" is resembling the existed array and document above. I've created this for checking for equivalent the exact document or for exact array. But I just noticed that this document doesn't exactly match the existing array. I'll redo it. Do you have any suggestions or something to add or maybe it should to rename the document?

seeforschauer: I would suggest `composite` word.

ribaraka: `document-composite`?

seeforschauer: yep

ribaraka:
> check dirty and some tests you've added are not ok, can you please fix it?

Could you please rerun tests? My tests should pass successfully.

seeforschauer:
confirm, tests are ok
dirty is not ok.

ribaraka:
Excuse me, but what is dirty?

AlekSi:
https://github.com/FerretDB/FerretDB/runs/6009434112?check_suite_focus=true fails. That steps runs `git diff --exit-code` that exits with non-zero code if there are local changes. And those local changes are present because code was re-formatted on CI.

TL;DR Run `bin/task fmt`

integration/query_comparision_test.go
```diff
@@ -39,22 +39,37 @@ func TestQueryComparisonImplicit(t *testing.T) {
 		filter      bson.D
 		expectedIDs []any
 	}{
+		"Document": {
+			filter:      bson.D{{"value", bson.D{{"foo", int32(42)}, {"42", "foo"}, {"array", bson.A{int32(42), "foo", nil}}}}},
```
AlekSi, line 43: Let's also add a test for comparision with document with the same keys and values, but in different order. The same in other tests

integration/query_comparision_test.go
```diff
@@ -39,22 +39,37 @@ func TestQueryComparisonImplicit(t *testing.T) {
...
+			expectedIDs: []any{"document-composite"},
+		},
+
+		"Array": {
+			filter:      bson.D{{"value", bson.A{int32(42), "foo", nil}}},
```
AlekSi, line 48: Same there. The same in other tests

integration/query_comparision_test.go
```diff
@@ -39,22 +39,37 @@ func TestQueryComparisonImplicit(t *testing.T) {
...
+			expectedIDs: []any{"array-three"},
+		},
+
```
AlekSi, line 51: No empty line there; we group by argument type

internal/handlers/common/compare.go
```diff
@@ -188,15 +189,24 @@ func compare(docValue, filter any) compareResult {
 
 	switch docValue := docValue.(type) {
 	case *types.Document:
+		filter, isFilterDocument := filter.(*types.Document)
```
AlekSi, line 192:
```suggestion
		filter, ok := filter.(*types.Document)
```

"comma ok" is a Go idiom

internal/handlers/common/compare.go
```diff
@@ -188,15 +189,24 @@ func compare(docValue, filter any) compareResult {
...
+		if isFilterDocument {
+			return compareDocuments(filter, docValue)
+		}
 		return notEqual
 
 	case *types.Array:
+		switch filter.(type) {
```
AlekSi, line 199:
```suggestion
		switch filter := filter.(type) {
```

Then `filter` can be used without type assertion below.

But then why we use different type checking style there and for document above? Please use the same, which is more common in that file

seeforschauer (approved):
:+1:

internal/handlers/common/compare.go
```diff
@@ -188,15 +189,24 @@ func compare(docValue, filter any) compareResult {
...
+			return compareArray(filter.(*types.Array), docValue)
+		}
+
 		for i := 0; i < docValue.Len(); i++ {
 			arrValue := must.NotFail(docValue.Get(i))
 
 			switch arrValue.(type) {
 			case *types.Document, *types.Array:
-				return notEqual
+				continue
```
AlekSi, line 209: That looks very suspicious. `compare` checks if values are equal, not equal, less or greater. It does not checks conditions like "a is a subset of b". I'm not sure why we want to skip some values there.

internal/handlers/common/compare.go
```diff
@@ -253,3 +263,30 @@ func compareOrdered[T constraints.Ordered](a, b T) compareResult {
 func compareNumbers(a float64, b int64) compareResult {
 	return compareOrdered(a, float64(b))
 }
+
+// compareDocuments converts two documents to fjson and compares them.
```
AlekSi, line 267: The comparision of two values should compare them directly, not their serialized forms.
Let's add TODO itdem to improve that in the future.

internal/handlers/common/compare.go
```diff
@@ -253,3 +263,30 @@ func compareOrdered[T constraints.Ordered](a, b T) compareResult {
...
+	if string(must.NotFail(fjson.Marshal(a))) == string(must.NotFail(fjson.Marshal(b))) {
+		return equal
+	}
+
+	return notEqual
+}
+
+// compareArray compares two arrays and returns document where array equals exactly the specified array
+// or array contains an element that equals the array.
+func compareArray(filterArr, docArr *types.Array) compareResult {
```
AlekSi, line 278: That's a good description but a bad function name.

Let's make `compareArrays` that compares that arrays are exactly the same.

Let's then make `matchArrays` that does the "smart" thing

integration/query_comparision_test.go
ribaraka: Will do.

internal/handlers/common/compare.go
ribaraka: This func actually is going to do the smart thing when it process cases with $gt, $lt, etc. It's just has not been implemented yet. I'm gonna add more logic to this func when I finish with simple $eq and implicit filter. Or I'm adding handler for $gt operator to this PR. WDYT?

AlekSi: I think all `compareXXX` functions should not have any smarts. We almost could call them `equalXXX` except that they also check if a is less or greater than b. We need a separate function when we need to compare that documents and arrays are strictly equal. For example, we might end up deleting our custom functions that do that in tests: https://github.com/FerretDB/FerretDB/blob/main/internal/util/testutil/equal.go

MongoDB documentation calls those smarts "matching", so that could be a good prefix for functions with smarts.

AlekSi: Another options is to move `compare` function for arrays and documents to types.Array and types.Document methods, but that's not possible for scalar types like int64.

Anyway, let's do all that later. Right now, we should separate simple comparision and smart comparision functions

internal/handlers/common/compare.go
AlekSi: I guess we should apply the same logic as to array then with `compare` and `match` functions

integration/query_comparision_test.go
ribaraka: Done

integration/query_comparision_test.go
ribaraka: Added test with different order of values

integration/query_comparision_test.go
ribaraka: fixed

internal/handlers/common/compare.go
ribaraka: Noted. The lines have been removed. I will handle this case in another PR.

internal/handlers/common/compare.go
ribaraka: Noted and fixed

internal/handlers/common/compare.go
ribaraka: I have processed this case, pls have a look.

internal/handlers/common/compare.go
```diff
@@ -253,3 +257,45 @@ func compareOrdered[T constraints.Ordered](a, b T) compareResult {
...
+	}
+	if b == nil {
+		log.Panicf("%v is nil", b)
+	}
+
+	keys := a.Keys()
+	if !slices.Equal(keys, b.Keys()) {
+		return false
+	}
+	return reflect.DeepEqual(a.Map(), b.Map())
```
seeforschauer, line 274: no, reflect is less effective, can we use here our `deepEqual`, some code is already written for it.

ribaraka: I can't find our `deepEqual`, could you please guide me?

seeforschauer: it needs to be written

ribaraka: got it. I am going to come up with something :)

AlekSi: It is less about effectiveness and more about correctness. For example, reflect.DeepEqual doesn’t work like we need it to work for our beloved NaNs.

There is a helper in testutil. Please copy and adopt it; I will refactor those copies later

AlekSi: (you can’t use it as-is because testutil depends on testing package, and we don’t want to import in non-test code)

ribaraka: OK! Thanks

seeforschauer (changes requested):
`deepEqual`  instead of reflect

seeforschauer (approved):
let's merge it and after make a separate PR with `deepEquals`

AlekSi:
#481

AlekSi (approved):

AlekSi:
I will fix tests in a separate PR ASAP

PR #483, opened 2022-04-18
Support querying with dot notation for documents

### LINKED ISSUE #437: Support querying with dot notation for documents
Support querying nested fields using dot notation (including query operators in values).

`db.inventory.find( { "size.h": { $lt: 15 } } )`

Arrays are not in scope.

GinGin3203:
Closes #437.

### DIFF
--- a/internal/handlers/common/filter.go
+++ b/internal/handlers/common/filter.go
@@ -666,6 +666,11 @@ func filterFieldValueByTypeCode(fieldValue any, code typeCode) (bool, error) {
 				panic(err)
 			}
 
+			// Skip embedded arrays.
+			if _, ok := value.(*types.Array); ok {
+				continue
+			}
+
 			res, err := filterFieldValueByTypeCode(value, code)
 			if err != nil {
 				return false, err
--- a/internal/handlers/common/filter.go
+++ b/internal/handlers/common/filter.go
@@ -50,6 +50,21 @@ func FilterDocument(doc, filter *types.Document) (bool, error) {
 
 // filterDocumentPair handles a single filter element key/value pair {filterKey: filterValue}.
 func filterDocumentPair(doc *types.Document, filterKey string, filterValue any) (bool, error) {
+	if strings.ContainsRune(filterKey, '.') {
+		// {field1./.../.fieldN: filterValue}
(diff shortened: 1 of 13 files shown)


internal/handlers/handler_test.go
```diff
@@ -122,6 +122,26 @@ func TestFind(t *testing.T) {
 	// Do not use sentences, spaces, or underscores in subtest names
 	// to make it easier to run individual tests with `go test -run test/name` and for consistency.
 	testCases := map[string]testCase{
+		"DotNotationQuery": {
```
AlekSi, line 125: We started moving away from those tests to integration tests that test both FerretDB and MongoDB. Please add a new test there instead: https://github.com/FerretDB/FerretDB/blob/c4d88cfd9d5c03ffc48026f59116aec5be2a6c84/integration/query_comparision_test.go#L42-L49

internal/handlers/common/filter.go
```diff
@@ -50,6 +50,21 @@ func FilterDocument(doc, filter *types.Document) (bool, error) {
 
 // filterDocumentPair handles a single filter element key/value pair {filterKey: filterValue}.
 func filterDocumentPair(doc *types.Document, filterKey string, filterValue any) (bool, error) {
```
AlekSi, line 52: If I understand correctly, dot notation can be used at the deeper level too. For example, `{"$or": [{"value.foo": 42}, {"value.42": "foo"}]}` should be a valid query. But I did not check it :) If it works, then checking for dot should not be done there

AlekSi (commented):
You can run linters locally with `bin/task lint`

seeforschauer (changes requested):
above +
could you please also make dot notation recursive?

internal/handlers/common/filter.go
GinGin3203: You are right, the example you provided is a valid query and it works for both Mongo and the proposed implementation -- I've just checked both

AlekSi: I spent a significant amount of time trying to come up with an example that does not work – and could not. :) Great job!

AlekSi (approved):

seeforschauer (approved):

PR #488, opened 2022-04-18
Fix bitwise operators

### LINKED ISSUE #442: Bitwise operators behave differently

w84thesun:
Closes #442.


integration/query_bitwise_test.go
```diff
@@ -15,41 +15,517 @@
...
+		})
+	}
+}
+
+func TestQueryBitwiseAnySet(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setup(t, shareddata.Scalars)
+
+	for name, tc := range map[string]struct {
+		filter      any
```
seeforschauer, line 426: `filter` is a `value`, isn't it?

integration/query_bitwise_test.go
```diff
@@ -15,41 +15,517 @@
...
+		})
+	}
+}
+
+func TestQueryBitwiseAnyClear(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setup(t, shareddata.Scalars)
+
+	for name, tc := range map[string]struct {
+		filter      any
```
seeforschauer, line 302: `filter` is a `value`, also, WDYT?

internal/handlers/common/filter.go
```diff
@@ -594,76 +594,231 @@ func filterFieldExprSize(fieldValue any, sizeValue any) (bool, error) {
...
+		if err != nil {
+			return false, formatBitwiseOperatorErr(err, "$bitsAllClear", maskValue)
+		}
+		return (^uint64(value) & bitmask) == bitmask, nil
 
-	for i := 0; i < len(fieldBinary.B); i++ {
-		if (fieldBinary.B[i] & maskBinary.B[i]) != 0 {
+	case float64:
+		// TODO check float negative zero
+		if value != math.Trunc(value) || math.IsNaN(value) || math.IsInf(value, 0) || value > math.MaxInt64 {
```
seeforschauer, line 614: can it be in several strings? 
why not to check negative zero now?

seeforschauer (changes requested):
Some naming and linters, others seems like ok. Thank you.

internal/handlers/common/filter.go
w84thesun: Why do you want to split it?
Negative zero seems to be another tough thing. It's not so easy to check for that as far as I get.

integration/query_bitwise_test.go
w84thesun: I think we should stick with the same naming across tests.

internal/handlers/common/filter.go
seeforschauer: > Why do you want to split it? 

different conditions on separate lines.

> Negative zero seems to be another tough thing. It's not so easy to check for that as far as I get.

ok

integration/query_bitwise_test.go
AlekSi: Yes, we should. But that's not a filter, that's only a value for it. Compare and contrast:

Whole filter: https://github.com/FerretDB/FerretDB/blob/373e79c3b929304df30b7d9e18a578d003d60843/integration/query_comparision_test.go#L157

Value for filter: https://github.com/FerretDB/FerretDB/blob/373e79c3b929304df30b7d9e18a578d003d60843/integration/query_comparision_test.go#L383

integration/query_bitwise_test.go
```diff
@@ -15,41 +15,517 @@
...
-				"binary-empty", "double-negative-zero", "double-zero",
-				"int32-min", "int32-zero", "int64-min", "int64-zero",
+				"binary",
+				"double-whole",
+				"int32", "int32-max",
+				"int64", "int64-max",
+			},
+		},
+		"FloatNegativeValue": {
+			filter: int32(-1),
```
AlekSi, line 448: That's not float. Also, there is no "float" at all; BSON and we call it "double"

w84thesun: Fixed

integration/query_bitwise_test.go
w84thesun: Fixed

integration/query_bitwise_test.go
```diff
@@ -15,41 +15,528 @@
...
+		"Int32NegativeValue": {
+			value: int32(-1),
+			err: mongo.CommandError{
+				Code:    9,
+				Name:    "FailedToParse",
+				Message: "Expected a positive number in: $bitsAllClear: -1",
+			},
+		},
+
+		"Int64": {
```
seeforschauer, line 130:
```suggestion
		"MaxInt64": {
```

integration/query_bitwise_test.go
```diff
@@ -15,41 +15,528 @@
...
+		"Int32NegativeValue": {
+			value: int32(-1),
+			err: mongo.CommandError{
+				Code:    9,
+				Name:    "FailedToParse",
+				Message: "Expected a positive number in: $bitsAllSet: -1",
+			},
+		},
+
+		"Int64": {
```
seeforschauer, line 247:
```suggestion
		"MaxInt64": {
```

integration/query_bitwise_test.go
```diff
@@ -15,41 +15,528 @@
...
+		"Int32NegativeValue": {
+			value: int32(-1),
+			err: mongo.CommandError{
+				Code:    9,
+				Name:    "FailedToParse",
+				Message: "Expected a positive number in: $bitsAnyClear: -1",
+			},
+		},
+
+		"Int64": {
```
seeforschauer, line 364:
```suggestion
		"MaxInt64": {
```

integration/query_bitwise_test.go
```diff
@@ -15,41 +15,528 @@
...
+		"Int32NegativeValue": {
+			value: int32(-1),
+			err: mongo.CommandError{
+				Code:    9,
+				Name:    "FailedToParse",
+				Message: "Expected a positive number in: $bitsAnySet: -1",
+			},
+		},
+
+		"Int64": {
```
seeforschauer, line 491:
```suggestion
		"MaxInt64": {
```

internal/handlers/common/filter.go
seeforschauer, line 614:
```suggestion
		if value != math.Trunc(value) || 
		   math.IsNaN(value) || 
		   math.IsInf(value, 0) || 
		   value > math.MaxInt64 {
```

internal/handlers/common/filter.go
```diff
@@ -594,76 +594,231 @@ func filterFieldExprSize(fieldValue any, sizeValue any) (bool, error) {
...
+		if err != nil {
+			return false, formatBitwiseOperatorErr(err, "$bitsAllSet", maskValue)
+		}
+		return (uint64(value) & bitmask) == bitmask, nil
 
-	for i := 0; i < len(fieldBinary.B); i++ {
-		if (fieldBinary.B[i] & maskBinary.B[i]) != maskBinary.B[i] {
+	case float64:
+		// TODO check float negative zero
+		if value != math.Trunc(value) || math.IsNaN(value) || math.IsInf(value, 0) || value > math.MaxInt64 {
```
seeforschauer, line 671:
```suggestion
		if value != math.Trunc(value) || 
		math.IsNaN(value) || 
		math.IsInf(value, 0) || 
		value > math.MaxInt64 {
```

internal/handlers/common/filter.go
seeforschauer: When editing if condition: adding or deleting, it will clearly and explicitly visible what was added, deleted or what condition was edited. 

Having single row for all condition it would be more difficult to catch the change in a row.

w84thesun:
Tests for BinData would be in #508

seeforschauer (approved):

integration/query_bitwise_test.go
```diff
@@ -15,41 +15,503 @@
...
+		"Int32NegativeValue": {
+			value: int32(-1),
+			err: mongo.CommandError{
+				Code:    9,
+				Name:    "FailedToParse",
+				Message: "Expected a positive number in: $bitsAllClear: -1",
+			},
+		},
+
+		"MaxInt64": {
```
AlekSi, line 97: Int64 should be a prefix

integration/query_bitwise_test.go
```diff
@@ -15,41 +15,503 @@
...
+		"Int64NegativeValue": {
+			value: int64(-1),
+			err: mongo.CommandError{
+				Code:    9,
+				Name:    "FailedToParse",
+				Message: "Expected a positive number in: $bitsAllClear: -1",
+			},
+		},
+
+		"Array": {
```
AlekSi, line 114: Composite types should be before scalar types: https://pkg.go.dev/github.com/FerretDB/FerretDB@v0.1.1/internal/types#hdr-Mapping

internal/handlers/common/params.go
```diff
@@ -99,104 +100,76 @@ func GetWholeNumberParam(value any) (int64, error) {
...
+			if byteAt == 0 {
+				continue
+			}
+
+			if b < 8 {
+				bitmask |= uint64(byteAt) << b
+			} else {
+				bitmask |= 1 << 63
+			}
+		}
```
AlekSi, line 166: Let's remove that code if it is not tested (and, I guess, does not work the same way as MongoDB)

w84thesun: I've added tests for that code.

integration/query_bitwise_test.go
w84thesun: Fixed

integration/query_bitwise_test.go
w84thesun: Fixed

seeforschauer (approved):

internal/handlers/common/filter.go
```diff
@@ -594,76 +594,170 @@ func filterFieldExprSize(fieldValue any, sizeValue any) (bool, error) {
...
-	for i := 0; i < len(fieldBinary.B); i++ {
-		if (fieldBinary.B[i] & maskBinary.B[i]) != 0 {
+	case int64:
+		bitmask, err := getBinaryMaskParam(maskValue)
+		if err != nil {
+			return false, formatBitwiseOperatorErr(err, "$bitsAllClear", maskValue)
+		}
+		return (^uint64(value) & bitmask) == bitmask, nil
+
+	case float64:
```
AlekSi, line 612: Types order

internal/handlers/common/filter.go
```diff
@@ -594,76 +594,170 @@ func filterFieldExprSize(fieldValue any, sizeValue any) (bool, error) {
...
-	for i := 0; i < len(fieldBinary.B); i++ {
-		if (fieldBinary.B[i] & maskBinary.B[i]) != maskBinary.B[i] {
+	case int64:
+		bitmask, err := getBinaryMaskParam(maskValue)
+		if err != nil {
+			return false, formatBitwiseOperatorErr(err, "$bitsAllSet", maskValue)
+		}
+		return (uint64(value) & bitmask) == bitmask, nil
+
+	case float64:
```
AlekSi, line 654: Types order

AlekSi (approved):

PR #500, opened 2022-04-20
Mention force pushes

AlekSi:


.github/PROCESS.md
```diff
@@ -12,6 +12,6 @@ We will do all that for you so you can focus on what is really interesting for y
 2. Pull request **description** typically just mentions the issue number like "Refs #123." or "Closes #456." (do not use `https://github.com/org/repo/issue/123` even if it is rendered the same on GitHub).
    End sentences with a dot.
 3. Commits messages (both titles and bodies) are not important and can be "WIP" or anything else.
-   All commits are squashed on merge.
-   Autogenerated squash commit's title and body should be manually replaced by PR's title and description; that includes both manual merges and auto-merges.
+   All commits are always squashed on merge by GitHub, so there is no need to squash them manually and/or do force pushes.
```
DoodgeMatvey, line 15: So should I only do a new commit when I make some fixes?

DoodgeMatvey (dismissed):

ribaraka:
Maybe it's worth mentioning that you don't need to use `amend` for most of the fixed code, so as not to lose the essence of the correction?

AlekSi:
Ah, mentioning --amend is a good idea

seeforschauer (approved):

PR #506, opened 2022-04-21
Correctly override `FUZZTIME` on CI

AlekSi:


seeforschauer (approved):

.github/workflows/go.yml
```diff
@@ -130,12 +130,10 @@ jobs:
       # TODO https://github.com/FerretDB/FerretDB/issues/238
       # TODO https://github.com/FerretDB/FerretDB/issues/344
       - name: Run collected fuzzing corpus
-        run: bin/task fuzz
+        run: bin/task fuzz FUZZTIME=1s
```
seeforschauer, line 133: why 1s?

AlekSi: The plan was to run inputs from the collected corpus of interesting inputs, not search for new interesting inputs on CI for each PR. 1s is a hack to do that.

But that was broken due to way task variables and their defaults work, and confusing documentation: `env FUZZTIME=1s task fuzz` uses the default from the Taskfile.yml, but `task fuzz FUZZTIME=1s` actually uses 1s.

AlekSi: All that to do CI for PRs faster

seeforschauer: thank you.

PR #509, opened 2022-04-22
Improve `hostInfo` command's `os` response

### LINKED ISSUE #447: Improve `hostInfo` command's `os` response
We could improve our handling of `hostInfo` command by adding `os.name` and `os.version` fields as best effort. That's ok to do that only on Linux.

The simplest way to do that seems to be to parse `/etc/os-release` file. More info: https://www.freedesktop.org/software/systemd/man/os-release.html

DoodgeMatvey:
Closes #447.


seeforschauer (changes requested):
@DoodgeMatvey tests are failing, please update `go mod`

internal/handlers/pg/msg_hostinfo.go
```diff
@@ -22,6 +22,8 @@ import (
 	"strings"
 	"time"
 
+	"gopkg.in/ini.v1"
```
AlekSi, line 25: Let's not use an additionaly dependency for something simple as this task

DoodgeMatvey: Fixed

seeforschauer (changes requested):
can you please add some tests? 
also, the parsing host info func itself can be moved to common functions, to be reusable by other handlers which may appear in the future.

AlekSi:
I'm all for adding tests with `/etc/os-release` file content from different Linux distribution, but I'm not sure what common function could be extracted there. We could always extract common code when it is, well, common.

AlekSi (changes requested):
See the comment above

integration/parse_test.go
```diff
@@ -0,0 +1,101 @@
...
+ID_LIKE=debian
+PRETTY_NAME=Ubuntu 18.04 LTS
+VERSION_ID=18.04
+HOME_URL=https://www.ubuntu.com/
+SUPPORT_URL=https://help.ubuntu.com/
+BUG_REPORT_URL=https://bugs.launchpad.net/ubuntu/
+PRIVACY_POLICY_URL=https://www.ubuntu.com/legal/terms-and-policies/privacy-policy
+VERSION_CODENAME=bionic
+UBUNTU_CODENAME=bionic
+`,
```
AlekSi, line 55: Can we get one more from Red Hat / RPM-based distro?

internal/handlers/common/parse.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"bufio"
+	"io"
+	"strings"
+
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+)
+
+// ParseOSRelease parses the etc/os-release file.
+func ParseOSRelease(reader io.Reader) (map[string]string, error) {
```
AlekSi, line 26: I think that function could return just name and version, so caller should not have to know what valid keys are

AlekSi: Also, let's move it to msg_hostinfo.go and unexport

integration/parse_test.go
```diff
@@ -0,0 +1,101 @@
...
+			"PRIVACY_POLICY_URL": "https://www.ubuntu.com/legal/terms-and-policies/privacy-policy",
+			"VERSION_CODENAME":   "bionic",
+			"UBUNTU_CODENAME":    "bionic",
+		},
+	}
+
+	for key, testCase := range testCases {
+		osReleaseReader := bytes.NewReader([]byte(osReleaseFilesData[key]))
+
+		osRelease, err := common.ParseOSRelease(osReleaseReader)
```
AlekSi, line 92: That's not an integartion test, that's [unit test](https://github.com/FerretDB/FerretDB/blob/main/CONTRIBUTING.md#code-overview), it should be in `internal/`

We should also update this test to check that name and version are set on Linux: https://github.com/FerretDB/FerretDB/blob/3b082cd08832682aee502f7b972742cbc35c71fb/integration/commands_diagnostic_test.go#L44

AlekSi (changes requested):

internal/handlers/common/parse.go
```diff
@@ -0,0 +1,42 @@
...
+	scanner := bufio.NewScanner(reader)
+
+	configParams := map[string]string{}
+
+	for scanner.Scan() {
+		str := strings.Split(scanner.Text(), "=")
+
+		configParams[str[0]] = str[1]
```
seeforschauer, line 34: too many line breaks here and it seems it's made not to logically separate, WDYT of changing this?

DoodgeMatvey: I didn't catch what you mean. Do you mean indent or what?

integration/parse_test.go
DoodgeMatvey: Yes, added

internal/handlers/common/parse.go
DoodgeMatvey: Now `ParseOSRelease` return just name and version and moved to `msg_hostinfo.go` but in order to be able to use this function in tests, I left it exported

integration/parse_test.go
DoodgeMatvey: Fixed

internal/handlers/parse_osrelease_test.go
```diff
@@ -0,0 +1,101 @@
...
+		},
+	}
+
+	for key, testCase := range testCases {
+		osReleaseReader := bytes.NewReader([]byte(osReleaseFilesData[key]))
+
+		osName, osVersion, err := pg.ParseOSRelease(osReleaseReader)
+		require.NoError(t, err)
+
+		assert.Equal(t, testCase["NAME"], osName, "expected: %s\nactual  : %s", testCase["NAME"], osName)
```
AlekSi, line 98:
```suggestion
		assert.Equal(t, testCase["NAME"], osName)
```

The standard message should be good enogh, no?

internal/handlers/pg/msg_hostinfo.go
```diff
@@ -60,3 +77,26 @@ func (h *Handler) MsgHostInfo(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg
 
 	return &reply, nil
 }
+
+// ParseOSRelease parses the etc/os-release file.
+func ParseOSRelease(reader io.Reader) (string, string, error) {
```
AlekSi, line 82: We should not have that function and unit test for it in different packaages.

I think the best place for both could be internal/handlers/common

AlekSi (changes requested):

internal/handlers/pg/msg_hostinfo.go
DoodgeMatvey: If I place them in `internal/handlers/common` and make it unexportable I cannot use it in pg package for `MsgHostInfo` func

internal/handlers/parse_osrelease_test.go
DoodgeMatvey: Elsewhere in the code I have seen `"expected: %s\nactual  : %s"`, but what you suggested is also ok, I replaced

internal/handlers/pg/msg_hostinfo.go
DoodgeMatvey: I placed them in `internal/handlers/common` but left the function exported. What do you think about it?

internal/handlers/common/parse.go
seeforschauer: I meant empty lines, a lines w/o code

seeforschauer (changes requested):
lgfm, just some formatting

internal/handlers/common/parse_osrelease.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+)
+
+// ParseOSRelease parses the etc/os-release file.
+func ParseOSRelease(reader io.Reader) (string, string, error) {
+	scanner := bufio.NewScanner(reader)
+
+	configParams := map[string]string{}
+
```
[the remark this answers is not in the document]
DoodgeMatvey: fixed

internal/handlers/common/parse_osrelease.go
```diff
@@ -0,0 +1,46 @@
...
+	for scanner.Scan() {
+		str := strings.Split(scanner.Text(), "=")
+
```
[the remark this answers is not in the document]
DoodgeMatvey: fixed

internal/handlers/common/parse_osrelease.go
```diff
@@ -0,0 +1,46 @@
...
+		if len(str) == 1 {
+			continue
+		}
+
```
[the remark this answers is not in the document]
DoodgeMatvey: fixed

internal/handlers/common/parse_osrelease.go
```diff
@@ -0,0 +1,46 @@
...
+		configParams[str[0]] = str[1]
+	}
+
```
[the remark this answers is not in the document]
DoodgeMatvey: Thank you, fixed

seeforschauer (approved):
Thank you!

AlekSi (approved):

PR #518, opened 2022-04-25
Support `$slice` projection query operator

### LINKED ISSUE #362: Support `$slice` projection query operator

GinGin3203:
Closes #362.


internal/handlers/common/projection.go
```diff
@@ -233,3 +252,81 @@ func filterFieldArrayElemMatch(k1 string, doc, conditions *types.Document, docVa
 	}
 	return
 }
+
+// filterFieldArraySlice is a function that implements $slice projection query.
+func filterFieldArraySlice(docValue *types.Array, projectionValue any) error {
+	var elementsToSkip, elementsToReturn int
+	switch projectionValue.(type) {
+	// TODO can it have type int64?
```
AlekSi, line 260: I think PRs like that are where test-driver development really shines. We could write a simple integration test first, note how MongoDB works, adjust the test to pass against MongoDB, then make FerretDB behave the same way.

integration/projection_array_test.go
```diff
@@ -0,0 +1,210 @@
...
+	require.NoError(t, err)
+
+	type testCase struct {
+		projection    bson.D
+		expectedArray bson.A
+		err           *mongo.CommandError
+	}
+
+	for name, tc := range map[string]testCase{
+		//"SingleArgString": {
```
GinGin3203, line 29: See my comment below

GinGin3203:
I've commented out some tests because there is an issue with printing types and values the same way as MongoDB. For example, Ferret prints a string `"someString"` as `someString`, while Mongo prints it as `"someString"`; a similar issue with `types.Array` and `types.Document` (I've implemented `String()` for `types.Array`, we should probably make the same for `types.Document`), but doing a `String()` for `string` type would require introducing a type alias, and I'm not sure how you feel about that 😄 

After it's resolved, these tests can be uncommented

internal/types/types.go
```diff
@@ -188,3 +192,20 @@ func deepCopy(value any) any {
 		panic(fmt.Sprintf("types.deepCopy: unsupported type: %[1]T (%[1]v)", value))
 	}
 }
+
+// String is needed to conveniently inform about type or value
+// mismatch when returning an error. For example,
+// when informing that `int32` is an invalid type, in the
+// error message we would like to get `int` and not `int32`.
+func String(v any) string {
```
GinGin3203, line 200: I guess we're going to need this or something similar when implementing more operators

w84thesun: I've implemented this method in #544

integration/projection_array_test.go
```diff
@@ -0,0 +1,210 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+func TestProjectionQuerySlice(t *testing.T) {
```
w84thesun, line 14: I think this test should be in `query_projection_test.go`. And it should be called `TestQueryProjectionSlice`.

integration/projection_array_test.go
```diff
@@ -0,0 +1,210 @@
...
+	t.Parallel()
+	ctx, collection := setup(t)
+	_, err := collection.InsertOne(ctx,
+		bson.D{{"_id", "array"}, {"value", bson.A{1, 2, 3, 4}}},
+	)
+	require.NoError(t, err)
+
+	type testCase struct {
```
w84thesun, line 22: We don't use that structure anywhere outside the for loop below. Let's make embedded as in other tests.

internal/handlers/common/projection.go
```diff
@@ -25,7 +26,7 @@ import (
 	"github.com/FerretDB/FerretDB/internal/util/must"
 )
 
-// isProjectionInclusion: projection can be only inclusion or exclusion. Validate and return true if inclusion.
+// isProjectionInclusion: projection can be only inclusion or exlusion. Validate and return true if inclusion.
```
w84thesun, line 29:
```suggestion
// isProjectionInclusion: projection can be only inclusion or exclusion. Validate and return true if inclusion.
```

internal/handlers/common/projection.go
```diff
@@ -234,3 +258,140 @@ func filterFieldArrayElemMatch(k1 string, doc, conditions *types.Document, docVa
 	}
 	return
 }
+
+// filterFieldArraySlice is a function that implements $slice projection query.
```
w84thesun, line 262:
```suggestion
// filterFieldArraySlice implements $slice projection query.
```

internal/handlers/common/projection.go
```diff
@@ -234,3 +258,140 @@ func filterFieldArrayElemMatch(k1 string, doc, conditions *types.Document, docVa
...
+func filterFieldArraySlice(docValue *types.Array, projectionValue any) (*types.Array, error) {
+	switch projectionValue.(type) {
+	// TODO what do we do with int64?
```
w84thesun, line 265: You could add a test for the `int64` case.

internal/handlers/common/projection.go
```diff
@@ -234,3 +258,140 @@ func filterFieldArrayElemMatch(k1 string, doc, conditions *types.Document, docVa
...
+					"First argument to $slice must be an array, but is of type: %s",
+					types.String(must.NotFail(arr.Get(0))),
+				),
+			)
+		}
+		var skip, limit int
+		arg := [2]int{}
+		for i := range arg {
+			switch v := must.NotFail(arr.Get(i)).(type) {
+			// TODO what do we do with int64?
```
w84thesun, line 329: Same as above.

integration/projection_array_test.go
seeforschauer: I suggest renaming the file also, to be in the same style as in other places, i.e `to query_slice_test.go`

internal/handlers/common/projection.go
```diff
@@ -234,3 +258,140 @@ func filterFieldArrayElemMatch(k1 string, doc, conditions *types.Document, docVa
...
+					"Invalid $slice syntax. The given syntax { $slice: %s } "+
+						"did not match the find() syntax because :: Location31272: "+
+						"$slice array argument should be of form [skip, limit] :: "+
+						"The given syntax did not match the expression "+
+						"$slice syntax. :: caused by :: "+
+						"Expression $slice takes at least 2 arguments, and at most 3, but %d were passed in.",
+					projectionValue,
+					arr.Len(),
+				))
+		} else if arr.Len() == 3 {
```
seeforschauer, line 316:
```suggestion
		} 
		if arr.Len() == 3 {
```

internal/types/array.go
```diff
@@ -55,6 +60,22 @@ func MustNewArray(values ...any) *Array {
 
 func (a *Array) compositeType() {}
 
+func (a *Array) String() string {
+	b := new(strings.Builder)
+	b.WriteString("[")
+	for i := 0; i < a.Len(); i++ {
+		// TODO obviously this won't work correctly with structs
+		b.WriteString(fmt.Sprintf(" %v", must.NotFail(a.Get(i))))
```
seeforschauer, line 68: No, it is better to create a recursive function `.String()` for all types, including a document.

integration/projection_array_test.go
w84thesun: We already have `query_projection_test.go`

GinGin3203: @w84thesun @seeforschauer so which one is it? 😄

GinGin3203: I guess I'm either for @w84thesun's suggestion (I agree that there aren't so many tests for `$slice` that it deserves it's own file) or for storing document projection tests and array projection tests in separate files; @seeforschauer WDYT?

seeforschauer: it's ok to store it in the existing file.

internal/types/array.go
GinGin3203: I agree! 
But if I understand correctly what you are suggesting, are you sure it should be in this PR? I've addressed this issue in a comment: https://github.com/FerretDB/FerretDB/pull/518#issuecomment-1114250719

I'm just saying that we can't call `.String()` recursively because of the use of in-built types

I would suggest leaving it as is and later refactoring it, WDYT?

internal/types/types.go
AlekSi: Let's cut it when #544 is merged

internal/types/array.go
AlekSi: I guess we first should decide what this function _does_, as there is no documentation comment. It returns some string representation of array… for what? If that's for debugging, then can we use `testutil.Dump` or some helper like `testutil.AssertEqual` that uses it instead? If that's for not-testing code, then what makes this string representation more important than others, given that this is a method that can be only one?

internal/handlers/common/projection.go
GinGin3203: An issue with `int64` is that `types.Array.Get(i int)` only operate with `int` 
WDID?

internal/types/array.go
GinGin3203: There are situations where Mongo returns an error, that contains user input, for example:
```
values> db.values.find({name:"array-embedded"}, {value:{$slice:[1,2,3,4]}})
MongoServerError: Invalid $slice syntax. The given syntax { $slice: [ 1, 2, 3, 4 ] } did not match the find() syntax because :: Location31272: $slice array argument should be of form [skip, limit] :: The given syntax did not match the expression $slice syntax. :: caused by :: Expression $slice takes at least 2 arguments, and at most 3, but 4 were passed in.
```

In this PR, I emulated this particular error with the following snippet:

```
	if arr.Len() < 2 || arr.Len() > 3 {
			return nil, NewErrorMsg(ErrInvalidArg,
				fmt.Sprintf(
					"Invalid $slice syntax. The given syntax { $slice: %s } "+
						"did not match the find() syntax because :: Location31272: "+
						"$slice array argument should be of form [skip, limit] :: "+
						"The given syntax did not match the expression "+
						"$slice syntax. :: caused by :: "+
						"Expression $slice takes at least 2 arguments, and at most 3, but %d were passed in.",
					projectionValue,
					arr.Len(),
				))
		}
```

So, attaching `.String()` to FerretDB types would make it convenient to print these values when returning an error message to the user

AlekSi: Well, the end-user does not care whatever this code is in method or function, or how it is named :)

The problem with this code being in the `String` method is that `fmt.Stringer` is being used implicitly in many cases, and that might be problematic. For example, that might be confusing in tests where simple assert.Equal uses `fmt.Sprint`. So let's not introduce a method with that name.

So let's use a name that does not match any common interface. `JSONSyntax` might be a good name. And let's add a documentation comment.

AlekSi:
@GinGin3203 May you please resolve merge conflicts?

GinGin3203:
@AlekSi done

GinGin3203:
Please note that I've only added basic support of `JSONSyntax`
I though at first of also including `types.Document`, but I've decided to stay away for now because of possible complications with key ordering

integration/query_projection_test.go
```diff
@@ -15,8 +15,11 @@
 package integration
 
 import (
+	"math"
 	"testing"
 
+	"go.mongodb.org/mongo-driver/mongo"
+
```
seeforschauer, line 22: imports sorting. It's strange that linters didn't report.

internal/handlers/common/projection.go
```diff
@@ -234,3 +258,154 @@ func filterFieldArrayElemMatch(k1 string, doc, conditions *types.Document, docVa
...
+		case float64:
+			if math.IsNaN(pr) {
+			} else if math.IsInf(pr, -1) || n < math.MinInt {
+				n = math.MinInt
+			} else if math.IsInf(pr, +1) || n > math.MaxInt {
+				n = math.MaxInt
+			} else {
+				n = int(pr)
+			}
+		case int64:
```
seeforschauer, line 278: Is the question of what to do with `in64` still actual? How it is sounds?

internal/types/types.go
```diff
@@ -189,3 +192,31 @@ func deepCopy(value any) any {
...
+// JSONSyntax formats a FerretDB value as it's JSON counterpart.
+func JSONSyntax(value any) string {
+	switch value := value.(type) {
+	case int32, int64, bool, float64:
+		return fmt.Sprintf("%v", value)
+	case string:
+		return fmt.Sprintf("%q", value)
+	case NullType:
+		return "null"
+	case *Array:
```
seeforschauer, line 205: what about `*Document` marshaling?

internal/types/types.go
```diff
@@ -189,3 +192,31 @@ func deepCopy(value any) any {
 		panic(fmt.Sprintf("types.deepCopy: unsupported type: %[1]T (%[1]v)", value))
 	}
 }
+
+// JSONSyntax formats a FerretDB value as it's JSON counterpart.
```
seeforschauer, line 196: can you please write a note about how it differs from `fjson` package, which also makes the thing `from-to`? and clarify why `fjson` cannot be used.

internal/types/types_test.go
```diff
@@ -52,3 +54,45 @@ func TestDeepCopy(t *testing.T) {
...
+		},
+		"Int64": {
+			input:    int64(12345),
+			expected: "12345",
+		},
+		"Bool": {
+			input:    true,
+			expected: "true",
+		},
+		"Array": {
```
seeforschauer, line 86: would you please add a Document to the test data?

seeforschauer:
> Please note that I've only added basic support of `JSONSyntax` I though at first of also including `types.Document`, but I've decided to stay away for now because of possible complications with key ordering

So it will fail if a Document will be in data. Maybe let's describe the complications and find a way to resolve them?

internal/handlers/common/projection.go
GinGin3203: No, I believe it's not relevant anymore
I made a stupid mistake there and worked with `int32` instead of `int`

internal/handlers/common/projection.go
GinGin3203: done

internal/handlers/common/projection.go
GinGin3203: done

integration/projection_array_test.go
GinGin3203: done

internal/types/types.go
```diff
@@ -189,3 +192,58 @@ func deepCopy(value any) any {
...
+// 	db.values.find( { *someFilter* }, { arr: { $slice: { a: { b: 3 }, b: [ 1, 2 ], x: 3 } } } )
+//
+// This query returns an error that starts with:
+// 	"Invalid $slice syntax. The given syntax { $slice: { a: { b: 3 }, b: [ 1, 2 ], x: 3 } } did not match the find()
+// 	syntax because :: /.../"
+//
+// This function makes sure that, continuing with the example, the value in the { $slice: /.../ } clause gets formatted
+// correctly, which isn't the case by default. It differs from fjson package in its purpose: while fjson package aims
+// to format values according FJSON standard, JSONSyntax is needed to visually match the output of MongoDB in the
+// particular case of embedding a provided value into an error message.
```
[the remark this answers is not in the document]
GinGin3203: Could you make it a suggested change?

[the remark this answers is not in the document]
seeforschauer: It's about several rows, can I? I thought I can do a suggested change if it's 1 row only. If you show me I will be grateful :)

internal/types/types.go
```diff
@@ -189,3 +192,58 @@ func deepCopy(value any) any {
 		panic(fmt.Sprintf("types.deepCopy: unsupported type: %[1]T (%[1]v)", value))
 	}
 }
+
+// JSONSyntax formats a FerretDB value as it's JSON counterpart. This is useful when embedding an invalid value in an
+// error.
```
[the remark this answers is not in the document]
seeforschauer: doesn't work for me :|

internal/types/types.go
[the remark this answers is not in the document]
GinGin3203: Well, this function is designed to output the exact match of what Mongo outputs in the described case (i.e. embedding user input into an error), so I'm not sure how `altMessage` can be of use here

integration/query_projection_test.go
```diff
@@ -130,3 +132,210 @@ func TestQueryProjectionElemMatch(t *testing.T) {
...
+				return
+			}
+			require.NoError(t, err)
+
+			var actual bson.D
+			err = res.Decode(&actual)
+			require.NoError(t, err)
+
+			if tc.expectedArray == nil {
+				assert.Nil(t, actual.Map()["value"])
```
seeforschauer, line 335: return; 
so next statement doesn't have to be in  `{}`

internal/handlers/common/projection.go
```diff
@@ -158,37 +161,58 @@ func projectDocument(inclusion bool, doc *types.Document, projection *types.Docu
...
+				return
+			}
+			projectionVal := must.NotFail(projectionVal.Get(projectionType))
+			res, err := filterFieldArraySlice(arr, projectionVal)
+			if err != nil {
+				return err
+			}
+			if res == nil {
+				must.NoError(doc.Set(k1, types.Null))
+			} else {
```
seeforschauer, line 206: same as above: cyclomatic complexity here can be reduced.

internal/handlers/common/projection.go
```diff
@@ -234,3 +258,152 @@ func filterFieldArrayElemMatch(k1 string, doc, conditions *types.Document, docVa
...
+
+// filterFieldArraySlice implements $slice projection query.
+func filterFieldArraySlice(docValue *types.Array, projectionValue any) (*types.Array, error) {
+	switch projectionValue.(type) {
+	case int32, int64, float64:
+		var n int
+		switch pr := projectionValue.(type) {
+		case float64:
+			if math.IsNaN(pr) {
+			} else if math.IsInf(pr, -1) || n < math.MinInt {
```
seeforschauer, line 270: also, break for the switch can be used here

integration/query_projection_test.go
GinGin3203: done

internal/types/types.go
```diff
@@ -189,3 +192,58 @@ func deepCopy(value any) any {
...
+//
+// This query returns an error that starts with:
+// 	"Invalid $slice syntax. The given syntax { $slice: { a: { b: 3 }, b: [ 1, 2 ], x: 3 } } did not match the find()
+// 	syntax because :: /.../"
+//
+// JSONSyntax makes sure that the value in the { $slice: /.../ } clause gets formatted correctly.
+// It differs from fjson package in its purpose.
+// JSONSyntax is needed to match the output of MongoDB in the visually particular case of embedding a
+// provided value into an error message. In contrast, fjson package aims to format values according to FJSON standard.
+func JSONSyntax(value any) string {
```
seeforschauer, line 209: Ono more thing (sorry didn't say it before)
Please see https://pkg.go.dev/github.com/FerretDB/FerretDB/internal/types#hdr-Mapping and note the order of the types ( summarised `Composite types` and `Scalar types`. 
And see that in the `fjson` package in the switch of func `fromFJSON` types are ordered like in the doc.
It helps not to omit any type. 

Could you please reorder types in this switch to the standard?

GinGin3203:
@seeforschauer to be honest, I feel a bit overwhelmed by the scope of this PR, I'm afraid it may never end 😅 
WDYT about calling it a day and addressing the work that needs to be done on `JSONSyntax` in a separate issue? 
For now I've added placeholders for unsupported types and sorted everything in the correct order

seeforschauer:
@GinGin3203 let it be a day  :-)  I fully understand this :) Some of my own PRs are also very big.

internal/types/types.go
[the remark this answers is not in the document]
AlekSi: > Well, this function is designed to output the exact match of what Mongo outputs in the described case

We decided not to do that if that's complicated or requires more work just to produce the same message. And I completely forgot to mention that in Slack or anywhere else.

So let's drop the whole `JSONSyntax` thing and produce an error message with simple `json.Marshal`'s result in the middle. `AssertEqualAltError` is designed for cases like this one.

seeforschauer (dismissed):

GinGin3203:
@AlekSi as I understand, this PR can be merged?

internal/handlers/common/projection.go
```diff
@@ -234,3 +258,161 @@ func filterFieldArrayElemMatch(k1 string, doc, conditions *types.Document, docVa
 	}
 	return
 }
+
+// filterFieldArraySlice implements $slice projection query.
+func filterFieldArraySlice(docValue *types.Array, projectionValue any) (*types.Array, error) {
```
AlekSi, line 263: Overall, looks overcomplicated. Can we somehow simplify that function by splitting it or using earlier returns? For example, two cases of the `switch projectionValue.(type)` below look almost unrelated, so probably could be extracted into their own functions

AlekSi (changes requested):

internal/handlers/common/projection.go
```diff
@@ -234,3 +258,161 @@ func filterFieldArrayElemMatch(k1 string, doc, conditions *types.Document, docVa
...
+	switch projectionValue.(type) {
+	case int32, int64, float64:
+		var n int
+		switch pr := projectionValue.(type) {
+		case float64:
+			if math.IsNaN(pr) {
+				break // because n == 0 already
+			}
+			if math.IsInf(pr, -1) || n < math.MinInt {
```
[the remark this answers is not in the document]
GinGin3203: Shoot, it's a typo
Thanks for pointing out

integration/query_projection_test.go
```diff
@@ -130,3 +137,249 @@ func TestQueryProjectionElemMatch(t *testing.T) {
...
+				Code: 28667,
+				Name: "Location28667",
+				Message: "Invalid $slice syntax. The given syntax { $slice: { a: { b: 3 }, b: \"string\" } } " +
+					"did not match the find() syntax because :: Location31273: " +
+					"$slice only supports numbers and [skip, limit] arrays " +
+					":: The given syntax did not match the expression $slice syntax. " +
+					":: caused by :: Expression $slice takes at least 2 arguments, and at most 3, " +
+					"but 1 were passed in.",
+			},
+			altMessage: "Invalid $slice syntax. The given syntax { $slice: {} } " +
```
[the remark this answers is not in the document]
GinGin3203: I guess I don't understand the role of `altMessage`: is it a "better error" or "as close as possible to Mongo without going out of the way error"?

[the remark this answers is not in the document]
AlekSi: It is "as close as possible to Mongo without lying" :) "The given syntax { $slice: {} }" tells the user that they submitted an empty document for `$slice` value. They may spend a significant amount of time inspecting their code looking for the bug.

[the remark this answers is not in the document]
AlekSi: #578

integration/query_projection_test.go
```diff
@@ -64,6 +66,11 @@ func TestQueryProjection(t *testing.T) {
 			projection: bson.D{{"_id", false}, {"array", int32(1)}},
 			expected:   bson.D{},
 		},
+		"ProjectionSliceNonArrayField": {
```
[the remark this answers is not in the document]
GinGin3203: Well, in my reasoning, in `TestQueryProjectionSlice` we are only interested in how the `$slice` operator transforms the array (that's why we have `expectedArray` there), but in the particular situation where `$slice` is not present we are interested in the document as a whole
So, I could put this test in `TestQueryProjectionSlice` outside the test table, but I think that it fits better where it is now
WDYT?

seeforschauer (approved):

AlekSi (approved):

PR #523, opened 2022-04-25
Move all handler test to integration tests

### LINKED ISSUE #417: Move more tests to integration tests

w84thesun:
Closes #417.


seeforschauer (approved):

integration/commands_administration_test.go
```diff
@@ -193,3 +198,132 @@ func TestCommandsAdministrationGetParameter(t *testing.T) {
 	assert.Contains(t, keys, "quiet")
 	assert.Equal(t, false, m["quiet"])
 }
+
+func TestStatisticsCommands(t *testing.T) {
+	t.Skip("TODO: response deltas investigation needed")
```
AlekSi, line 203: There should be an issue URL

integration/commands_administration_test.go
```diff
@@ -193,3 +198,132 @@ func TestCommandsAdministrationGetParameter(t *testing.T) {
...
+	ctx, collection := setup(t, shareddata.Scalars, shareddata.Composites)
+
+	for name, tc := range map[string]struct {
+		command  any
+		response map[string]any
```
AlekSi, line 208: Unless it is very hard, we should compare `bson.D`s to also check fields order

integration/query_projection_test.go
```diff
@@ -0,0 +1,122 @@
...
+	require.NoError(t, err)
+
+	for name, tc := range map[string]struct {
+		projection any
+		filter     any
+		expected   bson.D
+	}{
+		"FindProjectionInclusions": {
+			filter: bson.D{{"_id", "document-composite"}},
+			// TODO: fix projection
```
AlekSi, line 37: Issue URL

integration/query_projection_test.go
```diff
@@ -0,0 +1,122 @@
...
+		},
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			cursor, err := collection.Find(
+				ctx,
+				tc.filter,
+				options.Find().SetProjection(tc.projection),
```
AlekSi, line 61: That seems to be short enough to put on one line. One of the `expectred` documents above is longer

internal/handlers/handler_fuzz_test.go
```diff
@@ -17,14 +17,55 @@ package handlers
...
+	noLogging bool
+	poolOpts  *testutil.PoolOpts
+}
+
+// setup creates shared objects for testing.
+//
+// Using shared objects helps us spot concurrency bugs.
+// If some test is failing and the log output is confusing, and you are tempted to move setup call to subtest,
+// instead run that single test with `go test -run test/name`.
+func setup(t testing.TB, opts *setupOpts) (context.Context, *pg.Handler, *pgdb.Pool) {
```
AlekSi, line 43: I don't think that all should be in that file – that one is about fuzzing

AlekSi (commented):
Nice catch with a typo in file name!

integration/commands_administration_test.go
w84thesun: I used `map[string]any` to skip fields like `gitVersion` or `debug` field which requires me to use `util/version` package.
If it's ok to use it then I can change `response` field to `bson.D`.

internal/handlers/handler_fuzz_test.go
w84thesun: Should I leave that code in `handler_test.go`?

integration/query_projection_test.go
w84thesun: Fixed

integration/query_projection_test.go
w84thesun: Fixed

integration/commands_administration_test.go
w84thesun: Fixed

w84thesun:
> Looking into coverage report, I wonder if we deleted some tests without moving them. https://app.codecov.io/gh/FerretDB/FerretDB/compare/523/changes shows lines like that:
> 
> image
> 
> Did we forgot tests for negative results?

I've compared tests, and added one missing and two more new ones. The coverage shows those tests which now is rather disabled (stats) or not working correctly (projection).

integration/commands_administration_test.go
w84thesun: I've changed to `bson.D` for now.

internal/handlers/handler_fuzz_test.go
AlekSi: Ah, I see. Yeah, let's leave it there for now

AlekSi (approved):

PR #519, opened 2022-04-25
Implement `nameOnly` for `listDatabases`

seeforschauer:
Closes #301.

PR #524, opened 2022-04-25
Implement `nameOnly` for `listDatabases`

### LINKED ISSUE #301: Implement `nameOnly` for `listCollections` command
`listCollections` command should support `nameOnly` parameter. See references to this issue in the source code.

AlekSi:
Closes #301.

### DIFF
--- a/internal/handlers/pg/msg_listdatabases.go
+++ b/internal/handlers/pg/msg_listdatabases.go
@@ -43,8 +43,8 @@ func (h *Handler) MsgListDatabases(ctx context.Context, msg *wire.OpMsg) (*wire.
 		return nil, err
 	}
 
-	nameOnly, err := common.GetOptionalParam(document, "nameOnly", false)
-	if err != nil {
+	var nameOnly bool
+	if nameOnly, err = common.GetOptionalParam(document, "nameOnly", false); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


internal/handlers/pg/msg_listdatabases.go
```diff
@@ -66,27 +64,47 @@ func (h *Handler) MsgListDatabases(ctx context.Context, msg *wire.OpMsg) (*wire.
...
 			}
 		}
 	}
 
+	if nameOnly {
+		var reply wire.OpMsg
+		err = reply.SetSections(wire.OpMsgSection{
+			Documents: []*types.Document{must.NotFail(types.NewDocument(
+				"databases", databases,
+				"ok", float64(1),
```
AlekSi, line 98: Only "ok" was missing in #519

seeforschauer (changes requested):
how about tests? ^_^

AlekSi:
They exist already: https://github.com/FerretDB/FerretDB/blob/94b7d1025e9597e53d6cce9796507dba05eb8287/integration/commands_administration_test.go#L126-L128

internal/handlers/pg/msg_listdatabases.go
seeforschauer: thank you!

seeforschauer:
Then test must have failed, and they did not. Or?

AlekSi:
`ListDatabaseNames` uses only `name` fields in response documents. We returned did more queries and returned more data than was needed; that was a problem in performance, but that's valid behavior. Now we do fewer queries, but still work the same way as MongoDB.

seeforschauer (approved):

PR #532, opened 2022-04-26
Issues 449 get parameter

fenogentov:


.gitignore
```diff
@@ -17,3 +17,6 @@ internal/util/version/branch.txt
 **/testdata/fuzz
 
 docker-compose.override.yml
+go.wor
+go.wor.sum
+tst.txt
```
seeforschauer, line 22: could you please remove the private file which probably won't appear in another person's local file directory,  from the gitignore?

AlekSi:
I guess we can close that one in favor of #535?

fenogentov:
> I guess we can close that one in favor of #535?

Yes. `getParameter` is implemented there

PR #534, opened 2022-04-26
Set the handler to use via a command-line flag and remove debug handlers from interface

### LINKED ISSUE #455: Extract common interface for handlers
- [x] Make an interface that is common for all handlers:
```go
type Handler interface {
	MsgBuildInfo(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error)
	MsgCollStats(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error)
	// ...
	QueryCmd(ctx context.Context, query *wire.OpQuery) (*wire.OpReply, error)
}
```
- [x] Make a new `dummy` handler (in a `dummy` package) that implements this interface like this:
```go
// MsgBuildInfo returns a summary of the build information.
func (h *Handler) MsgBuildInfo(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
	return nil, NewErrorMsg(ErrNotImplemented, "I'm a dummy, not a handler")
}
```
- [x] Refactor the `Handle` method and related code to use that interface.
- [ ] Set the handler to use via a command-line flag: `-handler=dummy` or `-handler=pg` (default).
- [ ] Move "very" common command handlers listCommands, debug_error, debug_panic out of Handler interface to avoid implementing them in all interface implementations

seeforschauer:
Closes #455.


cmd/ferretdb/main.go
```diff
@@ -85,23 +90,43 @@ func main() {
...
-		PgPool:          pgPool,
+		Handler:         h,
 		Logger:          logger,
 		TestConnTimeout: *testConnTimeoutF,
 	})
 
 	prometheus.DefaultRegisterer.MustRegister(l)
 
+	var err error
 	err = l.Run(ctx)
```
AlekSi, line 130:
```suggestion
	err := l.Run(ctx)
```

internal/handlers/common/handler.go
```diff
@@ -17,11 +17,16 @@ package common
...
 
+	"github.com/prometheus/client_golang/prometheus"
+
 	"github.com/FerretDB/FerretDB/internal/wire"
 )
 
 // Handler interface represents common commands handlers.
 type Handler interface {
+	// Handle handles the message.
+	Handle(context.Context, *wire.MsgHeader, wire.MsgBody) (*wire.MsgHeader, wire.MsgBody, bool)
```
AlekSi, line 28: No, that method should not be here. Previously, we discussed if our common interface should have this one method or all methods for each individual command, and decided to stick with the latter. We should not have both, and we should force each handler to implement that quite complicated method.

internal/handlers/handler_test.go
```diff
@@ -102,6 +101,7 @@ func handle(ctx context.Context, t *testing.T, handler *pg.Handler, req *types.D
 
 func TestFind(t *testing.T) {
 	t.Parallel()
+
```
AlekSi, line 104: No, we don't have an empty line there; check out other tests.

(and we are going to delete all those tests anyway)

AlekSi (changes requested):

internal/handlers/common/handler.go
seeforschauer: extracted a common router

internal/handlers/handler_test.go
seeforschauer: removed

cmd/ferretdb/main.go
```diff
@@ -40,6 +44,7 @@ var (
 	proxyAddrF       = flag.String("proxy-addr", "127.0.0.1:37017", "")
 	versionF         = flag.Bool("version", false, "print version to stdout (full version, commit, branch, dirty flag) and exit")
 	testConnTimeoutF = flag.Duration("test-conn-timeout", 0, "test: set connection timeout")
+	handlerF         = flag.String("handler", "pg", "handler: can be pg or dummy")
```
w84thesun, line 47: I think we should provide information about the flag's purpose and then provide available options like `handler: set backend handler (pg, dummy)`

internal/clientconn/conn.go
```diff
@@ -250,3 +247,59 @@ func (c *conn) run(ctx context.Context) (err error) {
 		}
 	}
 }
+
+// route routes to common.Router and measures all the result.
```
w84thesun, line 251:
```suggestion
// route routes to common.Router and measures all the results.
```

internal/clientconn/conn.go
```diff
@@ -250,3 +247,59 @@ func (c *conn) run(ctx context.Context) (err error) {
...
+func (c *conn) route(h common.Handler, ctx context.Context, reqHeader *wire.MsgHeader, reqBody wire.MsgBody,
```
w84thesun, line 252: Looks like parameter `h common.Handler` is unused. Should we remove it?

cmd/ferretdb/main.go
```diff
@@ -85,24 +90,43 @@ func main() {
 
 	go debug.RunHandler(ctx, *debugAddrF, logger.Named("debug"))
 
-	pgPool, err := pgdb.NewPool(ctx, *postgresqlURLF, logger, false)
-	if err != nil {
-		logger.Fatal(err.Error())
+	var h common.Handler
+
+	startTime := time.Now()
```
w84thesun, line 95: There is only one usage of that variable. We can call time.Now() right in `handlerOpts` initialization as we do in `integration/setup.go`.

internal/clientconn/conn.go
seeforschauer: It's used on the line 290

cmd/ferretdb/main.go
seeforschauer: fixed

cmd/ferretdb/main.go
seeforschauer: moved

internal/clientconn/conn.go
```diff
@@ -250,3 +248,63 @@ func (c *conn) run(ctx context.Context) (err error) {
...
+		fallthrough
+	case wire.OP_DELETE:
+		fallthrough
+	case wire.OP_KILL_CURSORS:
+		fallthrough
+	case wire.OP_COMPRESSED:
+		fallthrough
+	default:
+		err = lazyerrors.Errorf("unexpected OpCode %s", reqHeader.OpCode)
+		result = pointer.ToString("unexpected")
```
seeforschauer, line 289: maybe extract to a named error... I don't like this much.

w84thesun (dismissed):

AlekSi (changes requested):
There are merge conflicts after handler_test.go removal

seeforschauer:
My hope is that if do not press the button I would merge **locally** faster.

internal/clientconn/conn.go
```diff
@@ -250,3 +248,63 @@ func (c *conn) run(ctx context.Context) (err error) {
 		}
 	}
 }
+
+// route routes to common.Router and measures all the results.
```
AlekSi, line 252:
```suggestion
// route routes to common.Route and measures all the results.
```

internal/handlers/common/route.go
```diff
@@ -0,0 +1,156 @@
...
+	"fmt"
+
+	"github.com/AlekSi/pointer"
+
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+	"github.com/FerretDB/FerretDB/internal/wire"
+)
+
+// Route routes the message.
```
AlekSi, line 28: I definitely like this name more!

But I'm not sure it belongs in this package. It contains code shared by handlers, but this function is not used by handlers. Placing it in `clientconn` package is probably a better idea?..

internal/handlers/common/route.go
```diff
@@ -0,0 +1,156 @@
...
+	// TODO Don't call MarshalBinary there. Fix header in the caller?
+	// https://github.com/FerretDB/FerretDB/issues/273
+	b, err := resBody.MarshalBinary()
+	if err != nil {
+		result = nil
+		panic(err)
+	}
+	resHeader.MessageLength = int32(wire.MsgHeaderLen + len(b))
+
+	// resHeader.RequestID = atomic.AddInt32(&h.lastRequestID, 1) // XXX: not set anywhere
```
AlekSi, line 137: We should set it though

internal/clientconn/listener_metrics.go
```diff
@@ -47,19 +49,41 @@ func NewListenerMetrics() *ListenerMetrics {
 			},
 			[]string{"error"},
 		),
+		requests: prometheus.NewCounterVec(
```
AlekSi, line 52: Well, they do not belong there. Those were listener's metrics, and requests and responses are on a different level

internal/handlers/common/route.go
seeforschauer: > I definitely like this name more!
>

Thank you. 

> But I'm not sure it belongs in this package. It contains code shared by handlers, but this function is not used by handlers. Placing it in `clientconn` package is probably a better idea?..

:+1:

internal/handlers/common/route.go
seeforschauer: Let's do this PR refactoring only.
Need to know how to generate it, who else should have access to it, in which logs to pass, probably let logger be in the context etc.
WDYT?

internal/handlers/common/route.go
seeforschauer: yes and... I run onto import cycling )

internal/clientconn/listener_metrics.go
seeforschauer: what about client conn?

internal/handlers/common/route.go
AlekSi: Sorry, I don’t follow. That field is being set now: https://github.com/FerretDB/FerretDB/blob/6f00c8dc8059c1d1707bdf5c23d61a19e29c19d7/internal/handlers/pg/handler.go#L192 I don’t see what prevents us from uncommenting this line back to keep it as it was before

seeforschauer: then the handler will have an interface LastRequestID(), it's ok?

seeforschauer: moved to `clentconn` as route now is a `clientconn`-s.

internal/clientconn/listener_metrics.go
seeforschauer: moved to clientconn.

internal/clientconn/route.go
```diff
@@ -0,0 +1,172 @@
...
+	case wire.OP_MSG:
+		msg := reqBody.(*wire.OpMsg)
+		var document *types.Document
+		document, err = msg.Document()
+		command = document.Command()
+
+		if err == nil {
+			resHeader.OpCode = wire.OP_MSG
+			resBody, err = c.handleOpMsg(ctx, msg, command)
+		}
```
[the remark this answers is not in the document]
seeforschauer: no ) it's on it's correct way ) I'll revert that change locally.

internal/clientconn/conn_metrics.go
```diff
@@ -52,18 +47,18 @@ func NewMetrics() *Metrics {
 }
 
 // Describe implements prometheus.Collector.
-func (lm *Metrics) Describe(ch chan<- *prometheus.Desc) {
+func (lm *ConnMetrics) Describe(ch chan<- *prometheus.Desc) {
```
AlekSi, line 50: It should be `cm` now, not `lm` (Listener Metrics)

internal/clientconn/route.go
```diff
@@ -0,0 +1,176 @@
...
+//
+// Specific message handlers should do one of the following:
+//  * return normal response body;
+//  * return protocol error (*common.Error, possibly wrapped) - it will be returned to the client;
+//  * return any other error - it will be returned to the client as InternalError before terminating connection.
+//
+// They should not panic on bad input, but may do so in "impossible" cases.
+// They also should not use recover(). That allows us to use fuzzing.
+// (Panics terminate the connection without a response on a different level.)
+func (c *conn) route(ctx context.Context, reqHeader *wire.MsgHeader, reqBody wire.MsgBody,
```
AlekSi, line 42: Not sure why we move that method away from all other `conn` methods. I think it belongs in `conn.go`, where the type is defined

internal/clientconn/route.go
```diff
@@ -0,0 +1,176 @@
...
+// Route routes the message.
+//
+// Specific message handlers should do one of the following:
+//  * return normal response body;
+//  * return protocol error (*common.Error, possibly wrapped) - it will be returned to the client;
+//  * return any other error - it will be returned to the client as InternalError before terminating connection.
+//
+// They should not panic on bad input, but may do so in "impossible" cases.
+// They also should not use recover(). That allows us to use fuzzing.
+// (Panics terminate the connection without a response on a different level.)
```
AlekSi, line 41: That's no longer a different level

internal/clientconn/route.go
```diff
@@ -0,0 +1,176 @@
...
+func (c *conn) route(ctx context.Context, reqHeader *wire.MsgHeader, reqBody wire.MsgBody,
+) (resHeader *wire.MsgHeader, resBody wire.MsgBody, closeConn bool) {
```
AlekSi, line 43: I think that's one case where we make linter happy and code worse. That closing `)` on a new line looks really bad.

`//nolint:lll` to the rescue!

internal/clientconn/route.go
```diff
@@ -0,0 +1,176 @@
...
+
+	"github.com/FerretDB/FerretDB/internal/handlers/common"
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+	"github.com/FerretDB/FerretDB/internal/wire"
+)
+
+// Route routes the message.
+//
+// Specific message handlers should do one of the following:
```
AlekSi, line 34: "Specific message handlers" was used to differentiate that `Handle` methods from other handlers. Since this method no longer named `Handle`, we should refresh that comment

internal/clientconn/conn.go
```diff
@@ -53,31 +53,34 @@ var AllModes = []Mode{NormalMode, ProxyMode, DiffNormalMode, DiffProxyMode}
...
-	mode    Mode
-	l       *zap.SugaredLogger
-	h       *pg.Handler
-	proxy   *proxy.Handler
+	netConn       net.Conn
+	mode          Mode
+	l             *zap.SugaredLogger
+	h             common.Handler
+	m             *ConnMetrics
+	proxy         *proxy.Handler
```
AlekSi, line 61: It always bothered me that `proxy.Handler` does not implement Handler interface. But now we have a better name – router. Let's rename that type

internal/clientconn/conn.go
```diff
@@ -86,18 +89,12 @@ func newConn(opts *newConnOpts) (*conn, error) {
...
-		Metrics:   opts.handlersMetrics,
-		StartTime: opts.startTime,
-	}
 	return &conn{
 		netConn: opts.netConn,
 		mode:    opts.mode,
 		l:       l.Sugar(),
-		h:       pg.New(handlerOpts),
+		h:       opts.handler,
+		m:       NewConnMetrics(),
```
AlekSi, line 97: That breaks metrics gathering since we only register a listener's collector in Prometheus registry, not all conneciton (and we can't register each connection individually).

That should be a shared instance passed via opts from listener.

internal/clientconn/conn.go
```diff
@@ -250,3 +241,13 @@ func (c *conn) run(ctx context.Context) (err error) {
 		}
 	}
 }
+
+// Describe implements prometheus.Collector.
+func (l *conn) Describe(ch chan<- *prometheus.Desc) {
```
AlekSi, line 246: Those methods are not called anywhere

internal/clientconn/listener.go
```diff
@@ -159,11 +160,9 @@ func (l *Listener) Addr() net.Addr {
 // Describe implements prometheus.Collector.
 func (l *Listener) Describe(ch chan<- *prometheus.Desc) {
 	l.metrics.Describe(ch)
-	l.handlersMetrics.Describe(ch)
```
AlekSi, line 162: That's why we no longer collect request/response metrics, but we should collect them

internal/handlers/common/commands.go
```diff
@@ -32,10 +32,6 @@ type Command struct {
 // Commands is a map of commands that common.Handler interface can support.
 // Order of entries matches the interface definition.
 var Commands = map[string]Command{
-	"listCommands": {
-		Help:    "Returns a list of currently supported commands.",
-		Handler: (Handler).MsgListCommands,
```
AlekSi, line 37: Handler for that command is in the same package, so we could wire it there, together with debug_error and debug_panic

internal/handlers/pg/handler.go
```diff
@@ -50,166 +43,20 @@ type Handler struct {
...
-			return cmd.Handler(h, ctx, msg)
-		}
-	}
-
-	errMsg := fmt.Sprintf("no such command: '%s'", cmd)
-	return nil, common.NewErrorMsg(common.ErrCommandNotFound, errMsg)
-}
-
 // MsgDebugError used for debugging purposes.
 func (h *Handler) MsgDebugError(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
```
AlekSi, line 61: There is no way to call those methods now.

internal/clientconn/route.go
```diff
@@ -0,0 +1,176 @@
...
+	// do not spend time dumping if we are not going to log it
+	if c.l.Desugar().Core().Enabled(zap.DebugLevel) {
+		c.l.Debugf("Response header: %s", resHeader)
+		c.l.Debugf("Response message:\n%s\n\n\n", resBody)
+	}
+	return
+}
+
+func (c *conn) handleOpMsg(ctx context.Context, msg *wire.OpMsg, cmd string) (*wire.OpMsg, error) {
+	if cmd == "listCommands" {
```
AlekSi, line 164: Let's move `listCommands`, `debug_error`, `debug_panic` into `common.Commands` too

AlekSi (changes requested):

integration/commands_diagnostic_test.go
```diff
@@ -86,8 +86,4 @@ func TestCommandsDiagnosticListCommands(t *testing.T) {
 
 	assert.Equal(t, 1.0, m["ok"])
 	assert.Equal(t, []string{"commands", "ok"}, CollectKeys(t, actual))
-
-	commands := m["commands"].(bson.D)
-	listCommands := commands.Map()["listCommands"].(bson.D)
-	assert.NotEmpty(t, listCommands.Map()["help"].(string))
```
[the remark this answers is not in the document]
seeforschauer: no need to change this after review fixes.

internal/clientconn/conn.go
seeforschauer: renamed.

internal/clientconn/conn.go
seeforschauer: moved to listerens' metrics.

internal/clientconn/conn.go
seeforschauer: it should be solved right now - call from the listener

internal/clientconn/conn_metrics.go
seeforschauer: done

internal/clientconn/listener.go
seeforschauer: added as conn metrics gathering

internal/clientconn/route.go
seeforschauer: changed.

internal/clientconn/route.go
seeforschauer: moved.

internal/clientconn/route.go
seeforschauer: > I think that's one case where we make linter happy and code worse. That closing `)` on a new line looks really bad.
> 
> `//nolint:lll` to the rescue!

Added. 

However, I think that here is rather the method to be changed since the parameter list and returning values is more than 4. 
For now, I don't want to refactor this, trying to concentrate on one change at a time (it can be big).

internal/clientconn/route.go
seeforschauer: done.

internal/handlers/common/commands.go
seeforschauer: added with nil's

internal/handlers/pg/handler.go
seeforschauer: added.

internal/clientconn/route.go
seeforschauer: removed.

AlekSi (approved):

PR #535, opened 2022-04-26
Support `getParameter`'s parameters

### LINKED ISSUE #449: Support `getParameter`'s parameters
Currently, FerretDB behaves as if `getParameter` command was invoked as `{"getParameter": "*"}`. We should also support the other form `{"getParameter": <something>, "<parameter>": <something>}` that returns the value of the named parameter.

We should also test the edge cases where something is something weird like NaN, Infinity, etc.

See also #533.

### LINKED ISSUE #533: Support protocol errors without code
`mongosh` query `db.adminCommand({getParameter: "invalid"})` makes MongoDB return:

```
{
  "Checksum": 0,
  "FlagBits": 0,
  "Sections": [
    {
      "Document": {
        "$k": [
          "ok",
          "errmsg"
        ],
        "ok": {
          "$f": 0
        },
        "errmsg": "no option found to get"
      },
      "Kind": 0
    }
  ]
}
```

That error does not have an error code at all.

FerretDB code assumes that all errors has code [there](https://github.com/FerretDB/FerretDB/blob/ad8acc08a70f64ff4e877e594653a9146b738b0e/internal/handlers/common/error.go#L59-L61), [there](https://github.com/FerretDB/FerretDB/blob/ad8acc08a70f64ff4e877e594653a9146b738b0e/integration/query_array_test.go#L126-L130) and in other similar places.

We should make FerretDB compatible with MongoDB even in that weird case.

See also #449.

fenogentov:
Closes #449.
Closes #533.


seeforschauer (changes requested):
@fenogentov check dirty is not clean: you need to run `task init` locally and see what changes.

seeforschauer:
> > check dirty is not clean: you need to run `task init` locally and see what changes.
> 
> local test was ok

yes, it is ok.. and go generate -x ./... will add some changes which it's necessary to commit.

internal/handlers/common/error.go
```diff
@@ -29,6 +29,8 @@ type ErrorCode int32
 
 const (
 	// For ProtocolError only.
+	ErrWeird = ErrorCode(0) // WeirdError
```
AlekSi, line 32: It is weird for sure, but let's call it `ErrUnset` or `ErrUnknown`

AlekSi (changes requested):
There are two more things we should do:

1. Add an integration test with `getParameter` call that shows the problem.
2. Do not set `code` and `codeName` for code == 0 there: https://github.com/FerretDB/FerretDB/blob/3b082cd08832682aee502f7b972742cbc35c71fb/internal/handlers/common/error.go#L98-L99

internal/handlers/common/error.go
AlekSi: Actually, `errUnset` – we don't have to export it

internal/handlers/common/error.go
```diff
@@ -29,6 +29,8 @@ type ErrorCode int32
 
 const (
 	// For ProtocolError only.
```
AlekSi, line 31: That comment is for `errInternalError`

internal/handlers/pg/msg_getparameter.go
```diff
@@ -25,25 +26,83 @@ import (
...
+		"quiet", false,
+		"ok", float64(1),
+	)
 
 	var reply wire.OpMsg
+	var errNoFound error
+	var resDoc *types.Document
+	if getPrm == "*" {
+		resDoc = resDB
+
```
AlekSi, line 52: CI complains about that empty line. It should be removed by `task fmt`

integration/commands_administration_test.go
```diff
@@ -175,21 +175,72 @@ func TestCommandsAdministrationCreateDropListDatabases(t *testing.T) {
 }
 
 func TestCommandsAdministrationGetParameter(t *testing.T) {
-	t.Parallel()
+	//	t.Parallel()
```
AlekSi, line 178: Let's keep it

integration/query_array_test.go
```diff
@@ -42,7 +42,7 @@ func TestQueryArraySize(t *testing.T) {
 	for name, tc := range map[string]struct {
 		filter      bson.D
 		expectedIDs []any
-		err         mongo.CommandError
+		err         *mongo.CommandError
```
AlekSi, line 45: We merged more tests where we should do the same changes

integration/commands_administration_test.go
```diff
@@ -175,21 +175,72 @@ func TestCommandsAdministrationCreateDropListDatabases(t *testing.T) {
...
+			err:    &mongo.CommandError{Code: 0, Name: "", Message: `no option found to get`},
+		},
+		"OnlyNonexistentParameters": {
+			filter: bson.D{{"getParameter", 1}, {"quiet_other", 1}},
+			err:    &mongo.CommandError{Code: 0, Name: "", Message: `no option found to get`},
+		},
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			//			t.Parallel()
```
AlekSi, line 224: Let's run those tests in parallel

AlekSi (changes requested):
I think it might be easier to make a separate PR from the current state of `main` branch that changes `err mongo.CommandError` to make it a pointer with matching `if tc.err.Code != 0` change, then return to that PR

integration/commands_administration_test.go
```diff
@@ -185,18 +185,69 @@ func TestCommandsAdministrationGetParameter(t *testing.T) {
 		databaseName: "admin",
 	})
 
-	var actual bson.D
-	err := collection.Database().RunCommand(ctx, bson.D{{"getParameter", "*"}}).Decode(&actual)
-	require.NoError(t, err)
+	for name, tc := range map[string]struct {
+		filter     bson.D
+		expected   map[string]any
+		noExpected []string
```
AlekSi, line 191:
```suggestion
		unexpected []string
```

internal/handlers/common/error.go
```diff
@@ -28,6 +28,8 @@ import (
 type ErrorCode int32
 
 const (
+	errUnset = ErrorCode(0) //
```
AlekSi, line 31:
```suggestion
	errUnset = ErrorCode(0) // Unset
```

And run `task gen` after that

AlekSi (changes requested):
Please also do the second point from https://github.com/FerretDB/FerretDB/pull/535#pullrequestreview-953855776

internal/handlers/common/error.go
seeforschauer: @fenogentov why didn't you commit that change? for stringer comments matter

fenogentov: > @fenogentov why didn't you commit that change? for stringer comments matter

mongo has this error with no name
*fixed for similar behavior

internal/handlers/common/error.go
```diff
@@ -92,12 +92,15 @@ func (e *Error) Unwrap() error {
 
 // Document returns wire protocol error document.
 func (e *Error) Document() *types.Document {
-	return types.MustNewDocument(
+	d := types.MustNewDocument(
```
seeforschauer, line 95: `	d := must.NotFail(types.NewDocument(`

internal/handlers/pg/msg_getparameter.go
```diff
@@ -25,25 +26,82 @@ import (
...
+		"acceptApiVersion2", false,
+		"authSchemaVersion", int32(5),
+		"quiet", false,
+		"ok", float64(1),
+	)
 
 	var reply wire.OpMsg
+	var errNoFound error
+	var resDoc *types.Document
+	if getPrm == "*" {
```
seeforschauer, line 50: let's create fully meaningful variables and name it `getParam` for example

internal/handlers/pg/msg_getparameter.go
```diff
@@ -25,25 +26,82 @@ import (
...
+			return nil, nil, err
+		}
+	}
+
+	if doc.Len() < 1 {
+		err := doc.Set("ok", float64(0))
+		if err != nil {
+			return nil, nil, err
+		}
+		errNoFound = common.NewErrorMsg(common.ErrorCode(0), "no option found to get")
```
seeforschauer, line 98: here you can make an early return and the other part of code will become more readable

internal/handlers/pg/msg_getparameter.go
fenogentov: ok

internal/handlers/pg/msg_getparameter.go
fenogentov: fix

internal/handlers/common/error.go
fenogentov: overlooked. fix

internal/handlers/pg/msg_getparameter.go
```diff
@@ -25,25 +26,82 @@ import (
...
+	)
 
 	var reply wire.OpMsg
+	var errNoFound error
+	var resDoc *types.Document
+	if getParameter == "*" {
+		resDoc = resDB
+	} else {
+		resDoc, errNoFound, err = selectParam(cmd, resDB)
+		if err != nil {
```
seeforschauer, line 54: I would suggest here:
```
	resDoc = resDB
	if getParameter != "*" {
	  resDoc, errNoFound, err = selectParam(cmd, resDB)
	  if err != nil {
```

integration/commands_administration_test.go
```diff
@@ -185,18 +185,69 @@ func TestCommandsAdministrationGetParameter(t *testing.T) {
...
+			},
+			unexpected: []string{"quiet_other"},
+		},
+		"EmptyParameters": {
+			command: bson.D{{"getParameter", 1}},
+			err:     &mongo.CommandError{Code: 0, Name: "", Message: `no option found to get`},
+		},
+		"OnlyNonexistentParameters": {
+			command: bson.D{{"getParameter", 1}, {"quiet_other", 1}},
+			err:     &mongo.CommandError{Code: 0, Name: "", Message: `no option found to get`},
```
AlekSi, line 224:
```suggestion
			err:     &mongo.CommandError{Message: `no option found to get`},
```

Zero values are good enough

integration/commands_administration_test.go
```diff
@@ -185,18 +185,69 @@ func TestCommandsAdministrationGetParameter(t *testing.T) {
...
-	err := collection.Database().RunCommand(ctx, bson.D{{"getParameter", "*"}}).Decode(&actual)
-	require.NoError(t, err)
+	for name, tc := range map[string]struct {
+		command    bson.D
+		expected   map[string]any
+		unexpected []string
+		err        *mongo.CommandError
+	}{
+		"AllParameters": {
+			command: bson.D{{"getParameter", "*"}},
```
AlekSi, line 195: Let's add a test for `bson.D{{"getParameter", "*"}, {"quiet", 1}, {"quiet_other", 1}}`

integration/commands_administration_test.go
```diff
@@ -185,18 +185,69 @@ func TestCommandsAdministrationGetParameter(t *testing.T) {
...
+			expected: map[string]any{
+				"acceptApiVersion2": false,
+				"authSchemaVersion": int32(5),
+				"quiet":             false,
+				"ok":                float64(1),
+			},
+		},
+		"ExistingParameters": {
+			command: bson.D{{"getParameter", 1}, {"quiet", 1}},
```
AlekSi, line 204: What would happen for `{"getParameter", 0}`? `{"getParameter", NULL}`? Should we add tests for those cases?

internal/handlers/pg/msg_getparameter.go
```diff
@@ -25,25 +26,80 @@ import (
 
 // MsgGetParameter OpMsg used to get parameter.
 func (h *Handler) MsgGetParameter(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
-	document, err := msg.Document()
+	cmd, err := msg.Document()
```
AlekSi, line 29: All handlers has `document, err := msg.Document()` as the first line. Let's keep it the same for consistency

internal/handlers/pg/msg_getparameter.go
```diff
@@ -25,25 +26,80 @@ import (
...
 	if err != nil {
 		return nil, lazyerrors.Error(err)
 	}
 
-	// TODO https://github.com/FerretDB/FerretDB/issues/449
+	getParameter, err := cmd.Get("getParameter")
```
AlekSi, line 34: That's typically written as `command := document.Command()`

internal/handlers/pg/msg_getparameter.go
```diff
@@ -25,25 +26,80 @@ import (
...
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
 
-	common.Ignored(document, h.l, "comment")
+	// SELECT * FROM 'admin'
```
AlekSi, line 39: What does this mean?

internal/handlers/pg/msg_getparameter.go
```diff
@@ -25,25 +26,80 @@ import (
...
 		return nil, lazyerrors.Error(err)
 	}
 
-	// TODO https://github.com/FerretDB/FerretDB/issues/449
+	getParameter, err := cmd.Get("getParameter")
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
 
-	common.Ignored(document, h.l, "comment")
```
AlekSi, line 35: We should keep that line

internal/handlers/pg/msg_getparameter.go
```diff
@@ -25,25 +26,80 @@ import (
...
-	return &reply, nil
+	comment, err := cmd.Get("comment")
+	if err == nil {
+		common.Ignored(cmd, h.l, fmt.Sprint(comment))
+	}
+
+	return &reply, errNoFound
+}
+
+// selectParam is makes a selection of the requested parameters received from database 'admin'.
```
AlekSi, line 72: Well, that's not what function does. We don't have that database

internal/handlers/pg/msg_getparameter.go
```diff
@@ -25,25 +26,80 @@ import (
...
+func selectParam(command, resDB *types.Document) (doc *types.Document, errNoFound error, err error) {
```
AlekSi, line 73: TBH, it is hard to tell what that function does with parameters, two returned errors, etc.

After all, all we need to do is to get values from fixed document: either one or all of them. It should not be that complicated

internal/handlers/pg/msg_getparameter.go
```diff
@@ -25,25 +26,80 @@ import (
...
+		Documents: []*types.Document{resDoc},
 	})
 	if err != nil {
 		return nil, lazyerrors.Error(err)
 	}
 
-	return &reply, nil
+	comment, err := cmd.Get("comment")
+	if err == nil {
+		common.Ignored(cmd, h.l, fmt.Sprint(comment))
```
AlekSi, line 66: There is not reason to make it that complicated, with `if`, error handling, etc. The old line of code is good

AlekSi (changes requested):

integration/commands_administration_test.go
fenogentov: fix

integration/commands_administration_test.go
fenogentov: fix

integration/commands_administration_test.go
fenogentov: fix

internal/handlers/pg/msg_getparameter.go
fenogentov: fix

internal/handlers/pg/msg_getparameter.go
fenogentov: that is, we ignore the comment from command?

internal/handlers/pg/msg_getparameter.go
fenogentov: fix

seeforschauer (changes requested):
@fenogentov 
would you be so kind to set the PR status to Draft if the work is in progress? 
Draft status can be changed under the reviewer's list.
Thank you.

internal/handlers/pg/msg_getparameter.go
fenogentov: `command := document.Command()` this is getting the command name
and in this case I get the value of the command

seeforschauer (changes requested):
LGTM, could you please resolve conflicts?

integration/commands_administration_test.go
seeforschauer: hmmm, looks like there are some more similar lines of code, could you please fix them also?

seeforschauer (changes requested):

integration/commands_administration_test.go
fenogentov: all tests are different.
can't find the same

seeforschauer: I mean, you wrote that it's fixed and it looks like you didn't push.  It is actually still in the code.

seeforschauer (dismissed):
Thank you, @fenogentov!

internal/handlers/pg/msg_getparameter.go
AlekSi: That's true, but then we should use `document.Get(command)`, not `command.Get("getParameter")`, for consistencty

AlekSi (changes requested):
Please remove `integration/integration.txt` from the PR, it should not be commited

internal/handlers/pg/msg_getparameter.go
AlekSi: We ignore comment either way, aren't we? That's what `common.Ignored` does

seeforschauer:
@fenogentov would you please resolve conflicts?

AlekSi (approved):

Issue #258, opened 2022-01-13
Support `$rand` query operator


AlekSi:
https://www.mongodb.com/docs/manual/reference/operator/query/rand/

It is not easy.

DoodgeMatvey:
I would like to work on this issue

seeforschauer:
I'll take it.

AlekSi:
It looks so simple, and yet it requires much more features in FerretDB than we have now.

PR #540, opened 2022-04-27
Support `$rand` query operator

ribaraka:
Refs #258.

$rand operator has been implemented for `update` command. 
For the `find` command, the `$expr` operator must first be implemented.


internal/handlers/common/params.go
```diff
@@ -57,6 +57,25 @@ func GetOptionalParam[T types.Type](doc *types.Document, key string, defaultValu
 	return res, nil
 }
 
+// GetCompositeParam returns doc's value of composite data type for key or protocol error for missing or invalid parameter.
+func GetCompositeParam(doc *types.Document, key string) (zero any, err error) {
```
w84thesun, line 61: Let's not use named result parameters as we don't really need them.

internal/handlers/common/params.go
```diff
@@ -57,6 +57,25 @@ func GetOptionalParam[T types.Type](doc *types.Document, key string, defaultValu
...
+	v, err := doc.Get(key)
+	if err != nil {
+		msg := fmt.Sprintf("required parameter %q is missing", key)
+		return zero, NewErrorMsg(ErrBadValue, msg)
```
w84thesun, line 65:
```suggestion
		return nil, NewErrorMsg(ErrBadValue, fmt.Sprintf("required parameter %q is missing", key))
```

internal/handlers/common/params.go
```diff
@@ -57,6 +57,25 @@ func GetOptionalParam[T types.Type](doc *types.Document, key string, defaultValu
...
+	}
+
+	switch v := v.(type) {
+	case *types.Document:
+		return v, nil
+	case *types.Array:
+		return v, nil
+	default:
+		msg := fmt.Sprintf("composite parameter %q has type %T", key, v)
+		return zero, NewErrorMsg(ErrBadValue, msg)
```
w84thesun, line 75:
```suggestion
		return nil, NewErrorMsg(ErrBadValue, fmt.Sprintf("composite parameter %q has type %T", key, v))
```

internal/handlers/common/params.go
ribaraka: fixed.

internal/handlers/common/params.go
ribaraka: fixed

internal/handlers/common/params.go
ribaraka: fixed

w84thesun (approved):

internal/handlers/common/update.go
```diff
@@ -16,16 +16,34 @@ package common
...
 )
 
 // UpdateDocument updates the given document with a series of update operators.
-func UpdateDocument(doc, update *types.Document) error {
-	for _, updateOp := range update.Keys() {
-		updateV := must.NotFail(update.Get(updateOp))
+func UpdateDocument(doc *types.Document, update any) error {
+	var updateDoc *types.Document
+	var isUpdateArray bool
+	for {
```
seeforschauer, line 30: Sorry what for here is an endless `for`? `update` is not changing.

seeforschauer (changes requested):
looks like endless for is extra

internal/handlers/common/update.go
ribaraka: Yep, I overdid it just a bit. :smile: 
Good catch. Thanks! Fixed

seeforschauer (changes requested):
Thank you!
can you please add tests if one wants to update something like
`update := bson.A{bson.D{{"$set", 
bson.D{{"foo", bson.D{{"bar": bson.D{{"baz", bson.M{"$rand": bson.D{}} }} }} }} } }}`

ribaraka:
> Thank you! can you please add tests if one wants to update something like `update := bson.A{bson.D{{"$set", bson.D{{"foo", bson.D{{"bar": bson.D{{"baz", bson.M{"$rand": bson.D{}} }} }} }} } }}`

This is the second time I have met the need for `deepEqual` in a document :smile:

seeforschauer:
> > Thank you! can you please add tests if one wants to update something like `update := bson.A{bson.D{{"$set", bson.D{{"foo", bson.D{{"bar": bson.D{{"baz", bson.M{"$rand": bson.D{}} }} }} }} } }}`
> 
> This is the second time I have met the need for `deepEqual` in a document smile Maybe this time is not exactly the same case as the first one, but something like that might help.
> 
> Added the test, it's failing. Working on fixing, current implementation is not flexible.

Use of `reflect.Deep` is unlikely, preferably write a recursive function.

AlekSi:
This ticket was specifically about `$rand` query operator, not about the aggregation pipeline operator. They are very separate.

ribaraka:
> This ticket was specifically about `$rand` query operator, not about the aggregation pipeline operator. They are very separate.

This comment relates to me? Bcs honestly, I do not understand what do you mean. I follow this https://www.mongodb.com/docs/manual/reference/operator/query/rand/

ribaraka:
> I guess there is no way to implement this issue now.
> 
> For queries, `$rand` can be used only inside `$expr`, and `$expr` requires a full aggregation pipeline support that we don't want to implement now (that's item 21 on our [roadmap](https://github.com/orgs/FerretDB/projects/2/views/1), and we are on item 2 now).
> 
> When used in updates, the array there is the same aggregation pipeline.
> 
> Sorry for confusion.

No worries. Thanks for answers :+1:

Issue #164, opened 2021-12-14
Support `findAndModify`


AlekSi:
Improve support for `findAndModify` command with the following arguments:
* `findAndModify` (collection name)
* `query`
* `sort`
* `remove`
* `update`
* `new`
* `upsert`

Ignore `bypassDocumentValidation`, `writeConcern`, `maxTimeMS`, `collation`, `hint`, `comment` arguments.
Fail if `arrayFilters`, `let`, `fields` argument is present.

We have a basic implementation already, but it misses support for many arguments and does not have any tests. We should add them.

PR #544, opened 2022-04-28
Add tests for `sort` and `find` parameters type

w84thesun:
Refs #164.

### DIFF
--- a/integration/query_test.go
+++ b/integration/query_test.go
@@ -1,3 +1,17 @@
+// Copyright 2021 FerretDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package integration
 
 import (
--- a/integration/query_test.go
+++ b/integration/query_test.go
@@ -260,11 +260,8 @@ func TestQueryBadSortType(t *testing.T) {
 
 			var actual bson.D
 			err := collection.Database().RunCommand(ctx, tc.command).Decode(&actual)
-			if tc.err != nil {
-				AssertEqualError(t, *tc.err, err)
-				return
-			}
-			require.NoError(t, err)
+			require.Error(t, err)
+			AssertEqualError(t, *tc.err, err)
 		})
 	}
 }
--- a/integration/query_test.go
+++ b/integration/query_test.go
@@ -70,7 +70,7 @@ func TestQueryCount(t *testing.T) {
 	}
 }
 
-func TestQueryFindType(t *testing.T) {
+func TestQueryBadFindType(t *testing.T) {
 	t.Parallel()
 	providers := []shareddata.Provider{shareddata.Scalars, shareddata.Composites}
 	ctx, collection := setup(t, providers...)
@@ -80,7 +80,7 @@ func TestQueryFindType(t *testing.T) {
 		response bson.D
 		err      *mongo.CommandError
 	}{
-		"BadFindTypeDocument": {
+		"Document": {
 			command: bson.D{
 				{"find", bson.D{}},
 				{"projection", bson.D{{"value", "some"}}},
@@ -91,7 +91,7 @@ func TestQueryFindType(t *testing.T) {
 				Message: "collection name has invalid type object",
 			},
 		},
-		"BadFindTypeArray": {
+		"Array": {
 			command: bson.D{
 				{"find", primitive.A{}},
 				{"projection", bson.D{{"value", "some"}}},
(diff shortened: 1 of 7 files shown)


internal/handlers/common/error.go
```diff
@@ -33,6 +33,7 @@ const (
 
 	ErrBadValue          = ErrorCode(2)     // BadValue
 	ErrFailedToParse     = ErrorCode(9)     // FailedToParse
+	ErrTypeMismatch      = ErrorCode(14)    // TypeMismatch
```
w84thesun, line 36: This error type also would be used in `findAndModify`.

seeforschauer (dismissed):

integration/query_test.go
```diff
@@ -67,3 +68,220 @@ func TestQueryCount(t *testing.T) {
...
+		"BadSortType": {
+			command: bson.D{
+				{"find", collection.Name()},
+				{"projection", bson.D{{"value", "some"}}},
+				{"sort", "123"},
+			},
+			err: &mongo.CommandError{
+				Code:    14,
+				Name:    "TypeMismatch",
+				Message: "Expected field sortto be of type object",
```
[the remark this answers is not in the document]
w84thesun: Yeah...

integration/query_test.go
```diff
@@ -67,3 +68,220 @@ func TestQueryCount(t *testing.T) {
...
+	}
+}
+
+func TestQuerySort(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setup(t, shareddata.Scalars, shareddata.Composites)
+
+	for name, tc := range map[string]struct {
+		command  bson.D
+		response bson.D
```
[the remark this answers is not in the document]
w84thesun: No, it's copy-paste. Removed it from both tests.

integration/query_test.go
```diff
@@ -67,3 +68,220 @@ func TestQueryCount(t *testing.T) {
...
 	}
 }
+
+func TestQueryBadFindType(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setup(t, shareddata.Scalars, shareddata.Composites)
+
+	for name, tc := range map[string]struct {
+		command  bson.D
+		response bson.D
```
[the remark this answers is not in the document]
w84thesun: Fixed.

integration/query_test.go
```diff
@@ -67,3 +68,206 @@ func TestQueryCount(t *testing.T) {
...
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			var actual bson.D
+			err := collection.Database().RunCommand(ctx, tc.command).Decode(&actual)
+			if tc.err != nil {
+				AssertEqualError(t, *tc.err, err)
+				return
+			}
+			require.NoError(t, err)
```
AlekSi, line 234:
```suggestion
			require.Error(t, err)
			AssertEqualError(t, *tc.err, err)
```

Since we always have error in that case

integration/query_test.go
```diff
@@ -67,3 +68,206 @@ func TestQueryCount(t *testing.T) {
...
+		})
+	}
+}
+
+func TestQuerySort(t *testing.T) {
```
AlekSi, line 239:
```suggestion
func TestQueryBadSortType(t *testing.T) {
```

Since we will have a separate test for sorting that could be named `TestQuerySort`

integration/query_test.go
```diff
@@ -67,3 +68,206 @@ func TestQueryCount(t *testing.T) {
...
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			var actual bson.D
+			err := collection.Database().RunCommand(ctx, tc.command).Decode(&actual)
+			if tc.err != nil {
+				AssertEqualError(t, *tc.err, err)
+				return
+			}
+			require.NoError(t, err)
```
AlekSi, line 270: The same as above

internal/handlers/common/typecode.go
```diff
@@ -138,3 +138,37 @@ func init() {
...
+	case *types.Document:
+		return "object"
+	case *types.Array:
+		return "array"
+	case float64:
+		return "double"
+	case string:
+		return "string"
+	case types.Binary:
+		return "binData"
```
AlekSi, line 154: I wonder if we should use `typeCodeBinData.String()` instead to ensure that those strings are the same everywhere. The same in `hasSameTypeElements`. That is, assuming that those aliases _are_ the same

AlekSi (commented):
LGTM with one question/suggestion and minor comments

internal/handlers/common/typecode.go
w84thesun: Looks like MongoDB these aliases are the same. I have some ideas on how to simplify `hasSameTypeElements` but I suggest doing it in a separate PR.

AlekSi: Ok, but can we simplify that new `AliasFromType` function? Because it looks like it started as a copy of `hasSameTypeElements` that was changed, and we went from two copies of "binData" string (one is generated by stringer, one is in `hasSameTypeElements`) to three. That breaks the Scout rule: "Always leave the code better than you found it". So no, we should replace "binData" with `typeCodeBinData.String()` in that PR

w84thesun: Updated

internal/handlers/common/filter.go
```diff
@@ -53,7 +53,7 @@ func filterDocumentPair(doc *types.Document, filterKey string, filterValue any)
 	if strings.ContainsRune(filterKey, '.') {
 		// {field1./.../.fieldN: filterValue}
 		path := strings.Split(filterKey, ".")
-		// we pass the path without the last key because we want {fieldN: *someValue*}, not just *someValue*
+		// we pass the path without the last key because we same {fieldN: *someValue*}, not just *someValue*
```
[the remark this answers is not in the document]
w84thesun: Refactored by mistake…

AlekSi (approved):

PR #546, opened 2022-05-01
Add more tests for `updateOne`

GinGin3203:


seeforschauer (changes requested):
> @seeforschauer yeah, but please pay attention that only FerretDB tests are failing and Mongo tests are passing, so there is a discrepancy between how Ferret and Mongo behave That's what I was trying to show with this PR initially

@GinGin3203 yes, and this exactly means that tests are written correctly and it is necessary to change how FerretDB behaves to make its response compatible with the MongoDB.

AlekSi (commented):
Let's fix #571 in this PR.

(@GinGin3203 we leave PR review comments on that PR just to remove it from our https://github.com/pulls/review-requested dashboards until re-review is needed 😃)

GinGin3203:
Moved to https://github.com/FerretDB/FerretDB/pull/609

PR #549, opened 2022-05-02
Add tests for `RemoveByPath`

### LINKED ISSUE #479: Add tests for `RemoveByPath`
It seems that there are cases when `RemoveByPath` does not work correctly. Add tests, fix problems.

Also, add it to `Array` and `CompositeTypeInterface`.

seeforschauer:
Closes #479.


seeforschauer:
A test for `RemoveByPath` is in the `testutil` module to avoid import cycle.

AlekSi:
> A test for RemoveByPath is in the testutil module to avoid import cycle.

Let's move it to `types` and use `assert.Equal` as other tests in that package do

seeforschauer:
updated with `assert.Equal`.

internal/types/document.go
```diff
@@ -321,39 +321,42 @@ func (d *Document) Remove(key string) {
 
 // RemoveByPath removes document by path, doing nothing if the key does not exist.
 func (d *Document) RemoveByPath(keys ...string) {
+	removeByPath(d, keys...)
+}
+
+func removeByPath(v any, keys ...string) {
```
w84thesun, line 327: Should we add comment to this function?

seeforschauer: it's private, the public one with similar name has.

AlekSi (approved):

PR #550, opened 2022-05-03
Add `altMessage` to `AssertEqualError`

w84thesun:


seeforschauer (commented):
As I understood from the call, we should compare the codes of errors, not the `any` of the error messages. 

How this would be used?

integration/helpers.go
```diff
@@ -124,6 +124,29 @@ func AssertEqualError(t testing.TB, expected mongo.CommandError, actual error) b
 	return assert.Equal(t, expected, a)
 }
 
+// AssertEqualErrorMessage asserts that expected error is the same as actual, ignoring the Raw part.
+func AssertEqualErrorMessage(t testing.TB, expected mongo.CommandError, altMessage string, actual error) bool {
```
w84thesun, line 128: I've added the new function just to make diff smaller.

integration/helpers.go
```diff
@@ -124,6 +124,29 @@ func AssertEqualError(t testing.TB, expected mongo.CommandError, actual error) b
...
+	t.Helper()
+
+	a, ok := actual.(mongo.CommandError)
+	if !ok {
+		return assert.Equal(t, expected, actual)
```
seeforschauer, line 133: here, it will fail always, isn't it? 
expected is type of `mongo.CommandError` and it's compared to actual which has another type in this if branch.

integration/helpers.go
```diff
@@ -124,6 +124,29 @@ func AssertEqualError(t testing.TB, expected mongo.CommandError, actual error) b
...
+	}
+
+	// raw part might be useful if assertion fails
```
seeforschauer, line 136: If the assertion fails, it's unreachable, isn't it? there is a return on 133 line.Or which one assertion do you mean?
Raw is not used below.

integration/helpers.go
w84thesun: Yes, it should fail here.

AlekSi: In pseudocode:
```go
if !assert.ObjectsAreEqual(expected, a) { // compare with expected
  // compare with expected with altMessage
  expected.Message = altMessage
  return assert.Equal(t, expected, a)
}
```

integration/helpers.go
```diff
@@ -109,7 +109,7 @@ func AssertEqualDocuments(t testing.TB, expected, actual bson.D) bool {
 }
 
 // AssertEqualError asserts that expected error is the same as actual, ignoring the Raw part.
-func AssertEqualError(t testing.TB, expected mongo.CommandError, actual error) bool {
+func AssertEqualError(t testing.TB, expected mongo.CommandError, altMessage string, actual error) bool {
```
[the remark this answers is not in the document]
w84thesun: @AlekSi WDYT?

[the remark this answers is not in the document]
AlekSi: Yeah, let's do two functions: keep `AssertEqualError` as is and add `AssertEqualAltError`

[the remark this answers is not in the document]
AlekSi: Yeah, let's do two functions: keep `AssertEqualError` as is and add `AssertEqualAltError`

AlekSi (changes requested):
Let's add a new function

integration/helpers.go
seeforschauer: If it's not used now, maybe could this code be removed?

integration/helpers.go
seeforschauer: This is covered by code from the lines 141-145, isn't it?

seeforschauer (changes requested):

integration/helpers.go
w84thesun: Why it's not used?

integration/helpers.go
w84thesun: We still need it to cast to mongo.CommandError

seeforschauer: this block can be moved to the 141-145 if entirely, WDYT?

integration/helpers.go
seeforschauer: I just don't see where it's used.. In the error message when it prints when it fails?

seeforschauer (dismissed):

integration/helpers.go
w84thesun: Yeah. When an assertion fails there would be a test output where the Raw part would be shown.

integration/helpers.go
AlekSi: It is also required for case when `actual` is `nil`

integration/helpers.go
```diff
@@ -124,6 +124,29 @@ func AssertEqualError(t testing.TB, expected mongo.CommandError, actual error) b
...
+
+	// raw part might be useful if assertion fails
+	require.Nil(t, expected.Raw)
+	expected.Raw = a.Raw
+
+	if !assert.ObjectsAreEqual(expected, a) { // compare with expected
+		// compare with expected with altMessage
+		expected.Message = altMessage
+		return assert.Equal(t, expected, a)
+	}
```
AlekSi, line 145: Looks like my pseudocode, which just showed that `assert.ObjectsAreEqual` exists, was copied as is, and now this code is more complicated than it could be. For example, we essential have this:

```go
if !equal {
  // ...
}

assertEqual()
```

At the time we call `assertEqual`, we already know that objects are equal.

It all could be simplified to:

```go
if equal(expected, actual) {
  return
}

expected.Message = altMessage
if equal(expected, actual) {
  return
}

// fail with a good message mentioning both messages
```

integration/query_test.go
```diff
@@ -254,7 +254,21 @@ func TestQueryBadSortType(t *testing.T) {
 	for name, tc := range map[string]struct {
 		command bson.D
 		err     *mongo.CommandError
+		message string
```
AlekSi, line 257: Let's call it `altMessage` if that's how we call the assertion parameter

AlekSi (changes requested):

integration/query_test.go
w84thesun: Fixed.

integration/helpers.go
w84thesun: Updated

AlekSi: Updated code now states: "if objects are equal, assert that they are equal"

AlekSi: If we already checked that they are equal, we can just return true and don't make future readers scratch their heads :)

w84thesun: Done.

AlekSi (approved):

PR #560, opened 2022-05-05
Add `RemoveByPath` for `Array` and `CompositeTypeInterface`

seeforschauer:
Refs #479

### DIFF
--- a/internal/types/path.go
+++ b/internal/types/path.go
@@ -19,6 +19,10 @@ import (
 	"strconv"
 )
 
+func RemoveByPath[T CompositeTypeInterface](comp T, keys ...string) {
+	removeByPath(comp, keys...)
+}
+
 // getByPath returns a value by path - a sequence of indexes and keys.
 func getByPath[T CompositeTypeInterface](comp T, path ...string) (any, error) {
 	var next any = comp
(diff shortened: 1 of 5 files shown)


internal/types/path.go
```diff
@@ -19,6 +19,10 @@ import (
 	"strconv"
 )
 
+func RemoveByPath[T CompositeTypeInterface](comp T, keys ...string) {
```
w84thesun, line 22: Please add function comment.

seeforschauer: done

internal/types/path_test.go
```diff
@@ -220,10 +178,63 @@ func TestRemoveByPath(t *testing.T) {
...
+			must.NotFail(NewDocument(
+				"document", "abc",
+				"score", float64(42.13),
+				"age", int32(1000),
+				"foo", deepDoc.DeepCopy(),
+			)),
+			must.NotFail(NewArray("1", "2", "3")),
+		))
+
+		type testCase struct {
```
w84thesun, line 200: We can embed this struct to make it similar to other tests

internal/types/path_test.go
```diff
@@ -220,10 +178,63 @@ func TestRemoveByPath(t *testing.T) {
...
+				expected: src.DeepCopy(),
+			},
+			{
+				name:     "array: index is not number",
+				path:     []string{"abcd"},
+				expected: src.DeepCopy(),
+			},
+		} {
+			tc := tc
+			t.Run(fmt.Sprint(tc.path), func(t *testing.T) {
```
w84thesun, line 229: We can use here `name` instead of `fmt.Sprintf(tc.path)` as test name

internal/types/path_test.go
seeforschauer: changed

internal/types/path_test.go
seeforschauer: nice catch! done for both cases.

AlekSi (approved):

PR #563, opened 2022-05-05
Support `$comment` query operator

### LINKED ISSUE #168: Support `$comment` query operator
Put `$comment`'s content into generated SQL query after the first statement. For example, `db.collection.find({$comment: "test"})` should generate `SELECT /* test */ * FROM ...`. See https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-COMMENTS.

To prevent SQL injection, we should replace occurrences of `/*` and `*/` in the comment with `/ *` and `* /` (add extra space).

ribaraka:
Closes #168.

### DIFF
--- a/integration/basic_test.go
+++ b/integration/basic_test.go
@@ -93,3 +93,22 @@ func TestInsertFind(t *testing.T) {
 		})
 	}
 }
+
+func TestFindCommentMethod(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setup(t, shareddata.Scalars)
+
+	var doc bson.D
+	opts := options.FindOne().SetComment("some test text")
+	err := collection.FindOne(ctx, bson.D{{"_id", "string"}}, opts).Decode(&doc)
+	require.NoError(t, err)
+}
+
+func TestFindCommentQuery(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setup(t, shareddata.Scalars)
+
+	var doc bson.D
+	err := collection.FindOne(ctx, bson.M{"_id": "string", "$comment": "some test text"}).Decode(&doc)
+	require.NoError(t, err)
+}
--- a/integration/basic_test.go
+++ b/integration/basic_test.go
@@ -99,7 +99,7 @@ func TestFindCommentMethod(t *testing.T) {
 	ctx, collection := setup(t, shareddata.Scalars)
 
 	var doc bson.D
-	opts := options.FindOne().SetComment("some test text")
+	opts := options.FindOne().SetComment("*/ 1; DROP DATABASE postgres --")
 	err := collection.FindOne(ctx, bson.D{{"_id", "string"}}, opts).Decode(&doc)
 	require.NoError(t, err)
 }
@@ -109,6 +109,6 @@ func TestFindCommentQuery(t *testing.T) {
 	ctx, collection := setup(t, shareddata.Scalars)
 
 	var doc bson.D
-	err := collection.FindOne(ctx, bson.M{"_id": "string", "$comment": "some test text"}).Decode(&doc)
+	err := collection.FindOne(ctx, bson.M{"_id": "string", "$comment": "*/ 1; DROP DATABASE postgres --"}).Decode(&doc)
 	require.NoError(t, err)
 }
--- a/integration/basic_test.go
+++ b/integration/basic_test.go
@@ -97,18 +97,28 @@ func TestInsertFind(t *testing.T) {
 func TestFindCommentMethod(t *testing.T) {
 	t.Parallel()
 	ctx, collection := setup(t, shareddata.Scalars)
+	name := collection.Name()
+	databaseNames, err := collection.Database().Client().ListDatabaseNames(ctx, bson.D{})
+	require.NoError(t, err)
+	comment := "*/ 1; DROP SCHEMA " + name + " CASCADE -- "
 
 	var doc bson.D
-	opts := options.FindOne().SetComment("*/ 1; DROP DATABASE postgres --")
-	err := collection.FindOne(ctx, bson.D{{"_id", "string"}}, opts).Decode(&doc)
+	opts := options.FindOne().SetComment(comment)
+	err = collection.FindOne(ctx, bson.D{{"_id", "string"}}, opts).Decode(&doc)
 	require.NoError(t, err)
+	assert.Contains(t, databaseNames, name)
 }
 
 func TestFindCommentQuery(t *testing.T) {
 	t.Parallel()
 	ctx, collection := setup(t, shareddata.Scalars)
+	name := collection.Name()
+	databaseNames, err := collection.Database().Client().ListDatabaseNames(ctx, bson.D{})
+	require.NoError(t, err)
+	comment := "*/ 1; DROP SCHEMA " + name + " CASCADE -- "
 
 	var doc bson.D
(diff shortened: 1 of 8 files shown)


internal/handlers/pg/fetch.go
```diff
@@ -73,3 +73,37 @@ func nextRow(rows pgx.Rows) (*types.Document, error) {
 
 	return doc.(*types.Document), nil
 }
+
+func (h *Handler) protoFetch(ctx context.Context, param fetchParam) ([]*types.Document, error) {
```
seeforschauer, line 77: could you please share with us your thought on why the function name is like this?

internal/handlers/pg/fetch.go
```diff
@@ -73,3 +73,37 @@ func nextRow(rows pgx.Rows) (*types.Document, error) {
...
+	var sql string
+	if param.comment != "" {
+		param.comment = fmt.Sprintf("/* %s */", param.comment)
+		sql = fmt.Sprintf(`SELECT %s _jsonb FROM %s`, param.comment, pgx.Identifier{param.db, param.collection}.Sanitize())
```
seeforschauer, line 81: why not to put the row above inline in the `Sprintf`?

internal/handlers/pg/msg_find.go
```diff
@@ -95,7 +76,31 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
 		}
 	}
 
-	fetchedDocs, err := h.fetch(ctx, db, collection)
```
seeforschauer, line 98: you've created a new `fetch` function abandoning the old one. Is there any reason for this?

seeforschauer (changes requested):

internal/handlers/pg/fetch.go
```diff
@@ -73,3 +73,37 @@ func nextRow(rows pgx.Rows) (*types.Document, error) {
 
 	return doc.(*types.Document), nil
 }
+
+func (h *Handler) protoFetch(ctx context.Context, param fetchParam) ([]*types.Document, error) {
+	var sql string
+	if param.comment != "" {
+		param.comment = fmt.Sprintf("/* %s */", param.comment)
```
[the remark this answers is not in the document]
AlekSi: From ticket description:

> To prevent SQL injection, we should replace occurrences of /* and */ in the comment with / * and * / (add extra space).

See also [there](https://xkcd.com/327/)

internal/handlers/pg/msg_find.go
```diff
@@ -95,7 +76,31 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+		return nil, common.NewErrorMsg(
+			common.ErrBadValue,
+			fmt.Sprintf("collection name has invalid type %s", common.AliasFromType(collectionParam)),
+		)
+	}
+	// comment set through options.FindOne().SetComment() method
+	if fp.comment, err = common.GetOptionalParam(document, "comment", fp.comment); err != nil {
+		return nil, err
+	}
+	// comment in query, e.g. db.collection.find({$comment: "test"})
```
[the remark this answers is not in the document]
ribaraka: It has been added more to catch a comment at all.

internal/handlers/pg/fetch.go
ribaraka: it is a good suggestion :)
but in this case I'm trying to avoid long lines and just explicitly expressed the operation (line 80)

internal/handlers/pg/msg_find.go
ribaraka: a comment parameter appeared and I created a new prototype function as a suggestion to replace the old one.. I should have mentioned it in the description of the PR. My bad, sorry

internal/handlers/pg/fetch.go
seeforschauer: yeah it's more visible though

internal/handlers/pg/fetch.go
seeforschauer: Hi, `protoFetch` is used in msg find and in tests and nowhere else.  
Why not rename `fetch` and add all code to this?

ribaraka: actually I'm going to leave `fetch` func.  `protoFetch` is a temporarily prototype as a future view of `fetch` func. I'm not sure if you all okay with this new implementation so I just replaced this in one in one place (for simplicity) and waiting for you approval and command to apply it in every occurrences.

internal/handlers/pg/fetch.go
```diff
@@ -73,3 +74,40 @@ func nextRow(rows pgx.Rows) (*types.Document, error) {
...
+		if err != nil {
+			return nil, lazyerrors.Error(err)
+		}
+		res = append(res, doc)
+	}
+
+	return res, nil
+}
+
+type fetchParam struct {
```
AlekSi, line 111: Let's add documentation comment and move it before function that uses it

internal/handlers/pg/fetch.go
```diff
@@ -73,3 +74,40 @@ func nextRow(rows pgx.Rows) (*types.Document, error) {
...
+
+func (h *Handler) protoFetch(ctx context.Context, param fetchParam) ([]*types.Document, error) {
+	var sql string
+	if param.comment != "" {
+		param.comment = strings.ReplaceAll(param.comment, "/*", "/ *")
+		param.comment = strings.ReplaceAll(param.comment, "*/", "* /")
+		param.comment = fmt.Sprintf("/* %s */", param.comment)
+
+		sql = fmt.Sprintf(`SELECT %s _jsonb FROM %s`, param.comment, pgx.Identifier{param.db, param.collection}.Sanitize())
+	} else {
```
AlekSi, line 86: I think that will be easier to read without `else`. In pseudocode:
```go
sql := `SELECT `
if param.comment != "" {
  // handle /*, */
  sql += `/* ` + params.comment + ` */ `
}
sql += `_jsonb FROM ` + pgx.Identifier{param.db, param.collection}.Sanitize()
```

AlekSi (commented):
`protoFetch` looks good, let's replace existing `fetch` with it

internal/handlers/pg/fetch.go
ribaraka: Done.

internal/handlers/pg/fetch.go
ribaraka: Done.

AlekSi (changes requested):

seeforschauer (changes requested):
@ribaraka would you please resolve conflicts

integration/basic_test.go
```diff
@@ -93,3 +93,22 @@ func TestInsertFind(t *testing.T) {
...
 }
+
+func TestFindCommentMethod(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setup(t, shareddata.Scalars)
+
+	var doc bson.D
+	opts := options.FindOne().SetComment("*/ 1; DROP DATABASE postgres --")
+	err := collection.FindOne(ctx, bson.D{{"_id", "string"}}, opts).Decode(&doc)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
ribaraka: the block has been refactored and added check for existing schema.

seeforschauer (approved):

AlekSi (approved):

PR #566, opened 2022-05-07
Add conform CI workflow, improve docs

AlekSi:


CONTRIBUTING.md
```diff
@@ -2,16 +2,33 @@
 
 Thank you for your interest in making FerretDB better!
 
+## Finding something to work on
+
+We are interested in all contributions, big or small, in code or documentation.
+But unless you are fixing a very small issue like a typo,
+we kindly ask you first to [create an issue](https://github.com/FerretDB/FerretDB/issues/new/choose),
+to leave a comment on an existing issue if you want to work on it,
+or to [join our Slack chat](./README.md#community) and leave a message for us there.
```
seeforschauer, line 11: Here an invite link can be given, to as not to send through the doc journey and get instantly what is necessary, WDYT?

AlekSi: I think linking to README section is better for two reasons:
* We will need to update Slack link in one place when it expires;
* We want to tell people about other places where our community lives.

seeforschauer (approved):

rumyantseva (commented):
LGTM! Extra thanks for updating the docs. Also, I noticed a tiny typo, will prepare a separate PR.

PR #584, opened 2022-05-10
Enable `gosimple`

AlekSi:


internal/handlers/common/filter.go
```diff
@@ -312,15 +312,18 @@ func filterFieldExpr(doc *types.Document, filterKey string, expr *types.Document
...
+				if found {
+					break
+				}
+
+				switch arrValue := must.NotFail(arr.Get(i)).(type) {
 				case *types.Array:
 					fieldValue, ok := fieldValue.(*types.Array)
 					if ok && matchArrays(fieldValue, arrValue) {
 						found = true
-						break
```
AlekSi, line 323: That was a bug… kinda. This `break` terminates `switch`, not a `for` loop

seeforschauer: ah, see

seeforschauer (approved):

Issue #457, opened 2022-04-08
Improve comparison/sorting


AlekSi:
We should make our comparison and sorting of scalar values, arrays, and documents work the same way as in MongoDB (excluding values we don't support, like NaNs and nested arrays).

* #1449
* #1450
* #1521
* #1522
* #1023

Compat newly discovered issues:
* #1539 
* #1542
* #1569
* #1577 
* #1578
* #1607
* #1637

---

* [x] comparison of documents without reflection;
* [x] comparison of `[]`, `[null]` arrays;
* [x] comparison of arrays with a different element order;
* [x] comparison of arrays with a different element types;
* [x] comparison of documents with missing fields and nulls;
* [x] comparison of documents with different fields order;
* [x] comparison of documents with different fields types;
* [x] ascending and descending sorting:
  * [x] scalars;
  * [x] documents;
  * [x] arrays;
* [x] check that changes are ported to `findAndModify`.

We might have some of that tested already, but we should check that, add missing tests, and fix bugs.

AlekSi:
The last point remains

ribaraka:
Sure!

AlekSi:
I agree. I think let's check scalar types first (both comparing and sorting by them), then documents, then arrays (that are most difficult)

ribaraka:
OK!

seeforschauer:
todo: check support in `findAndModify`

AlekSi:
On hold for now, until we decide about edge cases and validation

PR #589, opened 2022-05-11
Improve comparison of arrays and documents

ribaraka:
Refs #457 

- [x]  comparisons of [], [null] arrays;
- [x]  comparison of arrays with a different element order;
- [x]  comparison of documents with missing fields and nulls;
- [x]  comparison of documents with different fields order;


AlekSi:
This PR does not close that issue because the last point is not done

AlekSi (approved):

PR #596, opened 2022-05-12
Support `$inc` field `update` operator

### LINKED ISSUE #520: Support `$inc` field `update` operator
Support it for all types and edge cases. Support it with dot notation.

w84thesun:
Closes #520.


w84thesun:
> Draft state, tests fail. I assume it is not ready for review

I should set PR reviewers when it's ready?

seeforschauer:
> > Draft state, tests fail. I assume it is not ready for review
> 
> I should set PR reviewers when it's ready?

Agree, bcz when I check pull request review even if the state is draft I see it.

w84thesun:
I tried to implement dot notation support for `$inc` operator. It seems that it require a lot of changes to several structures including Document and Array. Maybe it's better to do it in separate PR? @AlekSi WDYT?

seeforschauer (approved):

internal/handlers/common/error.go
```diff
@@ -157,7 +170,92 @@ func formatBitwiseOperatorErr(err error, operator string, maskValue any) error {
...
+	d := must.NotFail(types.NewDocument(
+		"ok", float64(1),
+		"writeErrors", errs,
+	))
+	return d
+}
+
+// WriteError represents protocol write error.
+type WriteError struct {
+	index int64
```
seeforschauer, line 218: can you please add a comment what does `index` field?

internal/handlers/common/error.go
```diff
@@ -157,7 +170,92 @@ func formatBitwiseOperatorErr(err error, operator string, maskValue any) error {
...
+	code  ErrorCode
+	err   error
+}
+
+// Document returns wire protocol error document.
+func (we WriteError) Document() *types.Document {
+	d := must.NotFail(types.NewDocument(
+		"index", we.index,
```
seeforschauer, line 226: it seems like we can have `index` different from the index of the element of the array, and that is the question for like 218.

integration/update_test.go
```diff
@@ -72,3 +72,199 @@ func TestUpdateUpsert(t *testing.T) {
...
+			result: bson.D{{"_id", "int32"}, {"value", int64(43)}},
+		},
+
+		"FieldNotExist": {
+			filter: bson.D{{"_id", "int32"}},
+			update: bson.D{{"$inc", bson.D{{"foo", int32(1)}}}},
+			result: bson.D{{"_id", "int32"}, {"value", int32(42)}, {"foo", int32(1)}},
+		},
+
+		//"DotNotationInt": {
```
seeforschauer, line 232: may we remove tests as they are not used? 
maybe it's better to create a list of what is necessary to implement for `$inc` in a separate `md` document (and suggest tests there)?

internal/handlers/common/params.go
```diff
@@ -183,3 +185,43 @@ func parseTypeCode(alias string) (typeCode, error) {
...
+			return v2 + float64(v1), nil
+		case int32:
+			return v1 + int64(v2), nil
+		case int64:
+			return v1 + v2, nil
+		default:
+			return nil, errBadRightOperandType
+		}
+	default:
+		return nil, errBadLeftOperandType
```
seeforschauer, line 225: can it be rewritten in generics?

internal/handlers/common/error.go
w84thesun: Sorry, 218?

internal/handlers/common/params.go
w84thesun: I consider to left it as is right now.

internal/handlers/common/error.go
w84thesun: Added

integration/update_test.go
w84thesun: Removed commented tests.

internal/handlers/common/error.go
seeforschauer: 218 loc (line of code) where:

> // WriteError represents protocol write error.
> type WriteError struct {
>  index int64

internal/handlers/common/error.go
seeforschauer: ok, thank you, and would you be so kind to make a reference where to find the index and corresponding operation description? Thank you.

internal/handlers/common/error.go
seeforschauer: that's was just a note for https://github.com/FerretDB/FerretDB/pull/596/files#r874474258

internal/handlers/common/params.go
seeforschauer: ok

internal/handlers/common/error.go
w84thesun: I think I can remove this field as it's not used right now.

seeforschauer (dismissed):

internal/handlers/common/error.go
```diff
@@ -108,11 +108,19 @@ func (e *Error) Document() *types.Document {
 	return d
 }
 
+// ProtoErr represents protocol error type.
+type ProtoErr interface {
```
AlekSi, line 112: Why do we define a common interface between two implementations?

internal/handlers/common/error.go
```diff
@@ -108,11 +108,19 @@ func (e *Error) Document() *types.Document {
 	return d
 }
 
+// ProtoErr represents protocol error type.
```
AlekSi, line 111: It is not clear how that is different from `Error` type.

internal/handlers/common/error.go
```diff
@@ -108,11 +108,19 @@ func (e *Error) Document() *types.Document {
...
+type ProtoErr interface {
+	Error() string
+	Code() ErrorCode
+	Document() *types.Document
```
AlekSi, line 115: Methods should be documented

internal/handlers/common/error.go
```diff
@@ -108,11 +108,19 @@ func (e *Error) Document() *types.Document {
...
+}
+
 // ProtocolError converts any error to wire protocol error.
 //
 // Nil panics, *Error (possibly wrapped) is returned unwrapped with true,
+// *WriteErrors is returned unwrapped with true,
```
AlekSi, line 121:
```suggestion
// Nil panics, *Error or *WriteError (possibly wrapped) is returned unwrapped with true,
```

internal/handlers/common/error.go
```diff
@@ -158,7 +171,89 @@ func formatBitwiseOperatorErr(err error, operator string, maskValue any) error {
...
+func (we *WriteErrors) Error() string {
+	var err string
+	for _, e := range *we {
+		err += e.Error() + ","
+	}
+
+	return err
+}
+
+// Code returns error code.
```
AlekSi, line 187: Once methods are documented on the interface level, that comment could be just `Code implements XXX interface.`

internal/handlers/common/error.go
```diff
@@ -158,7 +171,89 @@ func formatBitwiseOperatorErr(err error, operator string, maskValue any) error {
...
+// Unwrap implements standard error unwrapping interface.
+func (we *WriteErrors) Unwrap() error {
+	for _, e := range *we {
+		return &e
+	}
+	return nil
+}
+
+// Document returns wire protocol error document.
+func (we WriteErrors) Document() *types.Document {
```
AlekSi, line 204: Why other methods are implemented on the pointer, but that one is one the value?

internal/handlers/common/error.go
```diff
@@ -158,7 +171,89 @@ func formatBitwiseOperatorErr(err error, operator string, maskValue any) error {
...
+	}
+
+	d := must.NotFail(types.NewDocument(
+		"ok", float64(1),
+		"writeErrors", errs,
+	))
+	return d
+}
+
+// WriteError represents protocol write error.
```
AlekSi, line 217: So is ProtoErr. We need better descriptions

internal/handlers/common/error.go
```diff
@@ -158,7 +171,89 @@ func formatBitwiseOperatorErr(err error, operator string, maskValue any) error {
...
+type WriteError struct {
+	code ErrorCode
+	err  error
+}
+
+// Document returns wire protocol error document.
+func (we WriteError) Document() *types.Document {
```
AlekSi, line 224: Any particular reason why methods order is different?

internal/handlers/common/params.go
```diff
@@ -213,3 +215,43 @@ func parseTypeCode(alias string) (typeCode, error) {
 
 	return code, nil
 }
+
+func addNumbers(v1, v2 any) (any, error) {
```
AlekSi, line 219: Function that accepts anything and returns anything _must_ be documented

internal/handlers/common/update.go
```diff
@@ -41,6 +42,66 @@ func UpdateDocument(doc, update *types.Document) error {
 				}
 			}
 
+		case "$inc":
+			incDoc, err := AssertType[*types.Document](updateV)
```
AlekSi, line 46: `AssertType` documentation says:
> If a custom error is needed, use a normal Go type assertion instead [...]

internal/handlers/common/params.go
```diff
@@ -213,3 +215,43 @@ func parseTypeCode(alias string) (typeCode, error) {
...
+	switch v1 := v1.(type) {
+	case float64:
+		switch v2 := v2.(type) {
+		case float64:
+			return v1 + v2, nil
```
AlekSi, line 224: That code would happily sum NaN and infinity, but there are no tests for that

AlekSi (changes requested):

internal/handlers/common/params.go
w84thesun: I've added a test for that case.

internal/handlers/common/update.go
w84thesun: Fixed

internal/handlers/common/error.go
w84thesun: In fact, `Error` type should be `CommandError`. 
Whereas `ProtoErr` represents a common protocol error type (it could be CommandError or WriteErrors).

internal/handlers/common/error.go
w84thesun: Added comments

integration/update_test.go
```diff
@@ -72,3 +73,186 @@ func TestUpdateUpsert(t *testing.T) {
...
 	AssertEqualDocuments(t, bson.D{{"_id", id}, {"foo", "qux"}}, doc)
 }
+
+func TestUpdateIncOperatorErrors(t *testing.T) {
+	t.Parallel()
+
+	for name, tc := range map[string]struct {
+		filter bson.D
+		update bson.D
+		err    *mongo.WriteError
```
seeforschauer, line 83: why not create bare value, not a pointer? 
Mongo driver returns not a pointer but a bare struct.

seeforschauer: ah I see now - to make nil pointer check

integration/update_test.go
```diff
@@ -72,3 +73,186 @@ func TestUpdateUpsert(t *testing.T) {
...
+			_, err := collection.UpdateOne(ctx, tc.filter, tc.update)
+			if tc.err != nil {
+				AssertEqualWriteError(t, tc.err, err)
+				return
+			}
+			require.NoError(t, err)
+
+			var actual bson.D
+			err = collection.FindOne(ctx, tc.filter).Decode(&actual)
+			require.NoError(t, err)
```
seeforschauer, line 141: must not reach here, it's a error test.

integration/update_test.go
```diff
@@ -72,3 +73,186 @@ func TestUpdateUpsert(t *testing.T) {
...
+			err: &mongo.WriteError{
+				Code:    14,
+				Message: `Cannot apply $inc to a value of non-numeric type. {_id: "string"} has the field 'value' of non-numeric type string`,
+			},
+		},
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			ctx, collection := setup(t, shareddata.Scalars)
```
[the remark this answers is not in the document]
w84thesun: We need separate collection names to ensure update operations are not executed on the same data.

integration/update_test.go
w84thesun: Nice catch, thanks!

integration/update_test.go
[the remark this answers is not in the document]
seeforschauer: In the `setup` there is an `InsertOne`. 
How does `insert` work? Does it add or replace? I assume, it adds but `_id`s are the same. 
If it adds then how it is still working. Are we trying to update all existing `_id`s? 
What happens there?

integration/helpers.go
```diff
@@ -184,3 +184,22 @@ func CollectKeys(t testing.TB, doc bson.D) []string {
 
 	return res
 }
+
+// AssertEqualWriteError compares expected mongo.WriteError Message and Code with actual error.
+func AssertEqualWriteError(t *testing.T, expected *mongo.WriteError, actual error) bool {
```
seeforschauer, line 189: I'm not sure, but it also can shadow the MongoDB errors checking for example when we expect in tests a certain error type, but MongoDB gives another type, it will bypass.

internal/handlers/common/error.go
```diff
@@ -108,21 +139,87 @@ func (e *Error) Document() *types.Document {
...
+}
+
+// Unwrap implements standard error unwrapping interface.
+func (we *WriteError) Unwrap() error {
+	return we.err
+}
+
+// Document implements ProtoErr interface.
+// Fields "code" and "errmsg" must always be filled in so that clients can parse the error message.
+func (we *WriteError) Document() *types.Document {
```
seeforschauer, line 217: Could you please add a comment that otherwise it would parse it as a command error?

integration/helpers.go
seeforschauer: No shadowing here - it can be seen in deep debug on mongo driver.

internal/handlers/common/error.go
```diff
@@ -108,21 +139,88 @@ func (e *Error) Document() *types.Document {
...
-	return NewError(errInternalError, err).(*Error), false
+// Unwrap implements standard error unwrapping interface.
+func (we *WriteErrors) Unwrap() error {
+	for _, e := range *we {
+		return &e
+	}
+	return nil
+}
+
+// Document implements ProtoErr interface..
```
seeforschauer, line 179: Could you please add here some more from your debugging adventures?

integration/update_test.go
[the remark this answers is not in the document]
seeforschauer: each time a new collection, ok.

internal/handlers/common/error.go
w84thesun: Added.

internal/handlers/common/error.go
w84thesun: Added.

rumyantseva (commented):
I still miss some context to provide a good code review here. So, I mostly ask questions, some of them might be completely irrelevant.

Overall, it seems that such choices (which types can be `$inc` and which cannot) need to be documented. I guess, it's a separate task to document those things better?

internal/handlers/common/error.go
```diff
@@ -161,4 +261,6 @@ func formatBitwiseOperatorErr(err error, operator string, maskValue any) error {
 // check interfaces
 var (
 	_ error = (*Error)(nil)
+	_ error = (*WriteError)(nil)
```
AlekSi, line 264: Is there an API that could return a single write error of this type? Should both `WriteError` and `WriteErrors` implement error interface?

And we should check that Error/CommandError and WriteErrors implement ProtoErr interface

internal/handlers/common/error.go
AlekSi: That was a question about ProtoErr interface definition in the source ode file – it was sandwiched between Error and WriteErrors definitions. That's not the case now

integration/update_test.go
```diff
@@ -72,3 +73,178 @@ func TestUpdateUpsert(t *testing.T) {
...
```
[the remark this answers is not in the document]
w84thesun: Added tests for documents and arrays.

integration/update_test.go
```diff
@@ -72,3 +73,178 @@ func TestUpdateUpsert(t *testing.T) {
...
+		"FieldNotExist": {
+			filter: bson.D{{"_id", "int32"}},
+			update: bson.D{{"$inc", bson.D{{"foo", int32(1)}}}},
+			result: bson.D{{"_id", "int32"}, {"value", int32(42)}, {"foo", int32(1)}},
+		},
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			ctx, collection := setup(t, shareddata.Scalars, shareddata.Composites)
```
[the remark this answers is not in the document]
w84thesun: Fixed.

internal/handlers/common/error.go
```diff
@@ -51,7 +51,38 @@ const (
 	ErrRegexMissingParen = ErrorCode(51091) // Location51091
 )
 
-// Error represents wire protocol error.
+// ProtoErr represents protocol error type.
```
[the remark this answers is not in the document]
w84thesun: We need `WriteErrors` to return correct errors to clients. In order to use CommandError and WriteError at `conn.go` we introduce this interface.

internal/handlers/common/error.go
w84thesun: The second sentence explains why it exists.

internal/handlers/common/params.go
```diff
@@ -105,10 +105,12 @@ func AssertType[T types.Type](value any) (T, error) {
...
 var (
-	errUnexpectedType = fmt.Errorf("unexpected type")
-	errNotWholeNumber = fmt.Errorf("not a whole number")
-	errNegativeNumber = fmt.Errorf("negative number")
-	errNotBinaryMask  = fmt.Errorf("not a binary mask")
+	errUnexpectedType      = fmt.Errorf("unexpected type")
+	errNotWholeNumber      = fmt.Errorf("not a whole number")
+	errNegativeNumber      = fmt.Errorf("negative number")
+	errNotBinaryMask       = fmt.Errorf("not a binary mask")
+	errBadLeftOperandType  = fmt.Errorf("bad left operand type")
```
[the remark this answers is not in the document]
w84thesun: > Is a similar wording used somewhere else, so this "bad" is consistent with it?

Nope, I don't think so.
Changed to `errUnexpectedLeftOpType`.

integration/update_test.go
```diff
@@ -72,3 +73,178 @@ func TestUpdateUpsert(t *testing.T) {
 	require.NoError(t, err)
 	AssertEqualDocuments(t, bson.D{{"_id", id}, {"foo", "qux"}}, doc)
 }
+
+func TestUpdateIncOperatorErrors(t *testing.T) {
+	t.Parallel()
+
+	for name, tc := range map[string]struct {
```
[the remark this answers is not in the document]
w84thesun: Added tests for documents and arrays.

integration/update_test.go
```diff
@@ -72,3 +73,178 @@ func TestUpdateUpsert(t *testing.T) {
 	require.NoError(t, err)
 	AssertEqualDocuments(t, bson.D{{"_id", id}, {"foo", "qux"}}, doc)
 }
+
+func TestUpdateIncOperatorErrors(t *testing.T) {
```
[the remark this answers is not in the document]
w84thesun: I thought that the test name reflects this by itself. Like we have `TestUpdateIncOperatorErrors` and `TestUpdateIncOperator`. The first one checks errors and the second one checks responses.

integration/helpers.go
```diff
@@ -184,3 +184,22 @@ func CollectKeys(t testing.TB, doc bson.D) []string {
...
+		return assert.Equal(t, expected, actual)
+	}
+
+	if len(writeException.WriteErrors) != 1 {
+		return assert.Equal(t, expected, actual)
+	}
+
+	actualWriteErr := writeException.WriteErrors[0]
+
+	return assert.Equal(t, expected.Message, actualWriteErr.Message) &&
```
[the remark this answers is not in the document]
w84thesun: `assert.Equal` would print message text or code if it's not equal. The potential problem here is that if the code and message are not equal it will print a message or a code not both of them, so you would require 2 test runs to fix it.

internal/handlers/common/error.go
seeforschauer: Thank you.

> Added.

internal/handlers/common/error.go
w84thesun: No, `WriteError` always returned as a part of `WriteErrors`. I made `WriteError` unexported.

integration/update_test.go
```diff
@@ -72,3 +73,194 @@ func TestUpdateUpsert(t *testing.T) {
...
+			err: &mongo.WriteError{
+				Code:    14,
+				Message: `Cannot apply $inc to a value of non-numeric type. {_id: "string"} has the field 'value' of non-numeric type string`,
+			},
+		},
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			ctx, collection := setup(t, shareddata.Scalars, shareddata.Composites)
```
seeforschauer, line 147: Would you consider adding here only document, string, and array? 
It requires each time a lot of connections / less unused data is better.

integration/update_test.go
```diff
@@ -72,3 +73,194 @@ func TestUpdateUpsert(t *testing.T) {
...
+		"FieldNotExist": {
+			filter: bson.D{{"_id", "int32"}},
+			update: bson.D{{"$inc", bson.D{{"foo", int32(1)}}}},
+			result: bson.D{{"_id", "int32"}, {"value", int32(42)}, {"foo", int32(1)}},
+		},
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			ctx, collection := setup(t, shareddata.Scalars)
```
seeforschauer, line 254: I would suggest the same here - double, double-nan, int34, and int64 are much less than shareddata.Scalars.

internal/handlers/common/error.go
```diff
@@ -108,21 +142,69 @@ func (e *Error) Document() *types.Document {
...
+// Unwrap implements standard error unwrapping interface.
+func (we *WriteErrors) Unwrap() error {
+	for _, e := range *we {
+		return errors.New(e.err)
+	}
+	return nil
+}
+
+// Document implements ProtoErr interface.
+// "writeErrors" field must present in result document so that clients could parse it as WriteErrors.
```
seeforschauer, line 184:
```suggestion
//  "writeErrors" field must be present in the result document so that clients can parse it as WriteErrors.
```

internal/handlers/common/error.go
```diff
@@ -108,21 +142,69 @@ func (e *Error) Document() *types.Document {
...
+// Fields "code" and "errmsg" must always be filled in so that clients can parse the error message,
+// otherwise mongo client would parse it as a CommandError.
```
seeforschauer, line 186:
```suggestion
// Otherwise, the mongo client would parse it as a CommandError.
```

internal/handlers/common/error.go
```diff
@@ -108,21 +142,69 @@ func (e *Error) Document() *types.Document {
...
+func (we *WriteErrors) Unwrap() error {
+	for _, e := range *we {
+		return errors.New(e.err)
+	}
+	return nil
+}
+
+// Document implements ProtoErr interface.
+// "writeErrors" field must present in result document so that clients could parse it as WriteErrors.
+// Fields "code" and "errmsg" must always be filled in so that clients can parse the error message,
```
seeforschauer, line 185:
```suggestion
// Fields "code" and "errmsg" must always be filled in so that clients can parse the error message.
```

internal/handlers/common/error.go
```diff
@@ -108,21 +142,69 @@ func (e *Error) Document() *types.Document {
...
+// Code implements ProtoErr interface.
+func (we *WriteErrors) Code() ErrorCode {
+	for _, e := range *we {
+		return e.code
 	}
+	return errUnset
+}
 
-	return NewError(errInternalError, err).(*Error), false
+// Unwrap implements standard error unwrapping interface.
```
seeforschauer, line 175:
```suggestion
// Unwrap implements a standard error unwrapping interface.
```

internal/handlers/common/error.go
```diff
@@ -108,21 +142,69 @@ func (e *Error) Document() *types.Document {
...
-// Nil panics, *Error (possibly wrapped) is returned unwrapped with true,
-// any other value is wrapped with InternalError and returned with false.
-func ProtocolError(err error) (*Error, bool) {
-	if err == nil {
-		panic("err is nil")
+// WriteErrors represents slice of protocol write errors.
+// It could be returned for Update, Insert, Delete, Replace operations.
+type WriteErrors []writeError
+
+// NewWriteErrorMsg creates new protocol write error with given ErrorCode and message.
```
seeforschauer, line 149:
```suggestion
// NewWriteErrorMsg creates a new protocol write error with given ErrorCode and message.
```

internal/handlers/common/error.go
```diff
@@ -108,21 +142,69 @@ func (e *Error) Document() *types.Document {
...
 
-// ProtocolError converts any error to wire protocol error.
-//
-// Nil panics, *Error (possibly wrapped) is returned unwrapped with true,
-// any other value is wrapped with InternalError and returned with false.
-func ProtocolError(err error) (*Error, bool) {
-	if err == nil {
-		panic("err is nil")
+// WriteErrors represents slice of protocol write errors.
+// It could be returned for Update, Insert, Delete, Replace operations.
```
seeforschauer, line 146:
```suggestion
// It could be returned for Update, Insert, Delete, and Replace operations.
```

internal/handlers/common/error.go
```diff
@@ -108,21 +142,69 @@ func (e *Error) Document() *types.Document {
...
 }
 
-// ProtocolError converts any error to wire protocol error.
-//
-// Nil panics, *Error (possibly wrapped) is returned unwrapped with true,
-// any other value is wrapped with InternalError and returned with false.
-func ProtocolError(err error) (*Error, bool) {
-	if err == nil {
-		panic("err is nil")
+// WriteErrors represents slice of protocol write errors.
```
seeforschauer, line 145:
```suggestion
// WriteErrors represents a slice of protocol write errors.
```

internal/handlers/common/error.go
```diff
@@ -51,7 +51,41 @@ const (
...
+		return writeErr, true
+	}
+
+	return NewError(errInternalError, err).(*Error), false
+}
+
+// CommandError represents wire protocol command error.
+type CommandError = Error
+
+// Error is a deprecated name for CommandError; use the later in the new code instead.
```
seeforschauer, line 88:
```suggestion
// Error is a deprecated name for CommandError; instead, use the later version in the new code.
```

internal/handlers/common/params.go
```diff
@@ -213,3 +215,46 @@ func parseTypeCode(alias string) (typeCode, error) {
 
 	return code, nil
 }
+
+// addNumbers returns the result of v1 and v2 addition and error if addition failed.
+// The v1 and v2 parameters could be float64, int32, int64.
+// The result would be the biggest type possible, ie int32 + int64 produce int64 result.
```
seeforschauer, line 221:
```suggestion
// The result would be the broader type possible, i.e. int32 + int64 produces int64.
```

integration/update_test.go
w84thesun: Scalars + composites are 49 documents in total. I don't think there should be any connection problem inserting 49 documents with one InsertMany request. Otherwise, I should extract those documents from sets or write my own for each test which makes those sets kinda useless.

integration/update_test.go
w84thesun: Same as above.

seeforschauer: I'm sorry I just saw in my own PR, where a lot of data was not necessary and it was taken too long for tests to retest.
And, my second argument, is that also we'd probably get rid someday of shareddata, it's in a comment of shared-data.

w84thesun: Updated.

seeforschauer (approved):
LGTM.

AlekSi (dismissed):

AlekSi:
@rumyantseva PTAL and merge with the right commit body if it looks good to you

internal/handlers/common/error.go
```diff
@@ -108,21 +142,69 @@ func (e *Error) Document() *types.Document {
...
+	for _, e := range *we {
+		return errors.New(e.err)
+	}
+	return nil
+}
+
+// Document implements ProtoErr interface.
+//  "writeErrors" field must be present in the result document so that clients can parse it as WriteErrors.
+// Fields "code" and "errmsg" must always be filled in so that clients can parse the error message.
+// Otherwise, the mongo client would parse it as a CommandError.
```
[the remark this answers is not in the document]
w84thesun: Fixed

internal/handlers/common/error.go
```diff
@@ -160,5 +242,8 @@ func formatBitwiseOperatorErr(err error, operator string, maskValue any) error {
 
 // check interfaces
 var (
-	_ error = (*Error)(nil)
+	_ error    = (*Error)(nil)
+	_ error    = (*WriteErrors)(nil)
```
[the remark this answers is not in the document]
w84thesun: Fixed

internal/handlers/common/update.go
```diff
@@ -41,6 +42,66 @@ func UpdateDocument(doc, update *types.Document) error {
 				}
 			}
 
+		case "$inc":
+			incDoc, ok := updateV.(*types.Document)
```
rumyantseva, line 46: Why don't we move that long construction to a separate function? So then it's easier to test and the cyclomatic complexity is a bit lower (and it's easier to read it). I don't want to block this PR merge by it, but I believe we should discuss this topic.

w84thesun: There would be no test impact if I move it to a separate function as we test with integration tests mostly.
I believe it's just because we don't have an agreement on when it's time to extract a function right now.

rumyantseva (approved):
LGTM. I have a question about long switch-case constructions, but as they are presented in various places, not only in this PR, I will raise it separately.

PR #599, opened 2022-05-12
Bump PostgreSQL and MongoDB versions

AlekSi:
To stay current.


rumyantseva (approved):
As those are minor/fix versions, that should be fine I guess.

Issue #323, opened 2022-02-16
Use multi-stage Docker build


AlekSi:
We should build FerretDB binary for Docker image inside Docker container using [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/).

* [x] Build all images with race detector and test coverage.
* [x] Build separate images for development (`ferretdb-dev` with race detector, test coverage, etc) and production (`ferretdb` image without race detector and test coverage).

See also #70.

AlphaB:
@AlekSi should we use `latest` tag for build step or it's better to fix minor version on 1.18?

AlekSi:
I think the best way would be to use `:1.18.2` and update Dependabot config to track changes

AlphaB:
@AlekSi We won't be able to use race detector with `scratch` image since it requires `cgo`.

AlekSi:
@AlphaB I think we should build two images:
* "release" image without race detector, code coverage, etc.
* "development" image with all that.

AlekSi:
We will do the second task closer to v1.0.0

PR #605, opened 2022-05-14
Use multi-stage docker build

AlphaB:
Refs #323.


AlphaB:
I've some doubts about this solution. First I have to use `golang` as a base image for the development image instead of using `alpine` because alpine doesn't support race detector without installing additional packages. So the dev image became huge ~ 840mb instead of ~50mb. Not sure if it's an issue since we won't push it anywhere.
Plus I'm not sure if it's a good idea to have a single `Dockerfile` for 2 different purposes. The other solution might be having 2 separate `Dockerfile` for dev and release images. That will allow us to use `alpine` as a base for the dev image with additional packages for race support and a more clean `Dockerfile` for the release image.

AlekSi:
Currently, we don't have separate release and development images. All our images, including releases, include FerretDB with test coverage enabled, for example. So using just one Dockerfile is fine for now.

I propose to use `golang` image in that PR that would be FerretDB with race detector enabled. Then we could create two separate issues for making an image smaller (probably by using Alpine) and for having separate release and development issues.

AlekSi (commented):
Currently, we don't have separate release and development images. All our images, including releases, include FerretDB with test coverage enabled, for example. So using just one Dockerfile is fine for now.

I propose to use `golang` image in that PR that would be FerretDB with race detector enabled. Then we could create two separate issues for making an image smaller (probably by using Alpine) and for having separate release and development issues.

AlekSi:
Well, ok, then you changed the meaning of those tasks :) Currently, there are two differences between them:
1. One of them does not push the image to the registry.
2. It also builds only one architecture.

So let's use `golang` base image in both tasks and don't use `scratch` image at all for now.

AlphaB:
Done!

Taskfile.yml
```diff
@@ -237,32 +237,27 @@ tasks:
...
-      - GOARCH=amd64 go test -c -o=bin/ferretdb-amd64 -trimpath -tags=testcover,tigris -coverpkg=./... ./cmd/ferretdb
-
   docker-local:
-    deps: [docker-build]
+    deps: [gen-version]
     cmds:
       - >
         docker buildx build --builder=ferretdb
+        --build-arg BUILD_ARGS="{{.BUILD_ARGS}}"
+        --build-arg CGO_ENABLED={{.CGO_ENABLED}}
```
AlekSi, line 246: Why we only add them there, but not in `docker-push`?

AlphaB: Those args have default values in [Dockerfile](https://github.com/FerretDB/FerretDB/pull/605/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R6-R7 ). Is it better to pass them explicitly?

Taskfile.yml
```diff
@@ -109,21 +109,21 @@ tasks:
     desc: "Run all integration tests in parallel"
     deps:
       - task: test-integration-db
-        vars: {DB: "FerretDB"}
+        vars: { DB: "FerretDB" }
```
AlphaB, line 112: My yaml formatter is insisting on this change :) Should I revert it?

AlphaB:
@AlekSi as we discussed: 
1. There are no differences between dev and production images
2. Multi-stage building
3. Both images have race detector and coverage enabled
4. Base image is `golang:1.18.3` instead of `scratch` for race detector to work

Taskfile.yml
AlekSi: `"yaml.format.bracketSpacing": false` in VS Code :)

AlekSi (approved):

AlekSi:
Thank you!

PR #606, opened 2022-05-16
Support getParameter's showDetails, allParameters

fenogentov:


seeforschauer:
looks good except that `.txt` file.

AlekSi (changes requested):
`showDetails` and `allParameters` parameters should be implemented, not just ignored. There is no implementation in that PR, only tests.

seeforschauer (changes requested):
above + are tests ok?- why they are not failing if it's not implemented? Is that field ignored?

fenogentov:
> above + are tests ok?- why they are not failing if it's not implemented? Is that field ignored?

_`The output may vary depending on the version and specific configuration of your MongoDB instance.`_

AlekSi:
> but no one ever actually does that

Why do you think so?

Is there a way to configure MongoDB to return details at least for some parameters? If yes, what should be done?

seeforschauer (approved):
this approach to tests naming looks both well documented and awkward since in go we don't use to use `_` in names. 
maybe here it's better to add a name for the test instead. but the tests look understandable. 
So ok.

internal/handlers/pg/msg_getparameter.go
```diff
@@ -101,3 +136,59 @@ func selectParam(document, resDB *types.Document) (doc *types.Document, err erro
...
+	} else {
+		if getPrm == "*" {
+			allParameters = true
+		}
+	}
+	return showDetails, allParameters, nil
+}
+
+// convertBool converts numeric types to bool type.
+func convertBool(v any) (val bool, errType string) {
```
w84thesun, line 175: We already have this kind of function — `common.GetBoolOptionalParam`

w84thesun (changes requested):

internal/handlers/pg/msg_getparameter.go
fenogentov: there is a double name `getParameter.allParameters` in the error
function only `allParameters`

w84thesun: That's ok, just use AssertEqualAltError to check it in tests.

w84thesun: The result error text would be kinda `BSON field 'allParameters' is the wrong type 'string', expected types '[bool, long, int, decimal, double]'`

fenogentov:
> The result error text would be kinda `BSON field 'allParameters' is the wrong type 'string', expected types '[bool, long, int, decimal, double]'` 

But it's not just about checking the error, but also about the exact match of the MongoDB error text
Shouldn't we "copy" MongoDB?

w84thesun:
> > The result error text would be kinda `BSON field 'allParameters' is the wrong type 'string', expected types '[bool, long, int, decimal, double]'`
> 
> But it's not just about checking the error, but also about the exact match of the MongoDB error text Shouldn't we "copy" MongoDB?

The original error text is already broken. (that part right before the end `BSON field 'getParameter.showDetails' is the wrong type 'string', expected types '[bool, long, int, decimal, double']` → `']`)
 We don't want to copy it in FerretDB.
And after all, it would take extra effort to support command name in `GetBoolOptionaParam`.

integration/commands_administration_test.go
```diff
@@ -181,7 +181,7 @@ func TestCommandsAdministrationCreateDropListDatabases(t *testing.T) {
 }
 
 func TestCommandsAdministrationGetParameter(t *testing.T) {
-	t.Parallel()
+	//	t.Parallel()
```
seeforschauer, line 184: why?

integration/commands_administration_test.go
```diff
@@ -563,17 +565,19 @@ func TestCommandsAdministrationGetParameter(t *testing.T) {
 				Name:    "TypeMismatch",
 				Message: `BSON field 'getParameter.allParameters' is the wrong type 'string', expected types '[bool, long, int, decimal, double']`,
 			},
+			altMessage: `BSON field 'allParameters' is the wrong type 'string', expected types '[bool, long, int, decimal, double]'`,
 		},
 	} {
 		name, tc := name, tc
 		t.Run(name, func(t *testing.T) {
-			t.Parallel()
+			//	t.Parallel()
```
seeforschauer, line 573: and here

integration/commands_administration_test.go
```diff
@@ -563,17 +565,19 @@ func TestCommandsAdministrationGetParameter(t *testing.T) {
...
 
 			var actual bson.D
 			err := collection.Database().RunCommand(ctx, tc.command).Decode(&actual)
 
 			if tc.err != nil {
-				AssertEqualError(t, *tc.err, err)
+				AssertEqualAltError(t, *tc.err, tc.altMessage, err)
+				//				AssertEqualError(t, *tc.err, err)
```
seeforschauer, line 580: There are should be no dead code. We have GitHub for this.

fenogentov: sorry, four parallel tasks in development is bad
Thanks

AlekSi (approved):

PR #607, opened 2022-05-16
Added support for sorting scalar data types

ribaraka:
Refs #457.

- improved regex comparison.
- fix `comment` bug.


seeforschauer (commented):
LGTM, seems like need a call on this.
Can be there any errors added in test?

internal/types/compare.go
```diff
@@ -255,3 +251,121 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+	doubleNegativeZero
+	doubleDT
+	int32DT
+	int64DT
+)
+
+// defineNumberDataType define which number type has value and returns a sequence the type has.
+func defineNumberDataType(value any) numberDataTypeOrderResult {
+	switch value := value.(type) {
+	case int64:
```
w84thesun, line 321: Please rearrange cases in order with [code style](https://github.com/FerretDB/FerretDB/blob/main/CONTRIBUTING.md#code-style-and-conventions)

internal/types/compare.go
```diff
@@ -255,3 +251,121 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+	dateDataType
+	timestampDataType
+	regexDataType
+	maxKeyDataType
+)
+
+// defineDataType define which type has value and returns a sequence the type has.
+func defineDataType(value any) dataTypeOrderResult {
+	switch value := value.(type) {
+	case NullType:
```
w84thesun, line 279: Please rearrange cases in order with [code style](https://github.com/FerretDB/FerretDB/blob/main/CONTRIBUTING.md#code-style-and-conventions)

w84thesun (changes requested):

internal/types/compare.go
ribaraka: good catch! thank you! the cases have been rearrangeg.

internal/types/compare.go
ribaraka: fixed.

ribaraka:
> @ribaraka would you please re-run `task gen` locally and commit that change? Then dirty check will pass. Thank you.

done

internal/handlers/common/error.go
```diff
@@ -41,7 +41,7 @@ const (
 	ErrNamespaceExists   = ErrorCode(48)    // NamespaceExists
 	ErrCommandNotFound   = ErrorCode(59)    // CommandNotFound
 	ErrNotImplemented    = ErrorCode(238)   // NotImplemented
-	ErrSortBadValue      = ErrorCode(15974) // Location15974
```
seeforschauer, line 44: hm, it means that some test cases are missing.  Would you please add them?

ribaraka: Could you please point to one?

w84thesun: There is one test we can add for this error type. Could you please add this test in `TestQueryBadSortType`?
```go
		"BadSortTypeValue": {
			command: bson.D{
				{"find", collection.Name()},
				{"projection", bson.D{{"value", "some"}}},
				{"sort", bson.D{{"asc", "123"}}},
			},
			err: &mongo.CommandError{
				Code:    15974,
				Name:    "Location15974",
				Message: `Illegal key in $sort specification: asc: "123"`,
			},
		},
```

seeforschauer: @w84thesun good example, thank you.

internal/types/compare.go
```diff
@@ -34,13 +34,10 @@ type CompareResult int8
...
-	Less    CompareResult = -1 // <
-	Greater CompareResult = 1  // >
-
-	// TODO Remove once it is no longer needed.
-	// It should not be needed.
-	NotEqual CompareResult = 127 // !=
+	Equal        CompareResult = 0   // ==
+	Less         CompareResult = -1  // <
+	Greater      CompareResult = 1   // >
+	Incomparable CompareResult = 127 // !=
```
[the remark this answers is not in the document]
AlekSi: Unfortunately, `<>` is exactly the same as `!=` in SQL. From [this page](https://www.postgresql.org/docs/current/functions-comparison.html):
> <> is the standard SQL notation for “not equal”. != is an alias, which is converted to <> at a very early stage of parsing

After looking into [that Wikipedia page](https://en.wikipedia.org/wiki/Mathematical_operators_and_symbols_in_Unicode) for way too long, I propose to use `≹` (U+2279). Bonus point for that symbol looking just like crossed-out Z.

AlekSi (changes requested):
> added support for storing negative zero

I added a "simple" integration test to see how MongoDB works with negative zero and noticed that our comparison function does not actually compare it correctly.
Please remove that from this PR for now.

internal/handlers/common/error.go
ribaraka: Thank you both for comments. Error case has been added. Error Code has been returned back, but slightly refactored.

AlekSi:
Added #612 to track negative zero support.

internal/types/compare.go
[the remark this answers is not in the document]
ribaraka: :smile: agreed.

w84thesun (approved):

seeforschauer (approved):

internal/types/compare.go
```diff
@@ -255,3 +251,123 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+		return int32DT
+	case int64:
+		return int64DT
+	}
+
+	panic(fmt.Sprintf("defineNumberDataType: value cannot be defined, value is %v, data type of value is %T", value, value))
+}
+
+// CompareOrder defines the data type for the two values and compares them. When the types are equal, it compares using Compare.
+func CompareOrder(a, b any, order string) CompareResult {
```
AlekSi, line 338: We should use some other type for `order` instead of string, and use typed constants.

I guess we have to move `common.sortType` there and export it

internal/types/compare.go
```diff
@@ -255,3 +251,123 @@ func compareNumbers(a float64, b int64) CompareResult {
 
 	return CompareResult(bigA.Cmp(bigB))
 }
+
+// dataTypeOrderResult represents the comparison order of data types.
+type dataTypeOrderResult uint8
+
+const (
+	_ dataTypeOrderResult = iota
+	minKeyDataType
```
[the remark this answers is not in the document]
AlekSi: I think yes; we need them for logging, debugging, error messages, etc.

internal/types/compare.go
ribaraka: > We should use some other type for `order` instead of string...

What do you mean by that? Could you please give an example?

internal/types/compare.go
```diff
@@ -255,3 +251,123 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+	case int32:
+		return int32DT
+	case int64:
+		return int64DT
+	}
+
+	panic(fmt.Sprintf("defineNumberDataType: value cannot be defined, value is %v, data type of value is %T", value, value))
+}
+
+// CompareOrder defines the data type for the two values and compares them. When the types are equal, it compares using Compare.
```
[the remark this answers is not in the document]
AlekSi: My understanding is that we will use `Compare` for filtering and `CompareOrder` for sorting. Is that correct?

If yes, then `Compare`'s function documentation says "filtering or sorting" which is not correct: it should not mention sorting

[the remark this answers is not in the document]
ribaraka: When in `CompareOrder` types are equal, it then compares their values using `Compare`.

internal/types/compare.go
ribaraka: Got it! Again, thank you for you time :smile:

ribaraka: Should I also move all those orders of data types (`dataTypeOrderResult`, `numberDataTypeOrderResult`) to common.sortType as well?

AlekSi: I think we should keep `CompareOrder` in `types` package. For that, we should move `common.sortType` to `types` package. To be able to use, we should export it.

I don't see why we need to move or export `dataTypeOrderResult` and `numberDataTypeOrderResult`

internal/types/compare.go
```diff
@@ -255,3 +251,123 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+		return regexDataType
+	case int32:
+		return numbersDataType
+	case Timestamp:
+		return timestampDataType
+	case int64:
+		return numbersDataType
+	}
+
+	panic(fmt.Sprintf("value cannot be defined, value is %v, data type of value is %T", value, value))
```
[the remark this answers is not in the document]
ribaraka: fixed.

integration/query_test.go
```diff
@@ -168,6 +168,129 @@ func TestQuerySort(t *testing.T) {
...
+				"int64-zero",
+				"int32-min",
+				"int64-min",
+				"double-negative-infinity",
+				"double-nan",
+				"null",
+			},
+		},
+		"BadSortValue": {
+			sort: bson.D{{"value", 11}},
```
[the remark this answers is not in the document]
ribaraka: The tests have been added.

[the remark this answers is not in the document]
ribaraka: Do I need to add tests with other data types?

internal/types/compare.go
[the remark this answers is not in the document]
ribaraka: Added.

internal/types/compare.go
ribaraka: migrated

integration/query_test.go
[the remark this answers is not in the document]
ribaraka: @AlekSi  ^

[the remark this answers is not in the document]
seeforschauer: @ribaraka which data types do you mean?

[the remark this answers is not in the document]
seeforschauer: The PR title is `scalar data types`, so suppose no need for documents and arrays. 

What about strings. Do you mean to have some scalar type lists, and try to sort them with different options? Definitely worth it.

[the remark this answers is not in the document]
ribaraka: > What about strings. Do you mean to have some scalar type lists, and try to sort them with different options? Definitely worth it.

I'd like to do it in another PR.

@AlekSi suggested to test sorting using BadValue as null and 0, now I've been suggesting adding a `BadValue` as another different data types.

internal/types/compare.go
```diff
@@ -27,40 +27,37 @@ import (
...
 
-// Compare compares any BSON values in the same way as MongoDB does it for filtering or sorting.
+// Compare compares any BSON values in the same way as MongoDB does it for filtering.
 //
 // It converts types as needed; that may result in different types being equal.
 // For that reason, it typically should not be used in tests.
 //
 // Compare and contrast with test helpers in testutil package.
 func Compare(v1, v2 any) CompareResult {
-	if v1 == nil {
```
seeforschauer, line 53: why this change?  as for me previous version was more readable.

internal/types/compare.go
```diff
@@ -255,3 +254,130 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+	arrayDataType
+	binDataType
+	objectIdDataType
+	booleanDataType
+	dateDataType
+	timestampDataType
+	regexDataType
+)
+
+// defineDataType define which type has value and returns a sequence the type has.
```
seeforschauer, line 277:
```suggestion
// defineDataType returns a sequence for build-in type.
```

seeforschauer: would you be so kind to add a comment about Document and Array types?

internal/types/compare.go
```diff
@@ -255,3 +254,130 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+
+const (
+	_ numberDataTypeOrderResult = iota
+	doubleNegativeZero
+	doubleDT
+	int32DT
+	int64DT
+)
+
+// defineNumberDataType define which number type has value and returns a sequence the type has.
```
seeforschauer, line 321:
```suggestion
// defineNumberDataType returns a sequence for float64, int32 and int64 types.
```

internal/handlers/pg/msg_find.go
```diff
@@ -92,13 +92,16 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
 		return nil, err
 	}
-	// comment in query, e.g. db.collection.find({$comment: "test"})
-	if sp.comment, err = common.GetOptionalParam(filter, "$comment", sp.comment); err != nil {
-		return nil, err
+	// get comment from query, e.g. db.collection.find({$comment: "test"})
+	if filter != nil {
+		if sp.comment, err = common.GetOptionalParam(filter, "$comment", sp.comment); err != nil {
+			return nil, err
+		}
```
AlekSi, line 104: Should we do the same in findAndModify?

internal/types/compare.go
```diff
@@ -255,3 +254,130 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+	}
+
+	panic(fmt.Sprintf("defineNumberDataType: value cannot be defined, value is %v, data type of value is %T", value, value))
+}
+
+// SortType represents sort type for $sort aggregation.
+type SortType int8
+
+const (
+	Ascending  SortType = 1  // asc
```
seeforschauer, line 342: Would you be so kind to add comments here for exported variables?

internal/types/compare.go
```diff
@@ -255,3 +254,130 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+			return doubleNegativeZero
+		}
+		return doubleDT
+	case int32:
+		return int32DT
+	case int64:
+		return int64DT
+	}
+
+	panic(fmt.Sprintf("defineNumberDataType: value cannot be defined, value is %v, data type of value is %T", value, value))
```
seeforschauer, line 335: Should we panic here?

internal/types/compare.go
```diff
@@ -255,3 +254,130 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+type SortType int8
+
+const (
+	Ascending  SortType = 1  // asc
+	Descending SortType = -1 // desc
+)
+
+// CompareOrder defines the data type for the two values and compares them.
+// When the types are equal, it compares their values using Compare.
+func CompareOrder(a, b any, order SortType) CompareResult {
```
AlekSi, line 348: Let's extract that function and everything about ordering into compare_order.go

internal/types/compare.go
```diff
@@ -255,3 +254,130 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+	if a == nil || b == nil {
+		panic(fmt.Sprintf("CompareOrder: values should not be nil, a is %v, b is %v", a, b))
+	}
+
+	aType := defineDataType(a)
+	bType := defineDataType(b)
+	if aType == bType {
+		res := Compare(a, b)
+		if res == Equal && aType == numbersDataType {
```
seeforschauer, line 357: could you please remove here cyclomatic complexity?

internal/types/compare.go
```diff
@@ -255,3 +254,130 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+			bNumberType := defineNumberDataType(b)
+			switch {
+			case aNumberType < bNumberType && order == Ascending:
+				return Less
+			case aNumberType > bNumberType && order == Ascending:
+				return Greater
+			case aNumberType < bNumberType && order == Descending:
+				return Greater
+			case aNumberType > bNumberType && order == Descending:
+				return Less
```
seeforschauer, line 368: default is a must for switch cases, could you please add it here?

internal/types/compare.go
```diff
@@ -255,3 +254,130 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+
+const (
+	Ascending  SortType = 1  // asc
+	Descending SortType = -1 // desc
+)
+
+// CompareOrder defines the data type for the two values and compares them.
+// When the types are equal, it compares their values using Compare.
+func CompareOrder(a, b any, order SortType) CompareResult {
+	if a == nil || b == nil {
```
seeforschauer, line 349: separate checks will be nicier.

internal/types/compare.go
```diff
@@ -255,3 +254,130 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+			}
+		}
+
+		return res
+	}
+
+	switch {
+	case aType < bType:
+		return Less
+	case aType > bType:
```
seeforschauer, line 378: would you be so kind to add default here?

seeforschauer (changes requested):

internal/types/compare.go
```diff
@@ -255,3 +254,130 @@ func compareNumbers(a float64, b int64) CompareResult {
 
 	return CompareResult(bigA.Cmp(bigB))
 }
+
+// dataTypeOrderResult represents the comparison order of data types.
+type dataTypeOrderResult uint8
```
AlekSi, line 259: I wonder if we could come up with a better name…

@rumyantseva thoughts?

internal/types/compare.go
```diff
@@ -255,3 +254,130 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+	binDataType
+	objectIdDataType
+	booleanDataType
+	dateDataType
+	timestampDataType
+	regexDataType
+)
+
+// defineDataType define which type has value and returns a sequence the type has.
+func defineDataType(value any) dataTypeOrderResult {
```
rumyantseva, line 278: As @AlekSi mentioned, `detectDataType` would be a better name here.

internal/types/compare.go
rumyantseva: Maybe one "pro" of dedicating a separate package is to decrease the cognitive load... Fewer lines in the file, simpler mental context... But still, it's rather a random idea.

internal/handlers/common/error.go
```diff
@@ -42,7 +42,8 @@ const (
 	ErrCommandNotFound   = ErrorCode(59)    // CommandNotFound
 	ErrInvalidNamespace  = ErrorCode(73)    // InvalidNamespace
 	ErrNotImplemented    = ErrorCode(238)   // NotImplemented
-	ErrSortBadValue      = ErrorCode(15974) // Location15974
+	ErrLocation15974     = ErrorCode(15974) // Location15974
```
seeforschauer, line 45: @ribaraka could you please give here some meaningful name? Thank you.

internal/types/compare.go
seeforschauer: > However I've moved panic to `default` case.

sorry I don't see your changes, could you please verify you have pushed?

internal/handlers/pg/msg_find.go
ribaraka: Not neccesary do it same in this PR.
This piece of code has been refactored due to a bug. The bug occurs when we do not provide a filter and at the same time try to get a value from it by key.
I can add `comment` in this PR or another. What do you think is the right thing to do?

internal/types/compare.go
ribaraka: > Something like `typeInComparisonOrder` looks pretty much similar to how it's already 

@rumyantseva, how about `sequenceDatatype`?

rumyantseva: @AlekSi I believe, as the main guard of code standards in this repo, you should be the decision-maker here ;)

internal/types/compare.go
ribaraka: returned to its original form

internal/types/compare.go
ribaraka: A comment for `detectType` has been changed.

A comment for Document and Array - do you mean like TODO?

internal/types/compare.go
ribaraka: fixed

internal/types/compare.go
ribaraka: added.

internal/types/compare.go
ribaraka: ooh, sorry, I had changed it locally. I only notified you about that to make sure that panic can be set by default.
A commit with changes has been already pushed. Please review.

internal/types/compare.go
ribaraka: fixed.

internal/types/compare.go
ribaraka: fixed

internal/types/compare.go
ribaraka: added

internal/types/compare.go
ribaraka: done

internal/types/compare.go
ribaraka: done

internal/handlers/common/error.go
ribaraka: sure! done.

internal/types/compare_order.go
```diff
@@ -0,0 +1,162 @@
...
+		panic("CompareOrder: b is nil")
+	}
+
+	aType := detectDataType(a)
+	bType := detectDataType(b)
+	switch {
+	case aType == bType:
+		res := Compare(a, b)
+
+		if res == Equal && aType == numbersDataType {
```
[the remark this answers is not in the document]
ribaraka: fixed.

internal/types/compare.go
ribaraka: @seeforschauer, did my changes helped or should I try and do more?

internal/types/compare_order.go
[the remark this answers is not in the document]
ribaraka: the code block has minor changes a little bit.

internal/types/compare.go
seeforschauer: I just didn't understand - the func was accepting `any` and the comment specified that the func is for build-in type, but `Document` and `Array`, which are in build-in type, were not processed in the function.

seeforschauer (commented):
LGTM, and let's add a TODO with an issue links.

internal/types/compare.go
ribaraka: I'm going to engage them (`Document` and `Array` types) in another PR.

internal/handlers/pg/msg_find.go
seeforschauer: @ribaraka I think if you are about to divide PR either you don't close the current issue and add some comments to it or create a new issue with the tas (depends on situation and description)

seeforschauer (dismissed):
`findAndModify` sorting of scalar data types can be done in a separate PR.
I wrote a comment in the related issue, hope that helps if it's not ok,  create a new one.

internal/handlers/pg/msg_find.go
ribaraka: Good thought! The issue has been added.

integration/query_test.go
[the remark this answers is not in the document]
AlekSi: #636

internal/handlers/common/sort.go
```diff
@@ -141,13 +130,16 @@ func (ds *docsSorter) Less(i, j int) bool {
 	return ds.sorts[k](p, q)
 }
 
-// getSortType determines sortType from input sort value.
-func getSortType(value any) (sortType, error) {
+// getSortType determines SortType from input sort value.
+func getSortType(key, value any) (types.SortType, error) {
```
AlekSi, line 134: `key` is always `string`

internal/types/compare.go
```diff
@@ -34,16 +34,13 @@ type CompareResult int8
 // Values match results of comparison functions such as bytes.Compare.
 // They do not match MongoDB `sort` values where 1 means ascending order and -1 means descending.
```
AlekSi, line 35: Let's mention `SortType` there

AlekSi (changes requested):

internal/types/compare_order.go
```diff
@@ -0,0 +1,167 @@
...
+	// Ascending is used for sort in ascending order.
+	Ascending SortType = 1
+
+	// Descending is used for sort in descending order.
+	Descending SortType = -1
+)
+
+// CompareOrder detects the data type for two values and compares them.
+// When the types are equal, it compares their values using Compare.
+func CompareOrder(a, b any, order SortType) CompareResult {
```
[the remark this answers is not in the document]
ribaraka: ascending

internal/types/compare_order.go
```diff
@@ -0,0 +1,167 @@
...
+// compareTypeOrderResult represents the comparison order of data types.
+type compareTypeOrderResult uint8
+
+const (
+	_ compareTypeOrderResult = iota
+	nullDataType
+	nanDataType
+	numbersDataType
+	stringDataType
+	documentDataType // TODO: https://github.com/FerretDB/FerretDB/issues/457
```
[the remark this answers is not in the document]
ribaraka: fixed

internal/types/compare.go
ribaraka: done

seeforschauer (approved):

AlekSi (approved):

PR #608, opened 2022-05-16
Support `maxTimeMS` argument for `find` and `findAndModify` methods

### LINKED ISSUE #497: Support `maxTimeMS` argument for `find` and `findAndModify`
`find` and `findAndModify` commands support `maxTimeMS` argument that specifies the command's time limit in milliseconds. We should support it.

The implementation should be relatively easy: get that optional parameter, and (if it is not zero) derive the Go context with a timeout and use it. In common handler code, check if the error is (possibly wrapped) context timeout, and return the right error in that case.

Of course, there are some small details that should be checked by integration tests: what types are allowed for that argument (can it be float64? int64? the document, array, null?), what error is returned on timeout, etc.

DoodgeMatvey:
Closes #497.


seeforschauer (changes requested):
@DoodgeMatvey would you please add some tests or modify old tests to make it visible that now it's supported? 
Thank you.

AlekSi (changes requested):
Yes, we need integration tests. For example:
* maxTimeMS=0
* maxTimeMS=-1
* maxTimeMS=1 (any very short value)
* etc

integration/query_test.go
```diff
@@ -426,6 +427,202 @@ func TestQueryBadSortType(t *testing.T) {
...
+				{"find", collection.Name()},
+				{"projection", bson.D{{"value", "some"}}},
+				{"maxTimeMS", 43.15},
+			},
+			err: &mongo.CommandError{
+				Code:    2,
+				Name:    "BadValue",
+				Message: "maxTimeMS has non-integral value",
+			},
+			altMessage: "maxTimeMS has non-integral value",
```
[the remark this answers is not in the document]
DoodgeMatvey: This message from mongodb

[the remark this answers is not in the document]
DoodgeMatvey: Thank you, fixed

AlekSi (changes requested):
There are merge conflicts

internal/handlers/pg/msg_find.go
```diff
@@ -69,6 +73,29 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+
+	if maxTimeMS < 0 {
+		return nil, common.NewErrorMsg(
+			common.ErrBadValue,
+			fmt.Sprintf("%v value for maxTimeMS is out of range", maxTimeMS),
+		)
+	}
+
+	// TODO: What time should I set for delay?
+	runCtx, runCancel := ctxutil.WithDelay(ctx.Done(), 40*time.Second)
```
AlekSi, line 93: We don't need it

internal/handlers/pg/msg_findandmodify.go
```diff
@@ -179,8 +208,12 @@ func (h *Handler) MsgFindAndModify(ctx context.Context, msg *wire.OpMsg) (*wire.
 			return &reply, nil
 		}
 
-		_, err = h.delete(ctx, params.sqlParam, resDocs)
+		_, err = h.delete(runCtx, params.sqlParam, resDocs)
 		if err != nil {
+			if errors.Is(err, context.DeadlineExceeded) {
+				return nil, fmt.Errorf("context deadline exceeded, maxTimeMS: %v", params.maxTimeMS)
```
AlekSi, line 214: Why do we need that? Why can't we return errors as they are?

internal/handlers/pg/msg_findandmodify.go
```diff
@@ -310,6 +344,23 @@ func prepareFindAndModifyParams(document *types.Document) (*findAndModifyParams,
...
+		return nil, err
+	}
+
+	if maxTimeMS < 0 {
+		return nil, common.NewErrorMsg(
+			common.ErrBadValue,
+			fmt.Sprintf("%v value for maxTimeMS is out of range", maxTimeMS),
+		)
+	}
+
```
AlekSi, line 363: I think getErrorForInvalidMaxTimeMS should be modified to return the correct value too, not only error. That way, we can reduce the amount of copy&pasted code

AlekSi (changes requested):

internal/handlers/pg/msg_findandmodify.go
```diff
@@ -56,13 +58,28 @@ func (h *Handler) MsgFindAndModify(ctx context.Context, msg *wire.OpMsg) (*wire.
 		return nil, err
 	}
 
-	fetchedDocs, err := h.fetch(ctx, params.sqlParam)
+	runCtx, runCancel := ctxutil.WithDelay(ctx.Done(), 10*time.Second)
```
seeforschauer, line 61: here also.

internal/handlers/pg/msg_find.go
DoodgeMatvey: fixed

internal/handlers/pg/msg_findandmodify.go
DoodgeMatvey: We can, I removed it

internal/handlers/pg/msg_findandmodify.go
DoodgeMatvey: Fixed

internal/handlers/pg/msg_findandmodify.go
DoodgeMatvey: fixed

internal/handlers/pg/msg_find.go
```diff
@@ -154,3 +169,65 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
 
 	return &reply, nil
 }
+
+func getMaxTimeMSParameter(document *types.Document) (int32, error) {
```
rumyantseva, line 173: Should we add some documentation like for the other unexported functions?

internal/handlers/pg/msg_find.go
```diff
@@ -154,3 +169,65 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+			return 0, getErrorForInvalidMaxTimeMS(document)
+		}
+
+		return 0, err
+	}
+
+	if maxTimeMSParam < 0 {
+		return 0, common.NewErrorMsg(
+			common.ErrBadValue,
+			fmt.Sprintf("%v value for maxTimeMS is out of range", maxTimeMSParam),
```
rumyantseva, line 188: "Out of range" doesn't sound specific enough to me. I'd suggest something like `maxTimeMS must not be negative (but %v is provided)`

internal/handlers/pg/msg_find.go
```diff
@@ -154,3 +169,65 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+		)
+	}
+
+	return maxTimeMSParam, nil
+}
+
+func getErrorForInvalidMaxTimeMS(document *types.Document) error {
```
rumyantseva, line 195: Same thing about documentation, let's provide something?

internal/handlers/pg/msg_find.go
```diff
@@ -154,3 +169,65 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+	v, err := document.Get("maxTimeMS")
```
rumyantseva, line 196: Why the code above used `common.GetOptionalParam(document, "maxTimeMS", maxTimeMS)` to get the value and here it's `document.Get`? What's the difference?

internal/handlers/pg/msg_find.go
```diff
@@ -154,3 +169,65 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+	if err != nil {
+		return nil
+	}
+
+	switch maxTimeMS := v.(type) {
```
rumyantseva, line 201: For all the error messages in this `switch`: is the difference between them so significant? Can't we just say something like `maxTimeMS must be a non-negative integer (but %v provided)` to cover all the cases in one? 

Also, the texts of those errors don't seem to be consistent. Some of them tell what maxTimeMS must be, for example, `maxTimeMS must be X`. But others are `%v value for maxTimeMS is out of range`, which doesn't give any exact details about the possible range.

To be honest, I like the  `maxTimeMS must be X` approach more because it gives the exact definition of what it should be, and the person who sees the error doesn't need to guess what's wrong with their parameter.

internal/handlers/pg/msg_find.go
```diff
@@ -154,3 +169,65 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+	if err != nil {
+		var commonErr *common.Error
+		if errors.As(err, &commonErr) {
+			return 0, getErrorForInvalidMaxTimeMS(document)
+		}
+
+		return 0, err
+	}
+
+	if maxTimeMSParam < 0 {
```
rumyantseva, line 185: Does it need any `if maxTimeMSParam > X` checks as well? Or do we accept any int32 numbers?

rumyantseva (changes requested):

internal/handlers/pg/msg_find.go
DoodgeMatvey: Because we need to check the type and return the corresponding value in the first case it is converted to type int32 and we need to check the type, so we need to get this value again and already determine the type

internal/handlers/pg/msg_find.go
DoodgeMatvey: Yes, it does, because maxTimeMS should be non-negative integer

internal/handlers/pg/msg_find.go
DoodgeMatvey: As i mentioned in the comment below, I tried to make the answers as much as possible the same as in mongodb, should I change this to what you suggested or can we leave the answer that would be returned when answering with mongodb?

internal/handlers/pg/msg_find.go
DoodgeMatvey: I agree, it was added

internal/handlers/pg/msg_find.go
DoodgeMatvey: Fixed

internal/handlers/pg/msg_find.go
seeforschauer: hey there, 
@ribaraka @rumyantseva 
you can use `AliasFromType()`  func to make it easier.  

Here is what I did in my PR for `$set` in `internal/handlers/common/update.go` file https://github.com/FerretDB/FerretDB/pull/634/commits/e274db0512c67bdf6413e20c74243af4569e6dbc

seeforschauer: `AliasFromType` was added in order to avoid unnecessary marshaling of `type.Document` in error messages. 
Here there are no documents in error messages.
However, it will decrease the complexity while the error code is the same and an error message will be still similar to mongo. 
So my opinion is that makes sense to use it here.

seeforschauer (dismissed):
there are some comments on `AliasFromType` func in errors, which is a question for discussion on how exactly our messages should be similar to MongoDB error messages.
PR looks good.

internal/handlers/pg/msg_find.go
ribaraka: Wouldn't it be better to rename `AliasFromType` func to `GetAliasFromType`? WDYT everyone?

seeforschauer: Function names should start with a verb. `alias` is not a verb.

ribaraka: Does this mean that you confirm my point of view?

seeforschauer: yep 
but I suggest listening what other people say)

internal/handlers/pg/msg_find.go
```diff
@@ -154,3 +169,67 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+
+	return maxTimeMSParam, nil
+}
+
+// getErrorForInvalidMaxTimeMS returns error for invalid maxTimeMS type.
+func getErrorForInvalidMaxTimeMS(document *types.Document) error {
+	v, err := document.Get("maxTimeMS")
+	if err != nil {
+		return nil
+	}
```
ribaraka, line 201: if you don't expect error here, you can replace with it with this:
```suggestion
	v := must.NotFail(document.Get("maxTimeMS"))
```

seeforschauer: mus.NotFail will panic.
previous solution - not.

internal/handlers/pg/msg_find.go
DoodgeMatvey: Okay, so let's discuss let's how similar our messages should be with the messages returned by mongodb. 
We can do this here or on Monday's Open Office House meeting. 
About `AliasFromType` func I prefer `GetAliasFromType` instead.

internal/handlers/pg/msg_find.go
DoodgeMatvey: Fixed

integration/query_test.go
```diff
@@ -426,6 +427,202 @@ func TestQueryBadSortType(t *testing.T) {
...
+				{"find", collection.Name()},
+				{"maxTimeMS", 0},
+			},
+		},
+		"MaxTimeMSInt32": {
+			command: bson.D{
+				{"find", collection.Name()},
+				{"maxTimeMS", 10000},
+			},
+		},
```
ribaraka, line 613: Please add test case:
```
		"MaxTimeMSInt64": {
			command: bson.D{
				{"find", collection.Name()},
				{"maxTimeMS", int64(42)},
			},
		}
```

DoodgeMatvey: I have `MaxTimeMSInt32` test case with `10000` value and it will be no difference with int64(42) because this value in the same range as int32. For int64 I have `BadMaxTimeMSMaxInt64` test case

internal/handlers/pg/msg_find.go
```diff
@@ -154,3 +169,67 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+				fmt.Sprintf("%v value for maxTimeMS is out of range", math.MaxInt64),
+			)
+		}
+
+		if maxTimeMS > math.MaxInt32 || maxTimeMS < math.MinInt32 {
+			return common.NewErrorMsg(
+				common.ErrBadValue,
+				fmt.Sprintf("%v value for maxTimeMS is out of range", int64(maxTimeMS)),
+			)
+		}
```
ribaraka, line 224: how about moving the infinite cases to the ".. out of range" case?

internal/handlers/pg/msg_find.go
```diff
@@ -154,3 +169,67 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+func getMaxTimeMSParameter(document *types.Document) (int32, error) {
+	var maxTimeMS int32
+	maxTimeMSParam, err := common.GetOptionalParam(document, "maxTimeMS", maxTimeMS)
+	if err != nil {
+		var commonErr *common.Error
+		if errors.As(err, &commonErr) {
+			return 0, getErrorForInvalidMaxTimeMS(document)
+		}
+
+		return 0, err
```
w84thesun, line 183: We disard the error we get from `common.GetOptionalParam`, so this could be simplified to:
```suggestion
		return 0, getErrorForInvalidMaxTimeMS(document)
```

DoodgeMatvey: Fixed

internal/handlers/pg/msg_find.go
DoodgeMatvey: In such case, our messages will be very different from what mongodb returns.

integration/query_test.go
DoodgeMatvey: Agree, added

integration/query_test.go
```diff
@@ -426,6 +427,208 @@ func TestQueryBadSortType(t *testing.T) {
...
+		"MaxTimeMSInt64": {
+			command: bson.D{
+				{"find", collection.Name()},
+				{"maxTimeMS", int64(1000)},
+			},
+		},
```
w84thesun, line 619:
```suggestion
		"Double": {
			command: bson.D{
				{"find", collection.Name()},
				{"maxTimeMS", float64(10000)},
			},
		},
		"DoubleZero": {
			command: bson.D{
				{"find", collection.Name()},
				{"maxTimeMS", float64(0)},
			},
		},
		"Int32": {
			command: bson.D{
				{"find", collection.Name()},
				{"maxTimeMS", int32(10000)},
			},
		},
		"Int32Zero": {
			command: bson.D{
				{"find", collection.Name()},
				{"maxTimeMS", int32(0)},
			},
		},
		"Int64": {
			command: bson.D{
				{"find", collection.Name()},
				{"maxTimeMS", int64(10000)},
			},
		},
		"Int64Zero": {
			command: bson.D{
				{"find", collection.Name()},
				{"maxTimeMS", int64(0)},
			},
		},
```

DoodgeMatvey: Good idea, added

w84thesun (approved):

internal/handlers/common/params.go
```diff
@@ -258,3 +258,64 @@ func addNumbers(v1, v2 any) (any, error) {
...
+func GetPositiveNumber(document *types.Document, key string) (int32, error) {
+	v, err := document.Get(key)
+	if err != nil {
+		return 0, nil
+	}
+
+	var value int32
+	switch param := v.(type) {
+	case float64:
+		if math.IsInf(param, -1) {
```
ribaraka, line 272: What if we here (for the case of floa64) use `switch` instead of multiple `if`?

internal/handlers/pg/msg_find.go
```diff
@@ -70,6 +70,17 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+	maxTimeMS, err := common.GetPositiveNumber(document, "maxTimeMS")
+	if err != nil {
+		return nil, err
+	}
+
+	runCtx, runCancel := context.WithTimeout(ctx, time.Duration(maxTimeMS)*time.Millisecond)
+	defer runCancel()
+	if maxTimeMS == 0 {
+		runCtx = ctx
+	}
```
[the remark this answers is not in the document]
DoodgeMatvey: But we need to change it in this case

internal/handlers/common/params.go
```diff
@@ -258,3 +258,64 @@ func addNumbers(v1, v2 any) (any, error) {
...
+				ErrBadValue,
+				fmt.Sprintf("%v value for %s is out of range", int64(param), key),
+			)
+		}
+
+		if param != math.Trunc(param) {
+			return 0, NewErrorMsg(ErrBadValue, fmt.Sprintf("%s must be an integer", key))
+		}
+
+		value = int32(param)
```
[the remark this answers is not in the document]
DoodgeMatvey: I think it doesn't matter here

internal/handlers/pg/msg_find.go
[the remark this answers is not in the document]
seeforschauer: @DoodgeMatvey WDYM? 
it could be also applied to the code below.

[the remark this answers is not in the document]
DoodgeMatvey: I see, fixed

ribaraka (approved):

internal/handlers/common/params.go
```diff
@@ -258,3 +258,61 @@ func addNumbers(v1, v2 any) (any, error) {
 		return nil, errUnexpectedLeftOpType
 	}
 }
+
+// GetPositiveNumber returns doc's value for key or protocol error for missing or invalid parameter.
+func GetPositiveNumber(document *types.Document, key string) (int32, error) {
+	v, err := document.Get(key)
+	if err != nil {
+		return 0, nil
```
AlekSi, line 266: > GetPositiveNumber returns doc's value for key or _protocol error for missing_ or invalid parameter.

Here we have a missing key but do not return a protocol error. Either description or code is wrong. I think we should return the error there. Do we have a test for that?

internal/handlers/common/params.go
```diff
@@ -258,3 +258,61 @@ func addNumbers(v1, v2 any) (any, error) {
 		return nil, errUnexpectedLeftOpType
 	}
 }
+
+// GetPositiveNumber returns doc's value for key or protocol error for missing or invalid parameter.
+func GetPositiveNumber(document *types.Document, key string) (int32, error) {
```
AlekSi, line 263: Let's use `GetWholeNumberParam` in that function

internal/handlers/pg/msg_find.go
```diff
@@ -70,6 +70,17 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
 		return nil, err
 	}
 
+	maxTimeMS, err := common.GetPositiveNumber(document, "maxTimeMS")
+	if err != nil {
+		return nil, err
+	}
+
+	runCtx, runCancel := context.WithTimeout(ctx, time.Duration(maxTimeMS)*time.Millisecond)
```
AlekSi, line 78: `run` looks out of place in this code. I think it was copied from some `run` function. It is a bad name

internal/handlers/pg/msg_find.go
```diff
@@ -70,6 +70,17 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+	defer runCancel()
+	if maxTimeMS != 0 {
+		ctx = runCtx
```
AlekSi, line 81: If `maxTimeMS` is zero we should not even create a new context that we will not use

AlekSi (changes requested):

internal/handlers/pg/msg_findandmodify.go
```diff
@@ -310,6 +317,11 @@ func prepareFindAndModifyParams(document *types.Document) (*findAndModifyParams,
 		return nil, err
 	}
 
+	maxTimeMS, err := common.GetPositiveNumber(document, "maxTimeMS")
+	if err != nil {
+		return nil, err
```
seeforschauer, line 322: Hey, @DoodgeMatvey  )
can you please add at least one test for `find and modify` to cover this `if` branch? Thank you.

seeforschauer:
@DoodgeMatvey would you please resolve conflicts? Thank you.

internal/handlers/common/params.go
DoodgeMatvey: Fixed, no we don't have a test

internal/handlers/common/params.go
DoodgeMatvey: Fixed

internal/handlers/pg/msg_find.go
DoodgeMatvey: Renamed

internal/handlers/pg/msg_find.go
DoodgeMatvey: Fixed

internal/handlers/pg/msg_findandmodify.go
DoodgeMatvey: Added

internal/handlers/common/params.go
DoodgeMatvey: Fixed

internal/handlers/common/params.go
```diff
@@ -258,3 +258,44 @@ func addNumbers(v1, v2 any) (any, error) {
 		return nil, errUnexpectedLeftOpType
 	}
 }
+
+// GetPositiveNumber returns doc's value for key or protocol error for missing or invalid parameter.
+func GetPositiveNumber(document *types.Document, key string, isOptional bool) (int32, error) {
+	v, err := document.Get(key)
+	if err != nil {
+		if isOptional {
```
w84thesun, line 266: Parameter `isOptional` is always true. I think it should be removed and the logic should be changed accordingly. Also it could be renamed from `GetPositiveNumber` to `GetOptionalPositiveNumber`

DoodgeMatvey: fixed, I decided to return nil as we return in `GetOptionalParam` method

internal/handlers/common/params.go
```diff
@@ -258,3 +258,40 @@ func addNumbers(v1, v2 any) (any, error) {
 		return nil, errUnexpectedLeftOpType
 	}
 }
+
+// GetOptionalPositiveNumber returns doc's value for key or protocol error for missing or invalid parameter.
```
w84thesun, line 262:
```suggestion
// GetOptionalPositiveNumber returns doc's value for key or protocol error for invalid parameter.
```

integration/query_test.go
```diff
@@ -579,6 +580,213 @@ func TestQueryBadSortType(t *testing.T) {
 	}
 }
 
+func TestQueryBadMaxTimeMSType(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setup(t, shareddata.Scalars, shareddata.Composites)
```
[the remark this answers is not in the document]
DoodgeMatvey: Fixed

integration/query_test.go
```diff
@@ -579,6 +580,213 @@ func TestQueryBadSortType(t *testing.T) {
...
+
+	for name, tc := range map[string]struct {
+		command    bson.D
+		err        *mongo.CommandError
+		altMessage string
+	}{
+		"BadMaxTimeMSTypeDouble": {
+			command: bson.D{
+				{"find", collection.Name()},
+				{"projection", bson.D{{"value", "some"}}},
```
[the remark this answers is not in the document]
DoodgeMatvey: Fixed

integration/query_test.go
```diff
@@ -579,6 +580,213 @@ func TestQueryBadSortType(t *testing.T) {
...
+				{"maxTimeMS", -14245345234123245.55},
+			},
+			err: &mongo.CommandError{
+				Code:    2,
+				Name:    "BadValue",
+				Message: "-14245345234123246 value for maxTimeMS is out of range",
+			},
+			altMessage: "-14245345234123246 value for maxTimeMS is out of range",
```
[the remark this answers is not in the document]
DoodgeMatvey: Ok, fixed

integration/query_test.go
```diff
@@ -579,6 +580,213 @@ func TestQueryBadSortType(t *testing.T) {
...
+				{"find", collection.Name()},
+				{"projection", bson.D{{"value", "some"}}},
+				{"maxTimeMS", 43.15},
+			},
+			err: &mongo.CommandError{
+				Code:    2,
+				Name:    "BadValue",
+				Message: "maxTimeMS has non-integral value",
+			},
+			altMessage: "maxTimeMS must be a whole number",
```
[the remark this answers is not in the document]
DoodgeMatvey: Yes, fixed

internal/handlers/pg/msg_find.go
```diff
@@ -70,6 +70,18 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
 		return nil, err
 	}
 
+	maxTimeMS, err := common.GetOptionalPositiveNumber(document, "maxTimeMS")
+	if err != nil {
+		return nil, err
+	}
+
+	if maxTimeMS != 0 {
+		ctxWithTimeout, runCancel := context.WithTimeout(ctx, time.Duration(maxTimeMS)*time.Millisecond)
```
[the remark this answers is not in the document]
DoodgeMatvey: renamed

internal/handlers/common/params.go
DoodgeMatvey: fixed

internal/handlers/common/params.go
DoodgeMatvey: fixed

w84thesun (dismissed):

AlekSi (commented):
On hold until #784 is merged. Then we should port those changes to Tigris handler.

AlekSi:
@DoodgeMatvey The linked PR was merged. Please port those changes to the new Tigris handler.

w84thesun (approved):

AlekSi (approved):

internal/handlers/pg/msg_find.go
```diff
@@ -73,6 +73,18 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
 		return nil, err
 	}
 
+	maxTimeMS, err := common.GetOptionalPositiveNumber(document, "maxTimeMS")
+	if err != nil {
+		return nil, err
+	}
+
+	if maxTimeMS != 0 {
+		ctxWithTimeout, cancel := context.WithTimeout(ctx, time.Duration(maxTimeMS)*time.Millisecond)
```
rumyantseva, line 82: Yeah, this is the exact thing I expect this param to do. Why don't we have the same in `msg_findandmodify.go`?

w84thesun: Now it's fixed! 😃

w84thesun (approved):

rumyantseva (approved):
LGTM

PR #609, opened 2022-05-17
Better compatibility for `updateOne` command

### LINKED ISSUE #571: `update`'s `nModified` is incorrect in some cases
@GinGin3203 added two test cases at #546 that pass with MongoDB, but fail with FerretDB. We should fix those issues.

GinGin3203:
Closes #571.


AlekSi:
@GinGin3203 Ping :) Do we still have this issue?

GinGin3203:
@AlekSi yep, sry for such a drag, it's life stuff
I was just going to do it today

GinGin3203:
I've added a test (`SetSameValueComposite`) that depends on `types.Compare` to work correctly to work with composite types, so this test fails for FerretDB right now.
Do I comment it out with something like `// TODO wait until types.Compare works with composite types` or wait until `types.Compare` is fixed?

Other than that, it's ready for review

AlekSi (commented):

GinGin3203:
@AlekSi done

AlekSi:
@ribaraka PTAL

internal/types/compare.go
```diff
@@ -259,6 +263,26 @@ func compareNumbers(a float64, b int64) CompareResult {
 	return CompareResult(bigA.Cmp(bigB))
 }
 
+// compareDocuments checks if two documents are equal.
+func compareDocuments(a, b *Document) CompareResult {
```
ribaraka, line 267: we have a function that checks if two documents are equal to each other, it's called `matchDocuments`. But since I plan to work on the `compareDocuments` function, you can leave your made function, and delete the match.go file where the `matchDocuments` is located

internal/types/compare.go
```diff
@@ -304,9 +328,20 @@ func compareArrays(filterArr, docArr *Array) CompareResult {
...
+			if err != nil {
+				return Incomparable
+			}
+			filterDoc, ok := filterValue.(*Document)
+			if !ok {
+				return Incomparable
+			}
+			if compareDocuments(arrValue, filterDoc) != Equal {
+				return Incomparable
+			}
```
ribaraka, line 344: try not to immediately return a `Incomparable`, because sometimes the value that satisfies the filter might be at the end of the array. Use `continue` instead of returns.

GinGin3203: Yeah, it's a silly mistake, thanks for noticing that

internal/types/compare.go
GinGin3203: Well, there is still too much depending on matchDocuments, so it can't be deleted because of backward compatibility

AlekSi:
A lot of code changed since then, so we re-did it in #949.

PR #613, opened 2022-05-18
Improve negative zero handling

### LINKED ISSUE #612: Improve negative zero handling
We should handle negative zero with the same care as NaNs, ±Infinity, etc.
* add it to bson tests;
* add it to fjson tests;
* add it to `testutil.AssertEqual`;
* etc.

AlekSi:
Closes #612.

### DIFF
--- a/internal/util/testutil/equal.go
+++ b/internal/util/testutil/equal.go
@@ -76,6 +76,7 @@ func diffValues[T types.Type](tb testing.TB, expected, actual T) (expectedS stri
 
 // equal compares any BSON values in a way that is useful for tests:
 //  * float64 NaNs are equal to each other;
+//  * float64 zero values are compared with sign (math.Copysign(0, -1) != math.Copysign(0, +1));
 //  * time.Time values are compared using Equal method.
 //
 // This function is for tests; it should not try to convert values to different types before comparing them.
@@ -173,6 +174,9 @@ func equalScalars(tb testing.TB, v1, v2 any) bool {
 		}
 		if math.IsNaN(s1) {
 			return math.IsNaN(s2)
+		}
+		if s1 == 0 && s2 == 0 {
+			return math.Signbit(s1) == math.Signbit(s2)
 		}
 		return s1 == s2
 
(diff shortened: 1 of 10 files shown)


AlekSi:
@ribaraka I think I moved everything relevant from #607 plus added tests. PTAL

ribaraka:
LGTM! Thanks for the notice!

ribaraka (approved):

internal/util/testutil/equal.go
```diff
@@ -174,6 +175,9 @@ func equalScalars(tb testing.TB, v1, v2 any) bool {
 		if math.IsNaN(s1) {
 			return math.IsNaN(s2)
 		}
+		if s1 == 0 && s2 == 0 {
+			return math.Signbit(s1) == math.Signbit(s2)
+		}
```
seeforschauer, line 180: why not: 
math.Signbit(s1) == math.Signbit(s2) && s1 == s2?

seeforschauer: it will be returning `equality` even if signs are not equal, according to the code below. So maybe it doesn't make sense. WDYT?

AlekSi: Assuming you mean

```go
if math.Signbit(s1) == math.Signbit(s2) && s1 == s2 {
  return true
}

return s1 == s2
```

that will not pass `AssertNotEqual(t, math.Copysign(0, +1), math.Copysign(0, -1))` test: condition under `if` will return `false`, but `==` is defined to return `true` for `math.Copysign(0, +1) == math.Copysign(0, -1)`. That's exactly why we need that PR at all

seeforschauer: ok, that unexpected behavior of `AssertNotEqual`

AlekSi: What's unexpected about it?

AlekSi: `AssertNotEqual(t, math.Copysign(0, +1), math.Copysign(0, -1))` passes with merged changes (values are not equal). Suggest change (`math.Signbit(s1) == math.Signbit(s2) && s1 == s2`) would fail that test

PR #615, opened 2022-05-18
Add log variable to integration tests

### LINKED ISSUE #470: Add optional log level to integration test runs
When we ran integration tests over FerretDB logger initialized with Debug level which results in big output when errors happens.
We could add optional parameter which would control logging level.

w84thesun:
Closes #470


Taskfile.yml
```diff
@@ -87,26 +87,26 @@ tasks:
     desc: "Run all integration tests in parallel"
     deps:
       - task: test-integration-db
-        vars: { DB: "FerretDB" }
+        vars: { DB: "FerretDB", LOG: "debug" }
```
seeforschauer, line 90: I would suggest default log level info, I set it all the time. WDYT?

w84thesun: I set debug level as I don't want to change the current behavior.

seeforschauer (approved):

w84thesun:
@AlekSi suggested another solution for #470, so I think we can just close this one.

PR #618, opened 2022-05-19
Support array querying

w84thesun:
Refs #73


integration/query_array_test.go
```diff
@@ -137,3 +140,96 @@ func TestQueryArraySize(t *testing.T) {
...
+			filter:      bson.D{{"value.array.2.foo", "bar"}},
+			expectedIDs: []any{},
+		},
+
+		"FieldPositionQueryRegex": {
+			filter: bson.D{{"value.array.0", bson.D{{"$lt", primitive.Regex{Pattern: "^$"}}}}},
+			err: &mongo.CommandError{
+				Code:    2,
+				Name:    "BadValue",
+				Message: "Can't have RegEx as arg to predicate over field 'value.array.0'.",
```
seeforschauer, line 214: It's already been covered, no?

internal/handlers/common/filter.go
```diff
@@ -58,11 +59,33 @@ func filterDocumentPair(doc *types.Document, filterKey string, filterValue any)
...
+				return false, nil
+			}
+
+			if docValue.Len() == 0 {
+				return false, nil
+			}
+
+			value, err := docValue.Get(index)
+			if err != nil {
+				return false, err
```
seeforschauer, line 79: Can you please add a test with an index value greater than array size to test this line?

integration/query_array_test.go
```diff
@@ -137,3 +140,96 @@ func TestQueryArraySize(t *testing.T) {
...
+	}{
+		"PositionDouble": {
+			filter:      bson.D{{"value.0", bson.D{{"$type", "double"}}}},
+			expectedIDs: []any{"array-double"},
+		},
+		"PositionString": {
+			filter:      bson.D{{"value.0", bson.D{{"$type", "string"}}}},
+			expectedIDs: []any{"array-strings"},
+		},
+		"PositionNull": {
```
seeforschauer, line 176: If it refers to `PositionRegex` and it was expected to put here nil, not `null` string value, then here is a typo. 
If it's not a type I suggest renaming the test.

internal/handlers/common/filter.go
```diff
@@ -58,11 +59,33 @@ func filterDocumentPair(doc *types.Document, filterKey string, filterValue any)
...
+			}
+
+			// We shouldn't compare with embedded array values.
```
seeforschauer, line 82: A comment is not clear to me. Why do we create an empty array here? Can you please add a code snippet to note? Thank you.

integration/query_array_test.go
w84thesun: It's not been covered with dot notation field.

internal/handlers/common/filter.go
w84thesun: Fixed it, we should just return `false`.

integration/query_array_test.go
```diff
@@ -137,3 +140,110 @@ func TestQueryArraySize(t *testing.T) {
...
+			filter:      bson.D{{"value.0", primitive.Regex{Pattern: "foo"}}},
+			expectedIDs: []any{"array-strings"},
+		},
+		"PositionArray": {
+			filter:      bson.D{{"value.0", primitive.A{}}},
+			expectedIDs: []any{},
+		},
+
+		"NoSuchFieldPosition": {
+			filter:      bson.D{{"value.some.0", bson.A{42}}},
```
rumyantseva, line 199: Do we also need a test when the filter is an array and we are able to find something? (for example, if a doc stores `[42]` array in `value.some.0`?

internal/handlers/common/filter.go
rumyantseva: +1! Me neither! I don't understand why we shouldn't compare!

rumyantseva (commented):
It's not super straightforward what's in the scope of that task... 

I remember we had `$unset`, `$set`, `$setOnInsert`, `$currentDate`, and maybe some other cases where the support of dot notation wasn't in scope. Is it a part of the current task to check all those cases? Or will they be checked separately?

seeforschauer:
> It's not super straightforward what's in the scope of that task...

Once upon a time tests were disabled (I already don't remember why). 

> I remember we had `$unset`, `$set`, `$setOnInsert`, `$currentDate`, and maybe some other cases where the support of dot notation wasn't in scope. Is it a part of the current task to check all those cases? Or will they be checked separately?

It's a separate task.

integration/query_array_test.go
w84thesun: I don't get the problem here. I'm not covering regex error, I'm covering dot notation. It turned out that you can't use the query as we have in that test.

internal/handlers/common/filter.go
w84thesun: What should the right comment be?

integration/query_array_test.go
```diff
@@ -137,3 +140,110 @@ func TestQueryArraySize(t *testing.T) {
...
+
+		"PositionTypeDouble": {
+			filter:      bson.D{{"value.0", bson.D{{"$type", "double"}}}},
+			expectedIDs: []any{"array-double"},
+		},
+		"PositionTypeString": {
+			filter:      bson.D{{"value.0", bson.D{{"$type", "string"}}}},
+			expectedIDs: []any{"array-strings"},
+		},
+		"PositionTypeNull": {
```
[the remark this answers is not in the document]
w84thesun: Those 4 tests shows that MongoDB doesn't check embedded arrays.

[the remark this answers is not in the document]
seeforschauer: > Those 4 tests shows that MongoDB doesn't check embedded arrays.

can it be done with just one test?

internal/handlers/common/filter.go
seeforschauer: > What should the right comment be?

`MongoDb compatibility`: <that comment> 

Can it be added?

AlekSi: I think that comment is a bit redundant in FerretDB: the whole project is about MongoDB compatibility. Being compatible is the norm; being incompatible is an exception that should be documented.

AlekSi:
@ribaraka PTAL, as there may be an overlap with your work

internal/handlers/common/filter.go
w84thesun: Removed.

integration/query_array_test.go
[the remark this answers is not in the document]
w84thesun: Fixed.

seeforschauer (approved):

ribaraka:
LGTM! 

> The scope of this task is to check how dot notation works with arrays, add tests and fix anything that is not working.

^ Considering this, I would probably call the PR like "Support array index position query" or something. But I guess it's not biggy. :smile:

ribaraka (approved):

rumyantseva (approved):
LGTM, discussed all the questions in person.

AlekSi (approved):

Issue #587, opened 2022-05-11
Enable linter that would ensure that all functions and methods are documented


AlekSi:
If one exists, use it. If not, make one (probably with go-ruleguard?)

PR #619, opened 2022-05-20
Add some comments to the functions and variables

seeforschauer:


w84thesun:
Does this linter checks unexported functions and methods?

AlekSi:
It checks all code. If you mean if it checks _documentation_ for unexpected methods and functions, then no, and currently, there is no such linter, according to Gophers Slack.

AlekSi:
Yeah, this PR does not refs 587

seeforschauer:
> Yeah, this PR does not refs #587

Removed #587. 

But it detects, isn't it?

seeforschauer:
> > Removed #587.
> 
> Both title and description still reference this issue.
> 
> > But it detects, isn't it?
> 
> Sorry, I don't understand the question. It (what it?) detects what?

It detects that comment absent for a variable or a function.

AlekSi:
Yes, but only for exported names. #587 is about all functions/methods, including unexpected ones.

seeforschauer:
> Yes, but only for exported names. #587 is about all functions/methods, including unexpected ones.

understood, thank you.

seeforschauer:
> Both title and description still reference this issue.

Renamed PR. Actually, it is just to make smaller PRs, here is not only comments that were added.

internal/handlers/common/error.go
```diff
@@ -34,19 +34,46 @@ const (
...
-	ErrProjectionInEx    = ErrorCode(31253) // Location31253
-	ErrProjectionExIn    = ErrorCode(31254) // Location31254
-	ErrRegexOptions      = ErrorCode(51075) // Location51075
+	// ErrBadValue for wrong input.
+	ErrBadValue = ErrorCode(2) // BadValue
+
+	// ErrFailedToParse failed to parse user input.
+	ErrFailedToParse = ErrorCode(9) // FailedToParse
+
+	// ErrTypeMismatch for $sort, when it is expected field sort to be of type object but is has other type.
```
rumyantseva, line 43: Are those codes some MongoDB's codes or something FerretDB-specific? The `it is expected field sort to be` part of the comment doesn't look clear to me. Is _"field sort"_ actually _field's sort_ or is it a field called _"sort"_? I was trying to find something Mongo-related for it in MongoDB docs, but I'm not sure where to look for it.

internal/handlers/pg/pgdb/placeholder.go
```diff
@@ -16,8 +16,10 @@ package pgdb
 
 import "strconv"
 
+// Placeholder is a numbered mark where to put variable in the postgresql query.
```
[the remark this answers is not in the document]
seeforschauer: both are obvious :D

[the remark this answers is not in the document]
seeforschauer: they are obvious to us, may I put here both definitions? WDYT?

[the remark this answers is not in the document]
seeforschauer: changed both comments (here and below)

internal/handlers/common/error.go
```diff
@@ -34,20 +34,46 @@ const (
...
-	ErrInvalidArg        = ErrorCode(28667) // Location28667
-	ErrSliceFirstArg     = ErrorCode(28724) // Location28724
-	ErrProjectionInEx    = ErrorCode(31253) // Location31253
-	ErrProjectionExIn    = ErrorCode(31254) // Location31254
-	ErrRegexOptions      = ErrorCode(51075) // Location51075
+	// ErrBadValue for wrong input.
+	ErrBadValue = ErrorCode(2) // BadValue
+
+	// ErrFailedToParse failed to parse user input.
+	ErrFailedToParse = ErrorCode(9) // FailedToParse
```
rumyantseva, line 41: Overall, here and below I miss verbs in comments. Usually, in the Go source code, there is a verb explaining what the variable is used for. For example, for errors, it could be "indicates" [like in the os package](https://cs.opensource.google/go/go/+/refs/tags/go1.18.2:src/os/error.go;l=18).

On the other hand, in the Go source code, not all the errors are documented as detailed as you did it here :) So, please treat my comment as an opinion, not a call for action.

seeforschauer: I welcome your feedback. I believe you added an example here to be more understandable to read when I add this. Anyway, this is a pure PR for comments.

internal/clientconn/conn.go
```diff
@@ -79,10 +79,6 @@ type newConnOpts struct {
 
 // newConn creates a new client connection for given net.Conn.
 func newConn(opts *newConnOpts) (*conn, error) {
-	if opts.handler == nil {
```
rumyantseva, line 82: Why is this deleted? Is this `opts.handler == nil` condition checked somewhere else?

internal/handlers/pg/pgdb/pool.go
```diff
@@ -41,7 +41,10 @@ const (
 )
 
 var (
-	ErrNotExist     = fmt.Errorf("schema or table does not exist")
+	// ErrNotExist is an error when schema or table do not exist whily attempt to access it.
```
rumyantseva, line 44:
```suggestion
	// ErrNotExist is an error when schema or table does not exist while attempting to access it.
```

internal/handlers/pg/pgdb/pool.go
```diff
@@ -41,7 +41,10 @@ const (
...
+	ErrNotExist = fmt.Errorf("schema or table does not exist")
+
+	// ErrAlreadyExist is an error when schema or table already exist while creating it.
```
rumyantseva, line 47:
```suggestion
	// ErrAlreadyExist is an error when schema or table already exists while creating it.
```

internal/util/version/version.go
```diff
@@ -40,6 +40,7 @@ var (
 	branch string
 )
 
+// Info contains this build details.
```
rumyantseva, line 43:
```suggestion
// Info contains current build details.
```

internal/util/version/version.go
```diff
@@ -51,6 +52,7 @@ type Info struct {
 
 var info *Info
 
+// Get returns info.
```
rumyantseva, line 55:
```suggestion
// Get returns current build's info.
```

rumyantseva (changes requested):
Added a couple of questions/suggestions, and a couple of typo corrections.
In general, looks good to me.

internal/clientconn/conn.go
seeforschauer: it's deleted bcz of linter error (cannot compare handler to nil)

AlekSi: Well, `opts.handler` is an interface, and it can be nil. Please revert that change

rumyantseva: @AlekSi should it panic though? Or should it return an error instead?

seeforschauer: > Well, `opts.handler` is an interface, and it can be nil. Please revert that change

done

AlekSi: > should it panic though? Or should it return an error instead?

It should not be possible for a user to do something that would trigger that case, so that's a programmer's error, and we "handle" it by panicking

rumyantseva: Ok! Sounds reasonable!

rumyantseva (approved):
LGTM

internal/handlers/common/error.go
```diff
@@ -25,7 +25,7 @@ import (
 
 //go:generate ../../../bin/stringer -linecomment -type ErrorCode
 
-// ErrorCode represents wire protocol error code.
+// ErrorCode represents FerretDB error codes.
```
AlekSi, line 28: I think it makes it less clear. Everything in the FerretDB code base is about FerretDB. And the value of that type represents a single error code, not many error codes.

internal/handlers/pg/pgdb/placeholder.go
```diff
@@ -16,8 +16,10 @@ package pgdb
 
 import "strconv"
 
+// Placeholder stores the number of the relevant placeholder of the query.
 type Placeholder int
 
+// Next increases the identifier value for the next variable in the postgresql query.
```
AlekSi, line 22:
```suggestion
// Next increases the identifier value for the next variable in the PostgreSQL query.
```

internal/types/binary.go
```diff
@@ -20,14 +20,29 @@ package types
 type BinarySubtype byte
 
 const (
-	BinaryGeneric    = BinarySubtype(0x00) // generic
-	BinaryFunction   = BinarySubtype(0x01) // function
+	// BinaryGeneric BSON generic binary subtype.
```
AlekSi, line 23: There should be a verb like "represents"

internal/types/types.go
```diff
@@ -48,6 +48,7 @@ import (
 	"time"
 )
 
+// MaxDocumentLen is maximum BSON object size.
```
AlekSi, line 51:
```suggestion
// MaxDocumentLen is the maximum BSON object size.
```

I think it is worth checking comments with Grammarly

AlekSi (changes requested):

internal/clientconn/conn.go
```diff
@@ -82,7 +82,6 @@ func newConn(opts *newConnOpts) (*conn, error) {
 	if opts.handler == nil {
 		panic("handler required")
 	}
-
```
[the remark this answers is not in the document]
seeforschauer: reverted.

internal/types/binary.go
seeforschauer: added.

internal/types/regex.go
```diff
@@ -21,17 +21,39 @@ import (
 )
 
 var (
+	// ErrOptionNotImplemented for unimplemented regex option.
```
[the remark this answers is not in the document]
seeforschauer: added verbs.

internal/handlers/common/error.go
seeforschauer: reverted.

rumyantseva (dismissed):
I don't have major questions about the comments. This is a significant improvement in documentation already!

seeforschauer:
> There are merge conflicts

resolved

rumyantseva (dismissed):
LGTM

internal/types/regex.go
```diff
@@ -21,17 +21,39 @@ import (
...
+	ErrTrailingBackslash = fmt.Errorf("Regular expression is invalid: \\ at end of pattern")
+
+	// ErrNothingToRepeat indicates invalid regex: nothing to repeat.
+	ErrNothingToRepeat = fmt.Errorf("Regular expression is invalid: nothing to repeat")
+
+	// ErrInvalidClassRange indicates that range out of order in character class.
+	ErrInvalidClassRange = fmt.Errorf("Regular expression is invalid: range out of order in character class")
+
+	//nolint:godot // false positive
+	// ErrUnsupportedPerlOp indicates for unrecognized character after (? or (?-.
```
w84thesun, line 52: This could be rephrased:
```suggestion
	// ErrUnsupportedPerlOp indicates unrecognized character after the grouping sequence start.
```

w84thesun (changes requested):

w84thesun (approved):

rumyantseva (dismissed):
LGTM

internal/util/testutil/testutil.go
```diff
@@ -20,6 +20,7 @@ import (
 	"testing"
 )
 
+// Ctx returns a new empty context.
```
AlekSi, line 23:
```suggestion
// Ctx returns test context.
```

In the future, it may return a context with a deadline/timeout for graceful cancelation, etc.

AlekSi (approved):
:shipit:

Issue #627, opened 2022-05-23
Support `$set` field update operator 


AlekSi:
Add tests for edge cases like `[]`, `null`, etc.

Dot notation is not in scope.

seeforschauer:
```go
"SetMalformedDoc": {
			id:     "string",
			update: bson.D{{"$set", bson.D{{"$k", "key"}}}},
			err: &mongo.WriteError{
				Code: 52,
				Message: "The dollar ($) prefixed field '$k' in '$k' is not allowed in the context " +
					"of an update's replacement document. Consider using an aggregation pipeline with $replaceWith.",
			},
		},
```
I would suggest to do in a separate PR.

seeforschauer:
Test case:
```go
		"SetMany": {
			id:     "string",
			update: bson.D{{"$set", bson.D{{"foo", int32(1)}, {"bar", bson.A{}}}}},
			stat: &mongo.UpdateResult{
				MatchedCount:  1,
				ModifiedCount: 1,
				UpsertedCount: 0,
			},
			result: bson.D{{"_id", "string"}, {"value", "foo"}, {"bar", bson.A{}}, {"foo", int32(1)}},
		},
```

Mongo givies:
`bson.D{{"_id", "string"}, {"value", "foo"}, {"bar", bson.A{}}, {"foo", int32(1)}},`

FerretDB gives: 
`bson.D{{"_id", "string"}, {"value", "foo"}, {"foo", int32(1)}, {"bar", bson.A{}},`

PR #633, opened 2022-05-24
Support `$set` field update operator - field order test case

seeforschauer:
Refs #627.

Test case:
```go
"SetMany": {
	id:     "string",
	update: bson.D{{"$set", bson.D{{"foo", int32(1)}, {"bar", bson.A{}}}}},
	stat: &mongo.UpdateResult{
		MatchedCount:  1,
		ModifiedCount: 1,
		UpsertedCount: 0,
	},
	result: bson.D{{"_id", "string"}, {"value", "foo"}, {"bar", bson.A{}}, {"foo", int32(1)}},
},
```
Please review after https://github.com/FerretDB/FerretDB/pull/634.


seeforschauer:
It's in https://github.com/FerretDB/FerretDB/pull/634 now.

PR #634, opened 2022-05-24
Support `$set` field update operator

### LINKED ISSUE #627: Support `$set` field update operator

seeforschauer:
Closes #627.


seeforschauer:
I need a `WriteErr` from 596 here, so added as a file.

seeforschauer:
State to draft: I'm about to add some tests for the func `TestAssertEqualWriteError`.

internal/handlers/common/update.go
```diff
@@ -18,33 +18,60 @@ import (
...
 				}
+				return true, nil
+
+			case *types.Array:
+				return false, NewWriteErrorMsg(ErrFailedToParse, fmt.Sprintf(msgFmt, "array", "[]"))
+
+			case float64:
+				return false, NewWriteErrorMsg(
+					ErrFailedToParse,
+					fmt.Sprintf(`Modifiers operate on fields but we found type double instead. `+
```
w84thesun, line 60: You could use `AliasFromType` to get the correct type name for all supported types.

internal/handlers/common/update.go
```diff
@@ -18,33 +18,60 @@ import (
...
 
 		switch updateOp {
 		case "$set":
-			setDoc, err := AssertType[*types.Document](updateV)
-			if err != nil {
-				return err
+			msgFmt := `Modifiers operate on fields but we found type %s instead. ` +
+				`For example: {$mod: {<field>: ...}} not {$set: %s}`
+
+			if updateV == nil ||
```
w84thesun, line 35: This check could be placed inside switch, isn't it?

seeforschauer: this is check for the value, not for the type. Can I?

internal/handlers/common/update.go
seeforschauer: It's shorter but it doesn't work...

seeforschauer: Maybe writing another function for type formatting in errors would help.

seeforschauer: alternatively, we can change the error msg for all types in order to not to marshal everything - I remember someone was about to create `.String()` method for the document only for the error message and we ended up with an alt message. 

I'll do it as we did before.

internal/handlers/common/update.go
seeforschauer: after I've changed the code for https://github.com/FerretDB/FerretDB/pull/634#discussion_r881583187 comment, it works in the `switch` case.

internal/handlers/common/update.go
```diff
@@ -16,35 +16,48 @@ package common
...
-			setDoc, err := AssertType[*types.Document](updateV)
-			if err != nil {
-				return err
-			}
 
-			for _, setKey := range setDoc.Keys() {
-				setValue := must.NotFail(setDoc.Get(setKey))
-				if err = doc.Set(setKey, setValue); err != nil {
-					return lazyerrors.Error(err)
+			switch setDoc := updateV.(type) {
```
rumyantseva, line 34: Does it make sense to move this switch to a separate function and call if from here? I guess, the idea is to support more `updateOp`s in the future, and then the parent switch-case will be really huge. Also, if it's a separate function it's possible (and easier) to write tests to cover this specific logic.

internal/handlers/common/error.go
```diff
@@ -51,7 +51,41 @@ const (
 	ErrRegexMissingParen = ErrorCode(51091) // Location51091
 )
 
-// Error represents wire protocol error.
+// ProtoErr represents protocol error type.
+type ProtoErr interface {
```
rumyantseva, line 55: Why is this type introduced? I feel like I've seen `ProtoErr` somewhere else before... 
Would it be possible to document the reason / the use cases for this type to be here? E.g. where/when does one need to use this interface? 

The only usage I see is in the `ProtocolError` function which seems to be quite a generic wrapper. Why this PR is the first time this function was introduced? How should it be used in the future (e.g. when more operators are implemented)?

rumyantseva (commented):
Not a full review yet, just a couple of questions to bring more clarity.

internal/handlers/common/error.go
seeforschauer: > I need a WriteErr from https://github.com/FerretDB/FerretDB/pull/596 here, so added as a file.

It will be just copied from the other PR #596 as is, I just need to go further using that code.

internal/handlers/common/update.go
seeforschauer: > Does it make sense to move this switch to a separate function and call if from here?

I think it's small enough, trying to keep it simple. 

> I guess, the idea is to support more `updateOp`s in the future, and then the parent switch-case will be really huge. 

when it becomes then it can be extracted. 

> Also, if it's a separate function it's possible (and easier) to write tests to cover this specific logic.

make sense. But it's covered in integration tests. WDYT of it?

internal/handlers/common/error.go
```diff
@@ -191,6 +192,8 @@ func (we *WriteErrors) Document() *types.Document {
 			"errmsg", e.err,
 		))))
 	}
+	return errUnset
+}
 
 	// "writeErrors" field must be present in the result document so that clients can parse it as WriteErrors.
 	d := must.NotFail(types.NewDocument(
```
AlekSi, line 199: Merge conflict, build fails

AlekSi (changes requested):

internal/handlers/common/error.go
seeforschauer: in progress.

rumyantseva (dismissed):
I don't have new comments. Codecov complains in a few places, but I guess it is considered?

integration/query_set_test.go
```diff
@@ -0,0 +1,287 @@
...
+	err    *mongo.WriteError
+	stat   *mongo.UpdateResult
+	alt    string
+}
+
+func TestSetOperatorOnString(t *testing.T) {
+	t.Parallel()
+
+	for name, tc := range map[string]testCase{
+		"SetMany": {
```
w84thesun, line 39: We already have `Set` in the test name so it's ok to skip it in the test case name. It's just easier to read test cases' names without it.

w84thesun (changes requested):

seeforschauer:
> I don't have new comments. Codecov complains in a few places, but I guess it is considered?

Tests to be added in this PR:
- dot notation is not supported
- unhandled operation $foo


Tests for `update.go:44`
```go
 if err = doc.Set(setKey, setValue); err != nil {
```
will be added in https://github.com/FerretDB/FerretDB/issues/643.

integration/query_set_test.go
seeforschauer: changed.

integration/helpers.go
```diff
@@ -186,20 +186,37 @@ func CollectKeys(t testing.TB, doc bson.D) []string {
 }
 
 // AssertEqualWriteError compares expected mongo.WriteError Message and Code with actual error.
-func AssertEqualWriteError(t *testing.T, expected *mongo.WriteError, actual error) bool {
+func AssertEqualWriteError(t *testing.T, expected *mongo.WriteError, alt string, actual error) bool {
```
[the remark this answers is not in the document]
seeforschauer: updated.

w84thesun (dismissed):

integration/helpers.go
```diff
@@ -203,3 +203,39 @@ func AssertEqualWriteError(t *testing.T, expected *mongo.WriteError, actual erro
...
+	t.Helper()
+
+	var actualWE mongo.WriteError
+	actualEx, ok := actual.(mongo.WriteException)
+	if ok {
+		if len(actualEx.WriteErrors) != 1 {
+			return assert.Equal(t, expected, actual)
+		}
+		actualWE = actualEx.WriteErrors[0]
+	} else {
```
AlekSi, line 218: I would expect `AssertEqualAltWriteError` to do the same as `AssertEqualWriteError` with additional check for alternative error message. But it does much more: checks for both WriteException and WriteError, logs a lot, etc. I think it should not do all that.

integration/helpers_test.go
```diff
@@ -0,0 +1,96 @@
...
+package integration
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+func TestAssertEqualAltWriteError(t *testing.T) {
```
AlekSi, line 25: I think test assertions should be simple enough so we don't have to write tests for tests. Otherwise, we will have to write tests for tests for tests.

integration/query_set_test.go
```diff
@@ -0,0 +1,287 @@
+// Copyright 2021 FerretDB Set.
```
AlekSi, line 1: That should be in update_test.go or at update_something.go. query_XXX is for query tests

integration/query_set_test.go
```diff
@@ -0,0 +1,287 @@
...
+import (
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+type testCase struct {
```
AlekSi, line 26: That should not be a package-level type. All other tests define their own test case types. Why disregard that here?

integration/query_set_test.go
```diff
@@ -0,0 +1,287 @@
...
+	id     string
+	update bson.D
+	result bson.D
+	err    *mongo.WriteError
+	stat   *mongo.UpdateResult
+	alt    string
+}
+
+func TestSetOperatorOnString(t *testing.T) {
```
AlekSi, line 35: There is a common pattern for test functions that we should follow

integration/query_set_test.go
```diff
@@ -0,0 +1,287 @@
...
+				Code: 9,
+				Message: "Modifiers operate on fields but we found type string instead. " +
+					"For example: {$mod: {<field>: ...}} not {$set: \"string\"}",
+			},
+			alt: "Modifiers operate on fields but we found type string instead. " +
+				"For example: {$mod: {<field>: ...}} not {$set: string}",
+		},
+		"Double": {
+			id:     "string",
+			update: bson.D{{"$set", float64(42.12345)}},
```
AlekSi, line 71: I think that this test, and several following it, does not add match compared to the string test above

internal/handlers/pg/msg_update.go
```diff
@@ -93,13 +93,13 @@ func (h *Handler) MsgUpdate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
 
 		var q, u *types.Document
-		var upsert bool
 		if q, err = common.GetOptionalParam(update, "q", q); err != nil {
 			return nil, err
 		}
 		if u, err = common.GetOptionalParam(update, "u", u); err != nil {
 			return nil, err
 		}
+		var upsert bool
```
AlekSi, line 102: I think we could move it back

AlekSi (changes requested):

integration/helpers.go
seeforschauer: Would you please propose your solution in that case?

integration/helpers_test.go
seeforschauer: > I think test assertions should be simple enough so we don't have to write tests for tests. Otherwise, we will have to write tests for tests for tests.

I don't mind writing tests for test functions. I want to feel I wrote a function that works as expected and to feel calm with tests proving it.

internal/handlers/pg/msg_update.go
seeforschauer: revert.

integration/query_set_test.go
seeforschauer: changed.

integration/query_set_test.go
seeforschauer: renamed.

integration/query_set_test.go
seeforschauer: removed.

integration/query_set_test.go
seeforschauer: changed.

integration/helpers.go
```diff
@@ -203,3 +203,39 @@ func AssertEqualWriteError(t *testing.T, expected *mongo.WriteError, actual erro
 	return assert.Equal(t, expected.Message, actualWriteErr.Message) &&
 		assert.Equal(t, expected.Code, actualWriteErr.Code)
 }
+
+// AssertEqualAltWriteError compares expected mongo.WriteError Message and Code with actual error.
+func AssertEqualAltWriteError(t *testing.T, expected *mongo.WriteError, alt string, actual error) bool {
```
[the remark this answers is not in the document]
seeforschauer: done.

integration/helpers.go
AlekSi: I propose to remove checking for error type that is not returned by the driver and logging. I propose to make this function just like AssertEqualWriteError with an additional check for the alternative error message and remove everything unrelated to it.

integration/helpers_test.go
AlekSi: But how do you prove that tests work as expected? Will you write tests for tests?
Let's make this test helper so simple, so we will not need tests for it

integration/update_test.go
```diff
@@ -279,3 +279,186 @@ func TestUpdateIncOperator(t *testing.T) {
...
+			})
+			require.NoError(t, err)
+
+			res, err := collection.UpdateOne(ctx, bson.D{{"_id", tc.id}}, tc.update)
+			if tc.err != nil {
+				if !AssertEqualAltWriteError(t, tc.err, tc.alt, err) {
+					t.Logf("%[1]T %[1]v", err)
+					t.FailNow()
+				}
+				return
```
AlekSi, line 452: Let's make this block the same as in other tests

integration/update_test.go
```diff
@@ -279,3 +279,186 @@ func TestUpdateIncOperator(t *testing.T) {
...
+			}
+			require.Equal(t, tc.stat, res)
```
AlekSi, line 454: `require.NoError(t, err)`

AlekSi (changes requested):

integration/helpers.go
seeforschauer: `Array` and `NilOperand` returning `mongo.WriteException`

seeforschauer: @AlekSi  I've changed the function, but it still can accept `mongo.Exception` and `mongo.WriteError`.

integration/helpers_test.go
seeforschauer: This test function has been removed.

integration/update_test.go
seeforschauer: added

integration/update_test.go
seeforschauer: log removed, added check for res.

integration/helpers.go
AlekSi: // WriteError is an error that occurred during execution of a write operation.
// **This error type is only returned as part of a WriteException or BulkWriteException.**

internal/handlers/common/update.go
```diff
@@ -16,36 +16,48 @@ package common
...
+			case *types.Document:
+				if setDoc.Len() == 0 {
+					return false, nil
+				}
+				var err error
```
w84thesun, line 40: `err` variable is not used anywhere outside the `for` block. We can remove it.

internal/handlers/pg/msg_update.go
```diff
@@ -160,15 +160,20 @@ func (h *Handler) MsgUpdate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 		matched += int32(len(resDocs))
 
 		for _, doc := range resDocs {
-			if err = common.UpdateDocument(doc, u); err != nil {
-				return nil, lazyerrors.Error(err)
+			var changed bool
```
w84thesun, line 163: Let’s change this declaration to `:=` to make all declarations the same inside `for` block.

w84thesun (changes requested):

internal/handlers/pg/msg_update.go
seeforschauer: done.

internal/handlers/common/update.go
seeforschauer: done.

internal/handlers/pg/msg_update.go
w84thesun: But it looks like nothing changed. 😃

seeforschauer: sorry, updated.

rumyantseva (approved):
I might be unaware of some cases, but so far the code looks good to me.

w84thesun (approved):

integration/helpers.go
AlekSi: > But it does much more: checks for both WriteException and WriteError

> propose to remove checking for error type that is not returned by the driver

> `Array` and `NilOperand` returning `mongo.WriteException`

Ok, what about the second type? Let's check documentation:

> // WriteError is an error that occurred during execution of a write operation.
> // **This error type is only returned as part of a WriteException or BulkWriteException.**

`actual` can never be `WriteError` because that error is never returned by the driver.

seeforschauer: removed, renamed. PTAL.

w84thesun (approved):

AlekSi (approved):

rumyantseva (approved):
LGTM

PR #551, opened 2022-05-03
Support `$elemMatch`  projection basic comparison query operators

seeforschauer:
Refs #439
- [x] `$eq`
- [x] `$ne`
- [x] `$gt`
- [x] `$gte`
- [x] `$lt`
- [x]  `$lte`
- [x] `$in`
- [x] `$nin`

PR #635, opened 2022-05-24
Make `Path` type

### LINKED ISSUE #631: Make `Path` type
Currently, we use `...string` or `[]string` for paths in documents. It is too easy to misuse.

Let's make a new type in `types` package:
```
type Path struct {
  s []string
}
```

That's a value type (it should be passed by values), just like `time.Time`.
It should be immutable; no methods should modify the internal state, just like `time.Time` methods.

It could be constructed from a dot-separated string or from a strings slice (it should make a copy of the slice).

All code that works with paths or dot notation should be updated to use that type.

w84thesun:
Closes #631.


seeforschauer:
https://app.codecov.io/gh/FerretDB/FerretDB/compare/635/diff#D7L246 and one more lines should be covered with tests somehow, WDYT?

internal/types/path.go
```diff
@@ -17,17 +17,67 @@ package types
...
+func (p Path) Remove(index int) Path {
+	if index < 0 || index > len(p.s) {
+		return p
+	}
+	path := NewPath(p.s...)
+	path.s = append(path.s[:index], path.s[index+1:]...)
+	return path
+}
+
+// NewPath returns Path from strings slice.
```
seeforschauer, line 58: It's important that there is `copy`. Could you mention this?

seeforschauer (dismissed):

internal/types/path.go
```diff
@@ -17,17 +17,68 @@ package types
...
+}
+
+// Slice returns path slice copy.
+func (p Path) Slice() []string {
+	path := make([]string, len(p.s))
+	copy(path, p.s)
+	return path
+}
+
+// Element returns path element specified by the index or empty string if index below 0 or greater than Path len.
```
rumyantseva, line 40: I'd suggest dividing this sentence into two. One might misread the "specified by the index or empty string" part :)

Maybe something like
```
Element returns path's element specified by the index.
If index is below 0 or greater than Path length, it returns an empty string.
```


P.S. I actually like how the doc for `Remove` looks like! Maybe something similar here?

internal/types/path.go
```diff
@@ -17,17 +17,68 @@ package types
...
+func (p Path) Element(index int) string {
+	if index < 0 || index > len(p.s) {
```
rumyantseva, line 42: What if `index == len(p.s)`? Wouldn't `p.s[index]` panic?

internal/handlers/common/filter.go
```diff
@@ -52,17 +52,17 @@ func FilterDocument(doc, filter *types.Document) (bool, error) {
 func filterDocumentPair(doc *types.Document, filterKey string, filterValue any) (bool, error) {
 	if strings.ContainsRune(filterKey, '.') {
 		// {field1./.../.fieldN: filterValue}
-		path := strings.Split(filterKey, ".")
+		path := types.NewPathFromString(filterKey)
 		// we pass the path without the last key because we want {fieldN: *someValue*}, not just *someValue*
-		docValue, err := doc.GetByPath(path[:len(path)-1]...)
+		docValue, err := doc.GetByPath(path.Remove(path.Len() - 1))
```
AlekSi, line 57: Do we actually have use cases for methods with indexes instead of just prefixes and suffixes?

I think that line, as well as `path.Element(path.Len() - 1)` below, are complex, and that's something we would like to hide with the good path's API

internal/types/path.go
```diff
@@ -17,17 +17,68 @@ package types
...
+		return p
+	}
+	path := NewPath(p.s...)
+	path.s = append(path.s[:index], path.s[index+1:]...)
+	return path
+}
+
+// NewPath returns Path from a strings slice. Provided strings slice would be copied into a new Path.
+func NewPath(path ...string) Path {
```
AlekSi, line 60: "Construct" functions should typically go before methods

internal/types/path.go
```diff
@@ -17,17 +17,68 @@ package types
...
+	if index < 0 || index > len(p.s) {
+		return ""
+	}
+	return p.s[index]
+}
+
+// Remove returns new path with removed path element specified by index.
+// If index below 0 or greater than Path len it would return the same value.
+func (p Path) Remove(index int) Path {
+	if index < 0 || index > len(p.s) {
```
rumyantseva, line 51: Same here, what if `index == len(p.s)`?

internal/handlers/common/filter.go
seeforschauer: > Do we actually have use cases for methods with indexes?

Some are in `$elemMatch` when it is necessary to delete an item in the middle of an array.

rumyantseva (commented):
I haven't finished the review yet, just want to"commit" the current comments/questions.

internal/handlers/common/filter.go
AlekSi: Please point me to the line of code

seeforschauer: https://github.com/FerretDB/FerretDB/pull/551/files#r881425664
It's a path however I don't see how the path that can be used there, it's overcomplication.

AlekSi: I'm not sure how that relates to my concern. I was saying that we don't have code that calls `path.Element(x)` with `x` being something other than `0` (prefix) or `path.Len()-1` (suffix). The linked line of code is not about path; the code below is about suffix.

internal/types/path.go
w84thesun: Fixed.

internal/handlers/common/filter.go
w84thesun: Replaced with TrimSuffix/TrimPrefix and Suffix/Prefix methods.

internal/types/path.go
w84thesun: Replaced this method with Suffix/Prefix.

internal/types/path.go
w84thesun: Replaced this method with TrimSuffix/TrimPrefix.

seeforschauer (dismissed):

internal/types/path.go
```diff
@@ -17,17 +17,81 @@ package types
 import (
 	"fmt"
 	"strconv"
+	"strings"
 )
 
+// Path represents field path type. It should be used wherever we work with paths or dot notation.
+type Path struct {
+	s []string
+}
```
AlekSi, line 26: I think we should add `String() string` method that would return a dot-separated string

AlekSi (changes requested):

internal/types/path.go
```diff
@@ -17,17 +17,81 @@ package types
...
+
+// Slice returns path slice copy.
+func (p Path) Slice() []string {
+	path := make([]string, len(p.s))
+	copy(path, p.s)
+	return path
+}
+
+// Suffix returns the first path element.
+func (p Path) Suffix() string {
```
AlekSi, line 55: Actually, let's keep those names, but fix descriptions

```suggestion
// Suffix returns the last path element.
func (p Path) Suffix() string {
```

internal/types/path.go
```diff
@@ -17,17 +17,81 @@ package types
...
+
+// Prefix returns the last path element.
+func (p Path) Prefix() string {
+	if len(p.s) == 0 {
+		return ""
+	}
+	return p.s[0]
+}
+
+// TrimSuffix returns a copy of path without suffix.
```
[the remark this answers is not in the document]
w84thesun: Fixed descriptions.

AlekSi (approved):

internal/types/path.go
```diff
@@ -17,17 +17,87 @@ package types
 import (
 	"fmt"
 	"strconv"
+	"strings"
 )
 
+// Path represents field path type. It should be used wherever we work with paths or dot notation.
```
seeforschauer, line 23:
```suggestion
// The path represents the field path type. It should be used wherever we work with paths or dot notation.
```

internal/types/path.go
```diff
@@ -17,17 +17,87 @@ package types
...
+// Path should be stored and passed as value. All of its methods return new values, not modifying the receiver's state.
```
seeforschauer, line 24:
```suggestion
// The path should be stored and passed as a value. Its methods return new values, not modifying the receiver's state.
```

internal/types/path.go
```diff
@@ -17,17 +17,87 @@ package types
...
+func (p Path) Prefix() string {
+	if len(p.s) == 0 {
+		return ""
+	}
+	return p.s[0]
+}
+
+// TrimSuffix returns a path without the last element.
+func (p Path) TrimSuffix() Path {
+	if len(p.s) == 0 {
```
seeforschauer, line 78: It is not expected, if I wanted to trim, I expect there will be at least 2 elements. 
My suggestion is panic.

internal/types/path.go
```diff
@@ -17,17 +17,87 @@ package types
...
+		return p
+	}
+	return NewPath(p.s[:p.Len()-1])
+}
+
+// TrimPrefix returns a copy of path without the first element.
+func (p Path) TrimPrefix() Path {
+	if len(p.s) == 0 {
```
seeforschauer, line 86: It is not expected, if I wanted to trim, I expect there will be at least 2 elements. 
My suggestion is panic.

internal/types/path.go
```diff
@@ -17,17 +17,87 @@ package types
...
+	return strings.Join(p.s, ".")
+}
+
+// Len returns path length.
+func (p Path) Len() int {
+	return len(p.s)
+}
+
+// Slice returns path values array.
+func (p Path) Slice() []string {
```
seeforschauer, line 54: Not obvious to me, that slice returns a slice (sic!).  Maybe rename it to `ToSlice`?

internal/types/path.go
rumyantseva: No, doc comment must start from the exact word used as type/var/function etc. So, in this particular case it's still just `Path` :)

internal/types/path.go
```diff
@@ -17,17 +17,87 @@ package types
...
+	}
+	return NewPath(p.s[:p.Len()-1])
+}
+
+// TrimPrefix returns a copy of path without the first element.
+func (p Path) TrimPrefix() Path {
+	if len(p.s) == 0 {
+		return p
+	}
+	return NewPath(p.s[1:])
```
rumyantseva, line 89: What if `len(p.s) == 1`?

internal/types/path.go
```diff
@@ -17,17 +17,87 @@ package types
...
+	copy(path, p.s)
+	return path
+}
+
+// Suffix returns the last path element.
+func (p Path) Suffix() string {
+	if len(p.s) == 0 {
+		return ""
+	}
+	return p.s[p.Len()-1]
```
rumyantseva, line 65: In some cases you use `len(p.s)` to get the length (e.g. three lines above). In others, like here, you use `p.(Len)`. Let's make it more consistent and use the same approach everywhere in this file?

rumyantseva (changes requested):
I see in some places codecov complains about coverage. And I would agree.
Example: `TrimPrefix` seems to panic if len == 1.

I guess len == 1 is not possible if `Path` is valid. Then, maybe, what @seeforschauer offers might be reasonable - a more obvious panic complaining about an invalid `Path`?

Also, should we give an example of a "typical" Path somewhere in the comments? Or is it straightforward?

AlekSi:
> I guess len == 1 is not possible if Path is valid. 

It is possible: "foo" is a valid path consisting of one field name

internal/types/path.go
rumyantseva: If it's a valid path, it needs to be covered with tests. I haven't seen test cases with len == 1.

internal/types/path.go
AlekSi: `String()`, not `ToString()`. The same logic there

internal/types/path.go
w84thesun: `strings` package works just like that. If you would try to trim the last element you get empty string → https://go.dev/play/p/f0qsrUBl9ES

seeforschauer: ok, them maybe the word `trim` is not good for this function.

internal/types/path.go
w84thesun: Fixed.

internal/types/path.go
seeforschauer: > No, doc comment must start from the exact word used as type/var/function etc. So, in this particular case it's still just `Path` :)

Changed.

internal/types/path.go
```diff
@@ -17,17 +17,87 @@ package types
...
+	if p.Len() == 0 {
+		return p
+	}
+	return NewPath(p.s[:p.Len()-1])
+}
+
+// TrimPrefix returns a copy of path without the first element.
+func (p Path) TrimPrefix() Path {
+	if p.Len() == 0 {
+		return p
```
seeforschauer, line 87: No it's misleading.

internal/types/path.go
```diff
@@ -17,17 +17,87 @@ package types
...
+	if p.Len() == 0 {
+		return ""
+	}
+	return p.s[0]
+}
+
+// TrimSuffix returns a path without the last element.
+func (p Path) TrimSuffix() Path {
+	if p.Len() == 0 {
+		return p
```
seeforschauer, line 79: Here is also misleading. I expect if there is nothing to trim I must not trim.

w84thesun: That's exactly what is happening here. I there is no `Path` elements we not trim anything and return the path itself. It's the caller responsibility to track if there is something in the path left to trim not the callee. The only option might be here IMO is to return a new path calling `NewPath([]string{})`

seeforschauer: To the function name, yes.
But looking at the usages in this PR - no, you suppose there must have been an element.

rumyantseva (dismissed):
As discussed in person, I approve this PR as the essential fixes were made.

internal/types/path.go
```diff
@@ -17,17 +17,87 @@ package types
...
+	if len(p.s) == 0 {
+		return ""
+	}
+	return p.s[p.Len()-1]
+}
+
+// Prefix returns the first path element.
+func (p Path) Prefix() string {
+	if p.Len() == 0 {
+		return ""
```
seeforschauer, line 71: `[""] != []`

internal/types/path.go
```diff
@@ -17,17 +17,87 @@ package types
...
+func (p Path) Slice() []string {
+	path := make([]string, p.Len())
+	copy(path, p.s)
+	return path
+}
+
+// Suffix returns the last path element.
+func (p Path) Suffix() string {
+	if len(p.s) == 0 {
+		return ""
```
seeforschauer, line 63: `[""] != []`

AlekSi (changes requested):
Let's disallow empty paths and paths where any element is empty

internal/types/path.go
```diff
@@ -17,17 +17,87 @@ package types
...
 
+// Path represents the field path type. It should be used wherever we work with paths or dot notation.
+// Path should be stored and passed as a value. Its methods return new values, not modifying the receiver's state.
+type Path struct {
+	s []string
+}
+
+// NewPath returns Path from a strings slice.
+func NewPath(path []string) Path {
+	p := Path{s: make([]string, len(path))}
```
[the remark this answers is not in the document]
w84thesun: Added.

internal/types/path.go
```diff
@@ -17,17 +17,87 @@ package types
...
+
+// Slice returns path values array.
+func (p Path) Slice() []string {
+	path := make([]string, p.Len())
+	copy(path, p.s)
+	return path
+}
+
+// Suffix returns the last path element.
+func (p Path) Suffix() string {
```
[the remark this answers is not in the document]
w84thesun: Added.

internal/types/path.go
```diff
@@ -17,17 +17,87 @@ package types
...
+	if len(p.s) == 0 {
+		return ""
+	}
+	return p.s[p.Len()-1]
+}
+
+// Prefix returns the first path element.
+func (p Path) Prefix() string {
```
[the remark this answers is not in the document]
w84thesun: Added.

internal/types/path.go
w84thesun: Fixed.

internal/types/path.go
w84thesun: Fixed.

seeforschauer (approved):

seeforschauer (changes requested):
now tests fail

seeforschauer (dismissed):

internal/types/path.go
```diff
@@ -54,22 +132,28 @@ func getByPath[T CompositeTypeInterface](comp T, path ...string) (any, error) {
...
+// removeByPath removes path elements for given value, which could be *Document or *Array.
+func removeByPath(v any, path Path) {
+	if path.Len() == 0 {
 		return
 	}
 
-	key := keys[0]
+	var key string
+	if path.Len() == 1 {
+		key = path.Slice()[0]
```
AlekSi, line 143:
```suggestion
		key = path.String()
```

AlekSi (approved):

seeforschauer (approved):

AlekSi (approved):

rumyantseva (commented):
LGTM

PR #641, opened 2022-05-26
Sync handlers

AlekSi:


internal/handlers/common/msg_listcommands.go
```diff
@@ -0,0 +1,188 @@
...
+	"createIndexes": {
+		Help:    "Creates indexes on a collection.",
+		Handler: (handlers.Interface).MsgCreateIndexes,
+	},
+	"dataSize": {
+		Help:    "Returns the size of the collection in bytes.",
+		Handler: (handlers.Interface).MsgDataSize,
+	},
+	"dbStats": {
+		Help:    "Returns the statistics of the database.",
```
seeforschauer, line 75: With the `<noun> of <noun>` the sentence sounds harder to read/speak. Instead, it's suggested to use `<adj> <noun>` here and after.

```suggestion
		Help:    "Returns the database statistics.",
```

internal/util/version/version.go
```diff
@@ -40,6 +40,12 @@ var (
 	branch string
 )
 
+// Fake MongoDB version for clients that check it.
+var (
+	MongoDBVersion      = "5.0.42"
```
seeforschauer, line 45: How is possible to return when in the docker-compose file the version is `mongo:5.0.8`?

seeforschauer (dismissed):

internal/util/version/version.go
seeforschauer: .42 is obviously a number that nobody will parse seriously.

internal/handlers/common/msg_hostinfo.go
```diff
@@ -0,0 +1,78 @@
...
+				"currentTime", now,
+				"hostname", hostname,
+				"cpuAddrSize", int32(strconv.IntSize),
+				"numCores", int32(runtime.NumCPU()),
+				"cpuArch", runtime.GOARCH,
+			)),
+			"os", must.NotFail(types.NewDocument(
+				"type", strings.Title(runtime.GOOS), //nolint:staticcheck // good enough for GOOS
+				"name", osName,
+				"version", osVersion,
```
rumyantseva, line 67: So, if OS is not Linux those are just empty strings, but I guess we are ok with it.

internal/handlers/common/msg_listcommands.go
```diff
@@ -0,0 +1,188 @@
...
+	"getLog": {
+		Help:    "Returns the most recent logged events from memory.",
+		Handler: (handlers.Interface).MsgGetLog,
+	},
+	"getParameter": {
+		Help:    "Returns the value of the parameter.",
+		Handler: (handlers.Interface).MsgGetParameter,
+	},
+	"hello": {
+		Help:    "Returns the role of the FerretDB instance.",
```
rumyantseva, line 115: Does it? I thought "ismaster" returns the role. Is this one an alias?

rumyantseva (commented):
I have a few questions (mostly about docs). They seem to be non-blocking, so their implementation is up to you.

internal/handlers/common/msg_listcommands.go
AlekSi: `ismaster` and `isMaster` are aliases. `hello` is very similar, but slightly different. We tests for that: https://github.com/FerretDB/FerretDB/blob/main/integration/commands_replication_test.go

internal/handlers/common/msg_hostinfo.go
AlekSi: We do not plan to support other OSes in the near future

internal/handlers/pg/msg_datasize.go
```diff
@@ -88,3 +65,26 @@ func (h *Handler) MsgDataSize(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg
 
 	return formatResponse(stats.SizeTotal, stats.Rows, millis, true)
 }
+
+func formatResponse(size, rows, millis int32, showEstimate bool) (*wire.OpMsg, error) {
```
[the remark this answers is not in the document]
AlekSi: Inlined that function instead

internal/util/version/version.go
```diff
@@ -40,6 +40,12 @@ var (
 	branch string
 )
 
+// Fake MongoDB version for clients that check it.
```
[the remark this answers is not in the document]
AlekSi: Improved comment

internal/handlers/handlers.go
```diff
@@ -12,5 +12,110 @@
...
+// Package handlers provides an interface for all handlers.
 package handlers
+
+import (
+	"context"
+
+	"github.com/FerretDB/FerretDB/internal/wire"
+)
+
+// Interface contains methods that should be implemented by all handlers.
```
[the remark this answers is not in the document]
AlekSi: Tried to add more info

seeforschauer (approved):
It fixes `revive` linter complain about the `ctx` variable in the `listCommand` function. I like it.

PR #642, opened 2022-05-26
Set linter header to the year 2022

seeforschauer:


seeforschauer:
https://ferretdb.slack.com/archives/C02NWN1K3MJ/p1653564471851389

AlekSi:
Removed the comment as we don't give legal advices

PR #644, opened 2022-05-26
Support `$setOnInsert` field update operator

### LINKED ISSUE #628: Support `$setOnInsert` field update operator
Dot notation is not in scope.

seeforschauer:
Closes #628.


integration/helpers.go
```diff
@@ -203,3 +203,50 @@ func AssertEqualWriteError(t *testing.T, expected *mongo.WriteError, actual erro
 	return assert.Equal(t, expected.Message, actualWriteErr.Message) &&
 		assert.Equal(t, expected.Code, actualWriteErr.Code)
 }
+
+// IsUnimplemented returns true if the error code is unimplemented.
```
rumyantseva: Do we have any agreement about negative vs positive statements for such cases? Double negatives cases often increase cognitive load (implemented == ! IsUnimplemented). E.g. we don't call functions `IsUnequal`, but we call them `IsEqual` instead. 

Is it on purpose to not call this function `IsImplemented` instead?

integration/helpers.go
```diff
@@ -203,3 +203,50 @@ func AssertEqualWriteError(t *testing.T, expected *mongo.WriteError, actual erro
...
+	actualCE, ok := actual.(mongo.CommandError)
+	if ok {
+		if actualCE.Code == 238 {
+			return true
+		}
+	}
+	return false
+}
+
+// AssertEqualAltWriteError compares expected mongo.WriteError Message and Code with actual error.
```
rumyantseva, line 177: If I'm not mistaken, I've seen this in another PR, so I don't review it here.

rumyantseva (commented):
I see many things are inherited from #634. So I guess, the current PR needs a review after #634 is merged.

integration/helpers.go
seeforschauer: Agree. Usually, I don't like it too. 
Here it's specifically true if it is not implemented. 
I understand this and it provides no mental mess for me. 
Of course, I'd change it if it's uncomfortable.

internal/handlers/pg/msg_update.go
```diff
@@ -100,6 +100,12 @@ func (h *Handler) MsgUpdate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 		if u, err = common.GetOptionalParam(update, "u", u); err != nil {
 			return nil, err
 		}
+		if u != nil {
```
seeforschauer, line 103: Before this PR, here the check of the upsert document was after the data was loaded to the client.
So if the data volume is big,  and `u` condition document is wrong, then the client unnecessarily waited for data.

integration/helpers.go
seeforschauer: removed the func.

internal/handlers/common/update.go
```diff
@@ -76,14 +78,16 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
 
 				if !doc.Has(incKey) {
 					must.NoError(doc.Set(incKey, incValue))
-					return true, nil
+					changed = true
```
seeforschauer, line 81: @AlekSi @w84thesun  here influences on $inc operator - some cases are uncovered, for example when there are several `$inc` in the update document.  WDYT?

integration/setup.go
```diff
@@ -220,3 +220,17 @@ func startup(t *testing.T) {
 
 	go debug.RunHandler(ctx, "127.0.0.1:0", zap.L().Named("debug"))
 }
+
+// dbByPort returns database name by port.
+func dbByPort() string {
```
rumyantseva, line 225: I see ports hardcoded in `cmd/envtool/main.go`, so I guess it's ok that we hard-code them here as well...
I believe, one day we might need to move them to consts/vars, but for now it seems to be ok.

internal/handlers/common/update.go
```diff
@@ -26,16 +26,18 @@ import (
...
 // Returns true if document was changed.
 func UpdateDocument(doc, update *types.Document) (bool, error) {
+	var changed bool
 	for _, updateOp := range update.Keys() {
 		updateV := must.NotFail(update.Get(updateOp))
 
 		switch updateOp {
-		case "$set":
+		case "$set",
+			"$setOnInsert":
```
rumyantseva, line 35: This formatting looks a bit unusual to me. But I'm not sure I know the right formatting :) Why are both conditions not on the same line? Is it something linter-specific?

internal/handlers/common/update.go
```diff
@@ -26,16 +26,18 @@ import (
...
 
 			switch setDoc := updateV.(type) {
 			case *types.Document:
 				if setDoc.Len() == 0 {
-					return false, nil
+					continue
```
rumyantseva, line 40: If I'm not mistaken, that changes the logic for the `$set` case. Why before it was ok to just return the result, and now it's not ok anymore, and we need to go to the next iteration?

rumyantseva (commented):
I don't feel confident approving as I don't understand the reasons to change the logic. Asked a question about it.

internal/handlers/common/update.go
seeforschauer: > This formatting looks a bit unusual to me. But I'm not sure I know the right formatting :) Why are both conditions not on the same line? Is it something linter-specific?

a new test case was added:
```go
                "set-setoninsert": {
			filter: bson.D{{"_id", "test"}},
			query: bson.D{
				{"$set", bson.D{{"foo", int32(12)}}},
				{"$setOnInsert", bson.D{{"value", math.NaN()}}},
			},
			res: bson.D{{"_id", "test"}, {"foo", int32(12)}, {"value", math.NaN()}},
		},
```
and it was found that when multiply changes in $set clause then `modified` count gives the wrong number.

internal/handlers/common/update.go
seeforschauer: > If I'm not mistaken, that changes the logic for the `$set` case. Why before it was ok to just return the result, and now it's not ok anymore, and we need to go to the next iteration?

> This formatting looks a bit unusual to me. But I'm not sure I know the right formatting :) Why are both conditions not on the same line? Is it something linter-specific?

a new test case was added:
```go
                "set-setoninsert": {
			filter: bson.D{{"_id", "test"}},
			query: bson.D{
				{"$set", bson.D{{"foo", int32(12)}}},
				{"$setOnInsert", bson.D{{"value", math.NaN()}}},
			},
			res: bson.D{{"_id", "test"}, {"foo", int32(12)}, {"value", math.NaN()}},
		},
```
and it was found that when multiply changes in $set clause then `modified` count gives the wrong number.

integration/update_test.go
```diff
@@ -460,3 +461,272 @@ func TestUpdateSet(t *testing.T) {
...
+		UpsertedCount: 1,
+	}
+
+	for name, tc := range map[string]struct {
+		filter         bson.D
+		query          bson.D
+		stat           *mongo.UpdateResult
+		res            bson.D
+		err            *mongo.WriteError
+		notImplemented bool
```
AlekSi, line 622: We should not add tests for things that are not implemented

AlekSi (changes requested):

internal/handlers/common/update.go
seeforschauer: Linters are ok with that.
As for me it's simpler to read and simpler to see changes during the time.
Imagine you adding a new operator.
If these operators were in one line, you should parse the line with your eyes and see the change.
But if they are each in the new line, then GitHub shows you the change.

That's it.

integration/update_test.go
seeforschauer: that's all removed.

integration/update_test.go
```diff
@@ -460,3 +461,272 @@ func TestUpdateSet(t *testing.T) {
...
+			_, err = collection.InsertMany(ctx, []any{
+				bson.D{{"_id", "string"}, {"value", "foo"}},
+			})
+			require.NoError(t, err)
+
+			opts := options.Update().SetUpsert(true)
+			var res *mongo.UpdateResult
+			res, err = collection.UpdateOne(ctx, tc.filter, tc.query, opts)
+
+			if tc.notImplemented && dbByPort() == "ferretdb" && err != nil {
```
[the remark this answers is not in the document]
seeforschauer: removed.

integration/update_test.go
```diff
@@ -460,3 +461,228 @@ func TestUpdateSet(t *testing.T) {
...
+	for name, tc := range map[string]struct {
+		filter      bson.D
+		setOnInsert any
+		stat        *mongo.UpdateResult
+		res         bson.D
+		err         *mongo.WriteError
+		alt         string
+	}{
+		"doc": {
+			filter:      bson.D{{"_id", "doc"}},
```
seeforschauer, line 483: here and with `array` - some versions of MongoDB can have different answers, so I suggest keeping those tests. 
Don't you mind?

integration/update_test.go
```diff
@@ -460,3 +461,213 @@ func TestUpdateSet(t *testing.T) {
...
+				{"$set", bson.D{{"foo", int32(12)}}},
+				{"$inc", bson.D{{"foo", int32(1)}}},
+				{"$setOnInsert", bson.D{{"value", math.NaN()}}},
+			},
+			err: &mongo.WriteError{
+				Code:    40,
+				Message: "Updating the path 'foo' would create a conflict at 'foo'",
+			},
+		},
+		"unknown-operator": {
```
w84thesun, line 629: I suggest extracting this to a separate test as it's not related to `$setOnInsert`. Just like with `TestQueryUnknownFilterOperator`.

integration/update_test.go
```diff
@@ -460,3 +461,213 @@ func TestUpdateSet(t *testing.T) {
...
+
+			var actual bson.D
+			err = collection.FindOne(ctx, tc.filter).Decode(&actual)
+			require.NoError(t, err)
+			AssertEqualDocuments(t, tc.res, actual)
+		})
+	}
+}
+
+func TestSetOnInsertMore(t *testing.T) {
```
w84thesun, line 592: Test name `TestSetOnInsertMore` doesn't reflect what's going on in it. Without `unknown-operator` it could be called `TestUpdateOperators`

integration/update_test.go
```diff
@@ -460,3 +461,213 @@ func TestUpdateSet(t *testing.T) {
...
+		UpsertedCount: 1,
+	}
+
+	for name, tc := range map[string]struct {
+		filter bson.D
+		query  bson.D
+		stat   *mongo.UpdateResult
+		res    bson.D
+		err    *mongo.WriteError
+		alt    string
```
w84thesun, line 607: We can remove it as we don't use it.

integration/update_test.go
```diff
@@ -460,3 +461,213 @@ func TestUpdateSet(t *testing.T) {
...
+
+	notModified := &mongo.UpdateResult{
+		MatchedCount:  0,
+		ModifiedCount: 0,
+		UpsertedCount: 1,
+	}
+
+	for name, tc := range map[string]struct {
+		filter bson.D
+		query  bson.D
```
w84thesun, line 603: I think it should be called `update` not `query`.

integration/update_test.go
```diff
@@ -460,3 +461,213 @@ func TestUpdateSet(t *testing.T) {
...
+		stat   *mongo.UpdateResult
+		res    bson.D
+		err    *mongo.WriteError
+		alt    string
+	}{
+		"setoninsert-set": {
```
w84thesun, line 609: I think we should reach an agreement on how to name test cases and use them everywhere. Right now most of the test cases' names are in PascalCase.

integration/update_test.go
```diff
@@ -460,3 +461,213 @@ func TestUpdateSet(t *testing.T) {
...
+			filter: bson.D{{"_id", "test"}},
+			query: bson.D{
+				{"$set", bson.D{{"foo", int32(12)}}},
+				{"$setOnInsert", bson.D{{"value", math.NaN()}}},
+			},
+			res: bson.D{{"_id", "test"}, {"foo", int32(12)}, {"value", math.NaN()}},
+		},
+		"trio": {
```
w84thesun, line 617: Why `trio`? Maybe `SetIncSetOnInsert`?

integration/update_test.go
```diff
@@ -460,3 +461,213 @@ func TestUpdateSet(t *testing.T) {
 		})
 	}
 }
+
+func TestSetOnInsert(t *testing.T) {
```
w84thesun, line 465:
```suggestion
func TestUpdateSetOnInsertOperator(t *testing.T) {
```

internal/handlers/common/update.go
```diff
@@ -118,5 +122,77 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
 		}
 	}
 
-	return true, nil
+	return changed, nil
+}
+
+// ValidateUpdateQuery validates update statement.
+func ValidateUpdateQuery(update *types.Document) (err error) {
```
w84thesun, line 129: This is a query actually. `ValidateUpdateOperators` might be the better name.

internal/handlers/common/update.go
```diff
@@ -118,5 +122,77 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
+				fmt.Sprintf("Unknown modifier: %s. Expected a valid update modifier or pipeline-style "+
+					"update specified as an array", updateOp,
+				))
+		}
+	}
+	return nil
+}
+
+// getDocument gets document by key `op` and returns WriteError error if it is not a document.
+func getDocument(op string, update *types.Document) (*types.Document, error) {
```
w84thesun, line 173: `getDocument` looks very common name for that function. Maybe something like `getUpdateOperatorDocument`?

integration/update_test.go
```diff
@@ -460,3 +461,213 @@ func TestUpdateSet(t *testing.T) {
...
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			var err error
+			ctx, collection := setup(t)
+
+			opts := options.Update().SetUpsert(true)
+			var res *mongo.UpdateResult
+			res, err = collection.UpdateOne(ctx, tc.filter, bson.D{{"$setOnInsert", tc.setOnInsert}}, opts)
```
w84thesun, line 566:
```suggestion
			ctx, collection := setup(t)

			opts := options.Update().SetUpsert(true)
			var res *mongo.UpdateResult
			res, err := collection.UpdateOne(ctx, tc.filter, bson.D{{"$setOnInsert", tc.setOnInsert}}, opts)
```

integration/update_test.go
```diff
@@ -460,3 +461,213 @@ func TestUpdateSet(t *testing.T) {
...
+				Message: "Unknown modifier: $foo. Expected a valid update modifier or pipeline-style update specified as an array",
+			},
+		},
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			ctx, collection := setup(t)
+			var err error
+			_, err = collection.InsertMany(ctx, []any{
```
w84thesun, line 643:
```suggestion

			_, err := collection.InsertMany(ctx, []any{
```

w84thesun (changes requested):

internal/handlers/common/update.go
seeforschauer: not much change and it's not about operators, it's about the entire query document.

integration/update_test.go
seeforschauer: the entire test set is renamed: it's related to tests when $set, $setOnInsert and $inc are used together.

integration/update_test.go
seeforschauer: renamed

internal/handlers/common/update.go
seeforschauer: renamed

integration/update_test.go
seeforschauer: actually, in this place, it's a query.

integration/update_test.go
seeforschauer: done

internal/handlers/common/update.go
seeforschauer: thank you, changed.

integration/update_test.go
```diff
@@ -460,3 +461,211 @@ func TestUpdateSet(t *testing.T) {
...
+			ctx, collection := setup(t)
+
+			_, err := collection.InsertMany(ctx, []any{
+				bson.D{{"_id", "string"}, {"value", "foo"}},
+			})
+			require.NoError(t, err)
+
+			opts := options.Update().SetUpsert(true)
+			var res *mongo.UpdateResult
+			res, err = collection.UpdateOne(ctx, tc.filter, tc.query, opts)
```
w84thesun, line 648: That's what I'm talking above. That's not the query, that's the update part of request.
```go
func (coll *Collection) UpdateOne(ctx context.Context, filter interface{}, update interface{},
	opts ...*options.UpdateOptions) (*UpdateResult, error)
```

integration/update_test.go
w84thesun: Please take a look at https://github.com/FerretDB/FerretDB/pull/644/files#r885582655

integration/update_test.go
seeforschauer: updated

w84thesun: Thanks!

w84thesun (dismissed):

internal/handlers/common/update.go
AlekSi: Let's not use that formatting. We have two existing options already, let's use one of them:

```go
case "$set", "$setOnInsert":
```

or

```go
case "$set":
  fallthrough
case "$setOnInsert":
```

internal/handlers/common/update.go
AlekSi: Let's add a test for `$set` that shows the problem in the previous code

internal/handlers/common/update.go
AlekSi: I agree, but that (and the one above) change in logic looks suspicious; let's add tests for that

internal/handlers/common/update.go
```diff
@@ -118,5 +122,77 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
+	if err != nil {
+		return err
+	}
+
+	inc, err := getUpdateOperatorDocument("$inc", update)
+	if err != nil {
+		return err
+	}
+
+	if set != nil && inc != nil {
```
AlekSi, line 144: How that would look like with more operators?

internal/handlers/common/update.go
```diff
@@ -44,7 +46,7 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
 						return false, err
 					}
 				}
-				return true, nil
+				changed = true
 			default:
 				msgFmt := fmt.Sprintf(`Modifiers operate on fields but we found type %[1]s instead. `+
```
AlekSi, line 51: We have the same error message in several places, including getUpdateOperatorDocument, which seems to be a common function that checks this case. Do we still need that check there?

AlekSi (changes requested):

internal/handlers/common/update.go
```diff
@@ -118,5 +122,77 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
 		}
 	}
 
-	return true, nil
+	return changed, nil
+}
+
+// ValidateUpdateOperators validates update statement.
+func ValidateUpdateOperators(update *types.Document) (err error) {
```
AlekSi, line 129: Let's not use a named result if we don't use it

internal/handlers/common/update.go
seeforschauer: > How that would look like with more operators?

1. what is your suggestion?
2. can it be solved when more operators appear?

internal/handlers/common/update.go
seeforschauer: This exact error is never reached if that check is present.
should we panic here?

internal/handlers/common/update.go
```diff
@@ -118,5 +122,77 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
+				ErrFailedToParse,
+				fmt.Sprintf("Unknown modifier: %s. Expected a valid update modifier or pipeline-style "+
+					"update specified as an array", updateOp,
+				))
+		}
+	}
+	return nil
+}
+
+// getUpdateOperatorDocument gets document by key `op` and returns WriteError error if it is not a document.
```
rumyantseva, line 173: This function is quite unclear to me. I don't understand what's the purpose - not by the doc, nor by the function name, nor by the code of this function. 

I see it receives a document called `update` and it returns some other document. What is the difference between the document it receives and the document it returns?

internal/handlers/common/update.go
seeforschauer: fixed

rumyantseva (commented):
I don't get some parts of this code. 
I asked questions about it.

internal/handlers/common/update.go
```diff
@@ -118,5 +122,77 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
+		return nil, nil
+	}
+	updateExpression := must.NotFail(update.Get(op))
+	switch doc := updateExpression.(type) {
+	case *types.Document:
+		for _, v := range doc.Keys() {
+			if strings.Contains(v, ".") {
+				return nil, NewError(
+					ErrNotImplemented,
+					fmt.Errorf("Dot notation is not implemented"),
```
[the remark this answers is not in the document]
seeforschauer: > Don't linter complain here about`Error`'s string being capitalized?

It argues only in the IDE >-<

internal/handlers/common/update.go
seeforschauer: > This function is quite unclear to me. I don't understand what's the purpose - not by the doc, nor by the function name, nor by the code of this function.
> 
> I see it receives a document called `update` and it returns some other document. What is the difference between the document it receives and the document it returns?

You are quite right. There is no difference. I just needed to extract a function to make a check.

seeforschauer: > 

I thought about naming the function starting `check` but it does return the document.

rumyantseva: I still don't understand :) Which function do you need to extract? What is the result of this extract?

seeforschauer: > I still don't understand :) Which function do you need to extract? What is the result of this extract?

the result of is extract is the `$set` or `$inc` expression document

rumyantseva: My initial question was what is the difference between the doc this function receives and returns? Maybe a small example of a document could help me understand.

seeforschauer: > My initial question was what is the difference between the doc this function receives and returns? Maybe a small example of a document could help me understand.

It's a smaller document.

The update document from the driver 
```go
bson.D{
	{"$set", bson.D{{"foo", int32(12)}}},
	{"$inc", bson.D{{"foo", int32(1)}}},
	{"$setOnInsert", bson.D{{"value", math.NaN()}}},
},
```

The document that returned by a function for `$setOnInsert` operator:
```go
bson.D{{"value", math.NaN()}
```
i.e. for `$inc`:
```go
bson.D{{"foo", int32(1)}}
```

internal/handlers/common/update.go
AlekSi: In that case, we can drop the whole type switch and replace it with `setDoc := updateV.(*types.Document)` that would panic

seeforschauer:
https://github.com/FerretDB/FerretDB/pull/644#discussion_r885611510

> Let's add a test for $set that shows the problem in the previous code

I added two tests `SetTwoFields` and `IncTwoFields`. It works in this PR.

internal/handlers/common/update.go
AlekSi: If we know how to scale that code for more update operators, I think we should do that now. If we don't know, well, then maybe we should write this code some other way? We know for sure that more operators will be added

internal/handlers/common/update.go
```diff
@@ -118,5 +123,78 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
+				)
+			}
+		}
+	}
+
+	for _, updateOp := range update.Keys() {
+		switch updateOp {
+		case "$set",
+			"$setOnInsert",
+			"$inc":
```
AlekSi, line 161: Please put all cases on the same line, or use 3 separate cases

internal/handlers/common/update.go
```diff
@@ -118,5 +123,78 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
+			// supported
+		default:
+			return NewWriteErrorMsg(
+				ErrFailedToParse,
+				fmt.Sprintf("Unknown modifier: %s. Expected a valid update modifier or pipeline-style "+
+					"update specified as an array", updateOp,
+				))
```
AlekSi, line 168:
```suggestion
				fmt.Sprintf(
				    "Unknown modifier: %s. Expected a valid update modifier or pipeline-style "+
					"update specified as an array", updateOp,
				))
```

seeforschauer:
And. Also, I found another case 

```go
"SetSet": {
	filter: bson.D{{"_id", "test"}},
	update: bson.D{
		{"$set", bson.D{{"foo", int32(12)}}},
		{"$set", bson.D{{"value", math.NaN()}}},
	},
	res: bson.D{{"_id", "test"}, {"foo", int32(12)}, {"value", math.NaN()}},
},
```


It works on MongoDB and doesn't work on Ferret now.

internal/handlers/common/update.go
AlekSi: And the same thing in other places

internal/handlers/common/update.go
seeforschauer: oops, I wish I had linter here. fixing.

seeforschauer: done

integration/update_test.go
```diff
@@ -107,6 +106,8 @@ func TestUpdateIncOperatorErrors(t *testing.T) {
 				Message: `Modifiers operate on fields but we found type string instead.` +
 					` For example: {$mod: {<field>: ...}} not {$inc: "string"}`,
 			},
+			alt: `Modifiers operate on fields but we found type string instead.` +
+				` For example: {$mod: {<field>: ...}} not {$inc: string}`,
```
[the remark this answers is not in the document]
seeforschauer: > Both $set and $inc use very similar code to generate very similar error messages, but both code and messages are different in non-significant ways. Can we streamline them?

I've changed to it if any, and I'm ready to modify it. 
Btw it seems to me it become less clear what was meant but I am sure people will be able to find what is wrong with their query.

internal/handlers/common/update.go
seeforschauer: nope, sorry not for this. adding...

seeforschauer: > nope, sorry not for this. adding...

nope, no tests needed to add here - it's already covered.

internal/handlers/common/update.go
rumyantseva: Based on this example, I'd suggest calling this function `FilterDocumentByOperator` unless the current name is already in accordance with some existing naming convention.

Also, what is the use case for this function? As a MongoDB user, what query do I send so you need to filter my query this way?

rumyantseva: Although, `Filter` is not a good word here. It might be misinterpreted as something relevant to `FilterDocument` function.

w84thesun: It's not about filtering here. You can think of it as it's a bigger update document that we need to break down into smaller parts to validate their correctness. For that purpose, we already have a set of functions called `GetOptionalParam`.

internal/handlers/common/update.go
```diff
@@ -118,5 +107,102 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
+// For example, for update document
+// ```go
+// bson.D{
+// 	{"$set", bson.D{{"foo", int32(12)}}},
+// 	{"$inc", bson.D{{"foo", int32(1)}}},
+// 	{"$setOnInsert", bson.D{{"value", math.NaN()}}},
+// },
+// ```
+// The result returned for $setOnInsert operator is `bson.D{{"value", math.NaN()}`.
+func extractValueFromUpdateOperator(op string, update *types.Document) (*types.Document, error) {
```
seeforschauer, line 191: @rumyantseva please see! I've added the comment and renamed the function.

w84thesun (approved):

AlekSi:
CI fails

seeforschauer:
> CI fails

re-run OK.

integration/update_test.go
AlekSi: #679

Previous MongoDB versions relied heavily on OP_QUERY messages that are deprecated in 5.0 and mostly not implemented by us. People just can't migrate from those versions to FerretDB without updating their expectations.

That case tests the edge case in the current behavior, and that's good. But we should not test older versions.

internal/handlers/common/update.go
```diff
@@ -118,5 +107,102 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
+			fallthrough
+		case "$setOnInsert":
+			// supported
+		default:
+			return NewWriteErrorMsg(
+				ErrFailedToParse,
+				fmt.Sprintf(
+					"Unknown modifier: %s. Expected a valid update modifier or pipeline-style "+
+						"update specified as an array", updateOp,
+				))
```
AlekSi, line 153:
```suggestion
				)
			)
```

internal/handlers/common/update.go
AlekSi: godoc does not support Markdown syntax for code block. Here how it looks:

image

The code block should be just indented to be rendered correctly

seeforschauer:
fixed.

internal/handlers/common/update.go
seeforschauer: removed quotes.

rumyantseva (dismissed):
LGTM

internal/handlers/common/update.go
seeforschauer: updated godoc.

w84thesun (approved):

AlekSi (approved):

PR #645, opened 2022-05-26
Fix incorrect test for `$mod` operator

### LINKED ISSUE #638: Incorrect test for `$mod`
https://github.com/FerretDB/FerretDB/blob/6f2ec6e8df6bbc42ff14b6a7c5014b9a740d2e7f/integration/query_evaluation_test.go#L278

fenogentov:
Closes #638.


seeforschauer:
@fenogentov code coverage didn't change it's strange.
w/o it looks like the test case was one and you changed it to another.
better to have both.

seeforschauer:
I think here a research on `$mod` operator is needed: what is covered and what is not, and add.

seeforschauer:
yes, and code coverage didn't change anything while you changed the test means that there is something wrong with test and the code is not covered.

AlekSi:
No, it means that code coverage in Go is not _that_ powerful. For example, it tracks source code lines, not code branches.

seeforschauer:
> No, it means that code coverage in Go is not _that_ powerful. For example, it tracks source code lines, not code branches.

Agree, and it might mean both opinions are correct.

AlekSi:
Without context – sure. Not in this PR, though. You can see that from the mismatch between case name and case filter, the same code generates the error message as for the positive zero case: https://github.com/FerretDB/FerretDB/blob/61c844b7082cdd4218bd3ba68d5f042014df6382/internal/handlers/common/filter.go#L858-L860

In Go, `== 0` matches both positive and negative zeros. The code coverage would be different if we had different code there for `math.Copysign(0, -1)` / `0` and `math.Copysign(0, -1)` – but we don't, and I think we should not write more code just to see the code coverage difference.

AlekSi (approved):

rumyantseva (approved):
LGTM

w84thesun (approved):

PR #650, opened 2022-05-27
Add more visibility for the router/proxy error log levels

### LINKED ISSUE #590: Log error responses with `error` log level
Currently, we log requests and responses with `debug` log level, and diffs with `info` level. We should make error responses more noticeable by logging them with the `error` log level.

* [x] log diffs with `debug` level, just like normal requests and responses;
* [x] log error responses that do not close connection with `warning` level in `route` method;
* [x] log error responses that do close connection with `error` level in `route` method;
* [x] log proxy's error responses with `warning`/`error` level too;
* [x] move proxy's response logging into `proxy.Router.Route` method for consistency.

rumyantseva:
# Description

This PR closes #590. This change aims to improve the visibility of errors in logs.

The change proposes a new function (`logResponse`) to handle the logic of log levels according to the requirements described in #590. The function is suitable for both "normal" and "proxy" responses. 

In addition to that, in case the response returns a warning/error, diffs are also logged with warning/error level accordingly.

# Testing

There are no special tests for this logging format change. One can run a FerretDB instance locally and send various queries from the client to see what the log levels are.


internal/clientconn/conn.go
```diff
@@ -228,7 +235,7 @@ func (c *conn) run(ctx context.Context) (err error) {
 		}
 
 		if resHeader == nil || resBody == nil {
-			c.l.Info("no response to send to client")
+			c.l.Debugf("no response to send to client")
```
rumyantseva, line 238: I'm not sure about this one, but the message itself doesn't look very informative.

internal/clientconn/conn.go
```diff
@@ -392,11 +390,42 @@ func (c *conn) handleOpMsg(ctx context.Context, msg *wire.OpMsg, cmd string) (*w
...
+// logResponse logs response's header and body and returns the log level that was used.
+// If op code is not OP_MSG, it always logs as a debug.
+// For the OP_MSG code, the level depends on the type of error.
+// If there is no errors in the response, it will be logged as a debug.
+// If there is an error in the response, and connection is closed, it will be logged as an error.
+// If there is an error in the response, and connection is not closed, it will be logged as a warning.
+func (c *conn) logResponse(resHeader *wire.MsgHeader, resBody wire.MsgBody, closeConn bool) zapcore.Level {
+	level := zap.DebugLevel
+
+	if resHeader.OpCode == wire.OP_MSG {
```
AlekSi, line 411: It was renamed in another PR

AlekSi (changes requested):

internal/clientconn/conn.go
rumyantseva: Fixed!

internal/clientconn/conn.go
seeforschauer: Is it true that If the header or body is nil there is no response sent to the client?

rumyantseva: @seeforschauer it seems so, yes. It looks a bit strange to me - there is not an error here and nothing else...

seeforschauer: the condition and the logging seem like an error - I expect that it was not sent only if both `resHeader` and `resBody` are nil.

rumyantseva: @AlekSi what do you think about this part of the code? It's not really straightforward which log level should be here, especially, if the client doesn't get any response... In which case it's possible that the response doesn't have any content? Is it a sign that something went wrong and we should panic? Or at least log an error?

internal/clientconn/conn.go
```diff
@@ -392,11 +390,42 @@ func (c *conn) handleOpMsg(ctx context.Context, msg *wire.OpMsg, cmd string) (*w
...
+// If there is an error in the response, and connection is closed, it will be logged as an error.
+// If there is an error in the response, and connection is not closed, it will be logged as a warning.
+func (c *conn) logResponse(resHeader *wire.MsgHeader, resBody wire.MsgBody, closeConn bool) zapcore.Level {
+	level := zap.DebugLevel
+
+	if resHeader.OpCode == wire.OpCodeMsg {
+		doc, err := resBody.(*wire.OpMsg).Document()
+		must.NoError(err)
+
+		ok, err := doc.Get("ok")
```
w84thesun, line 415: You could use `must.NotFail` here.

internal/clientconn/conn.go
```diff
@@ -392,11 +390,42 @@ func (c *conn) handleOpMsg(ctx context.Context, msg *wire.OpMsg, cmd string) (*w
...
+// If op code is not OP_MSG, it always logs as a debug.
+// For the OP_MSG code, the level depends on the type of error.
+// If there is no errors in the response, it will be logged as a debug.
+// If there is an error in the response, and connection is closed, it will be logged as an error.
+// If there is an error in the response, and connection is not closed, it will be logged as a warning.
+func (c *conn) logResponse(resHeader *wire.MsgHeader, resBody wire.MsgBody, closeConn bool) zapcore.Level {
+	level := zap.DebugLevel
+
+	if resHeader.OpCode == wire.OpCodeMsg {
+		doc, err := resBody.(*wire.OpMsg).Document()
```
w84thesun, line 412: You could use `must.NotFail` here.

internal/clientconn/conn.go
```diff
@@ -247,7 +254,7 @@ func (c *conn) run(ctx context.Context) (err error) {
 	}
 }
 
-// Route routes the message.
+// route sends request to a handler's command based on the op code provided in the request header.
```
w84thesun, line 257: I think we should change the next comment line `Route` -> `route` also.

w84thesun (changes requested):

internal/clientconn/conn.go
rumyantseva: @w84thesun I was thinking about it but I didn't get how :) Could you give me an example (e.g. where should I define a variable for `doc` etc)?

internal/clientconn/conn.go
rumyantseva: Well, the next like is the beginning of a new sentence, and I'm not sure if we should start it from a lowercased letter.

internal/clientconn/conn.go
AlekSi, line 412:
```suggestion
		doc := must.NotFail(resBody.(*wire.OpMsg).Document())
```

internal/clientconn/conn.go
rumyantseva: Rephrased.

internal/clientconn/conn.go
rumyantseva: Fixed

internal/clientconn/conn.go
rumyantseva: Fixed

internal/clientconn/conn.go
rumyantseva: Fixed. As agreed in call, throw panic instead.

seeforschauer (approved):

w84thesun (approved):

AlekSi (approved):

Issue #661, opened 2022-05-31
Add update expression checks


seeforschauer:
- [ ] Update document must contain key beginning with '$
- [ ] An empty update path is not valid

Details

# Update document must contain key beginning with '$

Refs #622 

```go
...
"test-case": {
		id: "double",
		update: bson.D{{
			"value",
			bson.D{{
				"array",
				bson.D{{"$currentDate", bson.D{{"unexsistent", bson.D{}}}}},
			}},
		}},
		err: fmt.Errorf("update document must contain key beginning with '$'"),
	},
...
...


res, err := collection.UpdateOne(ctx, bson.D{{"_id", tc.id}}, tc.update)
```

# An empty update path is not valid

Refs #662.

```go
"EmptyPath": {
	id:     "double",
	update: bson.D{{"$currentDate", bson.D{{"", true}}}},
	err: &mongo.WriteError{
		Code:    56,
		Message: "An empty update path is not valid.",
	},
},
```

fenogentov:
Related tasks #673?

PR #662, opened 2022-05-31
Support `$currentDate` field update operator

### LINKED ISSUE #622: Support `$currentDate` field update operator
Dot notation is not in scope. Use UnimplementedDot if available.

In a separate PR, timestamp:
```go
"Timestamp": {
        id:     "double",
	update: bson.D{{"$currentDate", bson.D{{"value", bson.D{{"$type", "timestamp"}}}}}},
	stat: &mongo.UpdateResult{
		MatchedCount:  1,
		ModifiedCount: 1,
		UpsertedCount: 0,
	},
        result: bson.D{{"_id", "double"}, {"value", nowTimestamp}},
},
```
where
```go
nowTimestamp := primitive.Timestamp{T: uint32(time.Now().UTC().Unix()), I: uint32(0)}
```

seeforschauer:
Closes #622.


seeforschauer:
I invite you not to review but to discuss. 
I need somehow to test data time values.
And all my ideas are overcomplicated.
Can we do it simply? 
Do we have an agreed way on how to test it?

internal/handlers/common/update.go
```diff
@@ -18,18 +18,175 @@ import (
...
+					)
+				}
+			}
+		default:
+			return NewWriteErrorMsg(
+				ErrFailedToParse,
+				"Modifiers operate on fields but we found another type instead",
+			)
+		}
+	}
```
seeforschauer, line 101: I see here 4 defaults here, and it's comfortable to compare with the original structure where data change applies. 
But maybe you have other suggestions. What are your thoughts?

internal/handlers/common/update.go
```diff
@@ -18,18 +18,175 @@ import (
...
+					panic("The '$type' string field is required to be 'date' or 'timestamp'")
+				}
+
+			default:
+				panic("not valid type for $currentDate. Please use a boolean ('true') " +
+					"or a $type expression ({$type: 'timestamp/date'}).",
+				)
+			}
+		}
+	default:
```
seeforschauer, line 169: here also.

internal/handlers/common/update.go
```diff
@@ -18,18 +18,175 @@ import (
...
+	"time"
 
 	"github.com/FerretDB/FerretDB/internal/types"
 	"github.com/FerretDB/FerretDB/internal/util/must"
 )
 
+// CheckCurrentdateFieldExpression checks $currentDate input on correctness.
+func CheckCurrentdateFieldExpression(update *types.Document) error {
+	for _, updateOp := range update.Keys() {
+		if updateOp != "$currentDate" {
```
seeforschauer, line 30: I need my PR for `$setOnInsert` here so when it will be merged, this code will change also.

seeforschauer:
`timestamp` is extracted as more complex - probably there will be more and more changes.

seeforschauer:
> `timestamp` is extracted as more complex - probably there will be more and more changes.

I've added it. Please see - there are some changes in types.

internal/types/timestamp.go
```diff
@@ -0,0 +1,50 @@
...
+	"sync/atomic"
+	"time"
+)
+
+type (
+	// Timestamp represents BSON type Timestamp.
+	Timestamp uint64
+)
+
+// timestampOperationCounter is an ordinal number for timestamps in the system.
```
rumyantseva, line 27: Can we explain in this comment why is this variable needed? I see it's increased every time when `NextTimestamp` is called. But why? I guess, when the app starts, this var is initialized with `0` (I don't see any special initialization)... And then every time when `NextTimestamp` is called, this var increases. What is the use case for it?

internal/handlers/common/update.go
```diff
@@ -208,3 +272,66 @@ func extractValueFromUpdateOperator(op string, update *types.Document) (*types.D
...
+			}
+
+			currentDateTypeString, ok := currentDateType.(string)
+			if !ok {
+				return NewWriteErrorMsg(
+					ErrBadValue,
+					"The '$type' string field is required to be 'date' or 'timestamp'",
+				)
+			}
+			if !slices.Contains([]string{"date", "timestamp"}, currentDateTypeString) {
```
rumyantseva, line 319: Do we need `slices.Contains` here? Why just simple `currentDateTypeString != "date" && currentDateTypeString != "timestamp"` wouldn't be enough?

internal/handlers/common/update.go
```diff
@@ -110,6 +120,55 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
 }
 
+// processCurrentdateFieldExpression changes document according to $currentDate operator.
+// If the document was changed it returns true.
+func processCurrentdateFieldExpression(doc *types.Document, currentDateVal any) (bool, error) {
+	var changed bool
+	var err error
+	currentDateExpression := currentDateVal.(*types.Document)
+
+	now := time.Now()
```
rumyantseva, line 130: I wonder if we should take care of timezones and other things somewhere... I guess, this is enough, but I just want us to be mindful about it :)

internal/handlers/common/update.go
```diff
@@ -110,6 +120,55 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
+	sort.Strings(currentDateExpression.Keys())
```
rumyantseva, line 131: Why do the keys need to be sorted?

internal/handlers/common/update.go
```diff
@@ -110,6 +120,55 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
+					return false, err
+				}
+				changed = true
+				continue
+			}
+
+			currentDateType = currentDateType.(string)
+			switch currentDateType {
+			case "timestamp":
+				if err := doc.Set(field, types.NextTimestamp(uint64(now.UTC().Unix()))); err != nil {
```
rumyantseva, line 156: Oh, here we do some conversion to UTC, but we didn't do it for `doc.Set` above. Is it because the timestamp is always stored in UTC (I've never thought about it)?

integration/update_test.go
```diff
@@ -673,3 +677,211 @@ func TestUpdateMany(t *testing.T) {
...
+			},
+			paths: []types.Path{
+				types.NewPathFromString("value"),
+				types.NewPathFromString("unexistent"),
+			},
+			result: bson.D{{"_id", "double"}, {"value", now}, {"unexistent", now}},
+		},
+		"False": {
+			id:     "double",
+			update: bson.D{{"$currentDate", bson.D{{"value", false}}}},
```
rumyantseva, line 766: So, both `true` and `false` are valid (I agree).

Does it make sense to add more corner case like `bson.D{{"$currentDate", bson.D{{"value", nil}}}}` and `bson.D{{"$currentDate", bson.D{{"", true}}}}`?

rumyantseva (commented):
In general, I don't have major concerns.
I asked a few questions where I couldn't understand some logic.
Also, I offered a couple of additional corner cases for tests.

integration/update_test.go
fenogentov: bson.D{{"$currentDate", bson.D{{"value", nil}}}}

seeforschauer: > Does it make sense to add more corner case like `bson.D{{"$currentDate", bson.D{{"value", nil}}}}` 

This test works as expected: no `$type' field gives an error.


> and `bson.D{{"$currentDate", bson.D{{"", true}}}}`?

This test case has been added in #661 issue.

internal/handlers/common/update.go
seeforschauer: > I wonder if we should take care of timezones and other things somewhere... I guess, this is enough, but I just want us to be mindful about it :)

Here we have to use UTC. Changed to UTC and tests didn't fail. Interesting.

internal/handlers/common/update.go
seeforschauer: > Why do the keys need to be sorted?

The addition of the new keys influences the order inside the document. 
The process of document fields has to be in lexicographic order.

internal/handlers/common/update.go
seeforschauer: > Oh, here we do some conversion to UTC, but we didn't do it for `doc.Set` above. Is it because the timestamp is always stored in UTC (I've never thought about it)?

For `Date` it's explicitly specified that `Date` has to be in UTC. For timestamp - is not. But I guess as the operator is the same, It's also should be in UTC.

internal/handlers/common/update.go
seeforschauer: > Do we need `slices.Contains` here? Why just simple `currentDateTypeString != "date" && currentDateTypeString != "timestamp"` wouldn't be enough?


I find the expression `currentDateTypeString != "date" && currentDateTypeString != "timestamp"` difficult to read. See, it has to be with a negation of two:

`! ( currentDateTypeString == "date" || 
    currentDateTypeString == "timestamp")`

internal/types/timestamp.go
seeforschauer: > Can we explain in this comment why is this variable needed? I see it's increased every time when `NextTimestamp` is called. But why? I guess, when the app starts, this var is initialized with `0` (I don't see any special initialization)... And then every time when `NextTimestamp` is called, this var increases. What is the use case for it?

It is to differentiate the timestamps between operations. Actually, it can be not for the entire system but inside a single one timestamp. It depends on our chosen way, @AlekSi WDYT?

internal/handlers/common/update.go
```diff
@@ -110,6 +120,55 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
 	return changed, nil
 }
 
+// processCurrentdateFieldExpression changes document according to $currentDate operator.
+// If the document was changed it returns true.
+func processCurrentdateFieldExpression(doc *types.Document, currentDateVal any) (bool, error) {
```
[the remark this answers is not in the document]
seeforschauer: > I suggest changing the function name to `processCurrentDateFieldExpression`.

renamed.

internal/types/timestamp.go
AlekSi: It must be process-wide if we ever implement MongoDB-like replication. And it does not hurt to be process-wide even if we don't.

It and some other arguments, however, should be uint32, not uint64.

And let's use some shorter names. For example, we use `objectIDCounter`, so that one can be `timestampCounter` and argument names could be `c` and `ts`.

AlekSi: Another option, that is a bit closer to https://bsonspec.org/spec.html: `timestampIncrement` and `inc`

seeforschauer: > And let's use some shorter names. For example, we use `objectIDCounter`, so that one can be `timestampCounter` and argument names could be `c` and `ts`.

variable renamed to `timestampCounter`.
variable name `ts` looks understandable.
variable name `c` looks not clear. It's not `sec` since it's `c`. What is it?

seeforschauer: ah, `c` - counter!

seeforschauer: > It and some other arguments, however, should be uint32, not uint64.

`timestampCounter` type changed to uint32.
In other places int64 for timestamp as an original type.

w84thesun (approved):

w84thesun (approved):

seeforschauer:
@Dmitry @rumyantseva PTAL, I've changed to `time.Time`

w84thesun (approved):

internal/types/timestamp.go
```diff
@@ -0,0 +1,51 @@
...
+func NextTimestamp(t time.Time) Timestamp {
+	sec := t.UTC().Unix()
+	c := atomic.AddUint32(&timestampCounter, 1)
+	sec <<= 32
+	sec |= int64(c)
+	return Timestamp(sec)
+}
+
+// DateTime returns time.Time ignoring increment.
+func DateTime(t Timestamp) time.Time {
```
AlekSi, line 48: Let's make it a method: `func (t Timestamp) Time() time.Time`

internal/types/timestamp.go
```diff
@@ -0,0 +1,51 @@
...
+	t >>= 32
+	return time.UnixMilli(int64(t)).UTC()
```
AlekSi, line 50: `.UTC()` does nothing there

internal/types/timestamp.go
```diff
@@ -0,0 +1,51 @@
...
+	sec <<= 32
+	sec |= int64(c)
+	return Timestamp(sec)
+}
+
+// NextTimestamp returns a timestamp from time and an internal ops counter.
+func NextTimestamp(t time.Time) Timestamp {
+	sec := t.UTC().Unix()
+	c := atomic.AddUint32(&timestampCounter, 1)
+	sec <<= 32
```
AlekSi, line 42: Looks like we could just call `NewTimestamp(t, c)` there to avoid copying the same code

internal/types/timestamp.go
```diff
@@ -0,0 +1,51 @@
...
+	// Timestamp represents BSON type Timestamp.
+	Timestamp int64
+)
+
+// timestampCounter is an ordinal number for timestamps in the system.
+var timestampCounter uint32
+
+// NewTimestamp returns a timestamp from time and an increment.
+func NewTimestamp(t time.Time, c uint32) Timestamp {
+	sec := t.UTC().Unix()
```
AlekSi, line 32: `.UTC()` does nothing there. Unix timestamps are always in UTC

integration/update_test.go
```diff
@@ -714,3 +718,211 @@ func TestUpdateMany(t *testing.T) {
...
+			for _, path := range tc.paths {
+				testutil.CompareAndSetByPathTime(
+					t,
+					expected,
+					actual,
+					maxDifference,
+					path,
+				)
+				expected.RemoveByPath(path)
+				actual.RemoveByPath(path)
```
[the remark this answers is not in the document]
seeforschauer: > We do we set values by path only to remove them on the next line?

There is  `assert.Equal(t, expected, actual)` below,  and it tests whether other document values are ok (not changed, not removed, not updated anyhow).

internal/types/timestamp.go
seeforschauer: > Looks like we could just call `NewTimestamp(t, c)` there to avoid copying the same code

amazing.

internal/types/timestamp.go
seeforschauer: > `.UTC()` does nothing there. Unix timestamps are always in UTC

changed.

internal/types/timestamp.go
seeforschauer: > Let's make it a method: `func (t Timestamp) Time() time.Time`

beautiful :heart:

internal/types/timestamp.go
seeforschauer: > `.UTC()` does nothing there

changed.

internal/types/timestamp.go
seeforschauer: > Let's make it a method: `func (t Timestamp) Time() time.Time`

changed.

seeforschauer (commented):
Thank you for the review,  @AlekSi, :+1: 
I've added changes in #701.

PR #663, opened 2022-05-31
Explicitly disallow duplicate keys in documents

fenogentov:


AlekSi:
We recently restarted our work on improving validation. We now understand that while data documents (that we would store in the database) should probably not include duplicate fields, command documents can. See #666, for example.

I will close that one because the implementation should be very different.

PR #670, opened 2022-05-31
Support `$rename` field update operator

### LINKED ISSUE #626: Support `$rename` field update operator
https://www.mongodb.com/docs/manual/reference/operator/update/rename/

### DoD

Dot notation should be supported and tested. We should also check if we need to add test cases for dot notation to other compat update tests.

Update documentation about supported operators.

fenogentov:
Closes #626.


integration/update_test.go
```diff
@@ -460,3 +461,158 @@ func TestUpdateSet(t *testing.T) {
...
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			ctx, collection := setup(t)
+
+			_, err := collection.InsertMany(ctx, []any{
+				bson.D{{"_id", "1"}, {"name", "alex"}, {"phone", "9012345678"}},
+				bson.D{{"_id", "2"}, {"name", "bob"}},
+			})
+			require.NoError(t, err)
```
AlekSi, line 596: I think that it becomes a problem that, unlike `find` tests, each `update` test uses its own data set. Let's put that PR on pause until we decide if we want to do something about that. Created #674 for tracking

AlekSi: Let's hook into update_field_compat_test.go instead of using test-specific data set

AlekSi (changes requested):
Please also resolve merge conflicts

internal/handlers/common/update.go
```diff
@@ -113,6 +113,46 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
 				}
 			}
 
+		case "$rename":
```
w84thesun, line 116: Please move that code to the new function `processRenameFieldExpression`.

w84thesun (changes requested):

rumyantseva (commented):
Let's make the necessary changes.

rumyantseva:
Supported in #1753.

PR #522, opened 2022-04-25
[WIP] Added handling of composite data types for comparison operator `$gt``$lt`, `$gte``$lte`

ribaraka:

PR #676, opened 2022-06-01
Array comparison substitution

ribaraka:
Refs #457, #522. 

Removed comparison of two arrays by serialization.
If this implementation is okay, I'm going to improve and apply this function in the next #522 PR. In any case, this PR is intended to replace serialization.


internal/handlers/common/filter.go
```diff
@@ -248,7 +248,7 @@ func filterFieldExpr(doc *types.Document, filterKey string, expr *types.Document
 				return false, nil
 			case *types.Array:
 				if fieldValue, ok := fieldValue.(*types.Array); ok {
-					return matchArrays(exprValue, fieldValue), nil
+					return types.CompareArrays(exprValue, fieldValue) == types.Equal, nil
 				}
 				return false, nil
```
AlekSi, line 253: Quick thought: can we use a regular `types.Compare` there instead of both type assertion and a separate function? Can we un-export `CompareArrays`?

internal/types/compare.go
```diff
@@ -254,3 +258,103 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+				res = handleSubArrayComparingResult(i, &res, &EntireArrayResult, &subArrayEquality)
+				continue
+
+			default:
+				res := CompareArrays(filterArr, arrValue)
+				res = handleSubArrayComparingResult(i, &res, &EntireArrayResult, &subArrayEquality)
+			}
+			continue
+
+		// TODO: case Document
```
seeforschauer, line 291: should be here a link to an issue?

internal/types/compare.go
```diff
@@ -254,3 +258,103 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+
+	for i := 0; i < docArr.Len(); i++ {
+		arrValue := must.NotFail(docArr.Get(i))
+		switch arrValue := arrValue.(type) {
+		case *Array:
+			filterArrValue := must.NotFail(filterArr.Get(i))
+			switch filterArrValue := filterArrValue.(type) {
+			case *Array:
+				res := CompareArrays(filterArrValue, arrValue)
+				res = handleSubArrayComparingResult(i, &res, &EntireArrayResult, &subArrayEquality)
```
seeforschauer, line 282: Sorry - it's not clear. 
First, we compare filter value and array - what do we get? what for? 
Then it's understandable name,  elements of the array are compared. But the result is overwritten by the following statement, isn't it? Can we do here an early exit: if one of the elements is not Equal, then return?

internal/types/compare.go
```diff
@@ -254,3 +258,103 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+				}
+				continue
+			default:
+				res := CompareOrder(arrValue, filterValue, Ascending)
+				if EntireArrayResult == Incomparable && i == 0 { // set first non-Incomparable result
+					EntireArrayResult = res
+				}
+
+				if EntireArrayResult != res {
+					EntireArrayResult = Incomparable
```
seeforschauer, line 316: here and there are also early returns. Isn't it?

internal/types/compare.go
ribaraka: we get a value from `doc array` by an index; if the value is another array(subarray) then we should iterate through it in case the subarray might fit filter array; but what if we got a value from `filter array` by the index it is another array (filter sub array) - then we comparing by this `filterSubArray` so to say. And I don't think earlier return/exit is possible bsc there is a chance on last itterattion we might get sub array that might fit filter array.

internal/types/compare.go
ribaraka: looking for next element/value is array that might fit filter array.

internal/types/compare.go
ribaraka: Not nessesary. I'm going to work on this `TODO` as soon as this PR merged. I just want to make small and simple PRs. baby steps :smile:

AlekSi (dismissed):

internal/types/compare.go
```diff
@@ -254,3 +258,96 @@ func compareNumbers(a float64, b int64) CompareResult {
...
+// returns Equal when a document array contains another array(subarray) that equals filter array.
+func CompareArrays(filterArr, docArr *Array) CompareResult {
+	if docArr.Len() == 0 && filterArr.Len() == 0 {
+		return Equal
+	}
+	if filterArr.Len() == 0 {
+		return Incomparable
+	}
+
+	EntireArrayResult, subArrayEquality := Incomparable, Incomparable
```
AlekSi, line 272:
```suggestion
	entireArrayResult, subArrayEquality := Incomparable, Incomparable
```

internal/types/compare.go
```diff
@@ -254,3 +258,96 @@ func compareNumbers(a float64, b int64) CompareResult {
 
 	return CompareResult(bigA.Cmp(bigB))
 }
+
+// CompareArrays compares indices of a filter array according to indices of a document array;
+// returns Equal when a document array contains another array(subarray) that equals filter array.
+func CompareArrays(filterArr, docArr *Array) CompareResult {
```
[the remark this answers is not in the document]
ribaraka: fixed

internal/types/compare.go
ribaraka: fixed

AlekSi (approved):

PR #681, opened 2022-06-01
Build `tigris` handler only if tag is present

AlekSi:
Building without it produces a much smaller binary.

### DIFF
--- a/cmd/ferretdb/dummy.go
+++ b/cmd/ferretdb/dummy.go
@@ -12,23 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build testcover
-// +build testcover
-
 package main
 
 import (
-	"os"
-	"testing"
+	"github.com/FerretDB/FerretDB/internal/handlers"
+	"github.com/FerretDB/FerretDB/internal/handlers/dummy"
 )
 
-// TestCover allows us to run FerretDB with coverage enabled.
-func TestCover(t *testing.T) {
-	main()
-}
-
-// TestMain ensures that command-line flags are initialized correctly when FerretDB is run with coverage enabled.
-func TestMain(m *testing.M) {
-	initFlags()
-	os.Exit(m.Run())
+func init() {
+	registeredHandlers["dummy"] = func(*newHandlerOpts) (handlers.Interface, error) {
+		return dummy.New(), nil
+	}
 }
--- a/cmd/ferretdb/main.go
+++ b/cmd/ferretdb/main.go
@@ -47,6 +47,7 @@ type newHandlerOpts struct {
 }
 
 // registeredHandlers maps handler names to constructors.
+// The values for `registeredHandlers` must be set through the `init()` functions of the corresponding handlers so that we can control which handlers will be included in the build with build tags.
 var registeredHandlers = map[string]newHandler{}
 
 var (
--- a/cmd/ferretdb/main.go
+++ b/cmd/ferretdb/main.go
@@ -35,18 +35,18 @@ import (
 	"github.com/FerretDB/FerretDB/internal/util/version"
 )
 
-// newHandlerOpts represents configuration for all handlers.
+// newHandler represents a function that constructs a new handler.
+type newHandler func(opts *newHandlerOpts) (handlers.Interface, error)
+
+// newHandlerOpts represents common configuration for constructing handlers.
 //
-// Some fields may be unused by some handlers.
+// Handler-specific configuration is passed via command-line flags directly.
 type newHandlerOpts struct {
 	ctx    context.Context
 	logger *zap.Logger
 }
 
-// newHandler is a function that constructs a new handler.
-type newHandler func(opts *newHandlerOpts) (handlers.Interface, error)
-
-// registeredHandlers map handler names for constructors.
+// registeredHandlers maps handler names to constructors.
 var registeredHandlers = map[string]newHandler{}
 
 var (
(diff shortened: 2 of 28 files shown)


cmd/ferretdb/dummy.go
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+
+package main
+
+import (
+	"github.com/FerretDB/FerretDB/internal/handlers"
+	"github.com/FerretDB/FerretDB/internal/handlers/dummy"
+)
+
+func init() {
```
rumyantseva, line 22: Should we comment on why we need `init` here?

cmd/ferretdb/main.go
```diff
@@ -19,37 +19,62 @@ import (
...
-	"github.com/FerretDB/FerretDB/internal/handlers/pg"
-	"github.com/FerretDB/FerretDB/internal/handlers/pg/pgdb"
 	"github.com/FerretDB/FerretDB/internal/util/debug"
 	"github.com/FerretDB/FerretDB/internal/util/logging"
+	"github.com/FerretDB/FerretDB/internal/util/must"
 	"github.com/FerretDB/FerretDB/internal/util/version"
 )
 
-//nolint:gochecknoglobals // flags are defined there to be visible in `bin/ferretdb-testcover -h` output
+// newHandlerOpts represents configuration for all handlers.
```
rumyantseva, line 38: For all handlers or for any handler?

cmd/ferretdb/pg.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"time"
+
+	"github.com/FerretDB/FerretDB/internal/handlers"
+	"github.com/FerretDB/FerretDB/internal/handlers/pg"
+	"github.com/FerretDB/FerretDB/internal/handlers/pg/pgdb"
+)
+
+func init() {
```
rumyantseva, line 25: Same here. I believe, this coding choice needs to be documented somewhere. Otherwise, one day someone might want to get rid of those `init`s :)

I guess, the idea is that we can exclude various handlers from the build, so the binary is smaller. But it needs to be documented that we use `init`s for that particular reason - so the main func doesn't need to know about handlers packages.

rumyantseva (changes requested):
Please document the reasons behind the `init` coding choices (see my comments).

cmd/ferretdb/pg.go
AlekSi: > I guess, the idea is that we can exclude various handlers from the build, so the binary is smaller.

I mean, yeah, that's in the PR description :)

cmd/ferretdb/main.go
```diff
@@ -19,37 +19,62 @@ import (
...
-	proxyAddrF       = flag.String("proxy-addr", "127.0.0.1:37017", "")
-	versionF         = flag.Bool("version", false, "print version to stdout (full version, commit, branch, dirty flag) and exit")
+	versionF = flag.Bool("version", false, "print version to stdout (full version, commit, branch, dirty flag) and exit")
+
+	listenAddrF = flag.String("listen-addr", "127.0.0.1:27017", "listen address")
+	proxyAddrF  = flag.String("proxy-addr", "127.0.0.1:37017", "proxy address")
+	debugAddrF  = flag.String("debug-addr", "127.0.0.1:8088", "debug address")
+	modeF       = flag.String("mode", string(clientconn.AllModes[0]), fmt.Sprintf("operation mode: %v", clientconn.AllModes))
+
+	handlerF       = flag.String("handler", "pg", "<set in initFlags()>")
```
seeforschauer, line 60: Can we add a comment that it's set in *`init` functions for the handler? 
And rename handler initialization files to `handler_dummy.go`, `handler_pg.go`,  `handler_tigris.go`?

cmd/ferretdb/main.go
AlekSi: tbh, I'm not sure what word should be used there. Does the next sentence help?

AlekSi: Reworded to remove both "any" and "all" :)

seeforschauer (dismissed):

cmd/ferretdb/main.go
```diff
@@ -19,37 +19,62 @@ import (
...
+
+// newHandlerOpts represents common configuration for constructing handlers.
+//
+// Handler-specific configuration is passed via command-line flags directly.
+type newHandlerOpts struct {
+	ctx    context.Context
+	logger *zap.Logger
+}
+
+// registeredHandlers maps handler names to constructors.
```
rumyantseva, line 49:
```suggestion
// registeredHandlers maps handler names to constructors.
// The values for `registeredHandlers` must be set through the `init()` functions of the corresponding handlers so that we can control which handlers will be included in the build with build tags.
```

rumyantseva (dismissed):
LGTM

cmd/ferretdb/main.go
```diff
@@ -19,37 +19,63 @@ import (
...
+// The values for `registeredHandlers` must be set through the `init()` functions of the corresponding handlers so that we can control which handlers will be included in the build with build tags.
```
rumyantseva, line 50:
```suggestion
// The values for `registeredHandlers` must be set through the `init()` functions of the corresponding handlers
// so that we can control which handlers will be included in the build with build tags.
```
Make linter happy

rumyantseva (approved):
LGTM

seeforschauer (approved):

PR #685, opened 2022-06-02
Support `$all` array query operator

### LINKED ISSUE #255: Support all array query operators
Support all query comparison operators for all data types.

- [x] #657
- [x] #699
- [x] $size

### LINKED ISSUE #657: Support `$all` array query operator
https://www.mongodb.com/docs/manual/reference/operator/query/all/

rumyantseva:
## Description

This PR closes #657.

The implementation is the same as for the other filter operators. We validate if the doc matches the conditions set in `$all`.

The implementation is simple and not optimal. For example, the function `ContainsAll` that checks if the left array contains all the values of the right array can be improved for better performance.

## Testing

Two types of tests are added:

- Integrational tests for `$all` operator
- Unit tests for new `Array` operations (`Contains`, `ContainsAll`)

## Checklist

- [x] Code is regenerated if needed (`task gen`).
- [x] Code is formatted (`task fmt`).
- [x] Docs are formatted (`task docs-fmt`).
- [x] Tests passed (`task test`).
- [x] Linters passed (`task lint`).
- [x] PR assignee is set.
- [x] PR labels are set.
- [x] Project and iternation are set.
- [x] Critical codecov complaints are considered.
- [x] PR title and body are in accordance with the Conform PR GitHub Action.
- [x] Other GitHub Actions passed.


internal/handlers/common/filter.go
```diff
@@ -450,6 +450,13 @@ func filterFieldExpr(doc *types.Document, filterKey string, expr *types.Document
 				return false, err
 			}
 
+		case "$all":
+			// {field: {$all: [value, another_value, ...]}}
```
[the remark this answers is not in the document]
rumyantseva: I added it to the `integration/query_array_test.go` file, as this case looks similar to the `$size` op case.

internal/types/array.go
```diff
@@ -111,3 +111,38 @@ func (a *Array) Append(values ...any) error {
...
+	return false
+}
+
+// ContainsAll checks if the Array contains all the given values of the Array b.
+// Currently, this algorithm is O(n^2) without any performance tuning.
+// Possible improvement: to sorts both arrays before comparing.
+// If sorting is implemented, calling of the Contains function will be O(n*log(n)) instead of O(n^2).
+func (a *Array) ContainsAll(b *Array) bool {
+	if a.Len() < b.Len() {
+		// TODO: How to validate that all the values in b are unique?
```
rumyantseva, line 131: A controversial example:
`a`: `{1, 2}`
`b`: `{1, 1, 1}`

I believe, in this case, `ContainsAll` should return true. But for that, we either need to remove this `if a.Len() < b.Len()` condition or make `b` unique. 

@AlekSi what do you think?

internal/types/array.go
```diff
@@ -111,3 +111,38 @@ func (a *Array) Append(values ...any) error {
...
+		return false
+	}
+
+	// TODO: This place can be significantly improved:
```
rumyantseva, line 135: @AlekSi as I mentioned before, currently this algorithm is super slow. If you think we need to optimize it right now, please choose what optimization we should use.

integration/query_array_test.go
```diff
@@ -137,3 +137,68 @@ func TestQueryArraySize(t *testing.T) {
...
+			expectedIDs: []any{"array-many", "array-one", "array-three", "array-two"},
+		},
+		"float": {
+			filter:      bson.D{{"value", bson.D{{"$all", bson.A{12.5}}}}},
+			expectedIDs: []any{"array-many", "float"},
+		},
+		"multi_all": {
+			filter:      bson.D{{"value", bson.D{{"$all", bson.A{"1", 2}}}}},
+			expectedIDs: []any{"array-many", "array-three"},
+		},
```
rumyantseva, line 176: One more case to add is 

```go
filter: bson.D{{"value", bson.D{{"$all", bson.A{"1", "1", "1"}}}}},
```

which should return all the docs where at least one `value: "1"` is presented.

internal/types/array.go
```diff
@@ -111,3 +111,38 @@ func (a *Array) Append(values ...any) error {
 func (a *Array) RemoveByPath(path Path) {
 	removeByPath(a, path)
 }
+
+// Contains checks if the Array contains the given value.
+func (a *Array) Contains(value any) bool {
+	for _, v := range a.s {
```
seeforschauer, line 117: Why don't use here `slices.Contains`?

internal/types/array.go
seeforschauer, line 131: > // TODO: How to validate that all the values in b are unique?

Do we need to validate that they are unique? The Array with non-unique array elements is a valid `*Array` type.

seeforschauer: > A controversial example: `a`: `{1, 2}` `b`: `{1, 1, 1}`
> 
> I believe, in this case, `ContainsAll` should return `true`. 

What MongoDB international test gives? There are some things that are non-usual in Mongo and for better compatibility and faster migration to the FerretDB it has to be the same.

internal/types/array_test.go
```diff
@@ -77,3 +77,78 @@ func TestArray(t *testing.T) {
...
 	})
 }
+
+func TestArrayContains(t *testing.T) {
+	for name, tc := range map[string]struct {
+		array    *Array
+		filter   any
+		expected bool
+	}{
+		"string": {
```
seeforschauer, line 87: WOuld you please use CamelCase for test keys for the code similarity? Thank you.

seeforschauer (changes requested):

internal/types/array.go
seeforschauer, line 135: To measure effectiveness, it's necessary to add other tests, loaded with data including big arrays and the number of arrays will be also big enough to do a high load-testing.

It's a separate task and definitely, a separate PR and it can become a larger task and cover not only this one `$all` operation but also other update operators. And it can be measured carefully the influence on other operators performance... So it's a big another task.

internal/types/array.go
rumyantseva: Ah! This is actually something new to me :) Yeah, I believe I should use it!

internal/types/array.go
rumyantseva: No, this is not the problem. The problem is this optimization `if a.Len() < b.Len()` doesn't work if `b` has repeated values. Probably, I should just delete the whole optimization to not make this confusion.

internal/types/array.go
rumyantseva: @seeforschauer well, I don't need to add other tests to tell that O(n^2) is worse than O(n*log n) :)

internal/types/array_test.go
rumyantseva: So, this https://github.com/FerretDB/github-actions/blob/7ace04553a01a45fb914cb3928bbcde55a4abe21/detect-matching-pr/main_test.go#L19 format is wrong then?

rumyantseva: Yeah, in this file it's even not `camelCase` but rather `TitleCase` :)

internal/types/array.go
seeforschauer: > @seeforschauer well, I don't need to add other tests to tell that `O(n^2)` is worse than `O(n*log n)` :)

Then why not just implement it in a better way? What is needed for it?

internal/types/array.go
seeforschauer: > Probably, I should just delete the whole optimization to not make this confusion.

If there is optimization, it still works for some arrays.
If you delete optimization, it stops work for that case. 
It doesn't worsen the processing, is it?

internal/types/array.go
rumyantseva: @seeforschauer time :) Basically, my main question is about the future strategy. If all the things like `$all` should be called on the backend level, it doesn't make sense to use fancy algorithms on FerretDB. But if we are going to leave with `$all` being a "post-filter", I'd rather invest some time to offer a better algorithm right now.

integration/query_array_test.go
rumyantseva: Added

integration/query_array_test.go
```diff
@@ -137,3 +137,84 @@ func TestQueryArraySize(t *testing.T) {
...
 	}
 }
+
+// TestQueryArrayAll covers the case where the $all operator is used on an array or scalar.
+func TestQueryArrayAll(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setup(t)
+
+	// Valid cases when the value for $all is an array.
+	_, err := collection.InsertMany(ctx, []any{
```
AlekSi, line 147: Any reason we don't use shared data?

integration/query_array_test.go
```diff
@@ -137,3 +137,84 @@ func TestQueryArraySize(t *testing.T) {
...
+
+			var actual []bson.D
+			err = cursor.All(ctx, &actual)
+			require.NoError(t, err)
+			assert.Equal(t, tc.expectedIDs, CollectIDs(t, actual))
+		})
+	}
+
+	// Handle a case when the value for $all is a scalar.
+	t.Run("Op$allNeedsAnArray", func(t *testing.T) {
```
AlekSi, line 214: Can that test be a part of the test above? `err *mongo.CommandError` seems to be unused otherwise.

internal/types/array.go
AlekSi: I think it is okay as it is for now. We will be able to see that performance problem during benchmark profiling – once we have it :)

internal/types/array.go
```diff
@@ -111,3 +111,33 @@ func (a *Array) Append(values ...any) error {
...
+		if Compare(v, value) == Equal {
+			return true
+		}
+	}
+	return false
+}
+
+// ContainsAll checks if the Array contains all the given values of the Array b.
+// Currently, this algorithm is O(n^2) without any performance tuning.
+// Possible improvement: to sorts both arrays before comparing.
```
AlekSi, line 127: I think we should document that either inside the function or outside of it, but not in both places that we would need to keep in sync

internal/types/array_test.go
```diff
@@ -77,3 +77,78 @@ func TestArray(t *testing.T) {
...
+		"Int32Negative": {
+			array:    must.NotFail(NewArray(int32(42), int32(43), int32(45))),
+			filter:   int32(44),
+			expected: false,
+		},
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			assert.Equal(t, tc.expected, tc.array.Contains(tc.filter))
```
AlekSi, line 111: We typically use `testutil.AssertEqual` that handles NaNs

integration/query_array_test.go
rumyantseva: Well, I just followed the approach used for `$size`. What do you mean by shared data? Should I make data between `$size` and `$all` shared, or should I reuse some other data?

AlekSi: See `integration/shareddata` package and other tests that use it. `$size` test was made before we had shared arrays.

internal/types/array.go
rumyantseva: Moved everything outside of the function.

internal/types/array_test.go
rumyantseva: Fixed. I'd say the other function in this file - `TestArray` needs to be refactored to follow the same standard. Also, I'd separate tests for different `Array` ops, then one doesn't need to run tests for everything if they want to test or play with a particular operation... But it's not in scope if this task :)

integration/query_array_test.go
rumyantseva: Ok, I had a look. Those examples wouldn't fit. I need a few combinations of arrays where `$all` could and couldn't find values of various types - from one array, from a few arrays etc. The examples from shared data don't contain such cases. I don't want to extend shared data because it might affect the other cases where shared data is used, and those tests will need refactoring too.

Overall, shared data looks like an antipattern to me. I believe data needed for a test should be within the scope of that test. Why? Because of that exact reason - as soon as a new test wants to extend shared data to have more cases, the other tests will need to be refactored too to reuse those cases. 

Probably, it's fine to use individual cases from shared data (like we do with fixtures when writing tests), but using the whole data set doesn't look reasonable to me. If it's unclear, let's discuss.

internal/types/array.go
rumyantseva: I removed this condition to simplify the case.

integration/query_array_test.go
rumyantseva: Fixed!

seeforschauer (approved):
with `testutil` and `types` I've also met some complexity before, let's discuss it and do it in a separate PR. 
LGTM.

integration/query_array_test.go
AlekSi: Let's add some new data to `shareddata.Composite` and use `shareddata.Scalar`, `shareddata.Scalar` and some additional test-specific data

internal/handlers/common/filter.go
```diff
@@ -603,6 +610,30 @@ func filterFieldExprSize(fieldValue any, sizeValue any) (bool, error) {
 	return true, nil
 }
 
+// filterFieldExprSize handles {field: {$all: [value, another_value, ...]}} filter.
+// The main purpose of $all is to filter arrays.
+// In principle, it is possible to filter non-arrays: {field: {$all: [value]}},
+// but such statement is equivalent to {field: value}.
```
w84thesun, line 616: I suggest removing it as it doesn't add anything to the function description.

```suggestion
// It is possible to filter non-arrays: {field: {$all: [value]}}, but such statement is equivalent to {field: value}.
```

w84thesun (changes requested):

PR #687, opened 2022-06-02
Make log level configurable

### LINKED ISSUE #680: Make log level configurable
We should add a new flag `-log-level` that would allow us to set the logging level.

Possible values should be [`zapcore`](https://pkg.go.dev/go.uber.org/zap/zapcore#Level)'s `DebugLevel`, `InfoLevel`, `WarnLevel`, and `ErrorLevel`. The usage text should show them all just like we do for `-handler` flag.

Additionally, only debug level should enable zap's [development mode](https://pkg.go.dev/go.uber.org/zap#Config).

fenogentov:
Closes #680.


AlekSi (changes requested):
The part about development mode is not done. See `zap.NewDevelopmentConfig()` and related functions

internal/util/logging/logging.go
```diff
@@ -24,8 +24,15 @@ import (
 
 // Setup initializes logging with a given level.
 func Setup(level zapcore.Level) {
-	config := zap.NewDevelopmentConfig()
-	config.Level = zap.NewAtomicLevelAt(level)
+	config := zap.Config{
+		Level:            zap.NewAtomicLevelAt(level),
+		Development:      true,
```
AlekSi, line 29: > Additionally, only debug level should enable zap's [development mode](https://pkg.go.dev/go.uber.org/zap#Config).

I think the easiest way to do that is something like:
```go
var config zap.Config
if level <= zapcore.DebugLevel {
  config = ...
} else {
  config = ...
}
```

cmd/ferretdb/main.go
```diff
@@ -45,13 +46,19 @@ var (
...
 )
 
 func main() {
-	logging.Setup(zap.DebugLevel)
-	logger := zap.L()
 	flag.Parse()
 
+	level, err := zapcore.ParseLevel(*logLevel)
+	if err != nil {
+		level = zapcore.DebugLevel
```
AlekSi, line 57: We should fail in that case

AlekSi (changes requested):

AlekSi:
Oh, and tests fail

cmd/ferretdb/main.go
```diff
@@ -62,6 +64,7 @@ var (
 	handlerF = flag.String("handler", "pg", "<set in initFlags()>")
 
 	testConnTimeoutF = flag.Duration("test-conn-timeout", 0, "test: set connection timeout")
+	logLevel         = flag.String("log-level", "info", "set backend handler (debug/info/warn/error)")
```
AlekSi, line 67: Let's set both the default value and usage text in `initFlags` function and use zap's constants instead of our own hard-coded values

internal/util/logging/buffer_test.go
```diff
@@ -130,7 +130,7 @@ func TestCircularBuffer(t *testing.T) {
 		name := fmt.Sprintf("ZapHooks_%d", n)
 		tc := tc
 		t.Run(name, func(t *testing.T) {
-			logger.Info(tc.addMsg)
+			logger.Debug(tc.addMsg)
```
AlekSi, line 133: Why this change?

internal/util/logging/logging.go
```diff
@@ -24,8 +24,30 @@ import (
...
+		config = zap.Config{
+			Level:            zap.NewAtomicLevelAt(level),
+			Development:      true,
+			Encoding:         "console",
+			EncoderConfig:    zap.NewDevelopmentEncoderConfig(),
+			OutputPaths:      []string{"stderr"},
+			ErrorOutputPaths: []string{"stderr"},
+		}
+	} else {
+		config = zap.Config{
```
AlekSi, line 38: For now, let's use the same configuration as for debug above except `Development` field value. We will change it later

AlekSi (changes requested):

cmd/ferretdb/main.go
```diff
@@ -70,15 +73,20 @@ func initFlags() {
 	handlers := maps.Keys(registeredHandlers)
 	slices.Sort(handlers)
 	flag.Lookup("handler").Usage = "backend handler: " + strings.Join(handlers, ", ")
+	flag.Lookup("log-level").Usage = "sets logging levels"
```
AlekSi, line 76: The idea was that we could use code to set the help text and the default value, something that is not possible in a variable declaration. In pseudocode:

```go
var levels []string
for _, l := range []{zap.Debug, ...} {
  levels = append(levels, l.String())
}
flag.Lookup("log-level").Usage = "log level: " + strings.Join(levels, ", ")
```

AlekSi (approved):

PR #691, opened 2022-06-02
Support `$unset` field update operator

### LINKED ISSUE #629: Support `$unset` field update operator
Dot notation is not in scope.

seeforschauer:
Closes #629.

### DIFF
--- a/integration/update_test.go
+++ b/integration/update_test.go
@@ -593,6 +593,7 @@ func TestUpdateMany(t *testing.T) {
 		stat   *mongo.UpdateResult
 		res    bson.D
 		err    *mongo.WriteError
+		alt    string
 	}{
 		"SetSetOnInsert": {
 			filter: bson.D{{"_id", "test"}},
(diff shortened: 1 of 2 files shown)


integration/update_test.go
```diff
@@ -628,6 +629,46 @@ func TestUpdateMany(t *testing.T) {
 				Message: "Updating the path 'foo' would create a conflict at 'foo'",
 			},
 		},
+		"UnsetString": {
```
w84thesun, line 632: I think tests for ’$unset’ should be in ’TestUpdateUnsetOperator’, not in ’TestUpdateMany’.

internal/handlers/common/update.go
```diff
@@ -50,6 +50,20 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
+			case *types.Document:
+				if unsetDoc.Len() == 0 {
+					continue
+				}
+				for _, key := range unsetDoc.Keys() {
+					doc.Remove(key)
+				}
+				changed = true
+			default:
+				panic("Modifiers operate on fields but we found another type instead")
```
w84thesun, line 64: Why ’panic’? Maybe just return error?

w84thesun (changes requested):

integration/update_test.go
seeforschauer: > I think tests for `$unset` should be in `TestUpdateUnsetOperator`, not in `TestUpdateMany`.

why? 
we were about to create another task to restructure update tests, let's do it once in that PR?

internal/handlers/common/update.go
w84thesun: It that case switch could be replaced with type assertion.

seeforschauer: > It that case switch could be replaced with type assertion.

Good to note! Changed here and will change in the near PR. Thank you, @w84thesun!

integration/update_test.go
```diff
@@ -628,6 +629,46 @@ func TestUpdateMany(t *testing.T) {
...
+				UpsertedCount: 1,
+			},
+			res: bson.D{{"_id", "document-composite"}},
+		},
+		"UnsetEmptyArray": {
+			filter: bson.D{{"_id", "document-composite"}},
+			update: bson.D{{"$unset", bson.A{}}},
+			err: &mongo.WriteError{
+				Code: 9,
+				Message: "Modifiers operate on fields but we found type array instead. " +
```
rumyantseva, line 667: Why? I see that `$unset` can be an array.

rumyantseva (changes requested):
I have two questions:

- I see that dot notation is not in scope. Should we give "not implemented" for queries that use dot notation? Or the check is not in scope either?
- I commented a test case where `$unset` complains about being an array. But as far as I can see, it's a valid case, something like `$unset: ["a", "b", "c"]` is possible.

w84thesun (approved):

integration/update_test.go
seeforschauer: > Why? I see that `$unset` can be an array.

That error on an empty array gives MongoDB. To specify an array, dot notation should be used. Dot notation is not in scope.

seeforschauer: >  Should we give "not implemented" for queries that use dot notation? Or the check is not in scope either?

It should and it's in the other PR by @w84thesun 
https://github.com/FerretDB/FerretDB/pull/660/files#diff-d5aa988135248001431214e8620db5b7b3b57b75f6b692216bc820a1cddb68a5R198

> $unset: ["a", "b", "c"] 


I've already had a test for an empty array `UnsetEmptyArray`. 
This is same check, the error is 
`Modifiers operate on fields but we found type array instead. For example: {$mod: {<field>: ...}} not {$unset: [ \"a\", \"b\", \"c\" ]}`

rumyantseva (approved):
LGTM

AlekSi (approved):

PR #695, opened 2022-06-03
Support `$mul` field update operator

### LINKED ISSUE #625: Support `$mul` field update operator
https://www.mongodb.com/docs/manual/reference/operator/update/mul/

DoD:

- [ ] Compat tests for various data types (numeric, non-numeric, int32 * int32 = int64 (e.g. for Tigris), infinity, max double value * max double value)
- [ ] diff tests if needed (if we can't support something the same way)
- [ ] Update docs about supported operators

fenogentov:
Closes #625.


rumyantseva:
Solved in #1760.

PR #701, opened 2022-06-07
`$currentDate` Timestamp fix `DateTime` seconds and milliseconds bug

seeforschauer:
Refs #662.

`types.Timestamp` has `bson` format.  
In `bson` time is stored as seconds and nonce. `First 4 bytes are an increment, second 4 are a timestamp`. The error was here that was not seconds but milliseconds.


rumyantseva:
This one is hard to review without explanation. @seeforschauer would you mind describing what the bug was and what is the fix?

seeforschauer:
> This one is hard to review without explanation. @seeforschauer would you mind describing what the bug was and what is the fix?

`timestamp` has `bson` format.
In `bson` time is stored as seconds and nonce `First 4 bytes are an increment, second 4 are a timestamp.` The error was here that was not seconds but milliseconds.

AlekSi (commented):
> In bson time is stored as seconds and nonce First 4 bytes are an increment, second 4 are a timestamp.

Please current me if I'm wrong, but it does not look like we actually test that. Our test tests that if we assume that timestamps store seconds, then we could get them back. If we assume that timestamps store milliseconds, the same test would pass – simply because we use our own code for both writing and reading.

The simplest way to test our assumption could be:
1. store the current timestamp with `$currentDate` operator;
2. read it, check that it is close to the current time;
3. write a new timestamp value with the same time;
4. read it back, and check that it is still close to the current time.

Additionally, I left a few comments on #662. Let's try to work on them.

seeforschauer:
> The simplest way to test our assumption could be:
> 
> 1. store the current timestamp with `$currentDate` operator;
> 2. read it, check that it is close to the current time;
> 3. write a new timestamp value with the same time;
> 4. read it back, and check that it is still close to the current time.

Added this test.

integration/update_test.go
```diff
@@ -466,6 +467,98 @@ func TestUpdateSet(t *testing.T) {
...
+	// read it, check that it is close to the current time;
+	// write a new timestamp value with the same time;
+	// read it back, and check that it is still close to the current time.
+	t.Run("currentDateReadBack", func(t *testing.T) {
+		maxDifference := time.Duration(2 * time.Second)
+		now := time.Now()
+		nowTimestamp := primitive.Timestamp{T: uint32(now.Unix()), I: uint32(0)}
+		id := "string-empty"
+
+		tc := struct {
```
AlekSi, line 483: Let's not use a type for a single value

integration/update_test.go
```diff
@@ -466,6 +467,98 @@ func TestUpdateSet(t *testing.T) {
...
+		require.Equal(t, tc.stat, res)
+
+		var actualB bson.D
+		err = collection.FindOne(ctx, bson.D{{"_id", tc.id}}).Decode(&actualB)
+		require.NoError(t, err)
+
+		expected := ConvertDocument(t, tc.result)
+		actualX := ConvertDocument(t, actualB)
+
+		for _, path := range tc.paths {
```
AlekSi, line 515: That's a single value, no need to loop over it

integration/update_test.go
```diff
@@ -466,6 +467,98 @@ func TestUpdateSet(t *testing.T) {
 	}
 }
 
+func TestUpdateTimestamp(t *testing.T) {
+	t.Parallel()
+
+	// store the current timestamp with $currentDate operator;
+	// read it, check that it is close to the current time;
+	// write a new timestamp value with the same time;
+	// read it back, and check that it is still close to the current time.
```
AlekSi, line 476: Let's move those comments to individual steps

integration/update_test.go
```diff
@@ -466,6 +467,98 @@ func TestUpdateSet(t *testing.T) {
...
+
+	// store bson.Timestamp
+	// read it back, and check if values equal
+	t.Run("timestamp", func(t *testing.T) {
+		t.Parallel()
+
+		ctx, collection := setup(t)
+
+		now := time.Now()
+		nowTimestamp := primitive.Timestamp{T: uint32(now.Unix()), I: uint32(0)}
```
AlekSi, line 542: I think this test does not test anything that is not covered by the previous test. Here we write some value and later get it back. It is not important for that test that this value is a Unix timestamp

integration/update_test.go
```diff
@@ -466,6 +467,98 @@ func TestUpdateSet(t *testing.T) {
...
+		var actual bson.D
+		err = collection.FindOne(ctx, bson.D{{"_id", "timestamp"}}).Decode(&actual)
+		require.NoError(t, err)
+
+		AssertEqualDocuments(t, expected, actual)
+		doc := ConvertDocument(t, actual)
+		v, ok := must.NotFail(doc.Get("value")).(types.Timestamp)
+		assert.True(t, ok)
+
+		// nanoseconds are not stored by timestamp
```
AlekSi, line 556: That might be or might not be true, but we don't test it with this test. We have to truncate to second because `now.Unix()` above truncates to second

seeforschauer: > That might be or might not be true, but we don't test it with this test. We have to truncate to second because `now.Unix()` above truncates to second

removed that test.

integration/update_test.go
seeforschauer: > I think this test does not test anything that is not covered by the previous test. Here we write some value and later get it back. It is not important for that test that this value is a Unix timestamp

removed.

integration/update_test.go
seeforschauer: > Let's move those comments to individual steps

done.

integration/update_test.go
seeforschauer: > That's a single value, no need to loop over it

done.

integration/update_test.go
seeforschauer: > Let's not use a type for a single value

done.

integration/update_test.go
```diff
@@ -466,6 +466,58 @@ func TestUpdateSet(t *testing.T) {
...
+		update := bson.D{{"$currentDate", bson.D{{"value", bson.D{{"$type", "timestamp"}}}}}}
+		stat := &mongo.UpdateResult{
+			MatchedCount:  1,
+			ModifiedCount: 1,
+			UpsertedCount: 0,
+		}
+		path := types.NewPathFromString("value")
+		result := bson.D{{"_id", id}, {"value", nowTimestamp}}
+
+		ctx, collection := setup(t, shareddata.Scalars, shareddata.Composites)
```
w84thesun, line 488: I think it's better to move this line to the top of the test as we do with all other tests.

integration/update_test.go
```diff
@@ -466,6 +466,58 @@ func TestUpdateSet(t *testing.T) {
...
 }
 
+func TestUpdateTimestamp(t *testing.T) {
+	t.Parallel()
+
+	// store the current timestamp with $currentDate operator;
+	t.Run("currentDateReadBack", func(t *testing.T) {
+		maxDifference := time.Duration(2 * time.Second)
+		now := time.Now()
+		nowTimestamp := primitive.Timestamp{T: uint32(now.Unix()), I: uint32(0)}
```
w84thesun, line 476: That's the only usage of `now` variable.
```suggestion
		nowTimestamp := primitive.Timestamp{T: uint32(time.Now().Unix()), I: uint32(0)}
```

integration/update_test.go
```diff
@@ -466,6 +466,58 @@ func TestUpdateSet(t *testing.T) {
...
+		id := "string-empty"
+
+		update := bson.D{{"$currentDate", bson.D{{"value", bson.D{{"$type", "timestamp"}}}}}}
```
w84thesun, line 479: Let's move this line next to the actual update operation right before its usage.

integration/update_test.go
```diff
@@ -466,6 +466,58 @@ func TestUpdateSet(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Equal(t, stat, res)
+
+		// read it, check that it is close to the current time;
+		var actualB bson.D
+		err = collection.FindOne(ctx, bson.D{{"_id", id}}).Decode(&actualB)
+		require.NoError(t, err)
+
+		expected := ConvertDocument(t, result)
+		actualX := ConvertDocument(t, actualB)
```
w84thesun, line 500: Those names `actualX`, and `actualB` is really confusing. Maybe there is a way to make them more meaningful?

integration/update_test.go
```diff
@@ -466,6 +466,58 @@ func TestUpdateSet(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Equal(t, stat, res)
+
+		// read it back, and check that it is still close to the current time.
+		err = collection.FindOne(ctx, bson.D{{"_id", id}}).Decode(&actualB)
+		require.NoError(t, err)
+
+		AssertEqualDocuments(t, expectedB, actualB)
+		actualY := ConvertDocument(t, actualB)
+		testutil.CompareAndSetByPathTime(t, actualY, actualX, maxDifference, types.NewPathFromString("value"))
```
w84thesun, line 517:
```suggestion
		testutil.CompareAndSetByPathTime(t, actualY, actualX, maxDifference, path)
```

integration/update_test.go
```diff
@@ -466,6 +466,58 @@ func TestUpdateSet(t *testing.T) {
...
+		require.NoError(t, err)
+
+		expected := ConvertDocument(t, result)
+		actualX := ConvertDocument(t, actualB)
+
+		testutil.CompareAndSetByPathTime(t, expected, actualX, maxDifference, path)
+
+		// write a new timestamp value with the same time;
+		updateB := bson.D{{"$set", bson.D{{"value", nowTimestamp}}}}
+		expectedB := bson.D{{"_id", id}, {"value", nowTimestamp}}
```
w84thesun, line 506: `expectedB` is equal to `result` defined above.

w84thesun (changes requested):

integration/update_test.go
seeforschauer: > `expectedB` is equal to `result` defined above.

It will be confusing if I'd be comparing to the variable named as `result` and threaten it as expected.

integration/update_test.go
seeforschauer: > Let's move this line next to the actual update operation right before its usage.

It seems to me that it's a code polishing, which is discussed the previous week. But ok, I'll fix it now...

w84thesun (approved):

PR #704, opened 2022-06-07
Re-enable `TestStatisticsCommands` tests

### LINKED ISSUE #536: Fix fluky statistics tests
Statistics tests fail because of different numbers in responses data.
We should find out why and fix `TestStatisticsCommands`

seeforschauer:
Closes #536.


AlekSi (commented):
I think it will be easier to have on PR for all tests instead of several PRs changing the same code

integration/commands_administration_test.go
```diff
@@ -299,126 +298,186 @@ func TestCommandsAdministrationGetParameter(t *testing.T) {
...
-			err := collection.Database().RunCommand(ctx, tc.command).Decode(&actual)
-			require.NoError(t, err)
+	doc := ConvertDocument(t, actual)
 
-			AssertEqualDocuments(t, tc.response, actual)
-		})
-	}
+	assert.Equal(t, float64(1), must.NotFail(doc.Get("ok")))
+	assert.Regexp(t, `^5\.0\.`, must.NotFail(doc.Get("version")))
+	assert.NotEmpty(t, must.NotFail(doc.Get("gitVersion")))
```
rumyantseva, line 314: Do cases like `gitversion` or `GITversion` need to be checked as well? Or this one is case-sensitive, and only `gitVersion` is allowed?

integration/commands_administration_test.go
```diff
@@ -299,126 +298,186 @@ func TestCommandsAdministrationGetParameter(t *testing.T) {
...
+	ctx, collection := setup(t)
+
+	var actual bson.D
+	command := bson.D{{"collStats", collection.Name()}}
+	err := collection.Database().RunCommand(ctx, command).Decode(&actual)
+	require.NoError(t, err)
+
+	doc := ConvertDocument(t, actual)
+	assert.Equal(t, float64(1), must.NotFail(doc.Get("ok")))
+	assert.Equal(t, collection.Database().Name()+"."+collection.Name(), must.NotFail(doc.Get("ns")))
```
rumyantseva, line 345: So, in the tests in this PR, we use empty tables, right? I see we compare almost all the values with zeros... Does it make sense to have cases when things like `storageSize` should be greater than zero?

integration/commands_administration_test.go
```diff
@@ -299,126 +298,186 @@ func TestCommandsAdministrationGetParameter(t *testing.T) {
...
+	assert.LessOrEqual(t, int32(0), must.NotFail(doc.Get("count")))
+	assert.LessOrEqual(t, int32(0), must.NotFail(doc.Get("size")))
```
rumyantseva, line 347: Why do all those things use `LessOrEqual`? Those values could be less than 0?

internal/handlers/pg/msg_dbstats.go
```diff
@@ -47,6 +47,11 @@ func (h *Handler) MsgDBStats(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 		return nil, lazyerrors.Error(err)
 	}
 
+	var avgObjSize float64
+	if stats.CountRows > 0 {
```
rumyantseva, line 51: Nice catch!

rumyantseva (commented):
Asked questions about some test cases. 
From the issue description, it's not fully clear what is DoD of this task.
I see currently we check stats for empty schemas/tables. I wonder if we need tests for non-empty cases...

integration/commands_administration_test.go
seeforschauer: > Do cases like `gitversion` or `GITversion` need to be checked as well? 

Not necessary as MongoDB has no of these fields.

> Or this one is case-sensitive, and only `gitVersion` is allowed?

It is case sensitive, the test checks that some of MongoDB status is implemented by FerretDB. 

Didn't understand your question: are you about to suggest checking whether FerretDB has no authentic document fields?

integration/commands_administration_test.go
seeforschauer: > So, in the tests in this PR, we use empty tables, right?

right.

> I see we compare almost all the values with zeros... Does it make sense to have cases when things like `storageSize` should be greater than zero?

Do you mean creating a table, adding data there, and testing whether storage size in the response increases?

integration/commands_administration_test.go
seeforschauer: > Why do all those things use `LessOrEqual`? Those values could be less than 0?

It's about `count` value to be bigger or equal to 0.

integration/commands_administration_test.go
rumyantseva: @seeforschauer well, I remember there were some examples where both `pampam` and `pamPam` were allowed, so I wanted to check about the ones in this file. But if `gitVersion`, `sysInfo` and other mixed case fields are case-sensitive, we are safe here.

integration/commands_administration_test.go
rumyantseva: @seeforschauer yes, maybe something like this. I'm not sure if it's reasonable and doable, but it seems to be important to check not only for the "init" values :)

integration/commands_administration_test.go
seeforschauer: > But if `gitVersion`, `sysInfo` and other mixed case fields are case-sensitive, we are safe here.

they are, as `types.Document` array keys and map.

integration/commands_administration_test.go
seeforschauer: >  I'm not sure if it's reasonable and doable, but it seems to be important to check not only for the "init" values :)

It's reasonable and doable. 
Here in the task, I need just to enable the previously working tests. 

I suggest creating a separate PR for adding that another test since I can find something and change the code lot  ;-)

integration/commands_administration_test.go
rumyantseva: @seeforschauer `LessOrEqual` is about `count` value to be bigger or equal to 0? Sound quite hard to understand (why not `GreaterOrEqual`?) And also, why can't we just use `Equal` if we expect an exact 0 here?

seeforschauer: > @seeforschauer `LessOrEqual` is about `count` value to be bigger or equal to 0? 

LessOrEqual asserts that the first element is less than or equal to the second

> Sound quite hard to understand (why not `GreaterOrEqual`?) 

probably. But if you put the variables in order, then it's more easier: `0 <= value`. Value has to be bigger.

> And also, why can't we just use `Equal` if we expect an exact 0 here?
I consider adding `LessOrEqual` everywhere as I and @AlekSi agreed on the call that the later implementation of the entire command will be later on for each separate command. 
So in this PR, I'm about to add some checks only, for the variables where I can - add some more checks. For example If I expect an array, check for an array, If I expect IP, check IP regexp. If I expect a `5.0` Mongo version and I can skip minor version, I check for 5.0. 


These tests are fluky.  Let's discuss it as we don't want to do fluky tests lightly.

seeforschauer: > These tests are fluky. Let's discuss it as we don't want to do fluky tests lightly.

So that's my suggestion - to add `LessOrEqual` where it's necessary.

w84thesun (approved):

internal/util/testutil/path.go
```diff
@@ -75,14 +76,21 @@ func SetByPath[T types.CompositeTypeInterface](tb testing.TB, comp T, value any,
...
+}
+
 // CompareAndSetByPathNum asserts that two values with the same path in two objects (documents or arrays)
 // are within a given numerical delta, then updates the expected object with the actual value.
 func CompareAndSetByPathNum[T types.CompositeTypeInterface](tb testing.TB, expected, actual T, delta float64, path types.Path) {
 	tb.Helper()
 
 	expectedV := GetByPath(tb, expected, path)
 	actualV := GetByPath(tb, actual, path)
-	assert.IsType(tb, expectedV, actualV)
```
seeforschauer, line 85: I consider as not necessary.

integration/commands_administration_test.go
```diff
@@ -602,126 +605,221 @@ func TestCommandsAdministrationGetParameter(t *testing.T) {
...
+
+	assert.Equal(t, float64(1), must.NotFail(doc.Get("ok")))
+	assert.Equal(t, collection.Database().Name(), must.NotFail(doc.Get("db")))
+	assert.Equal(t, int32(1), must.NotFail(doc.Get("collections")))
+	assert.Equal(t, int32(0), must.NotFail(doc.Get("views")))
+	assert.Equal(t, int32(0), must.NotFail(doc.Get("objects")))
+	assert.Equal(t, float64(0), must.NotFail(doc.Get("avgObjSize")))
+	assert.Equal(t, float64(0), must.NotFail(doc.Get("dataSize")))
+	assert.Equal(t, float64(1000), must.NotFail(doc.Get("scaleFactor")))
+	testutil.AssertInThreshold(t, doc, "indexes", float64(1), 1)
```
seeforschauer, line 760: On empty collection there is an index for MongoDB, which Postres has not.

integration/commands_administration_test.go
```diff
@@ -602,126 +605,221 @@ func TestCommandsAdministrationGetParameter(t *testing.T) {
...
+
+	doc := ConvertDocument(t, actual)
+
+	assert.Equal(t, float64(1), must.NotFail(doc.Get("ok")))
+	assert.Equal(t, collection.Database().Name(), must.NotFail(doc.Get("db")))
+	assert.Equal(t, int32(1), must.NotFail(doc.Get("collections")))
+	assert.Equal(t, int32(0), must.NotFail(doc.Get("views")))
+	assert.Equal(t, int32(0), must.NotFail(doc.Get("objects")))
+	assert.Equal(t, float64(0), must.NotFail(doc.Get("avgObjSize")))
+	assert.Equal(t, float64(0), must.NotFail(doc.Get("dataSize")))
```
seeforschauer, line 758: Now I see I can add data and check it has changed, in progress.

integration/commands_administration_test.go
seeforschauer: > I guess then we need a separate issue too? I'm ok with it.

On empty tables - added some tests for tables with data.

Also created #727 for the error `numObjects`

w84thesun (dismissed):

rumyantseva (dismissed):
LGTM. As agreed in person, let's keep it like this, and eventually, we can improve it if bugs appear.

w84thesun (dismissed):

w84thesun (approved):

rumyantseva (approved):
LGTM

Issue #364, opened 2022-03-17
Explicitly disallow duplicate keys in data documents


AlekSi:
BSON documents may have more than one field with the same name. However, that doesn't seem to be the feature that anyone uses. 

That's not true for command documents.

* https://www.mongodb.com/docs/manual/core/document/#field-names
* https://www.mongodb.com/docs/manual/reference/limits/#mongodb-does-not-support-duplicate-field-names

This is also not true for update documents. See #666.

DoD:
- [x] Let's explicitly disallow that in `types.Document` API, add tests and document that.
- [x] `ValidateData` for `types.Document` checks for duplicates and disallow them (check that d.keys doesn't contain duplicates).

rumyantseva:
Implementation explanation:

As agreed with the team, quite a big refactoring was made. Now `types.Document`'s internal representation consists of a slice of fields where each field has a `string` field and `any` value.
That approach allows having duplicated keys in documents in principle.

However, `Document.ValidateData` checks that there are no duplicates.

Minor refactoring: `Document.Set`'s signature doesn't return `error `anymore, as, in fact, the implementation always returns `nil`. With that, there is no need to make the code more complicated (with all the checks for errors).

PR #707, opened 2022-06-07
Support $elemMatch array query operator

### LINKED ISSUE #699: Support `$elemMatch` array query operator

w84thesun:
Closes #699.


w84thesun:
I would really appreciate any test case suggestions. 😄

seeforschauer:
> I would really appreciate any test case suggestions. smile

* What if there are two `$elemMatch` inside the filter?
* what if there are two similar operates inside one `$elemMatch`?
* Please see error `ErrElemMatchExclusionInInclusion` and ErrElemMatchInclusionInExclusion in #551 these errors also applicable to `$elemMatch` operator?
* `filterFieldExpr` does almost all stuff, And what if the expression field contains `$and` `$or` `$nor` how does it behave? 
* please add `$elemMatch cannot contain $where expression`

seeforschauer (changes requested):
LGTM and some more tests are required.

internal/handlers/common/filter.go
```diff
@@ -495,6 +495,13 @@ func filterFieldExpr(doc *types.Document, filterKey string, expr *types.Document
 				return false, err
 			}
 
+		case "$elemMatch":
```
rumyantseva, line 498: Should `$elemMatch` be before `$size` in this list?

integration/query_array_test.go
```diff
@@ -212,6 +212,90 @@ func TestQueryArrayDotNotation(t *testing.T) {
...
+			assert.Equal(t, tc.expectedIDs, CollectIDs(t, actual))
+		})
+	}
+}
+
+func TestQueryElemMatchOperator(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setup(t, shareddata.Scalars, shareddata.Composites)
+
+	for name, tc := range map[string]struct {
```
rumyantseva, line 233: All the test cases have `gt` or `lt`. Can we have one that is a mix of conditions, something like `{ value: { $elemMatch: { field1: "a", field2: { $gt: 42 } } } }`?

Also, it seems that `$ne` might give interesting cases, and we should add it to the tests.

internal/handlers/common/filter.go
```diff
@@ -1084,3 +1091,20 @@ func filterFieldValueByTypeCode(fieldValue any, code typeCode) (bool, error) {
 
 	return true, nil
 }
+
+// filterFieldExprElemMatch handles {field: {$elemMatch: value}}.
+// Returns false if doc value is not an array.
+func filterFieldExprElemMatch(doc *types.Document, filterKey string, exprValue any) (bool, error) {
```
rumyantseva, line 1097: Should we have anything special about `$text` and `$where` not being supported in `$elemMatch`?

rumyantseva (changes requested):
A few suggestions of what could be added.

internal/handlers/common/filter.go
w84thesun: Moved it.

internal/handlers/common/filter.go
w84thesun: Added correct response for `$text` and `$where`.

w84thesun:
I suggest moving `$ne`, `$not`, and `$and` out of scope as it seems that their implementation would drastically increase PR size.

integration/query_array_test.go
w84thesun: I've added a similar case to that one you suggest → `GtField`.
Also please see https://github.com/FerretDB/FerretDB/pull/707#issuecomment-1152037874

w84thesun:
> * What if there are two `$elemMatch` inside the filter?
> * What if there are two similar operators inside one `$elemMatch`?
> * Please see error `ErrElemMatchExclusionInInclusion` and `ErrElemMatchInclusionInExclusion` in [Support `$elemMatch`  projection basic comparison query operators #551](https://github.com/FerretDB/FerretDB/pull/551). Are these errors for `$elemMatch` query applicable to `$elemMatch` operator?
> * `filterFieldExpr` does almost all stuff. And what if the expression field contains
>   
>   * `$and`
>   * `$or`
>   * `$nor`
>   * `$not` - the result of the query must be similar to the result without `$elemMatch`
>   * `$ne` - the result of the query must be similar to the result without `$elemMatch`
>     How does it behave?
> * `$elemMatch` cannot contain `$where`, could you please add a test for it?
> * `$elemMatch` cannot contain `$text`, could you please add a test for it?
> * WDYT of adding the test where an operator and a scalar field expression are present?  `bson.D{{"value", bson.D{{"$elemMatch", bson.D{{"$gt", int32(0)}, {"somefield", int32(42)}}}}}}`

Thanks for the suggestions!

Two `$elemMatch` or duplicate `$gt` are out of scope → #364 
I've added similar query operators → `GtLt` case
Looks like errors are not applicable.
For `$and` etc. please see https://github.com/FerretDB/FerretDB/pull/707#issuecomment-1152037874
Errors for `$where` and `$text` where added.
Case `GtField` added.

seeforschauer:
> I suggest moving `$ne`, `$not`, `$and`, `$or` and `$nor` out of scope as it seems that their implementation would drastically increase PR size.

@w84thesun would you please add an issue link or create on if it doesn't exists?

internal/handlers/common/filter.go
```diff
@@ -1084,3 +1093,35 @@ func filterFieldValueByTypeCode(fieldValue any, code typeCode) (bool, error) {
...
+	expr, ok := exprValue.(*types.Document)
+	if !ok {
+		return false, NewErrorMsg(ErrBadValue, "$elemMatch needs an Object")
+	}
+
+	for _, key := range expr.Keys() {
+		if slices.Contains([]string{"$text", "$where"}, key) {
+			return false, NewErrorMsg(ErrBadValue, fmt.Sprintf("%s can only be applied to the top-level document", key))
+		}
+
```
seeforschauer, line 1116: Would you please add a TODO with the issue number here? Thank you.

w84thesun: Added

seeforschauer (approved):

rumyantseva (approved):
LGTM. I agree with the idea of limiting the scope of this task and creating a separate issue for `$ne, $not, $and, $or, $nor`.

AlekSi (approved):

PR #698, opened 2022-06-05
Handle `find`, `count` and `delete` correctly when collection doesn't exist

rumyantseva:
## Description

This PR doesn't solve #651 but covers a part of it.

Issue #651 is quite a big one, so I suggest dividing it into two PRs.

**In the scope of this PR:**
- `find` should return an empty response, with no errors.
- `count` should return 0.
- `delete` shouldn't return any errors.

**In a separate PR (to be done):**
- `findAndModify` (this method is tricky as it needs to create a doc that represents data from both `query` and `update` docs).
- `update` (same reason).

**Other fixes in the current PR:**
I made an ugly bug in #650, and in fact, proxy responses haven't been logged since then. I've just fixed that bug.

## Testing

- I added tests for `count`, `find`, and `delete` to cover cases when collection doesn't exist.
- In addition, I manually checked various cases to compare responses between "pg-based implementation" and "proxy" to make sure that the logic is the same.

## Checklist

- [x] Code is regenerated if needed (`task gen`).
- [x] Code is formatted (`task fmt`).
- [x] Docs are formatted (`task docs-fmt`).
- [x] Tests passed (`task test`).
- [x] Linters passed (`task lint`).
- [x] PR assignee is set.
- [x] PR labels are set.
- [x] Project and iternation are set.
- [ ] Critical codecov complaints are considered.
- [x] PR title and body are in accordance with the Conform PR GitHub Action.
- [x] Other GitHub Actions passed.

Issue #651, opened 2022-05-27
Calling a non-existing collection causes internal error


rumyantseva:
When querying a non-existing collection, we should do the same as MongoDB: return an empty result and do not create the collection/table. I think we used to do that, but now we don't.

We should also do the same thing as MongoDB for updates, deletes, etc.

* [x] Add tests for
  * [x] `count` in `query_test.go`
  * [x] `delete` in a new `delete_test.go`
  * [x] `find` in `query_test.go`
  * [x] `findAndModify` in `findandmodify_test.go`
  * [x] `update` in `update_test.go`
* [x] Fix code.

That should be done for both pg and tigris handlers. Use compatibility tests where possible.

The original report follows.

---

### How to reproduce

Version: [18daf68 on Fri May 27 20:51:37 2022 +0300](https://github.com/FerretDB/FerretDB/tree/18daf6891aeccbd507605402c01484299aaa2e2a)

1. Connect to the `test` database provided by `task env-up`.
2. Run the query `db.pampam.find();` (`pampam` collection doesn't exist in the `test` db)

### Actual result

The server returns an internal error and closes the connection.
Here is what I see in the logs:

```
2022-05-28T00:31:35.243+0200	DEBUG	// 127.0.0.1:59314 -> 127.0.0.1:27017 .// 127.0.0.1:59314 -> 127.0.0.1:27017 	clientconn/conn.go:166	Request header: length:    70, id: 3736, response_to:    0, opcode: OP_MSG
2022-05-28T00:31:35.243+0200	DEBUG	// 127.0.0.1:59314 -> 127.0.0.1:27017 .// 127.0.0.1:59314 -> 127.0.0.1:27017 	clientconn/conn.go:167	Request message:
{
  "Checksum": 0,
  "FlagBits": 0,
  "Sections": [
    {
      "Document": {
        "$k": [
          "find",
          "filter",
          "$db"
        ],
        "find": "pampam",
        "filter": {
          "$k": []
        },
        "$db": "test"
      },
      "Kind": 0
    }
  ]
}



2022-05-28T00:31:35.248+0200	ERROR	pg.Pool	v4@v4.16.1/conn.go:354	Query	{"err": "ERROR: relation \"test.pampam\" does not exist (SQLSTATE 42P01)", "sql": "SELECT _jsonb FROM \"test\".\"pampam\"", "args": [], "pid": 140}
github.com/jackc/pgx/v4.(*Conn).log
[...]

PR #710, opened 2022-06-07
Handle `find`, `count` and `delete` correctly when collection doesn't exist

rumyantseva:
**Important!** I had to recreate the PR as I had messed up everything in my previous one. Sorry for that!
 The previous one was #698.

## Description

This PR refs #651 (covers a part of it)

Issue #651 is quite a big one, so I suggest dividing it into two PRs.

**In the scope of this PR:**
- `find` should return an empty response, with no errors.
- `count` should return 0.
- `delete` shouldn't return any errors.

**In a separate issue and PR:**
- `findAndModify` (this method is tricky as it needs to create a doc that represents data from both `query` and `update` docs).
- `update` (same reason).

**Other fixes in the current PR:**
Refactored `TestCreateDrop` to use subtests

## Testing

- I added tests for `count`, `find`, and `delete` to cover cases when collection doesn't exist.
- In addition, I manually checked various cases to compare responses between "pg-based implementation" and "proxy" to make sure that the logic is the same.

## Checklist

- [x] Code is regenerated if needed (`task gen`).
- [x] Code is formatted (`task fmt`).
- [x] Docs are formatted (`task docs-fmt`).
- [x] Tests passed (`task test`).
- [x] Linters passed (`task lint`).
- [x] PR assignee is set.
- [x] PR labels are set.
- [x] Project and iternation are set.
- [x] Critical codecov complaints are considered.
- [x] PR title and body are in accordance with the Conform PR GitHub Action.
- [x] Other GitHub Actions passed.


internal/handlers/pg/pgdb/pool_test.go
```diff
@@ -187,3 +187,82 @@ func TestConcurrentCreate(t *testing.T) {
 		}
 	}
 }
+
+func TestPool_TableExists(t *testing.T) {
```
AlekSi, line 191: We don't use underscores in test names. We also don't create tests for each function if several functions could be tested together. I think this test the and the following one could be combined together with several subtests

internal/handlers/pg/pgdb/pool_test.go
```diff
@@ -187,3 +187,82 @@ func TestConcurrentCreate(t *testing.T) {
...
+	pool := testutil.Pool(ctx, t, nil, zaptest.NewLogger(t))
+
+	schemaName := testutil.SchemaName(t)
+	tableName := testutil.TableName(t)
+
+	t.Cleanup(func() {
+		pool.DropSchema(ctx, schemaName)
+	})
+
+	// Case 1: schema does not exist, table does not exist
```
AlekSi, line 204: We use subtests for that

AlekSi (changes requested):

seeforschauer (commented):
Would you please add some comments about testing naming and grouping? Thank you.

internal/handlers/pg/pgdb/pool_test.go
rumyantseva: Fixed

internal/handlers/pg/pgdb/pool_test.go
rumyantseva: Fixed

internal/handlers/pg/pgdb/pool_test.go
```diff
@@ -93,39 +93,107 @@ func TestCreateDrop(t *testing.T) {
...
+		err := pool.CreateSchema(ctx, schemaName)
+		require.NoError(t, err)
 
-	err = pool.CreateTable(ctx, schemaName, tableName)
-	require.Equal(t, pgdb.ErrAlreadyExist, err)
+		// Schema exists ->
+		// - schema creation is not possible
+		// - table drop is not possible
+		// - table creation is possible
+		// - schema drop is possible (only once)
```
seeforschauer, line 142: :+1:

internal/handlers/pg/pgdb/pool_test.go
```diff
@@ -93,39 +93,107 @@ func TestCreateDrop(t *testing.T) {
...
+
+		t.Cleanup(func() {
+			pool.DropSchema(ctx, schemaName)
+		})
+
+		// Schema does not exist ->
+		// - table drop is not possible
+		// - schema drop is not possible
+		// - table creation is not possible
+		// - shema creation is possible
```
seeforschauer, line 110: :+1:

seeforschauer (dismissed):

w84thesun (dismissed):

AlekSi (approved):
:shipit:

PR #713, opened 2022-06-08
Use separate encodings for ObjectID and binary

AlekSi:


rumyantseva (approved):
Ok, as far as I understood it, the changes make sense.

PR #718, opened 2022-06-08
Always install Go and skip test cache

AlekSi:
Explicitly install Go on CI before invoking any actions. Our actions will soon require Go 1.18, but the default version of Go on GitHub-hosted runners is 1.17. The installation step is still fast as 1.18 is also present, just not enabled by default, and `actions/setup-go@v3` is smart enough to use it.

Additionally, we skip the Go test cache entirely as it is not smart enough to notice changes in the environment. Our tests are fast, so that's not a problem. We might re-enable cache for local runs later, but we will not use it on CI.


seeforschauer (approved):
LGTM except Conform PR didn't passed.

w84thesun (approved):

w84thesun (approved):

rumyantseva (approved):
LGTM

Issue #527, opened 2022-04-26
Make linter to enforce our preferred types order in type switch


AlekSi:
In many places in our code, we have type switches that check all or some of BSON types:
```go
switch v := v.(type) {
case *types.Document:
  // ...
case *types.Array:
  // ...
case float64:
  // ...
```

We have a preferred types order that is a bit strange but follows the order in the BSON spec: https://pkg.go.dev/github.com/FerretDB/FerretDB/internal/types#hdr-Mapping

Sometimes, we forget about that order. We should make a linter that enforces it.

Can it be done with https://github.com/quasilyte/go-ruleguard?

PR #654, opened 2022-05-30
 Make linter to enforce our preferred types order in type switch

fenogentov:
Closes #527.

PR #722, opened 2022-06-08
Fix types order

### LINKED ISSUE #714: Enforce our preferred types order in type switch
Implement before including linter #654.

fenogentov:
Refs #527.


internal/handlers/pg/msg_getlog.go
```diff
@@ -85,6 +85,5 @@ func (h *Handler) MsgGetLog(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 	if err != nil {
 		return nil, lazyerrors.Error(err)
 	}
-
```
AlekSi, line 88: All other commands have an empty line there, let's keep it

internal/handlers/common/projection.go
```diff
@@ -142,7 +161,7 @@ func projectDocument(inclusion bool, doc *types.Document, projection *types.Docu
 				return err
 			}
 
-		case float64, int32, int64: // field: number
+		case float64: // field: number
```
AlekSi, line 164: Can we handle that case in our linter? I think if several types are in present in one `case`, they should be the in the right order, and the first type should be in the right place, but not second and following types

AlekSi (changes requested):

internal/handlers/common/projection.go
fenogentov: linter has been submitted, now such cases are valid

internal/handlers/pg/msg_getlog.go
fenogentov: fix

fenogentov:
code refactored #654.

PR #724, opened 2022-06-09
Support `$all` array query operator

### LINKED ISSUE #657: Support `$all` array query operator

rumyantseva:
## Description

This PR closes #657.

The implementation is the same as for the other filter operators. We validate if the doc matches the conditions set in `$all`.

## Testing

Two types of tests are added:

- Integration tests for `$all` operator
- Compat tests for `$all` operator

_If you see that some test cases might be missing, please let me know._

## Readiness checklist


internal/types/array.go
```diff
@@ -15,11 +15,17 @@
...
+	"errors"
 	"fmt"
+	"math"
+	"reflect"
 
 	"github.com/FerretDB/FerretDB/internal/util/must"
 )
 
+// ErrNaNIsNotImplemented should be used in operations where NaN support is not implemented.
+var ErrNaNIsNotImplemented = errors.New("NaN is not implemented")
```
rumyantseva, line 27: Please speak up if you have a better idea!

internal/handlers/common/filter.go
```diff
@@ -250,6 +250,25 @@ func filterFieldExpr(doc *types.Document, filterKey string, expr *types.Document
 			if _, ok := exprValue.(types.NullType); ok {
 				return true, nil
 			}
+
+			// comparing not existent field with {$all: [null, null, ..., null]} should return true
+			// (at least one null needs to be presented in the $all array)
+			if exprKey == "$all" {
```
rumyantseva, line 256: If you have an idea how to make this check better, please let me know.

integration/query_array_test.go
```diff
@@ -414,3 +414,185 @@ func TestArrayEquality(t *testing.T) {
...
+func TestQueryArrayAll(t *testing.T) {
+	t.Parallel()
+
+	ctx, collection := setup.Setup(t, shareddata.Composites, shareddata.Scalars)
+
+	// Insert additional data to check more complicated cases:
+	// - a longer array of ints;
+	// - a field is called differently and needs to be found with the {$all: [null]} case.
+	_, err := collection.InsertMany(ctx, []any{
+		bson.D{{"_id", "many-integers"}, {"customField", bson.A{42, 43, 45}}},
```
seeforschauer, line 428: Why not add it to shareddata?

rumyantseva: It doesn't seem to be needed in shared data.

internal/handlers/common/filter.go
```diff
@@ -468,6 +487,13 @@ func filterFieldExpr(doc *types.Document, filterKey string, expr *types.Document
 				return false, err
 			}
 
+		case "$all":
+			// {field: {$all: [value, another_value, ...]}}
+			res, err := filterFieldExprAll(fieldValue, exprValue)
+			if !res || err != nil {
```
seeforschauer, line 493: can it be replaced with?
```go
return filterFieldExprAll(fieldValue, exprValue)
```

rumyantseva: None of the other operators return it like this.

internal/types/array.go
```diff
@@ -149,3 +151,50 @@ func (a *Array) Max() any {
...
 	return max
 }
+
+// Contains checks if the Array contains the given value.
+func (a *Array) Contains(filterValue any) bool {
+	switch filterValue := filterValue.(type) {
+	case *Document, *Array:
+		// filterValue is a composite type, so either a and filterValue must be equal
+		// or at least one element of a must be equal with filterValue.
+		// TODO: Compare might be inaccurate for some corner cases, we might want to fix it later.
```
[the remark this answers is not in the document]
rumyantseva: If I knew the corner cases, I would cover them in this task. The problem is that I couldn't find them. So, creating a task is not reasonable as it's not straightforward what the task should be about and how much time it would take.

internal/handlers/common/filter.go
seeforschauer: > None of the other operators return it like this.

The question is equal to - if it can be used with other operators? 
If yes, then let's add a test with another operator. 
If no, then it can be returned.

rumyantseva: That seems to be out of the scope of the current task. If you think this approach needs to be refactored, please create a separate issue.

internal/types/array.go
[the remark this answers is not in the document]
seeforschauer: > If I knew the corner cases, I would cover them right here, not in a separate issue. 

I would suggest adding the word `unknown` to corner cases.

integration/query_array_test.go
seeforschauer: > It doesn't seem to be needed in shared data.

It should. 
IIRC we planned reordering and consistency for tests where all `InsertMany` is to be removed. Let's not add new work to that.

rumyantseva: @AlekSi as before we discussed that this doesn't go to shared data, please confirm or decline that @seeforschauer's statement.

w84thesun (approved):
LGTM

integration/query_array_test.go
AlekSi: The current problem with shared data is that we have to update all tests if we add something to it. On the one hand, it is a feature, as it allows us to notice things that were incompatible with MongoDB before, but we did not notice them. On the other hand – yeah, we have to update a lot of tests.

The plan is to move all query tests to compatibility tests that will not have a list of IDs to update. That will give us benefit without the downside. For now, let's not update shared data.

AlekSi (approved):

PR #728, opened 2022-06-09
Add arrays with `NaN`, `double` and nested empty array to tests' shared data

rumyantseva:
This PR provides additional examples for tests' shared data composites. 

These examples will help to cover more test cases. 
For example, they are helpful for #724.
Also, `array-empty-nested` reveals a bug described in #732.

### How to review?
- Take a look at the changes in `integration/shareddata/composites.go`.
- Take a look at the rest of the files. Various tests were rewritten to support new composites.

### DIFF
--- a/integration/shareddata/composites.go
+++ b/integration/shareddata/composites.go
@@ -38,7 +38,7 @@ var Composites = &Values[string]{
 		"array-embedded":      bson.A{bson.A{int32(42), "foo"}, nil},
 		"array-empty":         bson.A{},
 
-		// TODO: This case demonstrates some bugs.
+		// TODO: This case demonstrates a bug, see https://github.com/FerretDB/FerretDB/issues/732
 		// "array-empty-nested": bson.A{bson.A{}},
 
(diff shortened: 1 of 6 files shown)


integration/shareddata/composites.go
```diff
@@ -28,10 +32,14 @@ var Composites = &Values[string]{
 		"document-empty":             bson.D{},
 
 		"array":               bson.A{int32(42)},
+		"array-two":           bson.A{42.42, math.NaN()},
```
AlekSi, line 35: We use 42.13 in scalars; let's use the same value as it probably will trigger more cases

AlekSi, line 35:
```suggestion
		"array-two":           bson.A{42.13, math.NaN()},
```

We use 42.13 in scalars; let's use the same value as it probably will trigger more cases

AlekSi:
@rumyantseva 🛎️

seeforschauer (approved):

seeforschauer (approved):

PR #738, opened 2022-06-15
Handle null value for nameOnly in listDatabases handler

DoodgeMatvey:


w84thesun:
Can you please add a test that would show that issue solved?

Issue #388, opened 2022-03-23
Improve Windows (with WSL2) development support


AlekSi:
It should be possible to follow our contributing guidelines on Windows without ifs and buts.

* [x] Make Taskfile targets work without gcc for `-race`.
  * Use `{{if ne OS "windows"}}-race{{end}}`?
  * Use some kind of autoconf-like autodetection?
* [x] Update documentation to mention recommended WSL2 configuration.
* [x] Fix golangci lint task (add correct Windows path `internal\\wire` to .golangci.yml)
* [ ] ~Check that `task env-up` works (and close #297).~

CI / GitHub Actions changes are not in scope.

AlekSi:
@w84thesun I think we should have more documentation for contributors on Windows. At the very least, we should mention how to run commands, what works and what doesn't.

PR #753, opened 2022-06-21
Skip `-race` flag on Windows

w84thesun:
Refs #388.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -90,13 +90,13 @@ tasks:
   test-unit-short:
     desc: "Run short unit tests"
     cmds:
-      - go test -short -count=1 -race -shuffle=on -coverprofile=cover.txt ./...
+      - go test -short -count=1 {{if ne OS "windows"}} -race {{end}} -shuffle=on -coverprofile=cover.txt ./...
 
   test-unit:
     desc: "Run all unit tests"
     cmds:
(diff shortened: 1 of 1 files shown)


seeforschauer (approved):

AlekSi (changes requested):
That's not the place where we have `-race`

Taskfile.yml
```diff
@@ -168,7 +168,7 @@ tasks:
     desc: "Build bin/ferretdb-testcover"
     deps: [gen-version]
     cmds:
-      - go test -c -o=bin/ferretdb-testcover -trimpath -tags=testcover,tigris -race -coverpkg=./... ./cmd/ferretdb
+      - go test -c -o=bin/ferretdb-testcover -trimpath -tags=testcover,tigris {{if not (eq OS "windows")}} -race {{end}} -coverpkg=./... ./cmd/ferretdb
```
AlekSi, line 171: Can we use `ne` (not equal) operator?

w84thesun: Yes, changed to `ne`.

w84thesun:
> That's not the only place where we have `-race`

Added to all other places.

AlekSi (approved):

seeforschauer (approved):

noisersup (approved):

PR #757, opened 2022-06-21
Fix fluky test

AlekSi:
We run a lot of tests in parallel, so the number of collections is in the range between 1 (this test) and many, where "many" is 100 for now.
Uptime can also be as small as 0 seconds on a fast machine but can also be large if the environment wasn't restarted for a long time.

# Checklist

* [x] `task all` passes.


rumyantseva (approved):
Still looks a bit flaky, but let's try. The explanation about 100 makes sense to me.

seeforschauer (approved):
well, ok we a;ready discussed `GreaterOrEqual` before..

rumyantseva (approved):
LGTM

PR #778, opened 2022-06-23
Support all valid collection names

### LINKED ISSUE #750: Support all valid collection names
See #125.

We should store a mapping between PostgreSQL and MongoDB names in a settings table. While we don't necessarily need it for this issue, we will need it later, so let's add it now, in a relatively small task.

Let's create a table/collection `_ferretdb_settings` per MongoDB database / PostgreSQL schema. It will contain a single fjson-encoded `*types.Document` with the following structure:
```json5
{
  "collections": {
    "collection1": "table1",
    // ...
  }
}
```
(`$f` fields omitted for brevity but should be present – that's a normal fjson-encoding of a `*types.Document` value).

`_ferretdb_` prefix is reserved for internal FerretDB collections. Tables/collections with that prefix in the name should be omitted everywhere where we return a list of collections, etc.

Let's add a method to `pgdb.Pool` type to return a PostgreSQL table name for the MongoDB collection name. It should query `_ferretdb_settings` table and return the existing value if present. If not, it should produce a valid PostgreSQL table name for a given collection name. The result should be stable (i.e., should not depend on other collections, the total number of them, random numbers, etc.) and should be readable to make debugging easier. For example, it could use a shorted version of a name together with a checksum/hash of a full name (e.g. `values_a4gf`). That result then should be saved back in the `_ferretdb_settings` table in the same transaction.

We should then use that method in all places where it is needed.

Add a test for a very long collection name.

Validating collection names is not in the scope. Tigris is not in scope.

Create PR as early as possible to get early feedback.

w84thesun:
Closes #750.

# Checklist.

* [x] `task all` passes.


internal/handlers/pg/pgdb/pool.go
```diff
@@ -413,3 +440,97 @@ func (pgPool *Pool) SchemaStats(ctx context.Context, schema, collection string)
...
+	}
+
+	return "", nil
+}
+
+// getTableNameFormatted returns collection name in form <shortened_name>_<name_hash>.
+func getTableNameFormatted(name string) (string, error) {
+	hash32 := fnv.New32a()
+	_, err := hash32.Write([]byte(name))
+	if err != nil {
```
rumyantseva, line 525: Hmm... Should it panic here?

internal/handlers/pg/pgdb/pool.go
```diff
@@ -413,3 +440,97 @@ func (pgPool *Pool) SchemaStats(ctx context.Context, schema, collection string)
...
+		return "", lazyerrors.Error(err)
+	}
+
+	return "", nil
+}
+
+// getTableNameFormatted returns collection name in form <shortened_name>_<name_hash>.
+func getTableNameFormatted(name string) (string, error) {
+	hash32 := fnv.New32a()
```
rumyantseva, line 523: Is there any particular reason you decided to use FNV? I guess FNV should be enough for us, I'm just curious.

internal/handlers/pg/pgdb/pool.go
```diff
@@ -413,3 +440,97 @@ func (pgPool *Pool) SchemaStats(ctx context.Context, schema, collection string)
...
+		return "", lazyerrors.Error(err)
+	}
+	if !tableExists {
+		err = pgPool.CreateSettingsTable(ctx, db)
+		if err != nil {
+			return "", lazyerrors.Error(err)
+		}
+	}
+
+	sql := `SELECT _jsonb FROM ` + pgx.Identifier{db, collectionPrefix + "settings"}.Sanitize()
```
rumyantseva, line 474: So, in fact, in this table we have only one raw that stores all the settings for everything? Not like "one line - one collection"? 

I wonder how does it look like from the race condition perspective? Let's say we got a few parallel requests to update settings (e.g. create settings for a few different collections). How would it be handled?

internal/handlers/pg/pgdb/pool.go
```diff
@@ -413,3 +440,97 @@ func (pgPool *Pool) SchemaStats(ctx context.Context, schema, collection string)
...
+
+	if collections.Has(collection) {
+		return must.NotFail(collections.Get(collection)).(string), nil
+	}
+
+	tableName := must.NotFail(getTableNameFormatted(collection))
+	must.NoError(collections.Set(collection, tableName))
+	must.NoError(settings.Set("collections", collections))
+
+	sql = `UPDATE ` + pgx.Identifier{db, collectionPrefix + "settings"}.Sanitize() + `SET _jsonb = $1`
```
rumyantseva, line 512: Yeah, this is what I mean. It looks quite unsafe from the race condition perspective.

rumyantseva (commented):
I had a look. My biggest concern is race conditions. 
Should we discuss how are we going to avoid them?

internal/handlers/pg/pgdb/pool.go
w84thesun: First of all, I went to the wiki hash functions page searching for some stable algorithm were was a list of non-cryptographic hash functions and then I looked at the Go library which has either hashmap or FNV. I decided to peed FNV because it has the 32-bit version which requires slightly less space.

internal/handlers/pg/pgdb/pool.go
w84thesun: Yeah, I guess we should panic here. Added `NotFail`.

internal/handlers/pg/pgdb/settings.go
```diff
@@ -0,0 +1,223 @@
...
+	// Internal collections prefix.
+	collectionPrefix = "_ferretdb_"
+
+	// PostgreSQL max table name length.
+	maxTableNameLength = 63
+)
+
+// CreateSettingsTable creates FerretDB settings table if it doesn't exist.
+func (pgPool *Pool) CreateSettingsTable(ctx context.Context, db string) error {
+	tables, err := pgPool.tables(ctx, db)
```
rumyantseva, line 41: Another thing that needs a transaction? Same case: we have two queries that query tables and couldn't find `collectionPrefix+"settings"`. Will both of them try to create it?

internal/handlers/pg/pgdb/pool.go
```diff
@@ -359,7 +403,12 @@ func (pgPool *Pool) CreateTableIfNotExist(ctx context.Context, db, collection st
 		return false, lazyerrors.Error(err)
 	}
 
-	if err := pgPool.CreateTable(ctx, db, collection); err != nil {
+	table, err := pgPool.GetTableName(ctx, db, collection)
```
rumyantseva, line 406: Again, there is a moment between `GetTableName` and `GetTableName` and we can't guarantee that during that moment the table won't be created. 

I believe this is a relatively minor thing, but I wonder how it would work with our hashes? E.g. is it possible that we can create two tables with different hashes for the same name?

internal/handlers/pg/pgdb/pool.go
```diff
@@ -203,11 +205,53 @@ func (pgPool *Pool) Schemas(ctx context.Context) ([]string, error) {
 	return res, nil
 }
 
+// Collections returns a sorted list of FerretDB collection names.
+func (pgPool *Pool) Collections(ctx context.Context, schema string) ([]string, error) {
+	tx, err := pgPool.Begin(ctx)
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+	defer tx.Rollback(ctx)
```
rumyantseva, line 214: So, we always want to rollback? Why?

integration/basic_test.go
```diff
@@ -127,3 +127,18 @@ func TestFindCommentQuery(t *testing.T) {
 	require.NoError(t, err)
 	assert.Contains(t, databaseNames, name)
 }
+
+func TestLongCollectionName(t *testing.T) {
```
rumyantseva, line 131: I like this test! Should we make it a big more generic and cover a few cases (long names, short names)?

internal/handlers/pg/pgdb/settings.go
```diff
@@ -0,0 +1,223 @@
...
+	schemaExists, err := pgPool.schemaExists(ctx, db)
+	if err != nil {
+		return lazyerrors.Error(err)
+	}
+
+	if !schemaExists {
+		return nil
+	}
+
+	tx, err := pgPool.Begin(ctx)
```
rumyantseva, line 178: Why do we start this transaction here (and not above)?

rumyantseva (commented):
My biggest concern is about transactions... I don't fully understand if they help or we rather need something like `select for update` in some cases? This is probably something to discuss on our daily.

Another story that is outside of the scope of this task (but needs to be noted somewhere): when we have queries that create collections (e.g. we attempt to insert into a collection that doesn't exist), that needs to be handled carefully (so, in case of parallel requests we don't create two separate tables for the same collection). Before the current task, in such a case we would just have an error that the table already exists, but how does it behave now, if we create two tables with different hashes?

I suggest discussing all these things at our daily sync today.

internal/handlers/pg/msg_create.go
```diff
@@ -72,7 +72,16 @@ func (h *Handler) MsgCreate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+	if err := h.pgPool.CreateSettingsTable(ctx, db); err != nil && err != pgdb.ErrAlreadyExist {
+		return nil, lazyerrors.Error(err)
+	}
+
+	table, err := h.pgPool.GetTableName(ctx, db, collection)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = h.pgPool.CreateTable(ctx, db, table); err != nil {
```
seeforschauer, line 84: Here you are trying to `CreateTable` but error message is with `collection` variable name. 
Could you please change this?

internal/handlers/pg/pgdb/pool.go
```diff
@@ -203,11 +205,53 @@ func (pgPool *Pool) Schemas(ctx context.Context) ([]string, error) {
...
+			continue
+		}
+
+		filtered = append(filtered, table)
+	}
+
+	return filtered, nil
+}
+
+// tables returns a list of PostgreSQL table names.
```
seeforschauer, line 253: ... and it also creates settings table. 
Would you please add this to the function comment. 
And, probably, it's better to rename the function as it creates schema, creates settings table..

internal/handlers/pg/pgdb/pool.go
```diff
@@ -413,3 +462,23 @@ func (pgPool *Pool) SchemaStats(ctx context.Context, schema, collection string)
 	}
 	return &res, nil
 }
+
+// schemaExists returns true if given schema exists.
+func (pgPool *Pool) schemaExists(ctx context.Context, db string) (bool, error) {
```
seeforschauer, line 467: I constantly wonder, how the db relates to schema (db contains schema in pg, and schema relates to mongodb collection names, but what is table then?).

And now we have yet another entity - settings table.

Let's call clarify our scheme in the comments and write a proper documentation for it 
* what are the fields inside that settings table
* why the structure is like this
* why we need another one entity.
* what else important
* etc.

The documentation can be added i.e. in the settings.go file, where the Settings Table term introduced.

internal/handlers/pg/pgdb/settings.go
```diff
@@ -0,0 +1,223 @@
...
+
+	settings, err := pgPool.getSettingsTable(ctx, tx, db)
+	if err != nil {
+		return "", lazyerrors.Error(err)
+	}
+
+	collectionsDoc := must.NotFail(settings.Get("collections"))
+	collections, ok := collectionsDoc.(*types.Document)
+	if !ok {
+		return "", lazyerrors.Errorf("invalid collections document: %v", collectionsDoc)
```
seeforschauer, line 110: Let's create meaningful error messages: 
* it's not a document, but why it should be a document?

integration/basic_test.go
w84thesun: In general there are 3 possible cases: less than 63 symbols, 63 symbols, more than 63 symbols.
First one is already covered by all other integration tests as they are creating collections. Third case covered with this test. I think we can add the second case.

internal/handlers/pg/pgdb/settings.go
```diff
@@ -0,0 +1,223 @@
...
+
+// RemoveTableFromSettings removes collection from FerretDB settings table.
+func (pgPool *Pool) RemoveTableFromSettings(ctx context.Context, db, collection string) error {
+	schemaExists, err := pgPool.schemaExists(ctx, db)
+	if err != nil {
+		return lazyerrors.Error(err)
+	}
+
+	if !schemaExists {
+		return nil
```
seeforschauer, line 175: Is here should be an error `ns not found`?

integration/basic_test.go
w84thesun: Added.

internal/handlers/pg/msg_create.go
w84thesun: It should be `collection` variable as this is error would be returned to the user and it would break integration test.

internal/handlers/pg/pgdb/pool.go
w84thesun: We don't want to change anything here but load settings table. `getSettingsTable` accepts transaction, so I have to create one. I can probably change `Rollback` to `Commit` but `Rollback` looked more logically correct to me.

internal/handlers/pg/pgdb/settings.go
```diff
@@ -0,0 +1,223 @@
...
+)
+
+// CreateSettingsTable creates FerretDB settings table if it doesn't exist.
+func (pgPool *Pool) CreateSettingsTable(ctx context.Context, db string) error {
+	tables, err := pgPool.tables(ctx, db)
+	if err != nil {
+		return lazyerrors.Error(err)
+	}
+
+	if slices.Contains(tables, collectionPrefix+"settings") {
```
seeforschauer, line 46: let's move `settings` to the constants also.

internal/handlers/pg/pgdb/settings.go
w84thesun: Fixed

internal/handlers/pg/pgdb/settings.go
w84thesun: Done.

internal/handlers/pg/pgdb/settings.go
w84thesun: In that particular case we should just exit.

internal/handlers/pg/msg_create.go
seeforschauer: > It should be `collection` variable as this is error that would be returned to the user and it would break integration test.

Can we add a comment here then? Thank you.

internal/handlers/pg/pgdb/pool.go
seeforschauer: > We don't want to change anything here but load settings table. `getSettingsTable` accepts transaction, so I have to create one. I can probably change `Rollback` to `Commit` but `Rollback` looked more logically correct to me.

Let's add a comment here right as you wrote in the comment explanation.

internal/handlers/pg/pgdb/settings.go
```diff
@@ -0,0 +1,226 @@
...
+
+// updateSettingsTable updates FerretDB settings table.
+func (pgPool *Pool) updateSettingsTable(ctx context.Context, tx pgx.Tx, db string, settings *types.Document) error {
+	sql := `UPDATE ` + pgx.Identifier{db, settingsTableName}.Sanitize() + `SET settings = $1`
+	_, err := tx.Exec(ctx, sql, must.NotFail(fjson.Marshal(settings)))
+	return err
+}
+
+// getSettingsTable returns FerretDB settings table.
+func (pgPool *Pool) getSettingsTable(ctx context.Context, tx pgx.Tx, db string) (*types.Document, error) {
```
seeforschauer, line 140: The type of returning value is `types.Document`. Why / what for? 
*  It's not convenient as you later decomposing the document and receiving from `must.Get`
* * also it's a `pgdb` handler's package, it can use bare values, isn't it?

internal/handlers/pg/pgdb/settings.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// updateSettingsTable updates FerretDB settings table.
+func (pgPool *Pool) updateSettingsTable(ctx context.Context, tx pgx.Tx, db string, settings *types.Document) error {
+	sql := `UPDATE ` + pgx.Identifier{db, settingsTableName}.Sanitize() + `SET settings = $1`
+	_, err := tx.Exec(ctx, sql, must.NotFail(fjson.Marshal(settings)))
+	return err
+}
+
+// getSettingsTable returns FerretDB settings table.
```
seeforschauer, line 139: FerretDB settings table has a `type.Document` format, is it consistent regarding Postgres and internal usage (the function is not exported)?

internal/handlers/pg/pgdb/settings.go
```diff
@@ -0,0 +1,226 @@
...
+	if err != nil {
+		return "", lazyerrors.Error(err)
+	}
+
+	return tableName, nil
+}
+
+// updateSettingsTable updates FerretDB settings table.
+func (pgPool *Pool) updateSettingsTable(ctx context.Context, tx pgx.Tx, db string, settings *types.Document) error {
+	sql := `UPDATE ` + pgx.Identifier{db, settingsTableName}.Sanitize() + `SET settings = $1`
```
seeforschauer, line 134: It seems like there are no identifiers for settings. 

* so it's not an `update` it's `set` of settings - as all, entire (table?) changes.
* when a database will have a lot of collections/ tables, will it be still convenient?

internal/handlers/pg/pgdb/settings.go
```diff
@@ -0,0 +1,226 @@
...
+	_, err = pgPool.Exec(ctx, sql, must.NotFail(fjson.Marshal(settings)))
+	if err != nil {
+		return lazyerrors.Error(err)
+	}
+
+	return nil
+}
+
+// GetTableName returns the name of the table for given collection or error.
+func (pgPool *Pool) GetTableName(ctx context.Context, db, collection string) (string, error) {
```
seeforschauer, line 70: Here also: the function not only creates the enigmatic settings table but also updates it.

internal/handlers/pg/pgdb/settings.go
```diff
@@ -0,0 +1,226 @@
...
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+	"github.com/FerretDB/FerretDB/internal/util/must"
+)
+
+const (
+	// Internal collections prefix.
+	collectionPrefix = "_ferretdb_"
+
+	// Settings table name.
```
seeforschauer, line 35: Hm, I though it depends on collection name. But now I see it's a FerretDB - database wide settings table name? 
But you replace the settings value each time...

internal/handlers/pg/pgdb/settings.go
w84thesun: Please see #750

internal/handlers/pg/pgdb/settings.go
w84thesun: Yes, it's FerretDB database settings table name.

internal/handlers/pg/pgdb/settings.go
```diff
@@ -0,0 +1,226 @@
...
+	settingsTableName = collectionPrefix + "settings"
+
+	// PostgreSQL max table name length.
```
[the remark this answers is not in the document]
w84thesun: I think it's not a common case and we could support it when we have any request from our users.

internal/handlers/pg/pgdb/pool.go
rumyantseva: +1! With that explanation, rollback makes sense (that was also my guess), but let's document it for clarity.

rumyantseva (commented):
@w84thesun so, I resolved what was explained or clearly out of scope. I have two other comments left, if they are not reasonable / out of scope, please comment on them.

internal/handlers/pg/pgdb/pool.go
w84thesun: Yeah, we store a single row with a document that stores all the settings for a schema.

On a previous iteration, there would be a chance that we create several documents but right now all access to that table is going to be under the transaction, so I guess it should be ok.

internal/handlers/pg/pgdb/pool.go
w84thesun: Added.

internal/handlers/pg/pgdb/pool.go
w84thesun: No, it's not. Please see full function body.

internal/handlers/pg/pgdb/settings.go
w84thesun: I've added more details on the function behavior.

internal/handlers/pg/pgdb/settings.go
w84thesun: I think we can call `schemaExists` outside of the transaction as it won't affect anything.

internal/handlers/pg/pgdb/pool.go
w84thesun: I've added a comment to `CreateSettingsTable`. I think it's enough for now. We can't clarify anything in that PR, as it requires refactoring to `pgPool` and handlers methods.

internal/handlers/pg/msg_create.go
w84thesun: Why? We return the wire protocol error here which is used by mongo clients. It should be enough. Although if someone would change `collection` to `table` here integration test would show it immediately.

seeforschauer: > Why? We return the wire protocol error here which is used by mongo clients. It should be enough. Although if someone would change `collection` to `table` here integration test would show it immediately.

let's make a call.

internal/handlers/pg/pgdb/pool.go
seeforschauer: > Added.

I expected you to add exactly as in the comment:
> We don't want to change anything here but load settings table. getSettingsTable accepts transaction, so I have to create one.

^^ and this information are not obvious.

But you've added:
> call Rollback() here as we don't change anything in the database

but that's obvious to do a rollback if you don't want change!

internal/handlers/pg/pgdb/settings.go
```diff
@@ -0,0 +1,242 @@
...
+	maxTableNameLength = 63
+)
+
+// CreateSettingsTable creates FerretDB settings table if it doesn't exist.
+// Settings table is used to store FerretDB settings like collections names.
```
seeforschauer, line 43:
```suggestion
// Settings table is used to store FerretDB settings like collections names mapping.
```

internal/handlers/pg/pgdb/pool.go
w84thesun: Added comment to transaction creation.

internal/handlers/pg/pgdb/pool.go
w84thesun: Hashes must be stable, they don't depend on anything else just on the passed string itself.

internal/handlers/pg/pgdb/settings.go
```diff
@@ -0,0 +1,242 @@
...
+		return "", lazyerrors.Error(err)
+	}
+	if !slices.Contains(tables, settingsTableName) {
+		err = pgPool.CreateSettingsTable(ctx, db)
+		if err != nil {
+			return "", lazyerrors.Error(err)
+		}
+	}
+
+	settings, err := pgPool.getSettingsTable(ctx, tx, db)
```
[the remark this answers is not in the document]
seeforschauer: > Let's get settings at the beginning of the function, so `getSettingsTable` could be used w/o transaction SQL.

I found better solution:

`getSettingsTable` can use `pgPool` isn't it? 

```go
// getSettingsTable returns FerretDB settings table.
func (pgPool *Pool) getSettingsTable(ctx context.Context, tx pgx.Tx, db string) (*types.Document, error) {
	sql := `SELECT settings FROM ` + pgx.Identifier{db, settingsTableName}.Sanitize()
	rows, err := pgPool.Query(ctx, SQL)
```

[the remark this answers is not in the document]
seeforschauer: I was wrong, this comment is not actual anymore. The transaction is necessary.

internal/handlers/pg/pgdb/pool_test.go
```diff
@@ -198,41 +198,54 @@ func TestConcurrentCreate(t *testing.T) {
...
-			go func() {
-				<-start
-				if withTable {
-					res <- pool.CreateTable(ctx, schemaName, tableName)
-				} else {
-					res <- pool.CreateSchema(ctx, schemaName)
-				}
-			}()
-		}
+	for _, tc := range []struct {
```
w84thesun, line 201: @rumyantseva helped me refactor this test and make it work. Thanks a lot!
@AlekSi PTAL, maybe you have some thoughts on it.

w84thesun: Well, at least it worked on my machine..

rumyantseva (changes requested):
I have a worry now :)

It looks to me that on the level of the `handlers/pg` some functions work with "beautiful" collection names (e.g. `msg_drop`). In other places, we work with "hashed" table names (`msg_insert`).

It looks quite inconsistent, and also it's a place where it's easy to make a mistake.
Can we "encapsulate" hashed names on the `pgdb` library level? So, none of `handlers/pg` entities know about hashed table names? 

As I promised to record a loom for codereview, here it is: https://www.loom.com/share/baf18cec79094d8280d4896ec3c32fc5 :) 
Let me know if it helped

internal/handlers/pg/pgdb/pool_test.go
rumyantseva: This test was a bit strange initially... I wonder if we need to rewrite it.

internal/handlers/pg/pgdb/pool.go
```diff
@@ -279,63 +346,110 @@ func (pgPool *Pool) DropSchema(ctx context.Context, schema string) error {
...
-	case pgerrcode.UniqueViolation, pgerrcode.DuplicateObject:
-		// https://www.postgresql.org/message-id/CA+TgmoZAdYVtwBfp1FL2sMZbiHCWT4UPrzRLNnX1Nb30Ku3-gg@mail.gmail.com
-		// Reproducible by integration tests.
+	tx, err := pgPool.Begin(ctx)
+	if err != nil {
+		return lazyerrors.Error(err)
+	}
+	defer func() {
+		if err != nil {
+			_ = tx.Rollback(ctx)
```
seeforschauer, line 374: let's process all errors.

internal/handlers/pg/pgdb/pool.go
```diff
@@ -239,10 +294,22 @@ func (pgPool *Pool) Tables(ctx context.Context, schema string) ([]string, error)
...
+	tx, err := pgPool.Begin(ctx)
+	if err != nil {
+		return lazyerrors.Error(err)
+	}
+	defer func() {
+		if err != nil {
+			must.NoError(tx.Rollback(ctx))
+			return
+		}
+		must.NoError(tx.Commit(ctx))
```
seeforschauer, line 306: why panic here? should return error

internal/handlers/pg/pgdb/pool_test.go
```diff
@@ -198,41 +198,54 @@ func TestConcurrentCreate(t *testing.T) {
...
+		expectedErr int
+	}{
+		{
+			name: "CreateSchema",
+			f: func() error {
+				return pool.CreateSchema(ctx, schemaName)
+			},
+			expectedErr: n - 1,
+		},
+		{
```
AlekSi, line 213:
```suggestion
		}, {
```

internal/handlers/pg/pgdb/pool.go
```diff
@@ -279,63 +346,110 @@ func (pgPool *Pool) DropSchema(ctx context.Context, schema string) error {
...
+	tx, err := pgPool.Begin(ctx)
+	if err != nil {
+		return lazyerrors.Error(err)
+	}
+	defer func() {
+		if err != nil {
+			_ = tx.Rollback(ctx)
+			return
+		}
+		_ = tx.Commit(ctx)
```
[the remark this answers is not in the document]
w84thesun: Fixed.

internal/handlers/pg/pgdb/pool.go
```diff
@@ -279,63 +346,110 @@ func (pgPool *Pool) DropSchema(ctx context.Context, schema string) error {
...
+
+	err = pgPool.removeTableFromSettings(ctx, tx, schema, collection)
+	if err != nil && err != ErrTableNotExist {
+		return lazyerrors.Error(err)
+	}
+	if err == ErrTableNotExist {
+		return ErrTableNotExist
+	}
+
+	// TODO probably not CASCADE
```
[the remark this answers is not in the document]
w84thesun: I didn't add anything here, that line was copied as part of that code.

internal/handlers/pg/pgdb/pool.go
w84thesun: Fixed.

internal/handlers/pg/pgdb/pool.go
w84thesun: Fixed.

internal/handlers/pg/pgdb/pool.go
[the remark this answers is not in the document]
seeforschauer: > I didn't add anything here, that line was copied as part of that code.

Would you please create an issue?

seeforschauer (dismissed):

seeforschauer (approved):

seeforschauer (changes requested):
@w84thesun  sorry, tests

seeforschauer (approved):

AlekSi:
See #815

internal/handlers/pg/pgdb/pool_test.go
```diff
@@ -198,41 +198,57 @@ func TestConcurrentCreate(t *testing.T) {
...
+			f: func() error {
+				return pool.CreateDatabase(ctx, schemaName)
+			},
+			compareFunc: func(t *testing.T, errors int) bool {
+				return assert.Equal(t, n-1, errors)
+			},
+		}, {
+			name: "CreateCollection",
+			f: func() error {
+				return pool.CreateCollection(ctx, schemaName, tableName)
```
rumyantseva, line 217: Is this `tableName` should we a `collectionName` now?

rumyantseva (approved):
It looks to me that the last place where `tableName` and `collectionName` are mixed is in the tests. But there will be tons of tests to refactor! So, I think personally I'm ready to approve what we have now. Maybe we can refactor names in tests later?

@w84thesun you rock! That was a huge refactoring!

internal/handlers/pg/pgdb/pool_test.go
w84thesun: Yeah, I suggest working on those tests in a separate PR. We might want to add more tests for the Pool.

internal/util/testutil/pg.go
```diff
@@ -129,13 +128,13 @@ func Table(ctx context.Context, tb testing.TB, pool *pgdb.Pool, db string) strin
 	table := TableName(tb)
```
AlekSi, line 128: @rumyantseva Please clean-up those helpers in your PR. For example, it is not clean if this helper return table name (with a hash) or collection name. At very least, documentation should be updated. Maybe name too

AlekSi (approved):
LGTM to unblock Elena

internal/handlers/pg/pgdb/pool_test.go
AlekSi: Ha, I did not read it before review. Good to see us on the same page

internal/util/testutil/pg.go
rumyantseva: 👌

PR #780, opened 2022-06-23
Fix `golangci-lint` `depguard` rules for Windows systems

w84thesun:
This PR refs #388.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -131,6 +131,20 @@ issues:
       path: internal/util/testutil
       text: pgdb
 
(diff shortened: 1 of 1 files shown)


AlekSi:
I think we should finish #758 instead

seeforschauer (approved):

w84thesun:
> I think we should finish #758 instead

Maybe we can merge it now so I can run linter locally and remove it next week when #758 would be done?

AlekSi:
#758 is ready for review

PR #788, opened 2022-06-23
Drop `test_db`

AlekSi:
We are not using it much now, but it makes `task env-up` much slower.

# Checklist

* [x] `task all` passes.

### DIFF
--- a/cmd/envtool/main.go
+++ b/cmd/envtool/main.go
@@ -25,47 +25,17 @@ import (
 	"os/exec"
 	"runtime"
 	"strings"
-	"sync"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 
-	"github.com/FerretDB/FerretDB/internal/clientconn"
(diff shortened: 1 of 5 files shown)


cmd/envtool/main.go
```diff
@@ -400,7 +223,7 @@ func main() {
 
 	logger := setupLogger(*debugLevel)
 
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
seeforschauer, line 226: Would you mind to add add this `time.Minute` to the top of the file?

rumyantseva (approved):
LGTM

AlekSi:
We never had `-race` in `go run ./cmd/envtool/main.go` command; otherwise, we would notice that problem sooner

cmd/envtool/main.go
AlekSi: Why? We use in a single place and its value is not configurable

noisersup (approved):

seeforschauer (approved):

PR #790, opened 2022-06-26
Support `multi` `update`'s argument

### LINKED ISSUE #486: Support `multi` `update`'s argument
We should support `update`'s `multi` argument.
https://www.mongodb.com/docs/manual/reference/command/update/#update-statements

fcoury:
Closes #486

This PR accepts the additional optional flag for multiple updates straight on the `update` command:

https://www.mongodb.com/docs/manual/reference/method/db.collection.update/#std-label-update-multi

~~Since there is no high level API to call update (only UpdateOne or UpdateMany), I added a test that uses the lower level `RunCommand` API, providing the expected update command payload document.~~

My strategy on this particular test was to add two documents with a same field (`x: 1`) and then update using ~~`multi: false`~~ `UpdateOne` first, and checking that only the first document was updated; and then update using ~~`multi: true`~~ `UpdateMany` making sure both documents were modified.

# Checklist

* [x] `task all` passes.


AlekSi:
> Since there is no high level API to call update (only UpdateOne or UpdateMany)

I think they use `update` command under the hood, setting `multi` flag to `false` (or omitting it) and `true`

fcoury:
> > Since there is no high level API to call update (only UpdateOne or UpdateMany)
> 
> I think they use `update` command under the hood, setting `multi` flag to `false` (or omitting it) and `true`

Updated the test case to use UpdateOne and UpdateMany under the hood, and added an additional test case to make sure it falls back to false when absent.

Now the behavior is exactly the same when testing from the go driver and mongosh:

**FerretDB**
image

**MongoDB**
image

fcoury:
Rebased from new `main` branch.

fcoury:
Ready for review.

integration/update_test.go
```diff
@@ -770,6 +770,66 @@ func TestUpdateMany(t *testing.T) {
 	}
 }
 
+func TestMultiFlag(t *testing.T) {
+	t.Parallel()
+
+	for name, tc := range map[string]struct {
+		filter bson.D
+		update bson.D
+		multi  string
```
AlekSi, line 779:
```suggestion
		multi  bool
```

AlekSi (commented):
May you also port those changes to `tigris/msg_update.go`. No tests are necessary for that handler yet

AlekSi:
@GinGin3203 PTAL if you have a moment :)

integration/update_test.go
GinGin3203: @fcoury I'm not really sure why this parameter is needed at all
Shouldn't all the tests in the suite run the `update` command with `multi` set to `true`?

integration/update_test.go
```diff
@@ -770,6 +770,66 @@ func TestUpdateMany(t *testing.T) {
...
+
+			_, err := collection.InsertMany(ctx, []any{
+				bson.D{{"_id", "first"}, {"foo", "x"}},
+				bson.D{{"_id", "second"}, {"foo", "x"}},
+			})
+			require.NoError(t, err)
+
+			var actual bson.D
+
+			if tc.multi == "true" {
```
GinGin3203, line 808: Running `update` with `multi:true` via `RunCommand` was actually a good idea, otherwise I'm not sure what are we even testing here
@AlekSi WDYT?

integration/update_test.go
```diff
@@ -770,6 +770,66 @@ func TestUpdateMany(t *testing.T) {
...
+				collection.UpdateMany(ctx, tc.filter, tc.update)
+			} else {
+				collection.UpdateOne(ctx, tc.filter, tc.update)
+			}
+
+			err = collection.FindOne(ctx, bson.D{{"_id", "first"}}).Decode(&actual)
+			require.NoError(t, err)
+
+			require.Equal(t, bson.D{{"_id", "first"}, {"foo", "y"}}, actual)
```
GinGin3203, line 817: Could you make a field `res` of type `bson.D` in the test table and move the expected documents for each test there? It allows for better readability (you see the query and what you get out of it), and also it would be consistent with already existing `TestUpdate/.../` functions

GinGin3203: (Also it would be easier to add new tests cases if needed)

internal/handlers/pg/msg_update.go
```diff
@@ -154,6 +158,10 @@ func (h *Handler) MsgUpdate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 			continue
 		}
 
+		if len(resDocs) > 1 && !multi {
+			resDocs = resDocs[0:1]
```
GinGin3203, line 162: I have a bad feeling about this because I don't understand if `resDocs[0]` is _always_ going to be the same document MongoDB would choose for an update, but I've checked out this branch and played with it a bit and it seems like Mongo and FerretDB are actually updating the same document, so it seems to work fine

GinGin3203 (changes requested):

internal/handlers/pg/msg_update.go
fcoury: Thank you for the review! I agree with you and I also only have empirical evidence that it works. I will go ahead and add a FIXME note to the code along with the updates the feedback you provided above.

integration/update_test.go
AlekSi: UpdateMany and UpdateOne are Go driver's methods that send `update` with `multi:true` and `multi:false` under the hood

rumyantseva (commented):
It doesn't look correct to me, but maybe I just don't understand something.

internal/handlers/pg/msg_update.go
```diff
@@ -190,6 +194,10 @@ func (h *Handler) MsgUpdate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 			continue
 		}
 
+		if len(resDocs) > 1 && !multi {
```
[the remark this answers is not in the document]
w84thesun: Nope, there is that condition before `if len(resDocs) == 0` which is used to perform upsert. In that case, we don't care about multi as no result docs were found. Then we update the resulting docs slice to 1 if multi == false and perform the actual update.

integration/update_test.go
```diff
@@ -75,3 +75,100 @@ func TestUpdateUpsert(t *testing.T) {
...
+						bson.D{{"q", tc.filter}, {"u", tc.update}, {"multi", tc.multi}},
+					}},
+				}
+
+				var result bson.D
+				err := collection.Database().RunCommand(ctx, command).Decode(&result)
+				require.NoError(t, err)
+
+				AssertEqualDocuments(t, tc.stat, result)
+			})
```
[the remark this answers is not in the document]
w84thesun: Ok, I'll add that check.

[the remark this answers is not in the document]
rumyantseva: (for the case when multi = false)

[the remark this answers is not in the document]
w84thesun: I've added the check for filter results after the update operation. If we multi is false, then `Find` will return 5 docs (1 updated), if `multi` is true, then 0 (6 updated).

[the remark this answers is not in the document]
rumyantseva: Yeah, sounds reasonable!

AlekSi (approved):

rumyantseva (approved):
LGTM

PR #791, opened 2022-06-26
Add support for updates with replacement objects

### LINKED ISSUE #697: Add support for update with replacement documents
We must support update with replacement documents.

```go
func TestUpdateReplaceDocuments(t *testing.T) {
	t.Parallel()

		filter       bson.D
		update       bson.D
		updateResult bson.D
		query        bson.D
		res          bson.D
	}{
		"Replace": {
			filter:       bson.D{{"value.foo.bar", bson.D{{"$eq", int32(1)}}}},
			update:       bson.D{{"new-value", int32(1)}},
			updateResult: bson.D{{"n", int32(1)}, {"nModified", int32(1)}, {"ok", float64(1)}},
			query:        bson.D{{"_id", "document"}},
			res:          bson.D{{"_id", "document"}, {"new-value", int32(1)}},
		},
	} {
		name, tc := name, tc
			t.Parallel()
			ctx, collection := setup(t)

			_, err := collection.InsertMany(
				ctx,
				[]any{
					bson.D{
						{"_id", "document"},
						{"value", bson.D{{"foo", bson.D{{"bar", int32(1)}}}}},
					},
				},
			)
			require.NoError(t, err)

			res := collection.Database().RunCommand(
				ctx,
				bson.D{
					{"update", collection.Name()},
					{"updates", bson.A{bson.D{{"q", tc.filter}, {"u", tc.update}}}},
				},
			)
			require.NoError(t, res.Err())

			var actual bson.D
			err = res.Decode(&actual)
			require.NoError(t, err)

			AssertEqualDocuments(t, tc.updateResult, actual)

			err = collection.FindOne(ctx, tc.query).Decode(&actual)
			require.NoError(t, err)
			AssertEqualDocuments(t, tc.res, actual)
		})
	}
}
```

fcoury:
Closes #697

The MongoDB documentation states that when a update call is made with no operation on the `update` parameter, it is treated as a Replacement Object and the whole document is replaced by the contents of the `update` payload.

More details here:
https://www.mongodb.com/docs/manual/reference/method/db.collection.update/#std-label-update-parameter

# Checklist

* [x] `task all` passes.


fcoury:
Ready for review.

rumyantseva (changes requested):
If I understand correctly, this task is not ready for review.

AlekSi (changes requested):

internal/handlers/pg/msg_findandmodify.go
```diff
@@ -381,7 +382,16 @@ func prepareFindAndModifyParams(document *types.Document) (*findAndModifyParams,
...
+		if slices.Contains(
+			[]string{
+				"$currentDate",
+				"$set",
+				"$setOnInsert",
+				"$unset",
+				"$inc",
+			},
+			k,
+		) {
```
AlekSi, line 394: @w84thesun That's the third place, in addition to `UpdateOperators` and `checkAllModifiersSupported`. I'm pretty sure we don't need it :)

integration/update_test.go
```diff
@@ -188,3 +188,56 @@ func TestMultiFlag(t *testing.T) {
...
 }
+
+func TestUpdateReplaceDocuments(t *testing.T) {
+	setup.SkipForTigris(t)
+
+	t.Parallel()
+
+	stat := bson.D{{"n", int32(1)}, {"nModified", int32(1)}, {"ok", float64(1)}}
+
+	for name, tc := range map[string]struct {
```
rumyantseva, line 199: If I'm not mistaken, in all those test cases we add a new key called `"replacement-value"`, but we don't really replace the current keys. Let's add a test where the key in the update doc is not `replacement-value` but something that already was in the initial doc. Also, let's add a case where the update doc provides an empty object for update.

rumyantseva (changes requested):
I think a couple of test cases are missing (see the comments in the review).

integration/update_test.go
w84thesun: That `expected` document compared against the document that was retrieved from DB.
In `Replace` we have `bson.D{{"_id", "bool-true"}, {"v", true}}` that replaced with `bson.D{{"_id", "bool-true"}, {"replacement-value", int32(1)}}`. In the `ReplaceWithDotNotation` the whole `document-composite` is replaced with `bson.D{{"_id", "document-composite"}, {"replacement-value", int32(1)}}`.

The thing about replacement objects is that it would replace the whole document except the `_id`.

So the first case you described would result in the same thing that we have now. The second one is good one. I have to debug the error I get on it.

internal/handlers/common/update.go
```diff
@@ -93,8 +90,33 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
+			setDoc := update
+
+			if setDoc.Len() == 0 {
+				continue
+			}
+
+			sort.Strings(setDoc.Keys())
+
+			for _, setKey := range doc.Keys() {
+				if !setDoc.Has(setKey) && !(setKey == "_id") {
```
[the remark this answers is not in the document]
w84thesun: No, rewrote condition.

internal/handlers/common/update.go
```diff
@@ -93,8 +90,33 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
-			panic(fmt.Errorf("unhandled operation %q", updateOp))
+			if strings.HasPrefix(updateOp, "$") {
+				return false, NewError(ErrNotImplemented, fmt.Errorf("UpdateDocument: unhandled operation %q", updateOp))
+			}
+
+			// Treats the update as a Replacement object.
+			setDoc := update
+
+			if setDoc.Len() == 0 {
+				continue
```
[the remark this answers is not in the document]
w84thesun: Removed that check. Empty doc would result in an error at the parsing stage.

integration/update_test.go
w84thesun: So, the first case is possible:
```
		"ReplaceDotNotationWithEmptyDoc": {
			update: bson.D{
				{"q", bson.D{{"v.array.0", bson.D{{"$eq", int32(42)}}}}},
				{"u", bson.D{{}}},
			},
			expectedFilter: bson.D{{"_id", "document-composite"}},
			expected:       bson.D{{"_id", "document-composite"}},
		},
```

But we can't add it right now, as the empty doc would result in parsing error.

w84thesun: I created the issue for that case → https://github.com/FerretDB/FerretDB/issues/1000

rumyantseva: What a number! `:envy:`

rumyantseva: @w84thesun hmm, I'd suggest keeping this test case commented in the current branch with TODO that refers to the ticket you created.

AlekSi:
@w84thesun tests, linters fail

integration/update_test.go
w84thesun: Fixed.

rumyantseva (approved):
LGTM

w84thesun (approved):

noisersup (approved):
LGTM

AlekSi (approved):

PR #796, opened 2022-06-28
[DRAFT] Fix whatsmyuri command

rumyantseva:
# Description

## Readiness checklist

### DIFF
--- a/internal/handlers/handlers.go
+++ b/internal/handlers/handlers.go
@@ -28,6 +28,7 @@ import (
 // CmdQuery handles a limited subset of OP_QUERY messages.
 //
 // Please keep methods documentation in sync with commands help text in the handlers/common package.
+// TODO: update docs
 type Interface interface {
 	// Close gracefully shutdowns handler.
 	Close()


internal/handlers/common/msg_whatsmyuri.go
```diff
@@ -0,0 +1,44 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"context"
+
+	"github.com/FerretDB/FerretDB/internal/clientconn/info"
```
rumyantseva, line 20: @AlekSi this is not according to the issue description, but it looked quite strange to import the whole `clientconn` to handlers, so I came up with a totally separate package. What do you think?

AlekSi: Sounds good

AlekSi: But let's rename it to `clientconninfo` or something, because it will be used as `info.Something`, and that's too generic

AlekSi (commented):
Let's merge #799 first

internal/clientconn/conninfo/conn_info.go
```diff
@@ -0,0 +1,52 @@
...
+// and can be shared through context.
+package conninfo
+
+import (
+	"context"
+	"net"
+)
+
+// contextKey is a special type to represent context.WithValue keys a bit more safely.
+type contextKey string
```
AlekSi, line 25:
```suggestion
type contextKey struct{}
```

That makes it even safer

internal/clientconn/conninfo/conn_info.go
```diff
@@ -0,0 +1,52 @@
...
+
+// connInfoKey stores the key for withConnInfo context value.
+const connInfoKey = contextKey("connInfo")
```
AlekSi, line 28:
```suggestion
var connInfoKey = contextKey{}
```

internal/clientconn/conninfo/conn_info.go
```diff
@@ -0,0 +1,52 @@
...
+// WithConnInfo returns a new context with the given ConnInfo.
+func WithConnInfo(ctx context.Context, connInfo *ConnInfo) context.Context {
+	return context.WithValue(ctx, connInfoKey, connInfo)
+}
+
+// GetConnInfo returns the ConnInfo value stored in ctx, or empty connInfo if there is nothing stored there.
+func GetConnInfo(ctx context.Context) *ConnInfo {
+	value := ctx.Value(connInfoKey)
+	if value == nil {
+		return &ConnInfo{}
```
AlekSi, line 44: I would say let's panic in that case as it should not happen

rumyantseva: I disagree but commit :)

integration/commands_administration_test.go
```diff
@@ -793,16 +794,31 @@ func TestCommandsAdministrationServerStatus(t *testing.T) {
...
+	var ports []uint16
+	for _, collection := range []*mongo.Collection{collection1, collection2} {
+		var actual bson.D
+		command := bson.D{{"whatsmyuri", int32(1)}}
+		err := collection.Database().RunCommand(ctx, command).Decode(&actual)
+		require.NoError(t, err)
+
+		doc := ConvertDocument(t, actual)
+		assert.Equal(t, float64(1), must.NotFail(doc.Get("ok")))
+		assert.Regexp(t, `^(\d+)\.(\d+)\.(\d+)\.(\d+)(\:\d+)?`, must.NotFail(doc.Get("you")))
```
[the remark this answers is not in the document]
rumyantseva: Fixed

internal/clientconn/conninfo/conn_info.go
AlekSi: My point is that the value will be absent only if we, developers, forget to add `WithConnInfo` to all places where it should be. It is not possible to trigger this error by any user action/input. Therefore, that's a programmer's error. Therefore, that should be a panic. (And that panic will not crash the whole program as we recover from them at the connection level, but it will be very noticeable for users who would create an issue or for us.)

rumyantseva: Yes, I understand that point. But then should we go deeper and check that the structure is not a zero value structure, the address is actually set and not nil etc?.. Probably, not (preventing missing context seems to be enough), but it's just a bit hard to differentiate one dev mistake from another.

internal/clientconn/conninfo/conn_info.go
```diff
@@ -0,0 +1,52 @@
...
+	value := ctx.Value(connInfoKey)
+	if value == nil {
+		panic("connInfo is not set in context")
+	}
+
+	connInfo, ok := value.(*ConnInfo)
+	if !ok {
+		panic("connInfo is set in context with wrong value type")
+	}
+	return connInfo
```
AlekSi, line 51: I think that's fine to check if `connInfo == nil` and panic in that case

seeforschauer (dismissed):

internal/clientconn/conninfo/conn_info.go
rumyantseva: @AlekSi should we also add some checks to `withConnInfo` or changes something to avoid `nil` there?

w84thesun (approved):
LGTM.

w84thesun (approved):

seeforschauer (approved):

AlekSi (approved):
:shipit:

PR #784, opened 2022-06-23
Add basic CI for Tigris

AlekSi:
Closes #782.

# Checklist

* [x] Tests are added for new functionality or fixed bugs.
* [x] `task all` passes.

See CONTRIBUTING.md for more details.

PR #799, opened 2022-06-29
Do not import `pg` handler explicitly

AlekSi:
To avoid conflicts between #784 and #796.

## Readiness checklist


rumyantseva (dismissed):
Yeah, makes sense!

integration/go.mod
```diff
@@ -13,14 +13,26 @@ require (
 )
 
 require (
```
AlekSi, line 15: Deps were messed up by the previous commit

AlekSi:
@rumyantseva PTAL

integration/go.mod
rumyantseva: Do you understand where are all the new indirect dependencies from?

AlekSi: From the Tigris handler. That's not a problem in integration tests

rumyantseva (approved):
LGTM

PR #805, opened 2022-06-30
Fix Docker image build

AlekSi:


seeforschauer:
How could I test this?

AlekSi:
By comparing https://github.com/FerretDB/FerretDB/runs/7126764945 with https://github.com/FerretDB/FerretDB/runs/7126900458 and optionally pulling and running that Docker image

Taskfile.yml
```diff
@@ -263,10 +263,10 @@ tasks:
         --tag=ferretdb-local
         --load .
     vars:
-      COMMIT:
-        sh: cat internal/util/version/commit.txt
       VERSION:
-        sh: cat internal/util/version/version.txt
+        sh: cat internal/util/version/gen/version.txt
```
seeforschauer, line 267: Why did you change the variables list order? 
If we'd taken it in alphabetical order, then it's better to change it back.

AlekSi: Now they are in the same order as they are used above, and how they are referenced at https://github.com/FerretDB/FerretDB/blob/main/internal/util/version/generate.go and https://github.com/FerretDB/FerretDB/blob/main/internal/util/version/version.go

seeforschauer: > Now they are in the same order as they are used above, and how they are referenced at https://github.com/FerretDB/FerretDB/blob/main/internal/util/version/generate.go and https://github.com/FerretDB/FerretDB/blob/main/internal/util/version/version.go

Well, those are async usage. 
That's unclear: in which cases we use alphabetical and in which they are in the order they are used.

AlekSi: Sorry, I'm not sure I understand you.

https://github.com/FerretDB/FerretDB/blob/db9698f91fcb606caf7a14a6b404a837c2493cb1/internal/util/version/version.go#L40-L44

The order is version, commit, branch.

https://github.com/FerretDB/FerretDB/blob/db9698f91fcb606caf7a14a6b404a837c2493cb1/internal/util/version/version.go#L80-L92

The order is version, commit, branch.

https://github.com/FerretDB/FerretDB/blob/db9698f91fcb606caf7a14a6b404a837c2493cb1/internal/util/version/version.go#L94-L97

The order is version, commit, branch.

https://github.com/FerretDB/FerretDB/blob/db9698f91fcb606caf7a14a6b404a837c2493cb1/Taskfile.yml#L261-L262

The order is version, commit.

https://github.com/FerretDB/FerretDB/blob/db9698f91fcb606caf7a14a6b404a837c2493cb1/Taskfile.yml#L277-L278

The order is version, commit.

The order is from the most important value to the least important. I'm not sure why it should be alphabetical

AlekSi:
Oops, missed!

Taskfile.yml
seeforschauer: > The order is version, commit.

Is it written somewhere?

> The order is from the most important value to the least important. I'm not sure why it should be alphabetical.

In some cases, it's alphabetical, i.e. for linters.

seeforschauer (approved):

PR #809, opened 2022-06-30
Unify `update` tests

### LINKED ISSUE #674: Unify `update` tests
We should unify our update tests to unblock the implementation of more update operators.

* [ ] `update_test.go` tests should be split into four files:
* `update_field_test.go` for https://www.mongodb.com/docs/manual/reference/operator/update-field/ operators;
* `update_array_test.go` for https://www.mongodb.com/docs/manual/reference/operator/update-array/ operators;
* `update_bitwise_test.go` the sole https://www.mongodb.com/docs/manual/reference/operator/update-bitwise/ operator;
* and `update_test.go` may remain for some basic tests.
* [ ] Test names should have a file name as a prefix. For example, `TestUpdateFieldSet`.
* [ ] Tests should use shared data. They may insert test-specifical data if needed, but that should be an exception.
* [ ] Tests should generally look similar to query tests and each other.

seeforschauer:
# Description

This PR closes #674.


integration/update_array_test.go
```diff
@@ -0,0 +1,31 @@
...
+// - `$`
+// - `$[]`
+// - `$[<identifier>]`
+// - `$addToSet`
+// - `$pop`
+// - `$pull`
+// - `$push`
+// - `$pullAll`.
+//
+// And array update operator modifiers:
```
AlekSi, line 27: There is a single group, not two groups

image

.github/workflows/conform-pr.yml
```diff
@@ -38,7 +38,7 @@ env:
   GOLANGCI_LINT_CACHE: /home/runner/go/cache/lint
   GOMODCACHE: /home/runner/go/mod
   GOPROXY: https://proxy.golang.org
-  CONFORM_TOKEN: ${{ secrets.CONFORM_TOKEN }}  # GITHUB_TOKEN is not enough to query projects, so we use a special one
+  CONFORM_TOKEN: ${{ secrets.CONFORM_TOKEN }} # GITHUB_TOKEN is not enough to query projects, so we use a special one
```
AlekSi, line 41: That's unrelated

integration/update_field_test.go
```diff
@@ -0,0 +1,929 @@
...
+// - `$inc`
+// - `$min`
+// - `$max`
+// - `$mul`
+// - `$rename`
+// - `$set`
+// - `$setOnInsert`
+// - `$unset`.
+
+func TestUpdateFieldCurrentDateTimestamp(t *testing.T) {
```
AlekSi, line 45: The test name should be `TestUpdateField<OperatorName>`. All variations of it (TestUpdateFieldCurrentDateTimestamp and TestUpdateFieldCurrentDate, TestUpdateFieldIncErrors and TestUpdateFieldInc below, etc) should be subtests

integration/update_field_test.go
```diff
@@ -0,0 +1,929 @@
...
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			ctx, collection := setup(t, shareddata.Composites)
+
+			_, err := collection.InsertMany(ctx, []any{
+				bson.D{{"_id", "double"}, {"value", 42.13}},
+				bson.D{{"_id", "double-nan"}, {"value", math.NaN()}},
+				bson.D{{"_id", "int32"}, {"value", int32(42)}},
+				bson.D{{"_id", "int64"}, {"value", int64(42)}},
+			})
```
AlekSi, line 486: We have value that are good enough in shareddata.Scalars, we should use them

AlekSi (changes requested):

integration/update_field_test.go
seeforschauer: > We have value that are good enough in shareddata.Scalars, we should use them

done.

integration/update_field_test.go
seeforschauer: > The test name should be `TestUpdateField<OperatorName>`. All variations of it (TestUpdateFieldCurrentDateTimestamp and TestUpdateFieldCurrentDate, TestUpdateFieldIncErrors and TestUpdateFieldInc below, etc) should be subtests

done.

integration/update_array_test.go
seeforschauer: > There is a single group, not two groups
> 
> image

done.

integration/update_field_test.go
AlekSi: May you please slow down and update other place too?

integration/update_field_test.go
w84thesun: @AlekSi should it be done for all other update tests too? Did I get it right that the expected result is that we have one test per iterator?

rumyantseva (approved):
To me, it looks like all the requirements written in the issue are met.

- Tests are moved to functions based on operations and operators
- Tests use shared data
- Names are correct
- Tests look similar to other tests

noisersup (approved):
All of the test cases moved properly.

integration/update_field_test.go
AlekSi: I think we should have one top-level test per specific operator. There can be a separate test for a mix of operators.

---

@seeforschauer 

> TestUpdateFieldCurrentDateTimestamp and TestUpdateFieldCurrentDate

Those tests were merged, great

> TestUpdateFieldIncErrors and TestUpdateFieldInc below

Those two were not merged, so that's not done

We also have tests that do not use shareddata.Scalars.

AlekSi (changes requested):
https://github.com/FerretDB/FerretDB/pull/809#discussion_r911854476

integration/update_field_test.go
seeforschauer: > We also have tests that do not use shareddata.Scalars. 

Could you please clarify, 
`shareddata.Scalars` must be removed from tests that do not require them?

integration/update_field_test.go
```diff
@@ -0,0 +1,913 @@
...
+				},
+			},
+		} {
+			name, tc := name, tc
+			t.Run(name, func(t *testing.T) {
+				t.Parallel()
+				ctx, collection := setup(t, shareddata.Composites)
+				_, err := collection.InsertMany(ctx, []any{
+					bson.D{{"_id", "string-inc-error"}, {"value", "foo"}},
+				})
```
AlekSi, line 483: This test does not use shareddata.Scalars, but inserts own data – that is not that different from data in our shared set.
Instead, this test should use shareddata.Scalars and not insert own data.

Please check other tests too

w84thesun (approved):

noisersup (approved):

AlekSi (approved):

PR #812, opened 2022-06-30
Listen on a UNIX domain socket

### LINKED ISSUE #737: Listen on a Unix domain socket
There should be a way to configure FerretDB to listen and accept connections on the UNIX domain socket.

Let's add a flag `--listen-socket` that would contain a single file path to a UNIX socket FerretDB will listen on. Empty value (the default) will disable it. Both `--listen-addr` and `--listen-socket` could be given; in that case, FerretDB would listen on both.

* Integration test is in the scope.
* Embedded usage (public `ferretdb` package) is in the scope.

w84thesun:
# Description

This PR closes #737.

## Readiness checklist


AlekSi:
Too many things changed in the meantime.

PR #814, opened 2022-06-30
Use `github.head_ref` instead of `github.ref`

AlekSi:
The latter is the base reference (in most cases, `main` branch) for `pull_request_target`.
It canceled this workflow for other PRs.


rumyantseva (approved):
LGTM

PR #817, opened 2022-07-01
Bump deps

AlekSi:


w84thesun (approved):

rumyantseva (approved):
`gopkg.in/yaml.v3 v3.0.1` seems to be quite a major migration - from v2 to v3. But I guess as the tests pass (and we don't use this package directly, it's a transitive dependency), that should be fine.

AlekSi:
> `gopkg.in/yaml.v3 v3.0.1` seems to be quite a major migration - from v2 to v3

There is not such migration: yaml.v3 is used by testify and yaml.v2 is used by zap. As far as Go modules system is concerned, they are completely separate modules

PR #819, opened 2022-07-01
Support `gt` `lt` operator comparison for Array type

ribaraka:
refs #457.

Array comparison in some cases returns multiple values, so now `Compare` func return slice of results.

## Readiness checklist


AlekSi:
@ribaraka Please check linter's warnings and merge with main

ribaraka:
@AlekSi, done! Please review.

internal/types/compare.go
```diff
@@ -45,7 +46,7 @@ const (
 // For that reason, it typically should not be used in tests.
 //
 // Compare and contrast with test helpers in testutil package.
-func Compare(docValue, filterValue any) CompareResult {
+func Compare(docValue, filterValue any) []CompareResult {
```
AlekSi, line 49: I can't wrap my head about that change. There has to be a better way.

PR #839, opened 2022-07-04
Remove old style of `+build` tags where possible

AlekSi:


w84thesun (approved):

tools/old.go
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Keep both old and new styles of build tags.
```
rumyantseva, line 15: That would be great to explain why!

AlekSi: There is an explanation below :)

rumyantseva (approved):
LGTM

rumyantseva (approved):
LGTM

Issue #782, opened 2022-06-23
Basic Tigris development environment


AlekSi:
* [ ] ~Add a separate `task` target to start Tigris, or parametrize `env-up`.~
* [x] Add a very basic integrational smoke test.
* [x] Add CI configuration.
* [x] Update CONTRIBUTING.md.

PR #842, opened 2022-07-04
Create Tigris databases

AlekSi:
Refs #782.

### DIFF
--- a/cmd/envtool/main.go
+++ b/cmd/envtool/main.go
@@ -27,6 +27,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/tigrisdata/tigris-client-go/config"
+	"github.com/tigrisdata/tigris-client-go/driver"
 	"go.uber.org/zap"
 
 	"github.com/FerretDB/FerretDB/internal/handlers/pg/pgdb"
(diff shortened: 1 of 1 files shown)


cmd/envtool/main.go
```diff
@@ -216,14 +190,72 @@ func run(ctx context.Context, logger *zap.SugaredLogger) error {
...
+func run(ctx context.Context, logger *zap.SugaredLogger) error {
+	go debug.RunHandler(ctx, "127.0.0.1:8089", logger.Named("debug").Desugar())
+
+	if err := setupPostgres(ctx, logger); err != nil {
+		return err
+	}
+
+	if err := setupTigris(ctx, logger); err != nil {
+		return err
+	}
```
rumyantseva, line 233: Are those commands fast? Should we run them in goroutines with waitgroup to make startup a bit faster?

cmd/envtool/main.go
```diff
@@ -216,14 +190,72 @@ func run(ctx context.Context, logger *zap.SugaredLogger) error {
...
+// setupTigris configures Tigris.
+func setupTigris(ctx context.Context, logger *zap.SugaredLogger) error {
+	logger = logger.Named("tigris")
+
+	if err := waitForPort(ctx, logger, 8081); err != nil {
+		return err
+	}
+
+	cfg := &config.Driver{
+		URL: "127.0.0.1:8081",
```
rumyantseva, line 205: Is this port `8081` a Tigris default port (like `5432` for Postgres)? The number `8081` is quite a common number for some sample apps, and I wonder if it's a good choice to take it.

rumyantseva (approved):
I have a couple of minor (non-blocking) concerns, see the comments.
In general, LGTM.

cmd/envtool/main.go
AlekSi: It is, yeah:
* https://docs.tigrisdata.com/quickstart#startup-tigris-on-your-local-machine
* https://github.com/tigrisdata/tigris#building-with-docker-containers

We could map that port to some other port if there is a conflict, but let's wait for the conflict to happen

cmd/envtool/main.go
AlekSi: The waitgroup was there before, with a bug :-) https://github.com/FerretDB/FerretDB/pull/788#pullrequestreview-1017562742

Right now we do very little in those functions, they are fast

rumyantseva: 😅

PR #844, opened 2022-07-05
Validate collection names

### LINKED ISSUE #806: Validate collection names
After #750, we should validate that MongoDB collection names are valid.

MongoDB rules are documented here: https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Restriction-on-Collection-Names

Let's simplify them to two rules:
* collection names should match regular expression `^[a-zA-Z_][a-zA-Z0-9_]{0,119}$` (that excludes non-latin letters, spaces, dots, dollars, dashes, long names, etc);
* collection names can't start with a reserved prefix `_ferretdb_`.

We could relax those rules in the future.

* [x] implement collection name validation on collection/table creation;
* [x] add integration test;
* [x] document the difference in rules in README.md;
* [ ] add diff dance test.

Tigris is not in scope.

seeforschauer:
# Description

This PR closes #806.

### DIFF
--- a/internal/handlers/pg/pgdb/pool.go
+++ b/internal/handlers/pg/pgdb/pool.go
@@ -343,10 +343,19 @@ func (pgPool *Pool) DropDatabase(ctx context.Context, db string) error {
 	}
 }
 
+// validateCollectionName validate collection name.
+func (pgPool *Pool) validateCollectionName(collection string) error {
+	return nil
+}
+
 // CreateCollection creates a new FerretDB collection in existing schema.
 //
 // It returns ErrAlreadyExist if table already exist, ErrTableNotExist is schema does not exist.
 func (pgPool *Pool) CreateCollection(ctx context.Context, db, collection string) error {
+	if err := pgPool.validateCollectionName(collection); err != nil {
+		return err
+	}
+
 	schemaExists, err := pgPool.schemaExists(ctx, db)
 	if err != nil {
 		return lazyerrors.Error(err)


internal/handlers/pg/pgdb/pool.go
```diff
@@ -343,10 +347,38 @@ func (pgPool *Pool) DropDatabase(ctx context.Context, db string) error {
...
+// Collection name must not contain non-latin letters, spaces, dots, dollars, dashes,
+// name collection name length must be less than 119 characters.
+func (pgPool *Pool) validateCollectionName(collection string) error {
+	if len(collection) == 0 {
+		return ErrInvalidTableName
+	}
+
+	if len(collection) > 119 {
+		return ErrInvalidTableName
+	}
```
w84thesun, line 360: I think those are redundant checks, as they are included in regex. Therefore they can be removed.

w84thesun (changes requested):
LGTM, except collection length checks.

integration/basic_test.go
```diff
@@ -128,22 +128,74 @@ func TestFindCommentQuery(t *testing.T) {
...
+
+		cases := map[string]struct {
+			collection string
+			err        *mongo.CommandError
+			alt        string
+		}{
+			"TooLongForBoth": {
+				collection: "very_long_collection_name_that_fails_both_in_mongo_and_in_ferretdb_databases" +
+					"_for_ferretdb_it_fails_because_it_is_more_than_119_characters__for_mongo_it_fails_because_it_is_more_than_255_charachters_" +
+					"long_that_excludes_non_latin_letters_spaces_dots_dollars_dashes",
```
AlekSi, line 145: Unless we really want people to actually read it, it could be replaced with `const collectionName300 = strings.Repeat("a", 300)` and referring to it both there and below, in the error message

The same in other places

integration/basic_test.go
```diff
@@ -128,22 +128,74 @@ func TestFindCommentQuery(t *testing.T) {
...
+				alt: "Collection must not contain non-latin letters, spaces, dots, dollars, dashes and be longer than 119 characters.",
+			},
+			"WithADollarSign": {
+				collection: "collection_name_with_a-$",
+				err: &mongo.CommandError{
+					Name:    "InvalidNamespace",
+					Code:    73,
+					Message: `Invalid collection name: collection_name_with_a-$`,
+				},
+				alt: "Collection must not contain non-latin letters, spaces, dots, dollars, dashes and be longer than 119 characters.",
```
AlekSi, line 162: Can we return the same error message there and in other places?

integration/basic_test.go
```diff
@@ -128,22 +128,74 @@ func TestFindCommentQuery(t *testing.T) {
...
+			},
+		}
+
+		for name, tc := range cases {
+			name, tc := name, tc
+			t.Run(name, func(t *testing.T) {
+				err := collection.Database().CreateCollection(ctx, tc.collection)
+				assert.NotNil(t, tc.err)
```
AlekSi, line 179: Not required with the helper below

internal/handlers/pg/pgdb/pool.go
```diff
@@ -343,10 +347,38 @@ func (pgPool *Pool) DropDatabase(ctx context.Context, db string) error {
 	}
 }
 
+// validateCollectionName validates collection name.
+// Collection name must not contain non-latin letters, spaces, dots, dollars, dashes,
+// name collection name length must be less than 119 characters.
+func (pgPool *Pool) validateCollectionName(collection string) error {
```
AlekSi, line 353: Let's inline this method into the only place where we use it

internal/handlers/pg/pgdb/pool.go
```diff
@@ -343,10 +347,38 @@ func (pgPool *Pool) DropDatabase(ctx context.Context, db string) error {
...
+	if len(collection) == 0 {
+		return ErrInvalidTableName
+	}
+
+	if len(collection) > 119 {
+		return ErrInvalidTableName
+	}
+
+	regex := regexp.MustCompile("^[a-zA-Z_][a-zA-Z0-9_]{0,119}$")
```
AlekSi, line 362: We could compile it only once

internal/handlers/pg/pgdb/pool.go
```diff
@@ -343,10 +347,38 @@ func (pgPool *Pool) DropDatabase(ctx context.Context, db string) error {
...
+	if strings.HasPrefix(collection, collectionPrefix) {
+		return ErrInvalidTableName
+	}
+
+	return nil
+}
+
 // CreateCollection creates a new FerretDB collection in existing schema.
 //
 // It returns ErrAlreadyExist if table already exist, ErrTableNotExist is schema does not exist.
```
AlekSi, line 376: It can return other errors too now, we should document that

integration/basic_test.go
seeforschauer: > Can we return the same error message there and in other places?

there -> where?
other places - > which places do you mean?

integration/basic_test.go
seeforschauer: > Not required with the helper below

Sorry I didn't understand you. WDYM?

I meant by this line the following:
This:
```go
assert.NotNil(t, tc.err)
```
Means checking that we didn't forget to add an error in the test cases for errors.

integration/basic_test.go
AlekSi: As we discussed previously, we should return the same error message as MongoDB where we can. I this test case MongoDB returns `Invalid collection name: <collection name>`. That's a good correct error message. It should not be hard to make FerretDB return the same error message. So let's do that here and in other places in this PR.

integration/basic_test.go
AlekSi: That code does not stop the execution of the test. If `tc.err` is nil, then `*tc.err` below will panic

integration/basic_test.go
seeforschauer: > Unless we really want people to actually read it, it could be replaced with `const collectionName300 = strings.Repeat("a", 300)` and referring to it both there and below, in the error message
> 
> The same in other places

Changed here and in other places for collection name tests.

internal/handlers/pg/pgdb/pool.go
seeforschauer: > Let's inline this method into the only place where we use it

removed.

internal/handlers/pg/pgdb/pool.go
seeforschauer: > We could compile it only once

moved to package level private variables.

internal/handlers/pg/pgdb/pool.go
seeforschauer: > It can return other errors too now, we should document that

added to doc and checked how the doc looks like.

integration/basic_test.go
seeforschauer: > > That code does not stop the execution of the test. If `tc.err` is nil, then `*tc.err` below will panic
> 
> Thank you. Changed to `require`.

removed it all in all.

w84thesun (dismissed):

README.md
```diff
@@ -33,6 +33,12 @@ and [contributing guidelines](CONTRIBUTING.md).
 
 1. FerretDB uses the same protocol error names and codes, but the exact error messages may be different in some cases.
 
+2. Collection names restrictions:
+
+* collection name cannot start with the reserved prefix `_ferretdb_`.
+* collection name must not include non-latin letters, spaces, dots, dollars or dashes.
+* collection name length must be less than 119 symbols.
+
```
rumyantseva, line 41: There is one more condition - collection name must not start with a number.

internal/handlers/pg/pgdb/pool.go
```diff
@@ -345,8 +352,19 @@ func (pgPool *Pool) DropDatabase(ctx context.Context, db string) error {
...
+// It returns:
+//  - ErrInvalidTableName - if the table name doesn't conform to restrictions.
+//  - ErrAlreadyExist - if table already exist.
+//  - ErrTableNotExist - is schema does not exist.
 func (pgPool *Pool) CreateCollection(ctx context.Context, db, collection string) error {
+	if !validateCollectionNameRe.MatchString(collection) {
+		return ErrInvalidTableName
+	}
+
+	if strings.HasPrefix(collection, collectionPrefix) {
```
rumyantseva, line 364: With that, I wonder if the `collectionPrefix` constant should be called something like `reservedCollectionPrefix`. But I don't insist :)

seeforschauer:
@rumyantseva 

> 1. In the PR body there is a case `max length in FerretDB: 255, in MongoDB: 63.`.

Please pay attention to the caption - it's a difference between FerretDB and MongoDB. 
I listed these differences in the PR body to remember what cases to cover, document, and mention in the dance tests. 

>  Should it be `max length in FerretDB: 119`? or what 119 is about? :)

Not sure about 119, it's a magic number for now. And yes, this magic number is not in the constant somewhere, I follow the task guideline that we can change it later.  

> 2. Are dance cases in the scope of this task? Or is there a separate issue for it?

It is in the scope of the task and it is not in the scope of the PR since the repository is FerretDB. 

> 3. It's great that there is a new piece of doc for README. It's not 100% full/accurate (I added a couple of comments about it).

Added, thank you for noting this.

> 4. Special thanks for updating `CreateCollection` godoc.

You are welcome ^ - ^. I'm glad I have a command to check what docs look like (thanks to @AlekSi).

README.md
```diff
@@ -33,6 +33,12 @@ and [contributing guidelines](CONTRIBUTING.md).
 
 1. FerretDB uses the same protocol error names and codes, but the exact error messages may be different in some cases.
 
+2. Collection names restrictions:
+
+* collection name cannot start with the reserved prefix `_ferretdb_`.
+* collection name must not include non-latin letters, spaces, dots, dollars or dashes.
+* collection name length must be less than 119 symbols.
```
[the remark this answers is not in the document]
seeforschauer: > "Less" or "less or equal"? The regex is about "less or equal".

changed.

rumyantseva (approved):
LGTM

w84thesun (dismissed):

internal/handlers/pg/pgdb/pool.go
```diff
@@ -345,8 +352,19 @@ func (pgPool *Pool) DropDatabase(ctx context.Context, db string) error {
 
 // CreateCollection creates a new FerretDB collection in existing schema.
 //
-// It returns ErrAlreadyExist if table already exist, ErrTableNotExist is schema does not exist.
+// It returns:
+//  - ErrInvalidTableName - if the table name doesn't conform to restrictions.
```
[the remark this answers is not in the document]
AlekSi: https://github.com/golang/proposal/blob/master/design/51082-godocfmt.md#new-syntax-for-lists

> To satisfy both, all three bullets and both forms of numbers will be recognized, but gofmt (see below) will rewrite them to a single canonical form: dash for bullets, and “N.” for numbers.

In short, it does not matter now, but the next version of Go will reformat all comments to use dashes

README.md
```diff
@@ -33,6 +33,13 @@ and [contributing guidelines](CONTRIBUTING.md).
 
 1. FerretDB uses the same protocol error names and codes, but the exact error messages may be different in some cases.
 
+2. Collection names restrictions:
+
+* collection name cannot start with the reserved prefix `_ferretdb_`.
+* collection name must not include non-latin letters, spaces, dots, dollars or dashes.
+* collection name length must be less or equal than 119 symbols.
```
AlekSi, line 40: 120, actually. The length should be in the range [1-120] inclusively

integration/basic_test.go
```diff
@@ -128,22 +129,73 @@ func TestFindCommentQuery(t *testing.T) {
...
+
+	t.Run("Ok", func(t *testing.T) {
+		ctx, collection := Setup(t)
+
+		longCollectionName := strings.Repeat("a", 65)
+		err := collection.Database().CreateCollection(ctx, longCollectionName)
+		require.NoError(t, err)
+		sixtyThreeCharsCollectionName := strings.Repeat("a", 63)
+		err = collection.Database().CreateCollection(ctx, sixtyThreeCharsCollectionName)
+		require.NoError(t, err)
```
AlekSi, line 193: 65 and 63 look special, but it is not clear why, and both work. Do we need both? Could we use a simpler number like 100 that is also not special?

integration/basic_test.go
AlekSi: Let's keep sentinel error `ErrInvalidTableName` as is for now then, and not add a new type

integration/basic_test.go
seeforschauer: > 65 and 63 look special, but it is not clear why, and both work. Do we need both? Could we use a simpler number like 100 that is also not special?

I've left one only.

README.md
seeforschauer: > 120, actually. The length should be in the range [1-120] inclusively

gotcha. I thought what a strange number 119. fixed.

w84thesun (approved):

AlekSi (approved):

PR #850, opened 2022-07-05
Initial Aggregate Pipeline Support

fcoury:
# Introduction

As the result of a personal exercise for the past week and a half, this PR offers an early implementation of the Aggregate Pipeline.

The reason the PR was submitted (as draft) now is I saw the team is starting to tackle the problem of always retrieving all the documents from a collection. As a way to contribute to the discussion, this pull request was opened as a proposal since there is a great opportunity for code reuse, specially from the `$match` pipeline.

This was initially made as a learning opportunity, but it may be at a point where collecting feedback might be useful.

## Description

This branch goes through the process of parsing an aggregate pipeline document for pipeline stages. It currently supports `$match`, `$group`, `$count` and `$sort`, along with the `$project` one reusing existing code.

Each stage is then transformed into a `Stage` struct that contains information about its fields, filters, groups and sort orders. 

Once we have all the stages parsed, we go through a final _wrapping stage_, which basically goes through each parsed stage, transforming each resulting struct into a SQL statement that is then nested on the order it appears on the pipeline.

## Test coverage

It's important to note that there is a rather good coverage of all the implemented code: 

- integration tests are on the `aggregate_test.go` file
- unit tests for each stage are on the `aggregate/*_test.go` files

## High level details about the implemetation

This is the Stage struct:

```go
type Stage struct {
	name       string
	fields     []StageField
	groups     []string
	sortFields []StageSortField
	root       *FilterNode
}
```

The filters are kept as a tree of FilterNodes, in a way that a JSON document like:

```js
db.sample.aggregate([{ $match: { $or: [{ a: { $gt: 1 } }, { b: 2 }] } }])
```

It is roughly translated into nodes like:

```
FilterNode {
    op: "OR"
    children: [
        FilterNode {
            index: 1
            field: 'a'
            op: '>'
            value: int32(1)
        }
        FilterNode {
            index: 2
            field: 'b'
            op: '='
            value: int32(2)
        }
    ]    
}
```

And the generated SQL is:

```sql
SELECT *
FROM "test"."sample"
WHERE (((CASE
             WHEN (_jsonb->'a' ? '$f') THEN (_jsonb->'a'->>'$f')::numeric
             ELSE (_jsonb->'a')::numeric
         END) > $1
        OR (CASE
                WHEN (_jsonb->'b' ? '$f') THEN (_jsonb->'b'->>'$f')::numeric
                ELSE (_jsonb->'b')::numeric
            END) = $2)) [1 2]
```

Please note the complex `CASE` expression that is used to handle numeric floats, since we don't have the knowledge of the queried field types beforehand.

A similar process happens for the other stages, the `$group` stage for instance, only accumulates a list of filters that are then used for the `SELECT` fields and `GROUP BY` clauses. Here's one example:

For the collection:

```js
db.sample.insertMany([{item: 1, qty: 10}, {item: 2, qty: 20}, {item: 1, qty: 5}])
```

And the following group pipeline:

```js
test> db.sample.aggregate([{$group: { _id: '$item', sum: { $sum: '$qty' }}}])
[ { _id: 2, sum: 20 }, { _id: 1, sum: 15 } ]
```

The following SQL is created:

```sql
SELECT 
  json_build_object('$k', jsonb_build_array('_id', 'sum'), '_id', _id, 'sum', json_build_object('$f', sum)) AS _jsonb
FROM
  (SELECT _jsonb->'item' AS _id,
                  SUM((CASE
                           WHEN (_jsonb->'qty' ? '$f') THEN (_jsonb->'qty'->>'$f')::numeric
                           ELSE (_jsonb->'qty')::numeric
                       END)) AS sum
   FROM "test"."sample"
   GROUP BY _id) AS wrapped
```

You can also combine multiple pipelines as expected, like `$match` and `$group`:

```js
test> db.sample.aggregate([{$match: {item: 1}, $group: { _id: '$item', sum: { $sum: '$qty' }}}])
[ { _id: 1, sum: 15 } ]
```

And the generated SQL:

```sql
SELECT 
  json_build_object('$k', jsonb_build_array('_id', 'sum'), '_id', _id, 'sum', json_build_object('$f', sum)) AS _jsonb
FROM
  (SELECT _jsonb->'item' AS _id,
                  SUM((CASE
                           WHEN (_jsonb->'qty' ? '$f') THEN (_jsonb->'qty'->>'$f')::numeric
                           ELSE (_jsonb->'qty')::numeric
                       END)) AS sum
   FROM
     (SELECT *
      FROM "test"."sample"
      WHERE ((CASE
                  WHEN (_jsonb->'item' ? '$f') THEN (_jsonb->'item'->>'$f')::numeric
                  ELSE (_jsonb->'item')::numeric
              END) = $1)) AS query1
   GROUP BY _id) AS wrapped
```

Futhermore, after all the pipelines are "wrapped" and the resulting documents retrieved, we run them through the existing projection pipeline if a `$project` stage is provided.

## Readiness checklist


rumyantseva (changes requested):
Thank you for sending this PR.
I hope we will be able to come back to this implementation later.
I "reject" it for now to at least mark that it needs an update from the main branch.

AlekSi (commented):
On hold for now

w84thesun (commented):
I like general approach, however in order to provide complete review I need to dig more into aggregation pipelines. With the #1581 generation of the queries could be simplified.

fredericalix:
Hi super team !
When do you plan to add $match, $group, $count and $sort ?
With these, we'll can use last Meteor.JS release and stop to use mongodb :)

AlekSi:
Hi @fredericalix,

We are working on those stages, see:
* #1417
* #1423
* #1439
* #1894

All of them should be implemented in version 1.0.0 GA, which is planned for this month.

---

It is probably time to close this PR. While comparing and filtering BSON values completely with SQL is very tempting, unfortunately, it is very hard to make it work the same way as in MongoDB. See [this release note](https://github.com/FerretDB/FerretDB/releases/tag/v0.1.0) for an explanation. We are working on blog posts exploring this topic in more detail.

@fcoury, thank you for your time and for your previous contributions to FerretDB.

fredericalix:
So good news !!!!
Thx a lot ! 

> Hi @fredericalix,
> 
> We are working on those stages, see:
> 
> * [Implement `$count` aggregation pipeline stage #1417](https://github.com/FerretDB/FerretDB/issues/1417)
> * [Implement `$group` aggregation pipeline stage #1423](https://github.com/FerretDB/FerretDB/issues/1423)
> * [Implement `$sort` aggregation pipeline stage #1439](https://github.com/FerretDB/FerretDB/issues/1439)
> * [Implement `$match` aggregation pipeline stage #1428](https://github.com/FerretDB/FerretDB/issues/1428)
> * [Support query pushdown for `$match` aggregation stage #1894](https://github.com/FerretDB/FerretDB/issues/1894)
> 
> All of them should be implemented in version 1.0.0 GA, which is planned for this month.
>

PR #856, opened 2022-07-06
Enable `wsl` linter for new and changed code

AlekSi:
# Description

`fuzztool` was updated as an example.

Can't say I like this style very much, but hey, if it's going to remove at least some judgment about coding style, that's worth it.

## Readiness checklist


rumyantseva (approved):
LGTM.

Should we mention something special about `wsl` in the contributing docs? Or do we consider it as any other linter?

AlekSi:
I don't think it's worth mentioning; `task lint` (and `task all`) should pass anyway

PR #873, opened 2022-07-12
Support auth and user commands

w84thesun:
# Description

This PR is for the analysis task.

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -39,11 +39,11 @@ Update connectionStatus to include authInfo.
 
 ## Commands that require authentication
 
-| #   | command          | Description                                 |
-|-----|------------------|---------------------------------------------|
-| 1   | listCommands     | Returns a list of commands.                 |
-| 2   | connectionStatus | Returns connection status.                  |
-| 3   | adminCommand     | `getParameter` response should be extended. |
+| #   | command          | Description                                           |
+|-----|------------------|-------------------------------------------------------|
+| 1   | listCommands     | Returns a list of commands.                           |
+| 2   | connectionStatus | Returns connection status.                            |
+| 3   | getParameter     | Response should be extended with authentication data. |
 
 
 # Tasks
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -43,7 +43,7 @@ Update connectionStatus to include authInfo.
 |-----|------------------|---------------------------------------------|
 | 1   | listCommands     | Returns a list of commands.                 |
 | 2   | connectionStatus | Returns connection status.                  |
-| 3   | adminCommand     | `getParameter` response whould be extended. |
+| 3   | adminCommand     | `getParameter` response should be extended. |
 
 
 # Tasks
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -7,9 +7,9 @@ Update connectionStatus to include authInfo.
 
 ## Auth Commands
 
-| #   | command      | Description                  |
-|-----|--------------|------------------------------|
-| 1   | authenticate | Authenticate with PostgreSQL |
+| #   | command      | Description                        |
+|-----|--------------|------------------------------------|
+| 1   | authenticate | Authenticate with x.509 mechanism. |
 
 
 ## Users Management Commands
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -15,6 +15,7 @@ Update connectionStatus to include authInfo.
 [MongoDB auth docs](https://github.com/mongodb/mongo/blob/a06bc8bbced8f0c60b94ed784f5f105f2f01ed5d/src/mongo/db/auth/README.md)
 
 [Authentication reference](https://www.mongodb.com/docs/manual/core/authentication/)
+
 [SCRAM authentication](https://www.mongodb.com/docs/manual/core/security-scram/)
 
 ## Auth Commands
@@ -33,10 +34,8 @@ Update connectionStatus to include authInfo.
 | 1   | createUser               | Creates user.                          |
 | 2   | dropAllUsersFromDatabase | Deletes all users from a database.     |
 | 3   | dropUser                 | Remove a single user from database.    |
-| 4   | grantRolesToUser         | Grants roles and privileges to a user. |
-| 5   | revokeRolesFromUser      | Removes roles from a user.             |
-| 6   | updateUser               | Updates a user's data.                 |
-| 7   | usersInfo                | Returns information about the users.   |
+| 4   | updateUser               | Updates a user's data.                 |
+| 5   | usersInfo                | Returns information about the users.   |
 
 ## Commands that require authentication
 
@@ -50,7 +49,7 @@ Update connectionStatus to include authInfo.
 ## Add support for separate connection pools for each user 
 
 *We will try to not store the username and password*.
-We should try to use PostreSQL's authentication mechanism.
+We should try to use PostgreSQL's authentication mechanism.
 
 When a client connects with credentials specified in the connection string (username and password), authentication message would be sent to the server (`saslStart`).
 In order to create a new connection pool, we should use the username from the authentication message.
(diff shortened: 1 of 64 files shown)


docs/auth.md
```diff
@@ -0,0 +1,111 @@
...
+
+[Authentication reference](https://www.mongodb.com/docs/manual/core/authentication/)
+[SCRAM authentication](https://www.mongodb.com/docs/manual/core/security-scram/)
+
+## Auth Commands
+
+| #   | command      | Description                        |
+|-----|--------------|------------------------------------|
+| 1   | authenticate | Authenticate with x.509 mechanism. |
+| 2   | saslStart    | Start SASL authentication.         |
```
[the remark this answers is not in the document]
w84thesun: Yep, nice catch, thanks!

docs/auth.md
```diff
@@ -0,0 +1,111 @@
...
+We should try to use PostreSQL's authentication mechanism.
+
+When a client connects with credentials specified in the connection string (username and password), authentication message would be sent to the server (`saslStart`).
+In order to create a new connection pool, we should use the username from the authentication message.
+
+When a client disconnects, we should close the connection pool.
+We should also close the system pool and all user pools when the server is shut down.
+
+To separate connection pools we should:
+* Use current connection pool as the `system` pool (we will use it later to query the database for user's credentials).
```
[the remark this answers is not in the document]
w84thesun: The `system` here is questionable. I believe we might want to call it `ferretdb` or something like that and then deprecate that name in the `Differences` section of README.md.

noisersup (approved):

docs/auth.md
```diff
@@ -0,0 +1,112 @@
...
+* Handle `saslStart` message.
+* To distinguish between users we should add to the Handler instance a map of connection pools, where the key is the username.
+
+### Questions:
+* How to distinguish between users connected without authentication?
```
seeforschauer, line 67: Will it be possible still to connect without authentication? I thought it was a system-wide requirement setting to authenticate.

docs/auth.md
```diff
@@ -0,0 +1,112 @@
...
+**Note:** This one should be discussed with the team.
+
+Add flag `authenticationMechanisms` to the FerretDB server.
+It should be an array of strings, containing the names of the supported authentication mechanisms.
+Possible values are:
+* SCRAM-SHA-1 (*not sure about this one as PostgreSQL doesn't support it*)
+* SCRAM-SHA-256
+* X509
+* GSSAPI (Kerberos)
+* PLAIN (LDAP SASL)
```
seeforschauer, line 110: As we agreed on call, not all of them will be used. 
Could you please add here "why" and the list of mechanics left?

docs/auth.md
```diff
@@ -0,0 +1,112 @@
...
+* Handle `saslStart` message.
+* To distinguish between users we should add to the Handler instance a map of connection pools, where the key is the username.
+
+### Questions:
+* How to distinguish between users connected without authentication?
+* Should we use a different connection pool for each user (for example for each host+port pair)?
+
+### Tests:
+* Connect to the server with username and password in connection string twice.
+* Connect to the server with different usernames and passwords.
```
seeforschauer, line 72: Then as we saw above, also try to connect w/o authentication. It must not let the user in.

docs/auth.md
```diff
@@ -0,0 +1,112 @@
+# Task description
+
+* Instead of using one global PostgreSQL connection pool, separate pools should be shared by all client connections with the same username and password.
+
+* FerretDB should handle the client's authentication commands and use provided credentials to authenticate in PostgreSQL.
+
+* Implement createUser, dropUser, etc. commands.
```
seeforschauer, line 7: Would you please list all commands here?
For each command, let's define what parameter will be implemented and what is not.

docs/auth.md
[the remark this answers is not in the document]
seeforschauer: I believe here we have to choose:
- `system` is more general for naming
- `ferretdb` for settings if more Postgres - alike (looking at their tables starting with `pg_`

As we would have more backends in the future, I vote for `system`, and if there is something specific then think on naming.

docs/auth.md
```diff
@@ -0,0 +1,112 @@
...
+Create a PostgreSQL user with provided credentials and auth mechanisms.
+Grant user access for all databases.
+
+### Tests:
+
+* Create user with username and password.
+* Create user with username and password and mechanisms.
+* Create user with username and password and mechanism that not supported or not exists.
+
+## Add flag `authenticationMechanisms`
```
seeforschauer, line 99: What about the authentification configuration for FerretDB? 
Some parameters MongoDB stores in `/etc/mongodb.conf` and they are manageable by the user. 
What about them?

docs/auth.md
```diff
@@ -0,0 +1,112 @@
...
+
```
seeforschauer, line 100: Let's also list all parameters regarding the authentication that must appear in the response of `db.adminCommand( { getParameter : '*' } )` .

AlekSi (commented):
Please re-request review once updated

docs/auth.md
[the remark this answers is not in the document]
w84thesun: I removed that pool as we won't need it at all.

docs/auth.md
w84thesun: No, it would not be possible to connect without authentication if it's enabled in PostgreSQL.

docs/auth.md
```diff
@@ -0,0 +1,112 @@
...
+|-----|------------------|-----------------------------|
+| 1   | listCommands     | Returns a list of commands. |
+| 2   | connectionStatus | Returns connection status.  |
+
+
+# Tasks
+## Add support for separate connection pools for each user 
+
+*We will try to not store the username and password*.
+We should try to use PostreSQL's authentication mechanism.
```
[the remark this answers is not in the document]
w84thesun: Added comment about Tigris

docs/auth.md
w84thesun: Mentioned it in the list of commands that we would update once authentication is done.

docs/auth.md
w84thesun: There would be no configuration on the FerretDB side for now. All authentication configurations would be in PostgreSQL.

docs/auth.md
w84thesun: Added that one.

docs/auth.md
w84thesun: Moved it above and added comments.

docs/auth.md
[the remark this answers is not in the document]
seeforschauer: > Added comment about Tigris

So Tigris uses a token (already), can we use it?

docs/auth.md
seeforschauer: > No, it would not be possible to connect without authentication if it's enabled in PostgreSQL.

So if `trusted` is enabled on Postgres, and auth enabled, then the user cannon login?

docs/auth.md
seeforschauer: > There would be no configuration on the FerretDB side for now. All authentication configurations would be in PostgreSQL.

Could you please add it to the doc? (Sorry I can't find this in the profile).

docs/auth.md
seeforschauer: > Mentioned it in the list of commands that we would update once authentication is done.

Didn't find it in the doc. Can you please check it or point me out to this? Thanks.

w84thesun: https://github.com/FerretDB/FerretDB/pull/873/files#diff-d9cbd97d17d60b9a415f168c988e7af55ab4d6945e24b0fb570a0577eff6b884R46

docs/auth.md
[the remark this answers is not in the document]
w84thesun: We can't implement authentication without user management API and as far as I know that functionality is not implemented yet.

docs/auth.md
w84thesun: `trusted` means PostgreSQL server trusts everyone, which means authentication is disabled. When password authentication is enabled username and a password are required to authenticate.

docs/auth.md
w84thesun: Added a note about this.

docs/auth.md
```diff
@@ -0,0 +1,115 @@
...
+
+`saslContinue` would not be supported for now.
+
+When a client disconnects, we should close the connection pool.
+We should close all user pools when the server is shut down.
+
+### Tests:
+* Connect to the server with username and password in connection string twice.
+* Connect to the server with different usernames and passwords.
+* Connect to the server without username and password in connection string.
```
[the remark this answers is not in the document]
noisersup: any updates?

rumyantseva (changes requested):
I reject this task as it is not for review (yet?)

noisersup (changes requested):
Not ready.

AlekSi (commented):
On hold for now

PR #875, opened 2022-07-12
Extract two more helpers

AlekSi:
## Readiness checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -114,6 +114,10 @@ linters:
 issues:
   exclude-use-default: false
   exclude-rules:
+    # ignore deprecation errors in existing code; new code is checked by the other configuration
+    - linters: [staticcheck]
+      text: SA1019
+
     # only `wire` package can import `bson` package
     - linters: [depguard]
       path: internal/wire
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -78,7 +78,9 @@ linters-settings:
     severity: warning
     enableAllRules: true
   staticcheck:
-    checks: ["all"]
+    checks:
+      - all
+      - -SA1019 # ignore deprecation errors in existing code; new code is checked by the other configuration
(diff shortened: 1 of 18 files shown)


.golangci.yml
```diff
@@ -114,6 +114,10 @@ linters:
 issues:
   exclude-use-default: false
   exclude-rules:
+    # ignore deprecation errors in existing code; new code is checked by the other configuration
```
rumyantseva, line 117: Awwww! Thanks!

rumyantseva (dismissed):
LGTM

w84thesun (dismissed):

noisersup (dismissed):

noisersup:
golangci-lint fails but if I understand correctly it won't fail after we merge it.

AlekSi:
There was a problem with golangci-lint configuration. @noisersup @rumyantseva PTAL

internal/util/testutil/db.go
```diff
@@ -0,0 +1,48 @@
...
+package testutil
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// DatabaseName returns a stable database name for that test.
```
seeforschauer, line 24: What does mean stable here?

seeforschauer (approved):

internal/util/testutil/db.go
AlekSi: The same for every invocation, non-random

rumyantseva (approved):
LGTM (the linter still doesn't pass though). Actually, now it looks like there is some error there...

AlekSi:
The cache was reset after the Go version bump, and 2 minutes is not enough for golangci-lint with a clean cache. But normally, that's not a problem because cron pre-fills our caches at night.

PR #885, opened 2022-07-13
Set pprof label for client connections

AlekSi:
image

## Readiness checklist

### DIFF
--- a/internal/clientconn/listener.go
+++ b/internal/clientconn/listener.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"runtime/pprof"
 	"sync"
 	"time"
 
@@ -74,7 +75,7 @@ func (l *Listener) Run(ctx context.Context) error {
 	close(l.listening)
 	logger.Sugar().Infof("Listening on %s ...", l.Addr())
 
-	// handle ctx cancelation
+	// handle ctx cancellation
 	go func() {
 		<-ctx.Done()
 		l.listener.Close()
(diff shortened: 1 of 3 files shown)


internal/clientconn/listener.go
```diff
@@ -105,17 +106,30 @@ func (l *Listener) Run(ctx context.Context) error {
 
 		// run connection
 		go func() {
+			connID := fmt.Sprintf("%s -> %s", netConn.RemoteAddr(), netConn.LocalAddr())
```
rumyantseva, line 109: Hmm, should `connID` be a part of the `conn` struct then?

internal/clientconn/listener.go
```diff
@@ -105,17 +106,30 @@ func (l *Listener) Run(ctx context.Context) error {
...
+
+			runCtx, runCancel := ctxutil.WithDelay(ctx.Done(), delay)
+			defer runCancel()
+
+			if l.opts.TestConnTimeout != 0 {
+				runCtx, runCancel = context.WithTimeout(runCtx, l.opts.TestConnTimeout)
+				defer runCancel()
+			}
+
+			defer pprof.SetGoroutineLabels(runCtx)
```
rumyantseva, line 119: Ok, I guess you tested it :) 
(I mean, the fact that we both call `SetGoroutineLabels` right here and also defer it)

seeforschauer (approved):

rumyantseva (commented):
Well, I don't really see a difference in terms of context cancellation itself.
Where is something we start canceling that we didn't cancel before?

internal/clientconn/listener.go
AlekSi: We don't use it otherwise, and I don't think we should. Having it as a field will be too tempting

rumyantseva (approved):
LGTM

internal/clientconn/listener.go
AlekSi: It works because `runCtx` is evaluated before `pprof.SetGoroutineLabels` call is put on stack

PR #808, opened 2022-06-30
Optimize documents fetching / filtering

rumyantseva:
# Description

This PR closes #372.

## Readiness checklist

* [x] I set assignee, reviewers, labels, project and sprint.
* [x] I added tests for new functionality or bugfixes.
* [x] I ran `task all`, and it passed.
* [x] I added/updated documentation for exported and unexported functions, variables, types, etc.
* [x] I checked complex documentation rendering with `bin/godoc -index -play -timestamps -http=127.0.0.1:6060`.

PR #887, opened 2022-07-13
Cover more logic in transactions

rumyantseva:
# Description

This PR is preparation for #808.
It covers more cases with transactions, e.g. `msgInsert` and `msgCreate`.
It doesn't cover all the possible places, it's rather timeboxed.

## Readiness checklist

### DIFF
--- a/internal/handlers/pg/msg_create.go
+++ b/internal/handlers/pg/msg_create.go
@@ -72,7 +72,7 @@ func (h *Handler) MsgCreate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, 
 	}
 
 	err = h.pgPool.InTransaction(ctx, func(tx pgx.Tx) error {
-		if err := pgdb.CreateDatabase(ctx, tx, db); err != nil {
+		if err := pgdb.CreateDatabaseIfNotExists(ctx, tx, db); err != nil {
 			return lazyerrors.Error(err)
 		}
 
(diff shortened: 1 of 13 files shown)


internal/handlers/pg/msg_create.go
```diff
@@ -69,21 +71,26 @@ func (h *Handler) MsgCreate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
-			return nil, common.NewErrorMsg(common.ErrNamespaceExists, msg)
+	err = h.pgPool.InTransaction(ctx, func(tx pgx.Tx) error {
+		if err := pgdb.CreateDatabaseIfNotExists(ctx, tx, db); err != nil {
+			return lazyerrors.Error(err)
 		}
-		if errors.Is(err, pgdb.ErrInvalidTableName) {
-			msg := fmt.Sprintf("Invalid collection name: '%s.%s'", db, collection)
-			return nil, common.NewErrorMsg(common.ErrInvalidNamespace, msg)
+
+		if err = pgdb.CreateCollection(ctx, tx, db, collection); err != nil {
```
AlekSi, line 79:
```suggestion
		if err := pgdb.CreateCollection(ctx, tx, db, collection); err != nil {
```

Otherwise, we are assigning to the outer `err`.

Maybe there is a more general solution? For example, start each closure with `var err error`?

AlekSi (commented):
Not a full review yet, just one thing I noticed

seeforschauer (commented):
LGMT except for that outer error assignment.

seeforschauer (approved):

internal/handlers/pg/msg_create.go
AlekSi: No comment about more general solution?

PR #890, opened 2022-07-14
Disconnect client in embedded tests

AlekSi:
## Readiness checklist

### DIFF
--- a/integration/setup.go
+++ b/integration/setup.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 	"go.mongodb.org/mongo-driver/mongo"
@@ -58,7 +59,7 @@ type SetupOpts struct {
 }
 
 // SetupWithOpts setups the test according to given options,
-// and returns test-specific context (that is cancelled when the test ends), database collection
+// and returns test-specific context (that is canceled when the test ends), database collection
 // and the port of the running server.
 func SetupWithOpts(t *testing.T, opts *SetupOpts) (context.Context, *mongo.Collection, int) {
 	t.Helper()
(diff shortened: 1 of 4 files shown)


integration/setup.go
```diff
@@ -165,11 +166,12 @@ func setupListener(t *testing.T, ctx context.Context, logger *zap.Logger) int {
...
-		ProxyAddr:  proxyAddr,
-		Mode:       mode,
-		Handler:    h,
-		Logger:     logger,
+		ListenAddr:         "127.0.0.1:0",
+		ProxyAddr:          proxyAddr,
+		Mode:               mode,
+		Handler:            h,
+		Logger:             logger,
+		TestRunCancelDelay: time.Hour, // make it easier to notice missing client's disconnects
```
rumyantseva, line 174: Let's comment on how exactly it makes it easier? (The test will hang or what?)

internal/clientconn/listener.go
```diff
@@ -108,7 +107,12 @@ func (l *Listener) Run(ctx context.Context) error {
 		go func() {
 			connID := fmt.Sprintf("%s -> %s", netConn.RemoteAddr(), netConn.LocalAddr())
 
-			runCtx, runCancel := ctxutil.WithDelay(ctx.Done(), delay)
+			// give clients a few seconds to disconnect after ctx is canceled
+			runCancelDelay := l.opts.TestRunCancelDelay
```
rumyantseva, line 111: I don't understand this part. What's the difference between this and `context.WithTImeout`?

rumyantseva (approved):
It's ok'ish that we fix it this way... But!
It feels like our embedded client needs a proper `Shutdown` to stop gracefully. Simple context propagation is not enough.

seeforschauer (approved):

PR #894, opened 2022-07-15
Wait for `Tigris` backend to be ready

seeforschauer:
# Description

## This PR

* adds wait for Tigris backend in envtool.
* adds `docker buildx prune` in `init-clean`
* adds `docker system prune` including volume in `init-clean`

## About `docker system prune` and `docker buildx prune` 

It might seem too much for one.
If you don't want to prune all docker volumes, feel free to say `No` to that when it asks. 
Task continues to run without pruning docker volumes.

## Checklist:

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -48,7 +48,7 @@ tasks:
       - go mod verify
 
   init-clean:
-    desc: "Clean all caches and re-install development tools"
+    desc: "Clean all caches, including docker buildx and docker system volumes, and re-installs development tools"
     cmds:
       - docker buildx prune --all
       - docker system prune --all --volumes
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -48,7 +48,7 @@ tasks:
       - go mod verify
 
   init-clean:
-    desc: "Clean all caches, including docker buildx and docker system volumes, and re-installs development tools"
+    desc: "Clean all caches, including docker buildx and docker system volumes, and re-install development tools"
     cmds:
       - docker buildx prune --all
       - docker system prune --all --volumes
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -50,6 +50,8 @@ tasks:
   init-clean:
     desc: "Clean all caches and re-install development tools"
     cmds:
+      - docker buildx prune --all
+      - docker system prune --all --volumes
       - bin/golangci-lint cache clean
       - go clean -cache -testcache -modcache -fuzzcache
       - task: init
--- a/cmd/envtool/main.go
+++ b/cmd/envtool/main.go
@@ -109,7 +109,7 @@ func waitForTigrisPort(ctx context.Context, logger *zap.SugaredLogger, port uint
 	for ctx.Err() == nil {
 		driver, err := driver.NewDriver(ctx, cfg)
 		if err == nil {
-			_, err := driver.Info(ctx)
+			_, err = driver.Info(ctx)
 			driver.Close()
 			if err == nil {
 				return nil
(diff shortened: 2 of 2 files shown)


Taskfile.yml
```diff
@@ -50,6 +50,8 @@ tasks:
   init-clean:
     desc: "Clean all caches and re-install development tools"
     cmds:
+      - docker buildx prune --all
+      - docker system prune --all --volumes
```
rumyantseva, line 54: Oh, that's quite too much! But I guess, we don't store too many things, so it should be fast to recreate everything every time?

My biggest concern is that if apart from FerretDB images I have some other things too, they will be pruned as well. As a FerretDB developer, I'm ok with it, but as a community maintainer, I won't be happy if everything I have is deleted.

rumyantseva (commented):
I'm quite sceptical about `docker system prune --all --volumes`. As an open source maintainer, I don't want a project to run `docker system prune --all --volumes` on my machine.

Taskfile.yml
seeforschauer: > Oh, that's quite too much! 

😁 it helped in my case.

> But I guess, we don't store too many things, so it should be fast to recreate everything every time?

yes, it's fast enough.


> My biggest concern is that if apart from FerretDB images I have some other things too, they will be pruned as well.

Not silently. Only if you press `yes`.

> As a FerretDB developer, I'm ok with it, but as a community maintainer, I won't be happy if everything I have is deleted.

If you don't want to do that prunes, you can just answer No twice and the task continues to run without pruning docker volumes.

AlekSi: We already have "go clean -cache -testcache -modcache -fuzzcache" that is as destructive as those prune commands. `init-clean` is a not an every day function, and it is not even documented in out contributing guidelines. That's a tool of last resort that helps in complicated cases

rumyantseva: @AlekSi `go clean` won't delete my Docker volumes where I might store something totally irrelevant to FerretDB or even Go but really needed for something else.

AlekSi: (and yes, "go clean" does not ask for confirmation, so it is even more destructive)

cmd/envtool/main.go
```diff
@@ -105,16 +105,14 @@ func waitForTigrisPort(ctx context.Context, logger *zap.SugaredLogger, port uint
...
-	}
-	defer driver.Close()
 
 	for ctx.Err() == nil {
-		_, err := driver.Info(ctx)
+		driver, err := driver.NewDriver(ctx, cfg)
 		if err == nil {
-			return nil
+			if _, err := driver.Info(ctx); err == nil {
+				return nil
```
AlekSi, line 113: We should close connection there too.

Taskfile.yml
seeforschauer: > @AlekSi `go clean` won't delete my Docker volumes where I might store something totally irrelevant to FerretDB or even Go but really needed for something else.

WDYT of extracting docker prune commands to totally new task target? (please suggest target name).

AlekSi: > go clean won't delete my Docker volumes where I might store something totally irrelevant to FerretDB or even Go

Yes, but `go clean` would prune your Go modules cache that might contain modules that are not recoverable. It also cleans fuzzing cache that might contain entries that are not recoverable too. Both commands can cause irrecoverable data loss

AlekSi: I think we could make task description more scary :)

cmd/envtool/main.go
AlekSi: I think we can simplify it to something like
```go
err = driver.Info
driver.Close()
if err == nil {
  return nil
}
```

Taskfile.yml
rumyantseva: A scare task description sounds enough to me, yes!

seeforschauer: Well, I hope the description of what this command now does is scary enough for you :imp:

AlekSi (approved):

PR #901, opened 2022-07-16
Add setup for compatibility tests

AlekSi:
Tests themselves will be in the next PR.

## Readiness checklist


integration/setup.go
```diff
@@ -38,9 +39,10 @@ import (
 )
 
 var (
-	portF      = flag.Int("port", 0, "port to use; if 0, in-process FerretDB is used")
-	handlerF   = flag.String("handler", "pg", "handler to use for in-process FerretDB")
-	proxyAddrF = flag.String("proxy-addr", "", "proxy to use for in-process FerretDB")
+	portF       = flag.Int("port", 0, "first system's port for tests; if 0, in-process FerretDB is used")
```
rumyantseva, line 42: "First system" is a little bit unclear. Can we call it something like "target system" instead? I guess it's the main system we want to test, right?

integration/setup.go
```diff
@@ -38,9 +39,10 @@ import (
...
+	proxyAddrF  = flag.String("proxy-addr", "", "proxy to use for in-process FerretDB")
+	handlerF    = flag.String("handler", "pg", "handler to use for in-process FerretDB")
+	compatPortF = flag.Int("compat-port", 37017, "second system's port for compatibility tests; if 0, they are skipped")
```
rumyantseva, line 45: And this one is something like "reference system" I guess? Or is it the other way around?

integration/setup.go
```diff
@@ -50,110 +52,115 @@ var (
...
+// SetupResult represents setup results.
+type SetupResult struct {
+	Ctx              context.Context
+	Collection       *mongo.Collection
+	Port             uint16
+	CompatCollection *mongo.Collection
+	CompatPort       uint16
+}
+
+// SetupWithOpts setups the test according to given options.
```
rumyantseva, line 77: I'd say we need to document setup for tests better. I guess, the place to do it is here.
I expect some extended info here. For example, something like "If ports for both target and reference system are set, `ComatCollection` and `CompatPort` are defined as xyz, otherwise they are not set and musnt' be used".

integration/setup.go
```diff
@@ -50,110 +52,115 @@ var (
...
 	Providers []shareddata.Provider
 }
 
-// SetupWithOpts setups the test according to given options,
-// and returns test-specific context (that is canceled when the test ends), database collection
-// and the port of the running server.
-func SetupWithOpts(t *testing.T, opts *SetupOpts) (context.Context, *mongo.Collection, int) {
-	t.Helper()
+// SetupResult represents setup results.
+type SetupResult struct {
```
rumyantseva, line 69: Maybe something like this? In case we need to extend such a config later, that could simplify/speed up dev process.

```go
struct System {
	Collection       *mongo.Collection
	Port             uint16
}

type SetupResult struct {
	Ctx           context.Context
	TargetSystem        System
	ReferenceSystem  System
}
```

rumyantseva (commented):
The configuration itself makes sense to me. 
However, I think we can come up with a better wording than "first system", "second system".

For the "first system" I suggest something like "target system". If we don't run tests in compatibility mode, this system is still a "target".

For the "second system" it's a bit more complicated. I suggested the word "reference", but I'm not sure if "reference" is a good word. Maybe something opposite is actually a better word to describe this system. Maybe, something like "follower" or "following system"?

seeforschauer (commented):
LGTM, agree with Elena's comments on "first" and "second" and doc.

integration/setup.go
AlekSi: Well, I would not call it a reference system because that's FerretDB when the target system (a name I like) is MongoDB

integration/setup.go
AlekSi: I would apply YAGNI there: we have only one test that needs ports, and most tests use simple `Setup` or `SetupCompat` functions

integration/setup.go
AlekSi: That will be in CONTRIBUTING.md in the future PR

integration/setup.go
AlekSi: Done

AlekSi:
Let's think about names a bit later, when we would also document those tests. Right now, my goal is to move forward with all unimplemented update operators and tests for them that depend on that and the next PR

integration/setup.go
rumyantseva: Ok(ish)!

rumyantseva (approved):
LGTM

seeforschauer (approved):

PR #909, opened 2022-07-18
Support `explain` diagnostic command

### LINKED ISSUE #167: Support `explain` diagnostic command
After #372.

We should implement `explain` command for query commands like `find`, `findAndModify`, `count`, etc.

Non-query commands like `update`, `delete`, etc are not in scope.
Aggregation pipeline is not in scope.
Tigris is not in scope.

See:
* https://www.mongodb.com/docs/manual/reference/command/explain/
* https://www.mongodb.com/docs/manual/reference/explain-results/
* https://www.postgresql.org/docs/current/sql-explain.html

We don't have to replicate MongoDB output precisely; instead, we will use what `EXPLAIN (FORMAT JSON)` would return. But some similarity is needed.

* [x] convert returned JSON document into `types.Document` of the same form;
* [x] implement `explain` command:
  * [x] `verbosity` parameter should be ignored;
  * [x] `explainVersion`, `command`, `serverInfo` fields should be returned as in MongoDB;
  * [x] `queryPlanner` should contain converted `EXPLAIN (FORMAT JSON)` results;
  * [x] `executionStats` and `serverParameters` should be absent;
* [x] change `fetch` method to run `EXPLAIN (VERBOSE true, FORMAT JSON) <query>` instead of the query itself if parameter is given;
* [x] write integration test;
* [x] refactor `fetch` method (split it into two?) if it seems to be needed on review.

seeforschauer:
# Description

This PR closes #167.

## PR checklist

* [x] convert returned JSON document into `types.Document` of the same form;
* [x] implement `explain` command:
  * [x] `verbosity` parameter should be ingored;
  * [x] `explainVersion`, `command`, `serverInfo` fields should be returned as in MongoDB;
  * [x] `queryPlanner` should contain converted `EXPLAIN (FORMAT JSON)` results;
  * [x] `executionStats` and `serverParameters` should be absent;
* [x] change `fetch` method to run `EXPLAIN (VERBOSE true, FORMAT JSON) <query>` instead of the query itself if parameter is given;
* [x] write integration test;
* [x] refactor `fetch` method (split it into two?) if it seems to be needed on review.

## Example output
```sh
ferretdb=#
ferretdb=# EXPLAIN (FORMAT JSON) SELECT _jsonb FROM "testcommandsdiagnosticexplain"."testcommandsdiagnosticexplain_1dd9d237"
ferretdb-# ;
                            QUERY PLAN
------------------------------------------------------------------
 [                                                               +
   {                                                             +
     "Plan": {                                                   +
       "Node Type": "Seq Scan",                                  +
       "Parallel Aware": false,                                  +
       "Async Capable": false,                                   +
       "Relation Name": "testcommandsdiagnosticexplain_1dd9d237",+
       "Alias": "testcommandsdiagnosticexplain_1dd9d237",        +
       "Startup Cost": 0.00,                                     +
       "Total Cost": 23.60,                                      +
       "Plan Rows": 1360,                                        +
       "Plan Width": 32                                          +
     }                                                           +
   }                                                             +
 ]
(1 row)
```
Or
```sh
--------------------------------------[ RECORD 1 ]--------
QUERY PLAN 
| [                                        +
|   {                                      +
|     "Plan": {                            +
|       "Node Type": "Seq Scan",           +
|       "Parallel Aware": false,           +
|       "Async Capable": false,            +
|       "Relation Name": "values_34474c3b",+
|       "Schema": "test",                  +
|       "Alias": "values_34474c3b",        +
|       "Startup Cost": 0.00,              +
|       "Total Cost": 23.60,               +
|       "Plan Rows": 1360,                 +
|       "Plan Width": 32,                  +
|       "Output": ["_jsonb"]               +
|     }                                    +
|   }                                      +
| ]
```

## Readiness checklist


seeforschauer:
It's not ready yet but let's discuss some:
- I've added `explain ` to `iterateFetch`. `iterateFetch(ctx context.Context, fetched chan FetchedDocs, rows pgx.Rows, explain bool) error`. Do you have any thoughts on how can it be more beautiful? 
- the `Convert` func in internal/util/convert/convert.go. Should it be in that package or in another place?
- the test doesn't check any values. I'm about to add some, for example, from the example structures (see header)

integration/commands_diagnostic_test.go
```diff
@@ -106,3 +113,126 @@ func TestCommandsDiagnosticConnectionStatus(t *testing.T) {
...
+				{"verbosity", "queryPlanner"},
+			},
+		},
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			var actual bson.D
+			err = collection.Database().RunCommand(ctx, tc.command).Decode(&actual)
```
AlekSi, line 214:
```suggestion
			err := collection.Database().RunCommand(ctx, tc.command).Decode(&actual)
```

internal/handlers/common/msg_listcommands.go
```diff
@@ -95,6 +95,10 @@ var Commands = map[string]command{
 		Help:    "Drops production database.",
 		Handler: (handlers.Interface).MsgDropDatabase,
 	},
+	"explain": {
+		Help:    "Explain diagnostic command.",
```
AlekSi, line 99: Note that other command's help text actually describes the behavior

internal/handlers/handlers.go
```diff
@@ -78,6 +78,9 @@ type Interface interface {
 	// MsgDropDatabase drops production database.
 	MsgDropDatabase(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error)
 
+	// MsgExplain represent explain diagnostic command.
```
AlekSi, line 81: The same issue as with the help text

internal/handlers/pg/pgdb/query_documents.go
```diff
@@ -141,12 +148,21 @@ func iterateFetch(ctx context.Context, fetched chan FetchedDocs, rows pgx.Rows)
 				return writeFetched(ctx, fetched, FetchedDocs{Err: lazyerrors.Error(err)})
 			}
 
-			doc, err := fjson.Unmarshal(b)
-			if err != nil {
-				return writeFetched(ctx, fetched, FetchedDocs{Err: lazyerrors.Error(err)})
+			if explain {
+				var plans []map[string]any
+				if err := json.Unmarshal(b, &plans); err != nil {
```
AlekSi, line 153: Let's move `json.Unmarshal` to those new functions too

AlekSi (changes requested):

internal/handlers/pg/pgdb/query_documents.go
seeforschauer: > Let's move `json.Unmarshal` to those new functions too

extracted to function.

internal/handlers/handlers.go
seeforschauer: > The same issue as with the help text

changed to `Display the execution plan.`

internal/handlers/common/msg_listcommands.go
seeforschauer: > Note that other command's help text actually describes the behavior

changed to `Display the execution plan.`

internal/handlers/pg/pgdb/query_documents.go
```diff
@@ -126,7 +133,7 @@ func (pgPool *Pool) QueryDocuments(ctx context.Context, querier pgxtype.Querier,
 
 // iterateFetch iterates over the rows returned by the query and sends FetchedDocs to fetched channel.
 // It returns ctx.Err() if context cancellation was received.
-func iterateFetch(ctx context.Context, fetched chan FetchedDocs, rows pgx.Rows) error {
+func iterateFetch(ctx context.Context, fetched chan FetchedDocs, rows pgx.Rows, explain bool) error {
```
rumyantseva, line 136: Hmm, I wonder if it makes sense to have a totally separate function for `explain`. If I understand your code (and how Postgres works) correctly, we won't have the results, we will only have the plan. (It's not `explain analyze`, it's only `explain`). So then the whole story with iterating through lots of docs doesn't make a lot of sense. Also, there is not a lot of common code between the case with `explain` and without it.

internal/handlers/pg/pgdb/query_documents.go
```diff
@@ -97,6 +100,10 @@ func (pgPool *Pool) QueryDocuments(ctx context.Context, querier pgxtype.Querier,
 	}
 	sql += `FROM ` + pgx.Identifier{db, table}.Sanitize()
 
+	if sp.Explain {
```
rumyantseva, line 103: As `explain` doesn't return a huge list of docs (but rather an execution plan), does it make sense to include explain to this function (and deal with channels etc)? Or is it better to have a separate function for it?

rumyantseva (changes requested):
I don't think we need a channel for `explain` results, so I suggest not to use it (but implement a simpler approach than the one for `fetch`). This is also what was suggested in the #167.  Please correct me if I'm wrong.

Another thing that worries me: it looks like we implement `explain` for `find` but not for `update` or other operators. I see that this is how the initial issue is described, but I guess we will need to reuse at least a part of the current implementation for the other commands too.

internal/handlers/pg/pgdb/query_documents.go
seeforschauer: > Hmm, I wonder if it makes sense to have a totally separate function for `explain`. If I understand your code (and how Postgres works) correctly, we won't have the results, we will only have the plan. (It's not `explain analyze`, it's only `explain`). 

The plan is the result. We have results. WDYM? I don't understand.

> So then the whole story with iterating through lots of docs doesn't make a lot of sense. 

It's not a lot of docs, right, but when we would build an explain for queries where pushdown happens (not only for `_id` but for other complex requests, it will be more convenient to have it right where the SQL query creation happens. 

> Also, there is not a lot of common code between the case with `explain` and without it.

That's for now, please see the previous message.

internal/handlers/pg/pgdb/query_documents.go
seeforschauer: > As `explain` doesn't return a huge list of docs (but rather an execution plan), does it make sense to include explain to this function (and deal with channels etc)? Or is it better to have a separate function for it?

Answered in https://github.com/FerretDB/FerretDB/pull/909#discussion_r928533165

internal/handlers/pg/pgdb/query_documents.go
seeforschauer: I'll try to extract a func that builds query then probably it's possible to do it in a separate function.

seeforschauer: @rumyantseva please see the result if it satisfies you.

internal/handlers/pg/pgdb/query_documents.go
```diff
@@ -124,6 +106,82 @@ func (pgPool *Pool) QueryDocuments(ctx context.Context, querier pgxtype.Querier,
...
+	}
+	rows, err := querier.Query(ctx, sql)
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+	var res []*types.Document
+	for ctx.Err() == nil {
+		if !rows.Next() {
+			break
+		}
```
w84thesun, line 124: I think we can put `rows.Next()` inside `for` condition as they both will interrupt `for` execution.
```suggestion
	for ctx.Err() == nil && rows.Next() {
```

internal/handlers/pg/msg_explain.go
```diff
@@ -0,0 +1,177 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return h.buildExplainResult(ctx, document, resDocs)
+}
+
+// parseExplainParams validates document and returns pgdb.SQLParam.
+func (h *Handler) parseExplainParams(ctx context.Context, document *types.Document) (pgdb.SQLParam, error) {
```
w84thesun, line 60: Could you please add some new lines to this method? It's hard to read it as is.
It's easier to read code when variables declaration or `switch` or `if` blocks are separated with a new line.

w84thesun (changes requested):

internal/handlers/pg/msg_find.go
```diff
@@ -180,3 +114,99 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+// findParams are validated find query parameters.
+type findParams struct {
+	projection *types.Document
+	filter     *types.Document
+	sort       *types.Document
+	limit      int64
+	sp         pgdb.SQLParam
+}
+
+// parseFindParams validates document and returns findParams.
```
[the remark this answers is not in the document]
seeforschauer: > We can either change the method description, move out validation code, or split that method into 2: one for parsing and one for validating parameters.

Will it be OK for you if I just rename it to `validateFindParams`?

internal/handlers/pg/msg_explain.go
```diff
@@ -0,0 +1,177 @@
...
+		return err
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return h.buildExplainResult(ctx, document, resDocs)
+}
+
+// parseExplainParams validates document and returns pgdb.SQLParam.
```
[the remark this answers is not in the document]
seeforschauer: > Same here: the method is called `parseExplainParams` but the description says `validates and return`.

Could you please explain, I don't understand what do you want instead? What are your suggestions?

internal/handlers/pg/msg_explain.go
seeforschauer: > Could you please add some new lines to this method? It's hard to read it as is.

I've added, please see.

internal/handlers/pg/pgdb/query_documents.go
seeforschauer: changed

internal/handlers/pg/msg_explain.go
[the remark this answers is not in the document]
seeforschauer: > Same here: the method is called `parseExplainParams` but the description says `validates and return`.

I've renamed the func, please see if it OK to you.

seeforschauer:
I found that for `found` we need the same thing as for `find`. Let me add it.

w84thesun (approved):

Issue #787, opened 2022-06-23
Rework database and collection creation for Tigris


AlekSi:
Currently, we try to create databases and collections in too many places. We should clean that up.

Expand `tigrisdb`. Port `CreateTableIfNotExist` and `insert` from `pg` handler. Use them where needed.

Remove our own "filter" from tigrisdb.

rumyantseva:
+ same for database (see https://github.com/FerretDB/FerretDB/pull/887)

rumyantseva:
@AlekSi case: a few documents are inserted ("insert many") into a collection that didn't exist. The schema for the collection is created based on the first document. The second document doesn't match the schema. What should be returned in such a case? - In fact, we created a collection and inserted one doc, but then the second doc insertion failed.

PR #913, opened 2022-07-18
Validate database names

### LINKED ISSUE #807: Validate database names
After #750.

MongoDB database names restrictions are documented there: https://www.mongodb.com/docs/manual/reference/limits/#naming-restrictions
PostgreSQL restrictions – there: https://www.postgresql.org/docs/current/sql-syntax-lexical.html

For collection names, we use a special settings table in the schema matching database name. Let's not complicate things for database/schema names. Instead, let's limit database names to what PostgreSQL supports.

Let's validate that database/schema name:
* matches regular expression `^[a-z_][a-z0-9_]{0,62}$`;
* can't start with a reserved prefix `_ferretdb_` (just in case).

We could relax those rules in the future.

* [ ] implement database name validation on database/schema creation;
* [ ] add integration test;
* [ ] document the difference in rules in README.md;
* [ ] add diff dance test.

Tigris is not in scope.

seeforschauer:
# Description

This PR is a part of #807.

- removed `TestFindAndModifyEmptyCollectionName` test as the same test already done in `TestCollectionName`.
- database test names generation changed from sign `-` to sign `_` to conform the rules.

### Error codes
| Test Case | MongoDB | FerretDB |  Comment |
|--|--|--|--|
|too long name | `Invalid namespace specified`, quoted | `Invalid namespace`, not quoted | `specified` word in MongoDB, quotation differs | 
| `$^%*&` | `Invalid namespace`, not quoted | `Invalid namespace`, not quoted | equal | 

Dance PR https://github.com/FerretDB/dance/pull/176.

### Checklist


w84thesun (dismissed):
LGTM

README.md
```diff
@@ -36,11 +36,11 @@ and [contributing guidelines](CONTRIBUTING.md).
 
 2. FerretDB does not support NUL (`\0`) characters in strings.
 
-3. Collection names restrictions:
+3. Database and collection names restrictions:
 
 * collection name cannot start with the reserved prefix `_ferretdb_`.
 * collection name must not include non-latin letters, spaces, dots, dollars or dashes.
-* collection name length must be less or equal than 120 symbols.
+* collection name length must be less or equal than 120 symbols, database name length limit is 64 symbols.
```
rumyantseva, line 43: So, there is only one restriction for databases? It's not mentioned that database should match regex and shouldn't start from a reserved prefix. Based on this doc, only collection name has such restriction.

integration/basic_test.go
```diff
@@ -195,3 +197,59 @@ func TestCollectionName(t *testing.T) {
...
 	})
 }
+
+func TestDatabaseName(t *testing.T) {
+	t.Parallel()
+
+	t.Run("Err", func(t *testing.T) {
+		ctx, collection := Setup(t)
+
+		dbName300 := strings.Repeat("a", 300)
```
rumyantseva, line 207: Well, if we want to test boundary conditions, we should test with length 65 (or, at least, with length 65 too).
E.g. two cases: length 64 should pass validation, and length 65 should fail.

integration/basic_test.go
```diff
@@ -195,3 +197,59 @@ func TestCollectionName(t *testing.T) {
...
+
+		cases := map[string]struct {
+			db  string
+			err *mongo.CommandError
+			alt string
+		}{
+			"TooLongForBoth": {
```
rumyantseva, line 214: I don't understand this case name - `TooLongForBoth` - who are those `both`?

integration/findandmodify_test.go
```diff
@@ -152,34 +152,6 @@ func TestFindAndModifySimple(t *testing.T) {
 	}
 }
 
-func TestFindAndModifyEmptyCollectionName(t *testing.T) {
```
rumyantseva, line 155: Why is this test deleted? Empty DB name and empty collection name look like totally different cases two me and they will fail because of different reasons - `empty DB name + valid collection name` vs `valid DB name + empty collection name`

internal/handlers/pg/pgdb/databases.go
```diff
@@ -58,9 +58,20 @@ func Databases(ctx context.Context, querier pgxtype.Querier) ([]string, error) {
...
+//   * ErrInvalidDatabaseName if db name doesn't comply with the rules.
+// Use errors.Is to check the error.
 func CreateDatabase(ctx context.Context, querier pgxtype.Querier, db string) error {
+	if !validateDatabaseNameRe.MatchString(db) {
+		return ErrInvalidDatabaseName
+	}
+
+	if strings.HasPrefix(db, reservedPrefix) {
+		return ErrInvalidDatabaseName
+	}
```
rumyantseva, line 73: As it's the same error returned, why not both conditions together?

```suggestion
	if !validateDatabaseNameRe.MatchString(db) || strings.HasPrefix(db, reservedPrefix) {
		return ErrInvalidDatabaseName
	}
```

rumyantseva (changes requested):
A few things are not really clear to me, asked questions.

rumyantseva:
P.S. @seeforschauer could you also check `MsgUpdate`? I think it calls collection creation somewhere under the hood as well.

internal/handlers/pg/pgdb/databases.go
seeforschauer: > As it's the same error returned, why not both conditions together?

Well, it can be, but it's polishing.

README.md
seeforschauer: > P.S. @seeforschauer could you also check `MsgUpdate`? I think it calls collection creation somewhere under the hood as well.

WDYM? We here about the database validation. 
Anyway, `MsgUpdate` uses `insert` private handler function if necessary so it definitely has these checks.

integration/findandmodify_test.go
seeforschauer: > Why is this test deleted?

I've added the description in the PR before requesting review..
 
> Empty DB name and empty collection name look like totally different cases two me and they will fail because of different reasons - `empty DB name + valid collection name` vs `valid DB name + empty collection name`

Yes it is and `CollectionNameTest` does this test.

README.md
rumyantseva: Don't [this place](https://github.com/seeforschauer/FerretDB/blob/issue-807-validate-db-names/internal/handlers/pg/msg_update.go#L65) need a similar error check as the one you added for `MsgCreate` and `MsgInsert`?

seeforschauer: oh thank you, I understood why I didn't find this:
there is a deprecated usage.

Screenshot from 2022-07-19 13-08-58

internal/handlers/pg/pgdb/settings.go
```diff
@@ -33,10 +33,10 @@ import (
 
 const (
 	// Internal collections prefix.
```
AlekSi, line 35: The comment should be updated too

seeforschauer: > The comment should be updated too

fixed.

integration/findandmodify_test.go
seeforschauer: > > removed `TestFindAndModifyEmptyCollectionName` test as the same test already done in `TestCollectionName`.
> 
> `TestCollectionName` tries to create collection with an empty name. This test tries to find something in non-existing collection without creating it. I don't understand how they are the same

That's an argument, restored.

integration/basic_test.go
noisersup: a little thing but maybe `TooLongForBothDBs` could be more readable :)

seeforschauer: >  a little thing but maybe TooLongForBothDBs could be more readable :)

Good suggestion changed for collection name and database name tests.

README.md
```diff
@@ -36,12 +36,12 @@ and [contributing guidelines](CONTRIBUTING.md).
...
 2. FerretDB does not support NUL (`\0`) characters in strings.
 
-3. Collection names restrictions:
+3. Database and collection names restrictions:
 
-* collection name cannot start with the reserved prefix `_ferretdb_`.
-* collection name must not include non-latin letters, spaces, dots, dollars or dashes.
-* collection name length must be less or equal than 120 symbols.
-* collection name must not start with a number.
+* name cannot start with the reserved prefix `_ferretdb_`.
```
noisersup, line 41: We could also mention that it cannot start with capital letter
(^[**a-z**_][a-z0-9_]{0,62}$)

noisersup (changes requested):

README.md
seeforschauer: > We could also mention that it cannot start with capital letter (^[**a-z**_][a-z0-9_]{0,62}$)

added.

w84thesun (approved):

noisersup (approved):
LGTM

integration/basic_test.go
seeforschauer: > Well, at my first PR in #844 I also created such border cases. But, if you create this test, FerretDB would pass and fail in MongoDB.
> 
> To record that difference, I'll create a dance diff test as I did it for #844 PR. It's also specified in the issue description.

I misunderstood you.
Let me reformulate to highlight what.
Both databases have 64 letters max limit.
So you suggested testing 65 and 64-length names.

seeforschauer: but do we need to add positive test?

seeforschauer: > but do we need to add a positive test? we have it everywhere.

I checked and was surprised, you know, 
Database names MongoDB has max length 63, FerretDB has max length 64 :dancing_women: 

It's going to be in dance tests.

w84thesun (dismissed):

internal/handlers/pg/pgdb/settings.go
```diff
@@ -32,11 +32,11 @@ import (
 )
 
 const (
-	// Internal collections prefix.
-	reservedCollectionPrefix = "_ferretdb_"
+	// Reserved prefix.
```
rumyantseva, line 35:
```suggestion
	// Reserved prefix for database and collection names.
```

internal/handlers/tigris/msg_update.go
```diff
@@ -63,14 +63,6 @@ func (h *Handler) MsgUpdate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 		return nil, err
 	}
 
-	// created, err := h.pgPool.CreateTableIfNotExist(ctx, fp.db, fp.collection)
```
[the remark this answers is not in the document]
AlekSi: #787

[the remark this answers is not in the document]
AlekSi: I think it is safe to remove that commented out code now – we will not forget about it

internal/handlers/pg/msg_update.go
```diff
@@ -64,6 +65,11 @@ func (h *Handler) MsgUpdate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 
 	created, err := h.pgPool.CreateTableIfNotExist(ctx, sp.DB, sp.Collection)
 	if err != nil {
+		if errors.Is(pgdb.ErrInvalidTableName, err) ||
```
rumyantseva, line 68: If I understand correctly, in some similar cases, we use `switch - case` instead of `if`s. Peresonally, I'm ok with `if`s as well, so I leave it up to you to keep it as is or rewrite it to switch-case.

rumyantseva (approved):
I have a couple of minor comments, but in general it looks good to me.

internal/handlers/pg/msg_update.go
seeforschauer: > If I understand correctly, in some similar cases, we use `switch - case` instead of `if`s. Peresonally, I'm ok with `if`s as well, so I leave it up to you to keep it as is or rewrite it to switch-case.

well, let's use the following workflow:
1. we all agree on what to use, switch case or if-s
2. record it in our issue or somewhere else
3. I change.

For now, I'll leave it as it is.

seeforschauer:
that's fine, I committed the suggested change and all 3 approvals were dismissed. can we do anything with It?

rumyantseva (dismissed):
I don't have critical comments, but I guess this PR might need a final look from @AlekSi

AlekSi (commented):
https://github.com/FerretDB/FerretDB/pull/913/files#r924907975

AlekSi (approved):
:shipit:

Issue #803, opened 2022-06-30
Support dot notation for update operators


AlekSi:
Add dot notation support for:
* [x] `$inc`
* [x] `$set`
* [x] `$setOnInsert`
* [x] `$unset`

Add integration tests for each operator.

AlekSi:
`$inc` is done, 3 more operators to go

AlekSi:
IIUC, we still need to the same for `$unset`

PR #915, opened 2022-07-19
Add support for `update`'s `$inc` operator dot notation

w84thesun:
# Description

This PR refs #803.

In that PR I've added support for update's `$inc` operator. Several auxiliary functions were added to support it.

## Readiness checklist


internal/types/path.go
```diff
@@ -48,6 +50,15 @@ func NewPathFromString(s string) Path {
 	return NewPath(path)
 }
 
+// DeriveNewPath returns new Path constructed from given path and path element.
+func DeriveNewPath(path Path, elem string) Path {
```
seeforschauer, line 54: `DeriveNewPath` and `GetNewPath`: what is the difference in the expectation that functions should do?
It's not obvious. Can we use a `GetNewPath` instead?

internal/types/document.go
```diff
@@ -306,6 +306,16 @@ func (d *Document) RemoveByPath(path Path) {
 	removeByPath(d, path)
 }
 
+// SetByPath sets value by given path.
+func (d *Document) SetByPath(path Path, value any) error {
+	return setByPath(d, path, value)
+}
+
+// InsertByPath creates Document by given path if it's not exist and sets given value to that Document.
+func (d *Document) InsertByPath(path Path, value any) error {
```
seeforschauer, line 315: If it's always a `*types.Document`, let's set the variable name to `*types.Document`, not `any`.
Otherwise, let's change the comment?

integration/update_field_test.go
```diff
@@ -283,112 +283,142 @@ func TestUpdateFieldInc(t *testing.T) {
...
+			},
 		} {
 			name, tc := name, tc
 			t.Run(name, func(t *testing.T) {
 				t.Parallel()
 				ctx, collection := setup.Setup(t, shareddata.Scalars, shareddata.Composites)
 
-				_, err := collection.UpdateOne(ctx, tc.filter, tc.update)
+				_, err := collection.InsertMany(ctx, []any{
+					bson.D{{"_id", "document-nested"}, {"foo", bson.D{{"bar", bson.D{{"baz", int32(1)}}}}}},
```
seeforschauer, line 412: Why not add `document-nested` as `document-embedded` into `shareddata.Composites`?

integration/update_field_test.go
```diff
@@ -283,112 +283,142 @@ func TestUpdateFieldInc(t *testing.T) {
...
+					bson.D{{"_id", "array-nested"}, {"foo", bson.D{{"bar", bson.A{bson.D{{"baz", int32(1)}}}}}}},
```
seeforschauer, line 413: There is `array-middle-embedded` in `shareddata.Composites`. Let's use it?

internal/types/path.go
```diff
@@ -172,3 +183,50 @@ func removeByPath(v any, path Path) {
...
+		index, err := strconv.Atoi(path.Suffix())
+		if err != nil {
+			return err
+		}
+		err = inner.Set(index, value)
+		if err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("can't set value for %T type", inner)
```
rumyantseva, line 210: Is this case possible at all? If yes, let's write a test (in integration tests). If not, let's panic?

rumyantseva (commented):
@w84thesun I need some help to understand the changes :) 
The idea is to support `$inc` for dot notation, I got this. Now I go through test cases, and I don't understand them. I expect there are some changes in `integration/update_field_test.go`, but I don't fully understand what needs to be covered with tests and what doesn't. E.g. do the cases like a string value in a nested field with dot notation need to be covered? (I guess, $inc on a string is not possible). But maybe it was enough to cover it with a regular `$inc` test. So I really need some help to understand what we do test and what we don't.

internal/types/path.go
w84thesun: What is `GetNewPath`?

seeforschauer: > What is `GetNewPath`?

I mean, why did you name this function `DeriveNewPath` not i.e. `GetNewPath`?  
`GetNewPath` is more common.

internal/types/path.go
w84thesun: I think it's not possible, changed it to panic.

integration/update_field_test.go
w84thesun: Rewrite tests to use composite data.

integration/update_field_test.go
w84thesun: I rewrote tests to use composite data.

internal/types/path.go
w84thesun: I think we don't need a common name here. `GetNewPath` doesn't add any context. The context here is that we construct a new Path object from the given one and a path element passed to the function. Also `GetNewPath` would be confusing as we already have `NewPath` function.

build/deps/markdownlint.Dockerfile
```diff
@@ -1 +1 @@
-FROM ghcr.io/igorshubovych/markdownlint-cli:v0.32.0
+FROM ghcr.io/igorshubovych/markdownlint-cli:v0.31.1
```
AlekSi, line 1: That seems to be a mis-merge. Please check other places too

internal/handlers/common/update.go
```diff
@@ -77,6 +77,26 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
...
+					if err != nil {
+						err = doc.InsertByPath(path, incValue)
+						if err != nil {
+							return false, err
+						}
+						changed = true
+						continue
+					}
+
+					incremented, err := addNumbers(incValue, innerValue)
```
AlekSi, line 92: I think we will have the same issue with $dec.

Let's see if we have the same issue with `$set` (when we replace the value with the same value). If yes, I would say that code should be generalized

build/deps/markdownlint.Dockerfile
w84thesun: Fixed that.

internal/types/document.go
w84thesun: Rewrote the comment

internal/handlers/common/update.go
w84thesun: I think it could be refactored when I would add `$set` support. For now, I suggest leaving it as is. There is no `$dec` operator, btw :smiley:

internal/types/path.go
```diff
@@ -67,7 +76,7 @@ func (p Path) Slice() []string {
 
 // Suffix returns the last path element.
 func (p Path) Suffix() string {
-	if len(p.s) <= 1 {
+	if len(p.s) < 1 {
```
AlekSi, line 79: That makes it different from Prefix, TrimSuffix, TrimPrefix below. And I think filterDocumentPair depends in the current behavior. Can we avoid that change?

And let's add unit tests so we don't break that behavior

internal/types/path.go
AlekSi: I think that could be `func (p Path) Append(elem string) Path`

internal/types/document.go
```diff
@@ -306,6 +309,54 @@ func (d *Document) RemoveByPath(path Path) {
 	removeByPath(d, path)
 }
 
+// SetByPath sets value by given path.
```
AlekSi, line 312: We should document what happens when path doesn't exist in the document.

Can we merge SetByPath and InsertByPath?

And let's add unit tests so we don't break that behavior

seeforschauer (commented):
+ 1 on the name change on:
- Suffix
- InsertPath
- Append.

internal/types/document.go
```diff
@@ -316,6 +334,60 @@ func (d *Document) RemoveByPath(path Path) {
...
+	switch inner := innerComp.(type) {
+	case *Document:
+		err := inner.Set(path.Suffix(), value)
+		if err != nil {
+			return err
+		}
+	case *Array:
+		index, err := strconv.Atoi(path.Suffix())
+		if err != nil {
+			return err
```
rumyantseva, line 377: Is this error possible at all? Is it a valid case if `strconv.Atoi(path.Suffix())` returns an error? Or should it be impossible (so we can panic instead of returning an error)?

rumyantseva (commented):
I'm not sure I fully understand the code with dot notations.
I haven't noticed red flags, but I have one question (see the comment).

internal/types/document.go
w84thesun: Yeah, I guess if we got `innerComp` from the `GetByPath` it must be array there.

rumyantseva: If this is a possible case, let's cover it with a test to check the behavior?

rumyantseva (dismissed):
I don't fully understand all the cases, but I don't see any red flags.
So, from my side, I "approve" this code to be merged, and we can improve it later if needed.

internal/types/document.go
```diff
@@ -245,6 +254,11 @@ func (d *Document) Get(key string) (any, error) {
 		return value, nil
 	}
 
+	// we may have the key lay down inside the document
+	if strings.Contains(key, ".") {
+		return d.GetByPath(NewPathFromString(key))
```
AlekSi, line 259: At the very least, documentation should be updated.

But given that we have method GetByPath that uses Get already, and now have Get that uses GetByPath – I think this is too complex.

Let's stick with Has, Get, etc working with single keys and HasByPath, GetByPath, etc working with paths.

AlekSi (changes requested):

noisersup (dismissed):
LGTM

AlekSi (changes requested):
Let's add new compatibility tests to `update_field_compat_test.go`

w84thesun:
> Let's add new compatibility tests to `update_field_compat_test.go`

I suggest to work on compat tests in separate issue. I've added 2 cases just to not forget.

internal/types/path.go
w84thesun: Tests added, condition restored.

internal/types/document.go
w84thesun: Done.

internal/types/types.go
```diff
@@ -70,6 +70,7 @@ type Type interface {
 // TODO remove once we have go-sumtype equivalent?
 type CompositeTypeInterface interface {
 	CompositeType
+	fmt.Stringer
```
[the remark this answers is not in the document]
w84thesun: We use it for dumping data to format correct error output. That output a bit different from that `testutil.Dump` produces. I can turn those methods into functions like `formatDocumentError` and `formatArrayError` (with possibly better names than that).

internal/types/document.go
```diff
@@ -329,53 +329,41 @@ func (d *Document) GetByPath(path Path) (any, error) {
 // SetByPath sets value by given path. If the Path has only one element, it sets the value for the given key.
 // If some parts of the path are missing, they will be created.
 // The Document type will be used to create these parts.
-func (d *Document) SetByPath(path Path, value any) {
+func (d *Document) SetByPath(path Path, value any) error {
 	if path.Len() == 1 {
 		must.NoError(d.Set(path.Slice()[0], value))
```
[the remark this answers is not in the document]
w84thesun: Fixed.

internal/types/document.go
```diff
@@ -329,53 +329,41 @@ func (d *Document) GetByPath(path Path) (any, error) {
 // SetByPath sets value by given path. If the Path has only one element, it sets the value for the given key.
 // If some parts of the path are missing, they will be created.
 // The Document type will be used to create these parts.
-func (d *Document) SetByPath(path Path, value any) {
+func (d *Document) SetByPath(path Path, value any) error {
```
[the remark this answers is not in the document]
w84thesun: Updated

internal/types/types.go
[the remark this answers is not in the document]
AlekSi: Yeah, automatic fmt.Stringer might be too problematic

internal/types/path.go
```diff
@@ -181,3 +183,112 @@ func removeByPath(v any, path Path) {
...
+		}
+
+		next = v
+	}
+
+	return nil
+}
+
+// formatDocumentValue formats document value for error message output.
+func formatDocumentValue(v any) string {
```
rumyantseva, line 229: Minor note. Based on what this function does, I'd call it `formatAnyValue`, not just `Document` :)  I guess, the idea is something like `Format Document's Value` but I'm not sure how to name such a function.

internal/types/document.go
```diff
@@ -329,50 +329,35 @@ func (d *Document) GetByPath(path Path) (any, error) {
...
-
-				next = must.NotFail(d.GetByPath(insertedPath)).(*Document)
-			}
-
-			if doc, ok := v.(*Document); ok {
-				next = doc
-			}
+	if !d.HasByPath(path.TrimSuffix()) {
+		// we should insert the missing part of the path
+		if err := insertByPath(d, path); err != nil {
```
rumyantseva, line 339: Could you point me to the test that covers this case?

rumyantseva (commented):
The logics looks correct to me.

A couple of things:
- I have a few tiny questions about the error formatting (asked in comments). Those are quite minor, so feel free to ignore if it doesn't look helpful.
- Overall, I'm quite struggling with understanding how do we test dot notation and new features implemented in this PR (I asked for an example in comments).

internal/types/path.go
```diff
@@ -181,3 +183,112 @@ func removeByPath(v any, path Path) {
...
+	result := "{"
+
+	for i, key := range doc.keys {
+		if i > 0 {
+			result += ", "
+		}
+
+		switch value := doc.m[key].(type) {
+		case *Document:
+			result += fmt.Sprintf("%q: %s", key, value)
```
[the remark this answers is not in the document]
w84thesun: Yes, before that it should be done by `Sprintf` and `Stringer`, forgot to fix it. Thanks!

internal/types/path.go
w84thesun: `formatAnyValue` sounds reasonable!

internal/types/document.go
w84thesun: That case should be covered with `DotNotationDocumentFieldNotExist` case.

rumyantseva (approved):
LGTM

AlekSi (approved):

Issue #786, opened 2022-06-23
Add integration tests for Tigris


AlekSi:
We should have good integration tests for Tigris. We can't re-use our existing tests for the PostgreSQL handler because they use different value types in the same field. We also don't want to duplicate all our tests. What can we do?

rumyantseva:
I guess, one option would be to use "real" integration or "functional" tests, i.e. to come up with the same test set for all the backends and call it through mongodriver, not through calling exact handlers.

Like a "black-box" testing of FerretDB.

Issue #508, opened 2022-04-22
Add support for bitwise operators BitData type


w84thesun:

chilagrow:
This will make updating `Scalars` dataset easier, since it will allow moving Bitwise query tests to compat

PR #916, opened 2022-07-19
Use `v` instead of `value` in tests

AlekSi:
`value` is a reserved field name in Tigris.

Refs #786.

## Readiness checklist

### DIFF
--- a/integration/query_bitwise_test.go
+++ b/integration/query_bitwise_test.go
@@ -42,6 +42,7 @@ func TestQueryBitwiseAllClear(t *testing.T) {
 		value       any
 		expectedIDs []any
 		err         *mongo.CommandError
+		altMessage  string
 	}{
 		"Array": {
 			value: primitive.A{1, 5},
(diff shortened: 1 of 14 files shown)


integration/query_bitwise_test.go
```diff
@@ -98,8 +99,9 @@ func TestQueryBitwiseAllClear(t *testing.T) {
 			err: &mongo.CommandError{
 				Code:    2,
 				Name:    "BadValue",
-				Message: "value takes an Array, a number, or a BinData but received: $bitsAllClear: \"123\"",
+				Message: `v takes an Array, a number, or a BinData but received: $bitsAllClear: "123"`,
 			},
+			altMessage: `value takes an Array, a number, or a BinData but received: $bitsAllClear: "123"`,
```
AlekSi, line 104: That's a small bug in our implementation ("value" there should be a field name, but it is a fixed string in our code), but we are not fixing it now. See #508.

rumyantseva (approved):
Oh, that's a big one.

I haven't checked all the places, but I ack the change, and from first sight, it looks correct (and reasonable).

PR #919, opened 2022-07-19
Make integration tests work with Tigris

### LINKED ISSUE #786: Add integration tests for Tigris

AlekSi:
Closes #786.

## Readiness checklist


AlekSi:
The last one moved to #919.

PR #937, opened 2022-07-22
Support `$comment` `update`'s argument

### LINKED ISSUE #487: Support `comment` `update`'s argument
The same as #168, but for `update`.

noisersup:
# Description

This PR closes #487.

## Readiness checklist


integration/basic_test.go
```diff
@@ -106,16 +106,40 @@ func TestInsertFind(t *testing.T) {
 //nolint:paralleltest // we test a global list of databases
 func TestFindCommentMethod(t *testing.T) {
 	ctx, collection := setup.Setup(t, shareddata.Scalars)
-	name := collection.Name()
+	collName := collection.Name()
```
seeforschauer, line 109: why did you change it?

integration/basic_test.go
```diff
@@ -106,16 +106,40 @@ func TestInsertFind(t *testing.T) {
...
 	err = collection.FindOne(ctx, bson.D{{"_id", "string"}}, opts).Decode(&doc)
 	require.NoError(t, err)
-	assert.Contains(t, databaseNames, name)
+	assert.Contains(t, databaseNames, collName)
+}
+
+func TestUpdateCommentMethod(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setup.Setup(t, shareddata.Scalars)
+	collName := collection.Name()
```
seeforschauer, line 124: `collName` - column name, collection name? Let's have just name it's  OK.

integration/basic_test.go
noisersup: It's not ready for review yet. Sorry for false alarm, I wanted to create a draft :)

noisersup: Also I thought it would be more readable but if we prefer to have an old name I'll keep it in mind

seeforschauer (dismissed):
LGTM except naming.

w84thesun (dismissed):
LGTM

internal/handlers/pg/msg_update.go
```diff
@@ -103,6 +103,17 @@ func (h *Handler) MsgUpdate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
 			return nil, err
 		}
+
+		// get comment from options.Update().SetComment() method
+		if sp.Comment, err = common.GetOptionalParam(document, "comment", sp.Comment); err != nil {
+			return nil, err
+		}
+
+		// get comment from query, e.g. db.collection.UpdateOne({"_id":"string", "$comment: "test"},{$set:{"v":"foo""}})
+		if sp.Comment, err = common.GetOptionalParam(q, "$comment", sp.Comment); err != nil {
```
rumyantseva, line 113: If a comment is set for both update document and query, we replace the comment from the document? 
Is it on purpose? Or should we concatenate with a previous comment or something like that?

The current implementation is the same as for `msg_find.go` (the comment is rewritten there as well), so the question goes to @AlekSi.

internal/handlers/pg/pgdb/pool.go
```diff
@@ -262,15 +262,24 @@ func (pgPool *Pool) SchemaStats(ctx context.Context, schema, collection string)
 }
 
 // SetDocumentByID sets a document by its ID.
-func (pgPool *Pool) SetDocumentByID(ctx context.Context, db, collection string, id any, doc *types.Document) (int64, error) {
+func (pgPool *Pool) SetDocumentByID(
+	ctx context.Context, db, collection, comment string, id any, doc *types.Document,
```
rumyantseva, line 266: Maybe we can change with signature and use `sp SQLParam` instead of db, collection, and comment? So then, there are fewer parameters for this method.

integration/basic_test.go
```diff
@@ -132,6 +132,51 @@ func TestFindCommentQuery(t *testing.T) {
 	assert.Contains(t, databaseNames, name)
 }
 
+func TestUpdateCommentMethod(t *testing.T) {
```
rumyantseva, line 135: Would it make sense to combine this test and `TestUpdateCommentQuery` into one (with different test cases running as subtests)? Or is there a particular reason to keep those tests separately?

Also, none of the tests cover the case when `comment` is a part of the update doc, but such a code is implemented. If it's a possible case, let's write a test.

rumyantseva (changes requested):

integration/basic_test.go
AlekSi: > none of the tests cover the case when comment is a part of the update doc,

Please expand on that, because there are many "comment"s (`comment`, `$comment`) and many "update doc"s (`update`, `updates`, `u`) :)

internal/handlers/pg/pgdb/pool.go
noisersup: Yeah. I was thinking about that, but then you have an import loop. We could extract SQLParam to the seperate package or leave it as it is.

internal/handlers/pg/msg_update.go
rumyantseva: Is this case possible at all, or is `$comment` only for queries? (I can't find an example with `update` in docs).

integration/basic_test.go
rumyantseva: Ok, I take back my words :)

So, the full list of cases (as far as I understand).

For `db.collection.update(query, update, ..., comment)`
- `$comment` is a part of `query` (is it possible at all?)
- `comment` is a part of the list of `update` arguments

With that, there are 4 possible cases:
- none comments are set
- `$comment` as a part of `query` is set
- `comment` as a separate param is set
- both values are set

I believe we can write one test with four subtests (table test cases) to cover them all.

AlekSi: Let's rephrase it in terms of https://www.mongodb.com/docs/manual/reference/command/update/, not in terms of mongosh method

rumyantseva: Fixed

internal/handlers/pg/pgdb/pool.go
rumyantseva: Why an import loop? There are two packages - `pg` and `pgdb`. If I understand everything correctly, `SQLParam` is part of `pgdb`, `pg` imports `pgdb`, but `pgdb` doesn't impoty `pg` :)

noisersup: My bad :)

internal/handlers/pg/msg_update.go
noisersup: Doesn't that function run only for a single query?

integration/basic_test.go
noisersup: @rumyantseva then should we also implement something similar for the [find method](https://github.com/FerretDB/FerretDB/blob/main/integration/basic_test.go#L106-L133)?

noisersup: Nvm, found out the answer

internal/handlers/pg/msg_update.go
```diff
@@ -223,7 +234,7 @@ func (h *Handler) MsgUpdate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 func (h *Handler) update(ctx context.Context, sp pgdb.SQLParam, doc *types.Document) (int64, error) {
```
rumyantseva, line 234: I'd say here, and in the callers of that func we need to refactor it to `sp *pgdb.SQLParam` too.

rumyantseva (changes requested):
Last tiny correction is needed.

w84thesun (approved):

rumyantseva (approved):
LGTM

PR #944, opened 2022-07-26
`tjson`: Fix schema comparison

### LINKED ISSUE #907: `tjson`: Fix schema comparison

🎯  Currently, schema comparison method `func (s *Schema) Equal(other *Schema) bool` is implemented with `reflect.DeepEqual`. The goal of this task is to come up with a better approach.

The target code is located [here](https://github.com/FerretDB/FerretDB/blob/22f1fdbd0b80bdcaee4ae4486296b5fb7ef54612/internal/tjson/schema.go#L131).

Tigris' data types [consist of type and format](https://docs.tigrisdata.com/overview/schema#data-types). For `int64` and `double` format is not required but might be set.

Having that in mind, the following schemas must be equal:
```go
	doubleSchema = &Schema{
		Type: Number,
	}
	doubleSchemaWithFormat = &Schema{
		Type:   Number,
		Format: Double,
	}
```

As well as:
```go
	int64Schema = &Schema{
		Type: Integer,
	}
	int64SchemaWithFormat = &Schema{
		Type:   Integer,
		Format: Int64,
	}
```

We need to fix comparison to make such schemas equal.

Checklist:
- [x] Have a look at the [Tigris doc](https://docs.tigrisdata.com/overview/schema#data-types) once again to make sure that `int64` and `double` are the only cases that can cause different schemas need to be equal.
- [x] Implement the changes and add tests to check that the comparison works equally for the different schemas that describe the same data type.

rumyantseva:
# Description

This PR closes #907.

This implementation offers recursive calls of `schema.Equal` instead of `reflect.DeepEqual`.

Schemas are considered equal if:
- for arrays: Items' schemas are equal
- for objects: the keys are the same, and their schemas are equal
- for scalars: their types and formats are equal 

## Readiness checklist

### DIFF
--- a/internal/tjson/schema.go
+++ b/internal/tjson/schema.go
@@ -128,9 +128,6 @@ func (s *Schema) Equal(other *Schema) bool {
 		return true
 	}
 
-	// TODO compare significant fields only (ignore title, description, etc.)
-	// TODO compare format according to type (for example, for Number, EmptyFormat == Double)
-	// https://github.com/FerretDB/FerretDB/issues/683
 	return reflect.DeepEqual(s, other)
 }
 
--- a/internal/tjson/schema.go
+++ b/internal/tjson/schema.go
@@ -140,11 +140,11 @@ func (s *Schema) Equal(other *Schema) bool {
 			return false
 		}
 		for k, v := range s.Properties {
-			vother, ok := other.Properties[k]
+			vOther, ok := other.Properties[k]
 			if !ok {
 				return false
 			}
-			if eq := v.Equal(vother); !eq {
+			if eq := v.Equal(vOther); !eq {
 				return false
 			}
 		}
--- a/internal/tjson/schema.go
+++ b/internal/tjson/schema.go
@@ -133,8 +133,9 @@ func (s *Schema) Equal(other *Schema) bool {
 		return false
 	}
 
-	// If `s` and `other` are objects, compare their properties.
-	if s.Type == Object {
+	switch s.Type {
+	case Object:
+		// If `s` and `other` are objects, compare their properties.
 		if len(s.Properties) != len(other.Properties) {
 			return false
 		}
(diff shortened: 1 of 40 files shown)


internal/tjson/schema_test.go
```diff
@@ -67,3 +67,147 @@ func TestSchemaMarshalUnmarshal(t *testing.T) {
...
+		other:    &cArrayObjectsSchemaEqual,
+		expected: false,
+	}, {
+		name:     "ArrayObjectsDouble",
+		s:        &cArrayObjectsSchema,
+		other:    &cArrayDoubleSchema,
+		expected: false,
+	}} {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
```
rumyantseva, line 209: having `t.Parallel` here caused a data race. I believe it's because schemas are pointers, and some of them are reused.

internal/tjson/schema.go
```diff
@@ -123,15 +122,66 @@ var (
...
-	// TODO compare significant fields only (ignore title, description, etc.)
-	// TODO compare format according to type (for example, for Number, EmptyFormat == Double)
-	// https://github.com/FerretDB/FerretDB/issues/683
-	return reflect.DeepEqual(s, other)
+	if s.Type != other.Type {
+		return false
+	}
+
+	// If `s` and `other` are objects, compare their properties.
+	if s.Type == Object {
```
AlekSi, line 137: Let's use `switch s.Type` there, as it is much more common in our code base

internal/tjson/schema_test.go
AlekSi: We should fix a data race, not ignore it. Reading schema pointers in parallel should be safe, and we should not write to them

internal/tjson/schema_test.go
```diff
@@ -67,3 +67,168 @@ func TestSchemaMarshalUnmarshal(t *testing.T) {
...
+		Type:  Array,
+		Items: &cObjectSchemaEqual,
+	}
+	cArrayObjectsSchemaNotEqual := Schema{
+		Type:  Array,
+		Items: &cObjectSchemaNotEqual,
+	}
+
+	for _, tc := range []struct {
+		name     string
```
AlekSi, line 144: I think we should move to `map[string]struct` where key is a test name. Otherwise, we could have duplicate test names, and that's not very easy to spot

internal/tjson/schema.go
```diff
@@ -123,15 +122,66 @@ var (
...
+		return s.Items.Equal(other.Items)
+	}
+
+	if s.Format != other.Format {
+		// Normalizing schemas: empty format is equal to double for numbers and int64 for integers,
+		// see https://docs.tigrisdata.com/overview/schema#data-types.
+		switch s.Type {
+		case Number:
+			if s.Format == EmptyFormat {
+				s.Format = Double
```
AlekSi, line 167: We should not modify receiver in `Equal` method

AlekSi (changes requested):

internal/tjson/schema.go
rumyantseva: Ah, that answers the question of where data race is from! 🙈

internal/tjson/schema.go
rumyantseva: Fixed

internal/tjson/schema.go
rumyantseva: Fixed

internal/tjson/schema_test.go
rumyantseva: Fixed

internal/tjson/schema_test.go
rumyantseva: Fixed

internal/tjson/schema_test.go
```diff
@@ -67,3 +67,152 @@ func TestSchemaMarshalUnmarshal(t *testing.T) {
 
 	assert.Equal(t, expected, actual)
 }
+
+func TestSchemaEqual(t *testing.T) {
+	t.Parallel()
+
+	cInt64Schema := Schema{
```
w84thesun, line 74: What is this `c` prefix mean?

w84thesun (commented):
Overall looks good, got a small question about test.

internal/tjson/schema_test.go
rumyantseva: It means "case" (like "test case").  I didn't find a better word (but suggestions are welcome!)
The idea is to call it differently than just `int64Schema` because such a name already exists in the package.

w84thesun: I spent a few minutes, but nothing comes to mind. Maybe we can replace that prefix with the word `case` to make it more clear?

rumyantseva: Done!

w84thesun (dismissed):

internal/tjson/schema.go
```diff
@@ -123,15 +122,70 @@ var (
...
+	}
+
+	switch s.Type {
+	case Object:
+		// If `s` and `other` are objects, compare their properties.
+		if len(s.Properties) != len(other.Properties) {
+			return false
+		}
+		for k, v := range s.Properties {
+			vother, ok := other.Properties[k]
```
AlekSi, line 143: https://www.ancestry.com/name-origin?surname=vother ?

`vOther`? `otherV`? I think camel case will make it more readable

internal/tjson/schema_test.go
```diff
@@ -67,3 +67,152 @@ func TestSchemaMarshalUnmarshal(t *testing.T) {
...
+
+	for name, tc := range map[string]struct {
+		s        *Schema
+		other    *Schema
+		expected bool
+	}{"StringString": {
+		s:        stringSchema,
+		other:    stringSchema,
+		expected: true,
+	}, "StringNumber": {
```
AlekSi, line 151:
```suggestion
	},
	"StringNumber": {
```

That's not `}, {`

internal/tjson/schema.go
rumyantseva: Fixed

AlekSi (dismissed):

w84thesun (approved):

noisersup (approved):

w84thesun (approved):

PR #951, opened 2022-07-27
`tjson`: Check how we support `binary` and `object_id`

### LINKED ISSUE #900: `tjson`: Check how we support `binary` and `object_id`
This task should be taken after #899 is done.

🎯  The goal of this task is to check if we have a better way to work with `binary` and `object_id` data types.
💡 Tip: it might make sense to check your implementation idea and some other important details with @AlekSi. 

### ObjectID
 
Due to serialization reasons, to avoid double-serialization ObjectID is implemented [in a specific way](https://github.com/FerretDB/FerretDB/blob/fb56a54cee5160dfc641a9097f7b1fd33d2de27e/internal/tjson/tjson.go#L234). We need to check if it's possible to get rid of the `ObjectID` function and make sure that we reuse test cases from bson/fjson.

### Binary

Binary implementation is quite special as well. We need to check that it works correctly with our test cases. 

Definition of Done:

- [ ] Improvements in dealing with these data types are made (if possible).
- [ ] ObjectID and Binary types are covered with tests (unit, integration, fuzz). Test cases can be (and should be) ported from `fjson` and `bson` packages.
- [ ] If integration tests don't exist yet or don't cover these types, create a separate issue to add integration tests for ObjectID and Binary.

rumyantseva:
# Description

This PR closes #900.

## Readiness checklist


rumyantseva:
I close this draft as it was split into smaller PRs.

PR #968, opened 2022-07-30
Tigris: simplify `ObjectID` and filter usage

rumyantseva:
# Description

Ref #900.

In this PR:
- New package `internal/handlers/tigris/tigrisdb/filter` to replace Tigris' filter package which is based on generics and doesn't fit our custom types.
- The path `internal/handlers/tigris/tigrisdb` is chosen this way as it is supposed to be similar to `internal/handlers/pg/pgdb`.
- `tjson.ObjectID` function is deleted as it's not needed.

## Readiness checklist


integration/tigris/object_id_smoke_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+	"go.mongodb.org/mongo-driver/mongo"
+
+	"github.com/FerretDB/FerretDB/integration"
+	"github.com/FerretDB/FerretDB/integration/setup"
+)
+
+// TODO This is a temporary test to check how string identifiers would work.
```
rumyantseva, line 30: @AlekSi should I delete those smoke tests and write compat tests in this task? Or should I create a separate ticket for them?

w84thesun (commented):
Overall looks good. Just a small question.

Did I get it correct that we will use strings as `_id` values to insert them into Tigris?

rumyantseva:
@w84thesun sort of :) From Tigris' perspective it's [type: string, format: byte](https://docs.tigrisdata.com/overview/schema#string) to store binary data. From our perspective, it's still `[12]byte`.

w84thesun (dismissed):

internal/handlers/tigris/msg_update.go
```diff
@@ -196,9 +196,9 @@ func (h *Handler) MsgUpdate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 
 // update updates documents by _id.
 func (h *Handler) update(ctx context.Context, sp fetchParam, doc *types.Document) (int, error) {
-	id := must.NotFail(doc.Get("_id")).(types.ObjectID)
-	f := must.NotFail(filter.Eq("_id", tjson.ObjectID(id)).Build())
-	h.L.Sugar().Debugf("Filter: %s", f)
+	id := must.NotFail(doc.Get("_id"))
+	f := must.NotFail(filter.Eq("_id", id).Build())
```
AlekSi, line 200: I really think that having a separate package with three files, types, functions, builders, etc is too much for a single line: `json.Marshal(map[string]any{"_id": id})`

Let's keep it for now, but I think we should really simplify it when we will work on query pushdown

AlekSi (approved):

integration/tigris/object_id_smoke_test.go
AlekSi: Let's move them the `intergation` package

AlekSi (changes requested):

w84thesun (dismissed):

AlekSi (approved):

w84thesun (approved):

PR #973, opened 2022-08-01
Improve Document's Path API

w84thesun:
# Description

This PR refs #803.

In that PR I added the set of `Document` methods working with `Path`. They would be needed later to implement dot notation support for update operators.

Here we came up with the decision that a `Path` with one element is the correct path and it's a caller's responsibility to check where that case is applicable.

## Readiness checklist


rumyantseva (approved):
the `SetByPath` method is complicated but it looks correct from first sight.
I guess we will add integration tests as well as unit tests, so we will have some more cases / some real cases covered.

internal/types/document.go
```diff
@@ -311,8 +309,83 @@ func (d *Document) Remove(key string) any {
...
 
+// HasByPath returns true if the given path is present in the document.
+func (d *Document) HasByPath(path Path) bool {
+	_, err := d.GetByPath(path)
+
+	return err == nil
+}
+
+// GetByPath returns a value by path - a sequence of indexes and keys.
+// If the Path has only one element, it sets the value for the given key.
```
AlekSi, line 320: That does not sound right

internal/types/document.go
```diff
@@ -311,8 +309,83 @@ func (d *Document) Remove(key string) any {
...
+		}
+
+		must.NoError(inner.Set(index, value))
+	default:
+		panic(fmt.Errorf("can't set value for %T type", inner))
+	}
+}
+
 // RemoveByPath removes document by path, doing nothing if the key does not exist.
+// If the Path has only one element, it sets the value for the given key.
```
AlekSi, line 382: Looks like incorrect copy&paste

AlekSi (changes requested):

internal/types/document.go
w84thesun: Yeah... Updated.

internal/types/document.go
w84thesun: Updated

AlekSi (approved):

rumyantseva (approved):
LGTM

PR #978, opened 2022-08-02
Support `ordered` argument in `delete` command for postgres handler

### LINKED ISSUE #848: Support `ordered` `delete`s
https://www.mongodb.com/docs/manual/reference/command/delete/#std-label-delete-command-ordered

We should support that argument for both PostgreSQL and Tigris.

* write integration test first to check MongoDB behavior: `true`, `false`, with error, without error;
* let's not generalize it for `update` and `find` yet

## TODO:
- [x] Postgres handler.
- [x] Tigris handler.
- [x] Implement proper error handling.
- [x] compatibility tests

noisersup:
# Description

This PR closes #848.

_It needs another PR to make better error handling. For now it returns the last error that occured._

## Readiness checklist


.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
rumyantseva, line 1: I'd say this one should be in `~/.gitignore_global`, see https://github.com/FerretDB/FerretDB/blob/main/.gitignore#L1 ;)

noisersup: I'll remove that before review. I wanted to move this file between workstations and I'm too lazy to use a pendrive :)

noisersup:
### Done:
- Implemented functionality for postgres handler.
- Created tests for error simulation (thanks @rumyantseva!) 
- Splited PR to another two (tigris handler and more detailed tests for differrent errors). 

### Blockers:
- The FerretDB error messages differs by adding prefixes like `(BadValue)`.
- I'm not sure that nested function is the best approach.

build/deps/tigris.Dockerfile
```diff
@@ -1 +1 @@
-FROM tigrisdata/tigris-local:1.0.0-alpha.26
+FROM tigrisdata/tigris-local:1.0.0-alpha.22
```
rumyantseva: Please move it back to `alpha.26` as `alpha.22` has some bugs (e.g. we caught one with `delete`).

integration/delete_test.go
```diff
@@ -66,7 +66,7 @@ func TestDeleteOrdered(t *testing.T) {
 		deletes            bson.A
 		ordered            bool
 		expectedRemovedIDs []string
-		expectedErr        string
+		// TODO: add expected error
```
rumyantseva, line 69: If you plan to do it in a separate issue, let's mention this issue here?

internal/handlers/pg/msg_delete.go
```diff
@@ -153,20 +152,21 @@ func (h *Handler) MsgDelete(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
 				return nil, err
 			}
-			unorderedErrMsgs = append(unorderedErrMsgs, err.Error())
+			lastErr = err
 		}
 	}
 
-	if len(unorderedErrMsgs) > 0 {
-		return nil, fmt.Errorf("write exception: write errors: [%s]", strings.Join(unorderedErrMsgs, ","))
+	// TODO: handle all errors and return them in slice
```
rumyantseva, line 167: I guess, this needs to be done here.

rumyantseva (changes requested):
If I understand the PR correctly, it's not ready for review yet (there are two `todo`s here that look like something that needs to be implemented).

integration/delete_test.go
noisersup: I wanted to resolve it in the same issue, but different PR. Please let me know if you have any objections to this approach

internal/handlers/pg/msg_delete.go
noisersup: I wanted to do this in separate PR because it will probably require to change some code in error.go

integration/delete_test.go
rumyantseva: Yeah, that sounds reasonable!

I see codecov complains about some things, this might be something that needs to be fixed in this PR, or at least we need to describe test cases that would fix it (and we can implement test cases in a separate PR).

I'll take another look on Monday morning to give my final opinion.

(Please, don't work on weekends! I've just shown a really bad example of behavior with doing this review today, I'm sorry for that 🙈)

internal/handlers/pg/msg_delete.go
```diff
@@ -145,11 +149,26 @@ func (h *Handler) MsgDelete(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 			return nil
 		})
 
+		return err
```
rumyantseva, line 152: I believe, if we keep this `processQuery` func here, instead of this return `err` we can just use the `return` statement on the line 103 (where `inTransaction` is called).

internal/handlers/pg/msg_delete.go
```diff
@@ -38,61 +38,65 @@ func (h *Handler) MsgDelete(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
 	}
 
+	var ordered bool
+	if ordered, err = common.GetOptionalParam(document, "ordered", ordered); err != nil {
+		return nil, err
+	}
+
 	var deleted int32
-	for i := 0; i < deletes.Len(); i++ {
+	processQuery := func(i int) error {
```
rumyantseva, line 54: I see this is the function you mentioned. Yeah, I think it's ok to keep it here. It looks quite safe to me.

rumyantseva (dismissed):
I see some TODOs are meant to be implemented in a separate PR of the same issue. If it simplifies cognitive load, I'm ok with it.

w84thesun (approved):

internal/handlers/pg/msg_delete.go
noisersup: Correct! Thank you

AlekSi (changes requested):
Let's review and merge #1002 first, then add compatibility tests

rumyantseva (commented):
Waiting for a PR with error handling.

PR #1004, opened 2022-08-08
Support `ordered` argument for `delete` command

### LINKED ISSUE #848: Support `ordered` `delete`s

noisersup:
# Description

This PR closes #848.

## Readiness checklist


rumyantseva:
Let's close one of the PRs and make all the changes in one PR. 
The PR should contain:
- error handling
- the logic for pg handler
- the logic for tigris handler
- compatibility tests

integration/helpers.go
```diff
@@ -301,7 +302,8 @@ func CollectKeys(t testing.TB, doc bson.D) []string {
 
 // FindAll returns all documents from the given collection.
 func FindAll(t testing.TB, ctx context.Context, col *mongo.Collection) []bson.D {
-	cursor, err := col.Find(ctx, bson.D{})
+	opts := options.Find().SetSort(bson.D{{"_id", 1}})
```
AlekSi, line 305: 👍

integration/delete_compat_test.go
```diff
@@ -61,26 +119,38 @@ func testDeleteCompat(t *testing.T, testCases map[string]deleteCompatTestCase) {
...
 				compatCollection := compatCollections[i]
 				t.Run(targetCollection.Name(), func(t *testing.T) {
 					t.Helper()
 
-					dmm := mongo.NewDeleteManyModel().SetFilter(filter)
-
-					targetRes, targetErr := targetCollection.BulkWrite(ctx, []mongo.WriteModel{dmm}, opts)
-					compatRes, compatErr := compatCollection.BulkWrite(ctx, []mongo.WriteModel{dmm}, opts)
+					var targetRes bson.D
+					targetErr := targetCollection.Database().RunCommand(
```
AlekSi, line 131: Can we keep using BulkWrite?

integration/delete_test.go
```diff
@@ -55,3 +56,145 @@ func TestDeleteSimple(t *testing.T) {
 		})
 	}
 }
+
+// TestDeleteOrdered checks if ordered parameter works properly.
+func TestDeleteOrdered(t *testing.T) {
+	setup.SkipForTigris(t)
```
AlekSi, line 62: Do we still need that test? If yes, why we skip it for Tigris?

integration/delete_compat_test.go
noisersup: As far as I can see there's no Model for RunCommand, which is required to provide a couple of queries (We need a couple of them to easily simulate an error in one of them)

w84thesun (dismissed):
LTGM

integration/delete_compat_test.go
AlekSi: What specific queries we can't run with BulkWrite?

noisersup: We can't run a couple of queries in one request (at least I didn't find a way to do that).

integration/delete_test.go
noisersup: We skip it because tigris has an issues with some types that are in shareddata, however I think that this test might be unnecessary and we can remove it.

internal/handlers/tigris/msg_delete.go
```diff
@@ -47,55 +47,61 @@ func (h *Handler) MsgDelete(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
 		var ok bool
 		if fp.collection, ok = collectionParam.(string); !ok {
-			return nil, common.NewErrorMsg(
+			return common.NewErrorMsg(
 				common.ErrBadValue,
 				fmt.Sprintf("collection name has invalid type %s", common.AliasFromType(collectionParam)),
 			)
 		}
 
 		fetchedDocs, err := h.fetch(ctx, fp)
```
rumyantseva, line 95: It looks like with the current approach, you fetch all the documents for each iteration. I don't think it's reasonable to do it. Let's not fetch all the documents more than once per single query.

rumyantseva (changes requested):

internal/handlers/pg/msg_delete.go
```diff
@@ -38,61 +38,65 @@ func (h *Handler) MsgDelete(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
 	}
 
+	ordered := true
+	if ordered, err = common.GetOptionalParam(document, "ordered", ordered); err != nil {
+		return nil, err
+	}
+
 	var deleted int32
-	for i := 0; i < deletes.Len(); i++ {
+	processQuery := func(i int) error {
```
rumyantseva, line 54: Same here. Currently, inside `processQuery` we fetch the whole database. It doesn't look reasonable. Let's fetch the DB only once.

internal/handlers/pg/msg_delete.go
```diff
@@ -145,11 +149,26 @@ func (h *Handler) MsgDelete(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 			return nil
 		})
 
+		return err
+	}
+
+	unorderedErrMsgs := new(common.WriteErrors)
+
+	for i := 0; i < deletes.Len(); i++ {
```
rumyantseva, line 157: I'd question how optimal this approach is. If `ordered == true`, we don't need this cycle at all, it's enough to just call all the deletes at once (how we did before). We need one-by-one deletion only if `ordered == false`.

Do you think it's possible to refactor it? To me, it seems to be possible, but we might need something similar to how `h.delete` looks like in the Tigris handler. Thoughts @noisersup @w84thesun @AlekSi?

rumyantseva: Or maybe I'm actually wrong! If we call all the deletes in the same transaction and one of them fails, then in case of our pg handler, none of the docs will be deleted (because of transaction). But this is not what we want, right? So, I think, I need more documentation somewhere to understand this logic.

internal/handlers/tigris/msg_delete.go
noisersup: I didn't change any logic in the code inside the loop. All I've done was just extracting it to the nested function. However, as I'm looking at the code I think that it's reasonable to fetch documents for each iteration.

## Example

Let's say for an example that we have a collection with elements: 
```
[
  { _id: "a", v: "foo"}, 
  { _id: "b", v: "foo"}, 
  { _id: "c", v: "bar"}
]
```
We want to run a delete query with  the following filters: 1: `{"_id":"a"}` 2: `{"v":"foo"}`

If we implement document fetching outside the iteration it will execute in that way:
### fetch documents:
**docs**: `[{ _id: "a", v: "foo"}, { _id: "b", v: "foo"}, { _id: "c", v: "bar"}]`

### first iteration
**docs**: `[{ _id: "a", v: "foo"}, { _id: "b", v: "foo"}, { _id: "c", v: "bar"}]`
**query**: `{"_id":"a"}`
**delete will remove field:**: `{ _id: "a", v: "foo"}`

### second iteration
**docs**: `[{ _id: "a", v: "foo"}, { _id: "b", v: "foo"}, { _id: "c", v: "bar"}]`
**query**: `{"_id":"a"}`
**delete will **try to** remove fields:**: `{ _id: "a", v: "foo"}, { _id: "b", v: "foo"}`, but the first one will just try to remove the query that was already removed in the **first iteration**.

Fetching the document on every iteration solves the problem.

rumyantseva: Yeah, I think you are right.

rumyantseva (commented):
Ok, I'm not sure if the changes are needed, but I think we need a final approval from one more person.

internal/handlers/pg/msg_delete.go
noisersup: You right it doesn't look clear. I can try to add some comments, should I cover the whole `MsgDelete` function or just `processQuery` part?

rumyantseva: Let's cover `processQuery` part, yes.

internal/handlers/common/error.go
```diff
@@ -246,13 +246,44 @@ func (we *WriteErrors) Document() *types.Document {
...
+		*we = append(*we, *e)
+		return
+	}
+
+	if e, ok := parseErr[*CommandError](err); ok {
+		*we = append(*we, writeError{err: e.Unwrap().Error(), code: e.code})
+		return
+	}
+
+	*we = append(*we, writeError{err: err.Error(), code: errInternalError})
```
rumyantseva, line 271: I don't understand why it's exactly `errInternalError`. I guess, in the general case, it could be any type of error code?

internal/handlers/pg/msg_delete.go
```diff
@@ -145,11 +154,30 @@ func (h *Handler) MsgDelete(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+	unorderedErrMsgs := new(common.WriteErrors)
+
+	// process every delete filter
+	for i := 0; i < deletes.Len(); i++ {
+		err := processQuery(i)
 		if err != nil {
-			return nil, err
+			unorderedErrMsgs.Append(err)
+
+			// If ordered option is set then return the error,
```
rumyantseva, line 168: Maybe something like:

```
Delete statements in the `deletes` field are not transactional. 
It means that we run each delete statement separately. 
If `ordered` is set as `true`, we don't execute the remaining statements 
after the first failure.
If `ordered` is set as `false`,  we execute all the statements and return 
the list of errors corresponding to the failed statements.
```

rumyantseva (commented):
A couple of comments/questions.

internal/handlers/common/error.go
noisersup: The point is that errors that are not `writeError` or `commandError` don't have any error code attached. When I was talking with @w84thesun we agreed that internalError represents these errors the best. Of course we can try to find another solution.

rumyantseva (dismissed):
I'm still not 100% sure, but as I don't have any new suggestions, I approve the PR.

w84thesun (approved):

AlekSi (changes requested):
Let's merge #1030 first, then fix #1029, then return to that one

AlekSi:
May you post MongoDB's response in that case? It should be visible in the proxy output.

noisersup:
@AlekSi sure:

```
Body diff:
--- res body
+++ proxy body
@@ -7,4 +7,18 @@
         "$k": [
-          "ok",
-          "writeErrors"
+          "n",
+          "writeErrors",
+          "ok"
+        ],
+        "n": 0,
+        "writeErrors": [
+          {
+            "$k": [
+              "index",
+              "code",
+              "errmsg"
+            ],
+            "index": 1,
+            "code": 2,
+            "errmsg": "$all needs an array"
+          }
         ],
@@ -12,13 +26,3 @@
           "$f": 1
-        },
-        "writeErrors": [
-          {
-            "$k": [
-              "code",
-              "errmsg"
-            ],
-            "code": 2,
-            "errmsg": "$all needs an array"
-          }
-        ]
+        }
       },
```

rumyantseva:
@noisersup it's the diff between proxy and "normal" mode, and the response from the proxy should be a little bit above this, look for `Proxy response` in logs.

noisersup:
Yep sorry, I thought that diff would be more meaningful here :)
```
2022-08-16T20:39:14.553+0200    DEBUG   // 172.20.0.4:40642 -> 172.17.0.1:27017         clientconn/conn.go:437  Proxy response header: length:   304, id: 3830, response_to:   48, opcode: OP_MSG
2022-08-16T20:39:14.555+0200    DEBUG   // 172.20.0.4:40642 -> 172.17.0.1:27017         clientconn/conn.go:438  Proxy response message:
{
  "Checksum": 0,
  "FlagBits": 0,
  "Sections": [
    {
      "Document": {
        "$k": [
          "ismaster",
          "topologyVersion",
          "maxBsonObjectSize",
          "maxMessageSizeBytes",
          "maxWriteBatchSize",
          "localTime",
          "logicalSessionTimeoutMinutes",
          "connectionId",
          "minWireVersion",
          "maxWireVersion",
          "readOnly",
          "ok"
        ],
        "ismaster": true,
        "topologyVersion": {
          "$k": [
            "processId",
            "counter"
          ],
          "processId": {
            "$o": "62faef229e622afb8249f3c3"
          },
          "counter": {
            "$l": "0"
          }
        },
        "maxBsonObjectSize": 16777216,
        "maxMessageSizeBytes": 48000000,
        "maxWriteBatchSize": 100000,
        "localTime": {
          "$d": 1660675154552
        },
        "logicalSessionTimeoutMinutes": 30,
        "connectionId": 159,
        "minWireVersion": 0,
        "maxWireVersion": 13,
        "readOnly": false,
        "ok": {
          "$f": 1
        }
      },
      "Kind": 0
    }
  ]
}
```

AlekSi:
Both are from different requests

noisersup:
I've updated both comments with the actual diff and proxy response.

rumyantseva:
@AlekSi basically, the tricky part is that the response has all the `ok: {$f: 1}`, `n` and `writeError`. So, we either need to add `writeError` to the place where we write response (`msg_delete`), or we need to append (not overwrite) response with error in `route`.  I was advocating for doing it in `msg_delete`, but @noisersup reasonably noticed that we might have some other cases with similar behavior too.

AlekSi:
> we either need to add writeError to the place where we write response (msg_delete)

Let's do that, at least, for now

AlekSi (commented):
There are conflicts now

integration/delete_compat_test.go
```diff
@@ -41,18 +41,55 @@ func TestDeleteCompat(t *testing.T) {
...
+				{{"_id", "double"}},
+				{{"v", bson.D{{"$eq", 9}}}},
+			},
+			ordered: false,
+		},
 		"One": {
 			filters: []bson.D{
 				{{"v", int32(42)}},
 			},
-			skip: "https://github.com/FerretDB/FerretDB/issues/1029",
```
noisersup, line 48: @AlekSi should I leave skip fields in this issue?

internal/handlers/pg/msg_delete.go
```diff
@@ -29,6 +29,7 @@ import (
 )
 
 // MsgDelete implements HandlerInterface.
+// It can return both response and error if partial delete was made.
```
rumyantseva, line 32: Not anymore! 
```suggestion
```

internal/handlers/pg/msg_delete.go
```diff
@@ -145,17 +155,46 @@ func (h *Handler) MsgDelete(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+				break
+			}
 		}
 	}
 
-	var reply wire.OpMsg
+	// minimal reply document
+	replyDoc := must.NotFail(types.NewDocument(
+		"n", deleted,
+		"ok", float64(1),
```
rumyantseva, line 186: Is it on purpose that we always respond with `ok: 1` - even if none of docs were deleted / all the delete queries failed?

integration/delete_compat_test.go
```diff
@@ -41,18 +41,55 @@ func TestDeleteCompat(t *testing.T) {
...
+		},
+		"OrderedFalse": {
+			filters: []bson.D{
+				{{"_id", "string"}},
+				{{"v", bson.D{{"$all", 9}}}},
+				{{"_id", "double"}},
+			},
+			ordered: false,
+		},
+		"OrderedFalseTwoErrors": {
```
rumyantseva, line 74: I suggest adding two more cases:
 - Something like `OrderedFalseAllErrors` - when all the statements are not correct.
 - Something like `OrderedTrueAllErrors` - same but for `ordered: true`.

rumyantseva (commented):
Ask a couple of questions and suggest a couple of additional test cases.

internal/handlers/pg/msg_delete.go
noisersup: Nice catch!

internal/handlers/pg/msg_delete.go
noisersup: I was also thinking about that. Tests are passing, but I will add another testcase for that.

rumyantseva (approved):
As the current compat tests pass and I don't have new test cases in mind, I'm ready to approve this PR!

integration/delete_compat_test.go
```diff
@@ -101,7 +165,7 @@ func testDeleteCompat(t *testing.T, testCases map[string]deleteCompatTestCase) {
 						t.Logf("Target error: %v", targetErr)
 						targetErr = UnsetRaw(t, targetErr)
 						compatErr = UnsetRaw(t, compatErr)
-						assert.Equal(t, compatErr, targetErr)
+						assert.EqualError(t, compatErr, targetErr.Error())
```
AlekSi, line 168: We need to compare types too, not just error messages. Please revert this change

integration/delete_compat_test.go
```diff
@@ -41,18 +41,82 @@ func TestDeleteCompat(t *testing.T) {
 			filters:    []bson.D{},
 			resultType: emptyResult,
 		},
+		"OrderedTrue": {
+			filters: []bson.D{
+				{{"_id", "string"}},
```
AlekSi, line 46: The whole point of compatibility tests is to not have IDs hard-coded in the test. We run commands for all data sets and collections. We could depend on data staying the same, but we can't depend on IDs

internal/handlers/common/error.go
```diff
@@ -249,13 +249,44 @@ func (we *WriteErrors) Document() *types.Document {
 	return d
 }
 
+// parseArr converts err to a T type.
+// If it fails, the returned bool value is false.
+func parseErr[T any](err error) (T, bool) {
```
AlekSi, line 254: Please use a regular `errors.As` and remove that helper that mixes interfaces with generics

integration/delete_compat_test.go
```diff
@@ -41,18 +41,82 @@ func TestDeleteCompat(t *testing.T) {
...
+				{{"v", bson.D{{"$all", 9}}}},
+				{{"_id", "double"}},
+			},
+			ordered: true,
+		},
+		"OrderedFalse": {
```
AlekSi, line 52: I think that's already covered by the "Two" test case. Please check if there are other duplicates

internal/handlers/pg/msg_delete.go
```diff
@@ -145,17 +154,46 @@ func (h *Handler) MsgDelete(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+	))
+
+	// if there are delete errors append writErrors field
+	if len(*delErrors) > 0 {
+		protoErr, _ := common.ProtocolError(delErrors)
+		we := must.NotFail(common.AssertType[*types.Array](must.NotFail(protoErr.Document().Get("writeErrors"))))
+		replyDoc.Set("writeErrors", we)
```
AlekSi, line 192: That looks needlessly complicated: we convert `delErrors` to protocol error, then convert it to the document. Why do we need to `delError` to be protocol error at all? Why it can't be just a document or array?

AlekSi (changes requested):

internal/handlers/common/error.go
```diff
@@ -249,11 +255,47 @@ func (we *WriteErrors) Document() *types.Document {
...
+		return
+	}
+
+	if e, ok := parseErr[*CommandError](err); ok {
+		*we = append(*we, writeError{err: e.Unwrap().Error(), code: e.code, index: &index})
+		return
+	}
+
+	*we = append(*we, writeError{err: err.Error(), code: errInternalError, index: &index})
+}
```
noisersup, line 284:
```suggestion
func (we *WriteErrors) Append(err error, index int32) {
	var writeErr *writeError
	ok := errors.As(err, &writeErr)
	if ok {
		writeErr.index = &index
		*we = append(*we, *writeErr)

		return
	}

	var cmdErr *writeError
	ok = errors.As(err, &cmdErr)
	if ok {
		cmdErr.index = &index
		*we = append(*we, *cmdErr)

		return
	}

	*we = append(*we, writeError{err: err.Error(), code: errInternalError, index: &index})
}
```

@AlekSi Is it something you expect in place of `parseErr` function?
I've created it mainly to prevent the code repetition

AlekSi: I think it will look better with a `switch`:

```go
switch {
case errors.As(err, &writeErr):
  writeErr.index = &index
  fallthrough
case errors.As(err, &cmdErr):
  cmdErr.index = &index
  *we = append(*we, err)
  return
}

My main concern with `parseError` is that it mixes regular Go interfaces and type parameters that are represented by interfaces in Go but act very differently.

File: internal/handlers/tigris/msg_delete.go [hunk 19]
```diff
@@ -107,27 +118,61 @@ func (h *Handler) MsgDelete(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+	// minimal reply document
+	replyDoc := must.NotFail(types.NewDocument(
+		"n", deleted,
+		"ok", float64(1),
+	))
+
+	// if there are delete errors append writErrors field
+	if len(*delErrors) > 0 {
+		protoErr, _ := common.ProtocolError(delErrors)
+		we := must.NotFail(common.AssertType[*types.Array](must.NotFail(protoErr.Document().Get("writeErrors"))))
```
rumyantseva, line 170: Let's make it the way how we made it for the `pg` handler.

rumyantseva (changes requested):

internal/handlers/tigris/msg_delete.go
noisersup: nice catch, sorry for overlooking

w84thesun (approved):
LGTM

rumyantseva (approved):
I don't see critical things. We might want to refactor something later, but I think for now, it's good to go.

AlekSi (commented):
@noisersup sorry, there are merge conflicts now 😅

rumyantseva (approved):
LGTM

w84thesun (approved):

AlekSi (approved):

PR #1005, opened 2022-08-08
`tjson`: Support `null`

### LINKED ISSUE #906: `tjson`: Support `null`
This task should be taken after #899 is done.

🎯 The goal of the task is to support `null` type for `tjson`.
💡 One possible option to solve this task is to write integration tests first.

### Null in Tigris

This is how `null` values [work in Tigris](https://docs.tigrisdata.com/overview/schema#null-values):
- Primary key's fields can't be set as `null`.
- Any other field can be explicitly set as `null` in the JSON payload.
- If the field is set as `null` it will be stored as `null`.
- If the field is not set in the JSON payload, it is not stored in Tigris.
- If the field doesn't have existing schema but the value is set as null, the field will not occur in the schema

### Readiness checklist

- [x] Add `null`-related unit tests to the `tjson` package. Any field could be nullable if it's not a part of the primary key.
- [x] Add a nullable field to one of the test cases in the `document_test.go` file.
- [ ] The implemented cases (when `null` is possible and when `null` is not possible) need to be covered with integration tests. Dance tests are needed for the cases that are Tigris-specific.
    - [ ] Cases to cover:
        - [ ] Insert, upsert, update, unset
        - [ ] Queries
- [ ] Add special tests for the case when the field is not set in the payload (insert, upsert, queries etc)
    - [ ] Three documents in the collection: one has the field `A` and it's set as `string`, second has the field `A` and it's set as `null`, third doesn't have the field `A`.
- [x] Check that all shared data sets have Tigris schemas set (including composites!)

rumyantseva:
# Description

This PR closes #906.

## Test cases

These test cases represent the current logic implemented in Tigris.

### Insert

- Existing collection, existing field
  - [x] Can be set as `null` if the field is not a PK (compat test)
  - [ ] Can't be set as `null` if the field is PK  
  - [x] If the existing field is not set, it is not stored for the entry
- Existing collection, new field
  - [x] Can be set as `null`, will not occur in the schema
- New collection, new field
  - [x] Can be set as `null`, will not occur in the schema 

### Update

- Update entry and set existing field as null
  - [x] Currently, Tigris returns an error in such a case (bug on the Tigris side, see #1019) 
- Update entry and set a new field as null
  - [x] As the field is not present in the schema, Tigris will return an error.
- Update entry and don't set the field 
  - [x] The field is in the schema but is not present in the payload. Possible case, the field that is not present in the payload will not be modified.

## Readiness checklist


integration/commands_administration_test.go
```diff
@@ -682,7 +682,7 @@ func TestCommandsAdministrationDataSize(t *testing.T) {
 		assert.Equal(t, float64(1), must.NotFail(doc.Get("ok")))
 		assert.InDelta(t, float64(16_384), must.NotFail(doc.Get("size")), 16_384)
 		assert.InDelta(t, float64(4), must.NotFail(doc.Get("numObjects")), 4) // TODO https://github.com/FerretDB/FerretDB/issues/727
-		assert.InDelta(t, float64(100), must.NotFail(doc.Get("millis")), 100)
```
noisersup, line 685: Shouldn't that be in seperate PR?

rumyantseva: As the tests started failing in this PR, I believe it should be in this PR :)
I guess that those tests started failing because there is more data in this PR (all the extra collections and values for nil), so 100 is not enough anymore.

w84thesun (dismissed):
LGTM

integration/setup/setup_compat.go
```diff
@@ -163,15 +163,29 @@ func setupCompatCollections(tb testing.TB, ctx context.Context, client *mongo.Cl
...
+			switch err := err.(type) {
+			case nil:
+				// do nothing
+
+			case mongo.CommandError:
+				// If collection can't be created in MongoDB because MongoDB has a different validator format, it's ok.
+				// Otherwise, it's a problem:
+				if !strings.Contains(err.Message, `unknown top level operator: $tigrisSchemaString`) {
+					tb.Errorf("Failed to create collection %q: %v", collectionName, err)
+					tb.FailNow()
```
AlekSi, line 182:
```suggestion
					fallthrough
```

integration/shareddata/composites.go
```diff
@@ -52,39 +52,66 @@ var Composites = &Values[string]{
...
 		"document-double":          bson.D{{"v", 42.13}},
 		"document-double-whole":    bson.D{{"v", 42.0}},
 		"document-double-zero":     bson.D{{"v", 0.0}},
 		"document-double-max":      bson.D{{"v", math.MaxFloat64}},
 		"document-double-smallest": bson.D{{"v", math.SmallestNonzeroFloat64}},
 		"document-double-big":      bson.D{{"v", doubleBig}},
-		// TODO Dealing with empty doc needs a schema to be defined https://github.com/FerretDB/FerretDB/issues/772
-		// "document-empty":           bson.D{},
+		"document-double-empty":    bson.D{},
+		"document-double-nil":      nil,
```
AlekSi, line 68:
```suggestion
		"document-double-null":      nil,
```

That's the suffix we use in our data sets

integration/setup/setup_compat.go
rumyantseva: @AlekSi that doesn't compile with "cannot fallthrough in type switch".

AlekSi: `require.Contains` then?

rumyantseva: yeah, that looks better to me :)

rumyantseva: fixed

AlekSi (approved):

noisersup (approved):

PR #1035, opened 2022-08-12
support multiple modules for linters

### LINKED ISSUE #557: Make `golangci-lint` check integration tests
Our current configuration does not actually check `integration` module.

ravilushqa:
# Description

This PR closes #557 .

Run linters in all modules except tools (., ./integration)

## Readiness checklist


AlekSi (changes requested):
@ravilushqa Thank you. Unfortunately, that would require too many UNIX tools for our contributors on Windows. We would prefer to use per-module Taskfiles that already exist. Would you like to do that?

ravilushqa:
> @ravilushqa Thank you. Unfortunately, that would require too many UNIX tools for our contributors on Windows. We would prefer to use per-module Taskfiles that already exist. Would you like to do that?

Ah, didn't think about that 🙄. I can do it in a more "hard-code" way, so we won't need any Unix commands. WDYT?

AlekSi:
See:
* https://github.com/FerretDB/FerretDB/blob/8d3027ebd80a3a7400b4d996fe5d05290f91ba15/Taskfile.yml#L240-L244
* https://github.com/FerretDB/FerretDB/blob/main/integration/Taskfile.yml

ravilushqa:
I checked your example and implement it in the same way, I guess. Is it what you expected?
Also, do you want to get rid of .golangci-new.yml? and write exclude rules in .golangci.yml?

AlekSi:
> Also, do you want to get rid of .golangci-new.yml? and write exclude rules in .golangci.yml?

`.golangci-new.yml` contains new linters that check only new code: https://github.com/FerretDB/FerretDB/blob/2786a4d33e98be0bae6bcdb1810083704ee1f07b/.golangci-new.yml#L128

For example, we decided to enable `wsl` linter, but fixing all code will take a lot of time, so we do enforce it only for new code.

That's not about exclusion rules.

Taskfile.yml
```diff
@@ -228,6 +228,7 @@ tasks:
     cmds:
       - bin/golangci-lint{{exeExt}} run --config=.golangci-new.yml
       - go vet -vettool=./bin/checkswitch{{exeExt}} ./...
+      - bin/task{{exeExt}} -d integration lint
```
AlekSi, line 231: That should be dependency in the `deps` section above. Dependencies are run in parallel, so `task lint` works faster.

On that line we should check `.golangci-new.yml` rules for integration module

integration/Taskfile.yml
```diff
@@ -26,3 +26,10 @@ tasks:
       - ../bin/go-consistent{{exeExt}} -v -pedantic ./...
     sources:
       - "**/*.go"
+
+  lint:
+    cmds:
+      - go vet -vettool=../bin/checkswitch{{exeExt}} ./...
+      - ../bin/golangci-lint{{exeExt}} run ./... --config=../.golangci-new.yml
```
AlekSi, line 33: .golangci-new.yml and .golangci.yml should be in two separate tasks just like in the parent Taskfile

AlekSi (changes requested):

ravilushqa:
Depending on your review, the code will be:
Taskfile.yml
```make
...
  lint:
    desc: "Run linters"
    deps:
      - lint-golangci-lint
      - lint-golangci-lint-integration
      - lint-go-sumtype
      - lint-go-consistent
    cmds:
      - bin/golangci-lint{{exeExt}} run --config=.golangci-new.yml
      - go vet -vettool=./bin/checkswitch{{exeExt}} ./...
      - bin/task{{exeExt}} -d integration lint

  lint-golangci-lint-integration:
    cmds:
      - bin/task{{exeExt}} -d integration lint-golangci-lint
...
```

integration/Taskfile.yml
```make
...
  lint:
    cmds:
      - go vet -vettool=../bin/checkswitch{{exeExt}} ./...
      - ../bin/golangci-lint{{exeExt}} run --config=../.golangci-new.yml
    sources:
      - "**/*.go"

  lint-golangci-lint:
    cmds:
      - ../bin/golangci-lint{{exeExt}} run --config=../.golangci.yml
    sources:
      - "**/*.go"
...
```

running golangci-lint with ../.golangci.yml will raise a lot of errors, as you know. Do you expect in this PR to skip all these failing linters for integration?

AlekSi:
> Do you expect in this PR to skip all these failing linters for integration?

I think yes, as a temporary measure

ravilushqa:
> > Do you expect in this PR to skip all these failing linters for integration?
> 
> I think yes, as a temporary measure

Done, let me know if something need to be improved

.golangci-new.yml
```diff
@@ -3,6 +3,7 @@
 
 run:
   timeout: 3m
+  allow-parallel-runners: true
```
AlekSi, line 6: We don't use separate cache directories for separate runners, so using that option is not safe: golangci/golangci-lint#1769

That actually makes it clear that our complex way to run golangci-lint in parallel does not actually achieve that :) So let's simplify that by removing `deps` on other golangci-lint tasks

.golangci.yml
```diff
@@ -166,3 +167,43 @@ issues:
...
       text: "`go.mongodb.org/mongo-driver/bson/primitive.E` composite literal uses unkeyed fields"
+
+    # integration tests exclusions
+    - path: _test\.go
+      text: "composites: go.mongodb.org/mongo-driver/bson/primitive.E struct literal uses unkeyed fields"
+      linters: [govet]
+    - path: _test\.go
+      linters: [lll]
+    - path: _test\.go
+      text: "import of package `go.mongodb.org/mongo-driver/*"
```
AlekSi, line 178: That's very unfortunate. On the one hand, we don't want driver dependency in the main module: #853. On the other, it is not possible to specify "integration" directory there.

I'm not sure what the answer is there. Use separate configuration files and sync them manually? Do something with YAML includes?.. Meh.

@rumyantseva WDYT?

ravilushqa: @AlekSi yes, this is a little weird. IMO would be clean enough to create separate configuration file for the integration module.

AlekSi: Ok, let's use separate configuration files for the integration module without any templating for now.

@ravilushqa Wanna finish that one for Hacktoberfest? :)

ravilushqa: Ah, I missed that Hacktoberfest started 😄 
Sure, it would be great.
However looks like I have to create a new PR, isn't it?

> Your PR/MRs must be created between October 1 and October 31 (in any time zone, UTC-12 thru UTC+14).
Your PR/MRs must be made to a public, unarchived repository.

AlekSi: I think it will be easier, yeah

AlekSi:
Done in #1200

Issue #1036, opened 2022-08-13
Improve contribution guidelines


ravilushqa:
I started my contribution by looking at https://github.com/FerretDB/FerretDB/blob/main/CONTRIBUTING.md and moving step by step. And first step `task init` is failing by default.

The problem is with a command `git describe --tags --dirty > gen/version.txt` git describe return error on forks with default mark `Copy the main branch only`. I think would be helpful to describe this behavior in CONTRIBUTING.md, so that new contributors will not face that issues

AlekSi:
Sounds great, let's update our docs

AlekSi:
The checkbox in question:

![CleanShot 2022-08-22 at 17 43 38@2x](https://user-images.githubusercontent.com/11512/185949736-35d667d0-4f47-41f9-b4a2-905609f3c725.png)

PR #1039, opened 2022-08-15
[1036] Use --always to fallback to abbreviated hash

subkanthi:
closes: #1036 
Use --always with git describe to fallback to an abbreviated tag.

https://stackoverflow.com/questions/4916492/git-describe-fails-with-fatal-no-names-found-cannot-describe-anything

[* ] I ran `task all`, and it passed.


AlekSi (changes requested):
That would generate a version that does not look like a version at all. I think we should update our docs instead

PR #1050, opened 2022-08-17
`tjson`: Implement `regex`

### LINKED ISSUE #911: `tjson`: Implement `regex` type

 🎯 The goal of the task is to implement `regex` type.

As there is no special `regex` type in Tigris, the implementation would be similar to how `regex` is implemented in the `fjson` package.

To Do:
- [x] Learn or refresh your memory about how`regex` is implemented in `fjson`.
- [x] Reuse tests for `regex` from the `fjson` package.
- [ ] Add a regex field to one of the test cases in the `document_test.go` file of the tjson package.
- [x] Implement the necessary functionality.

noisersup:
# Description

This PR closes #911.

## Readiness checklist


w84thesun (approved):
LGTM.

internal/tjson/schema.go
```diff
@@ -57,6 +57,7 @@ const (
 	Byte     SchemaFormat = "byte"
 	UUID     SchemaFormat = "uuid"
 	DateTime SchemaFormat = "date-time"
+	Regex    SchemaFormat = "regex"
```
rumyantseva, line 60: Hmm, there is no "regex" format in Tigris' Schema Specification. Could you explain why it is here and how we are going to use it? I expect, if we try to create such a schema, Tigris will return an error. Have you tried it?

rumyantseva (changes requested):
I'm not sure about "regex format" (asked a question in a particular part of code).

internal/tjson/schema.go
noisersup: This may be resulted from my still little knowledge about Tigris :smile:. The tests are passing as far as I understand they need to create some schema. Should I remove that line or just leave it with empty quotes?

rumyantseva: I don't understand why you need this line at all :) We store regex as an object. Where is this format from, and how do you use it? If you don't use it at all, let's not invent it.

noisersup: You're right

internal/tjson/tjson.go
```diff
@@ -177,7 +177,7 @@ func Unmarshal(data []byte, schema *Schema) (any, error) {
 			fallthrough
 		case Int64, Int32:
 			fallthrough
-		case Byte, UUID, DateTime:
+		case Byte, UUID, DateTime, Regex:
```
rumyantseva, line 180: I guess, we don't need `Regex` here

internal/tjson/tjson.go
```diff
@@ -198,7 +198,7 @@ func Unmarshal(data []byte, schema *Schema) (any, error) {
 			res = &o
 		case UUID:
 			fallthrough
-		case Double, Float, Int64, Int32:
+		case Double, Float, Int64, Int32, Regex:
```
rumyantseva: I guess, we don't need `Regex` here either

internal/tjson/tjson.go
```diff
@@ -162,7 +162,7 @@ func Unmarshal(data []byte, schema *Schema) (any, error) {
 			res = &o
 		case Double, Float:
 			fallthrough
-		case Byte, UUID, DateTime:
+		case Byte, UUID, DateTime, Regex:
```
rumyantseva: I guess, we don't need `Regex` here

internal/tjson/tjson.go
noisersup: Yep I know it even cannot compile when it's called here :)

noisersup: Sometimes I push code and run `task all` in parallel to save a few minutes and after checks are passed on CI I rerequest for review :) Sorry for that

rumyantseva:
@noisersup let's maybe also add a regex to `tjson/document_test.go`? There is a test case called `all` that supposed to contain all the supported types, so regex should be there too.

rumyantseva (approved):
LGTM

w84thesun (approved):

PR #1055, opened 2022-08-18
Small `pgdb` cleanup

AlekSi:
Small cleanup before bigger changes.

## Readiness checklist


internal/handlers/pg/pgdb/databases.go
```diff
@@ -133,3 +133,26 @@ func CreateDatabaseIfNotExists(ctx context.Context, querier pgxtype.Querier, db
 		return lazyerrors.Error(err)
 	}
 }
+
+// DropDatabase drops FerretDB database.
+//
+// It returns ErrTableNotExist if schema does not exist.
```
rumyantseva, line 139:
```suggestion
// It returns ErrSchemaNotExist if schema does not exist.
```

rumyantseva (changes requested):
One piece of doc is not correct (see my suggestion) and linters fail, but overall the changes look reasonable.

internal/handlers/pg/pgdb/collections.go
```diff
@@ -160,6 +160,40 @@ func CreateCollection(ctx context.Context, querier pgxtype.Querier, db, collecti
...
+	}
+
+	// Table (or even schema) does not exist. Try to create it,
+	// but keep in mind that it can be created in concurrent connection.
+
+	if err := CreateDatabase(ctx, querier, db); err != nil && !errors.Is(err, ErrAlreadyExist) {
+		return false, lazyerrors.Error(err)
+	}
+
+	// TODO use a transaction instead of pgPool: https://github.com/FerretDB/FerretDB/issues/866
```
rumyantseva, line 185: Is this comment still actual? It seems that we already use querier here.

AlekSi: But we don't use a transation

rumyantseva: Ok, got it (now I see how transaction is used in update).

rumyantseva (approved):
LGTM

PR #1062, opened 2022-08-22
Improve `TestCommandsAdministrationServerStatus`

### LINKED ISSUE #1001: Improve `TestCommandsAdministrationServerStatus`
Currently, as part of `TestCommandsAdministrationServerStatus` we have the following check:
```go
assert.InDelta(t, float64(1), must.NotFail(catalogStats.Get("collections")), 1)
```

As tests are run in parallel, this check might fail. In some other tests, we create up to 10 collections (for a new data type each). If they are called first, in this case we have more than one collection.

To solve this problem, let's use a separate database in `TestCommandsAdministrationServerStatus`.

Also, check an order of arguments in `assert.InDelta` in this test.

rumyantseva:
# Description

This PR closes #1001.

It seems to me that `ServerStatus` should work across all the databases, and there is no such a param as `db` when you call `ServerStatus`.  So, the fix for this task is very different than what we thought in the beginning. Instead of only fixing the test, I fixed the logic of `ServerStatus` for the `pg` handler.

Please correct me if I'm wrong.

## Readiness checklist


internal/handlers/pg/pgdb/pool.go
```diff
@@ -199,13 +199,19 @@ func (pgPool *Pool) SchemaStats(ctx context.Context, schema, collection string)
...
       JOIN pg_indexes                AS i ON i.schemaname = t.table_schema
-                                         AND i.tablename = t.table_name
-     WHERE t.table_schema = $1`
+                                         AND i.tablename = t.table_name`
 
-	args := []any{schema}
-	if collection != "" {
-		sql = sql + " AND t.table_name = $2"
-		args = append(args, collection)
+	// TODO should we exclude service schemas and tables?
```
rumyantseva, line 204: @AlekSi, if this func is called for `serverStatus` (so we calculate the stats across the whole DB), should we exclude any service schemas/tables? Smth like ` WHERE  t.table_schema != 'magic_ferret_service_schema'?

integration/commands_administration_test.go
```diff
@@ -832,8 +832,8 @@ func TestCommandsAdministrationServerStatus(t *testing.T) {
 	catalogStats, ok := must.NotFail(doc.Get("catalogStats")).(*types.Document)
 	assert.True(t, ok)
 
-	// TODO This check needs to be fixed as the number of collections might vary https://github.com/FerretDB/FerretDB/issues/1001
-	assert.InDelta(t, float64(1), must.NotFail(catalogStats.Get("collections")), 1)
+	// catalogStats is calculated across all the databases, so there could be quite a lot of collections here.
```
rumyantseva, line 835: What I actually tried:
- run `serverStatus` on MongoDB with mongosh for a couple of different DBs with different number of collections
- the result in `catalogStats` was the same

So, yes, I didn't check if `$db` param is set, but my conclusion was that it's not respected.

internal/handlers/pg/msg_serverstatus.go
```diff
@@ -51,7 +40,8 @@ func (h *Handler) MsgServerStatus(ctx context.Context, msg *wire.OpMsg) (*wire.O
 
 	uptime := time.Since(h.startTime)
 
-	stats, err := h.pgPool.SchemaStats(ctx, db, "")
+	// TODO it needs to be calculated for all schemas
```
noisersup, line 43: Shouldn't we create an issue for that?

noisersup (approved):
Overall LGTM, I have only one question

internal/handlers/pg/msg_serverstatus.go
rumyantseva: We shouldn't because it's already implemented in this ticket :) This `TODO` is not needed, I was writing it for myself and forgot to delete :D

rumyantseva: Fixed

noisersup (approved):

integration/commands_administration_test.go
AlekSi: Interestingly, it also works with non-existing database ¯\_(ツ)_/¯

internal/handlers/pg/pgdb/pool.go
```diff
@@ -199,13 +199,20 @@ func (pgPool *Pool) SchemaStats(ctx context.Context, schema, collection string)
...
+
+	args := []any{}
+
+	if schema != "" {
```
AlekSi, line 208: We should update method's documentation

internal/handlers/pg/pgdb/pool.go
AlekSi: What is service schema?

For tables – yes, we should skip tables that start with `reservedPrefix`.

Not sure if we want to do in that PR

rumyantseva: @AlekSi well, it seems to me that `reservedPrefix` is used for both cases with db and collection creation. With that, it seems that there are some "service schemas" or something :D

internal/handlers/pg/pgdb/pool.go
rumyantseva: Fixed

internal/handlers/pg/pgdb/pool.go
rumyantseva: Reworded everything and fixed TODO.

w84thesun (approved):

PR #1066, opened 2022-08-23
Require `limit` parameter in `delete` command

### LINKED ISSUE #982: Require `limit` in delete query
The goal of this task is to require a `limit` field with numerical value of 0 or 1 while running the `delete` query. 

Currently we allow sending the delete query without this field, which is unexpected behavior. See: https://www.mongodb.com/docs/v4.2/reference/command/delete/

 [The Example of the test that is failing for MongoDB and not for FerretDB](https://github.com/FerretDB/FerretDB/runs/7652477027?check_suite_focus=true).

Don't make a test too complicated.

noisersup:
# Description

This PR closes #982.

## Readiness checklist


integration/delete_test.go
```diff
@@ -55,3 +55,18 @@ func TestDeleteSimple(t *testing.T) {
...
+	ctx, collection := setup.Setup(t)
+	cmd := bson.D{
+		{"delete", collection.Name()},
+		{"deletes", bson.A{bson.D{{"q", bson.D{{"v", "foo"}}}}}},
+	}
+
+	expectedErr := "(Location40414) BSON field 'delete.deletes.limit' is missing but a required field"
+
+	res := collection.Database().RunCommand(ctx, cmd)
+	assert.EqualError(t, res.Err(), expectedErr)
```
AlekSi, line 71: We should compare error values, both types and content, not error strings. See other tests that use `mongo.CommandError`

AlekSi (changes requested):

integration/delete_test.go
AlekSi: See also `AssertEqualError` helper

integration/delete_test.go
```diff
@@ -55,3 +56,22 @@ func TestDeleteSimple(t *testing.T) {
 		})
 	}
 }
+
+func TestDeleteLimitNotSet(t *testing.T) {
```
w84thesun, line 60: Do we want to add other cases like pass valid values, pass `float64`?

w84thesun (commented):
Overall looks good, asked the question about test cases.

integration/delete_test.go
noisersup: We could. I'm going to improve that then

internal/handlers/pg/msg_delete.go
```diff
@@ -63,10 +63,17 @@ func (h *Handler) MsgDelete(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+		l, err := d.Get("limit")
+		if err != nil {
+			return nil, common.NewErrorMsg(
+				common.ErrMissingField,
+				"BSON field 'delete.deletes.limit' is missing but a required field",
+			)
+		}
+
+		if limit, err = common.GetWholeNumberParam(l); err != nil {
+			return nil, err
```
rumyantseva, line 76: Is it enough to just return `err`? Shouldn't it be something like `NewErrorMsg(ErrBadValue, err)`?

Let's write a test that would generate this error? I guess, it could be something like `limit` is a string or a value like `42.13`.

internal/handlers/tigris/msg_delete.go
```diff
@@ -66,10 +66,17 @@ func (h *Handler) MsgDelete(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+
+		l, err := d.Get("limit")
+		if err != nil {
+			return nil, common.NewErrorMsg(
+				common.ErrMissingField,
+				"BSON field 'delete.deletes.limit' is missing but a required field",
+			)
+		}
+
+		if limit, err = common.GetWholeNumberParam(l); err != nil {
```
rumyantseva, line 78: Same comment as for `pg`.

rumyantseva (changes requested):
It seems to be that `bad_value` is missing if `limit` is not a correct number (see the comments).

internal/handlers/pg/msg_delete.go
```diff
@@ -68,11 +68,21 @@ func (h *Handler) MsgDelete(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+			return common.NewErrorMsg(
+				common.ErrMissingField,
+				"BSON field 'delete.deletes.limit' is missing but a required field",
+			)
+		}
+
+		if limit, err = common.GetWholeNumberParam(l); err != nil {
+			return common.NewErrorMsg(
+				common.ErrFailedToParse,
+				fmt.Sprintf("The limit field in delete objects must be 0 or 1. Got %v", l),
```
rumyantseva, line 84: And what if I set limit as 100? :)

internal/handlers/tigris/msg_delete.go
```diff
@@ -211,3 +233,13 @@ func (h *Handler) delete(ctx context.Context, fp tigrisdb.FetchParam, docs []*ty
 
 	return len(ids), nil
 }
+
+// respondAsBulkWriteException calls the fun. If fun returns
+// not-nil error then it is wrapped with lazyerrors.Error.
+func respondAsBulkWriteException(fun func() error) error {
```
rumyantseva, line 239: I don't like the word "Exception" here as I don't see what is "Exception" here.

rumyantseva (commented):
I'd suggest sitting together and rewriting some things as this is not quite an exact thing we discussed at the standup yesterday.

internal/handlers/pg/msg_delete.go
noisersup: Nice catch!

rumyantseva (approved):
LGTM for now. We might want to refactor something later (e.g. when we have transactions for Tigris), but I'd say for now this is enough.

w84thesun (approved):

PR #1086, opened 2022-08-25
(WIP) - refact: use iterator to fetch documents instead channels

### LINKED ISSUE #898: Offer a better design for fetching documents (iterator)

---

nicolascb:
# Description

This PR closes #898.

## Readiness checklist


nicolascb:
Hello @rumyantseva!

I'm working to implement this issue #898, but I wanted an opinion, the way I built it, always load all [documents](https://github.com/nicolascb/FerretDB/blob/refactor/issue/898/internal/handlers/pg/pgdb/query.go#L225) to memory.

That makes sense? This was also final behavior with channels, correct?

Suggestions on how I can improve are welcome! :D

nicolascb:
> @nicolascb thank you very much for contributing to FerretDB! Before we implemented a channel-based solution, we had been fetching all the documents and applying the filter after that. Obviously, it was taking too much memory. The idea of the channel was to fetch documents in small batches, apply the filter to those small batches, and only leave the documents that pass the filter. So, with that, I expect the iterator to work in a similar way - we should apply the filter right after we fetch the doc.

Thanks for the answer!

I'm already thinking about how to implement it, thank you very much!

internal/handlers/pg/pgdb/query_iterator.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+)
+
+type queryRowsIterator interface {
+	Close()
+	Next() bool
+	Err() error
+	Scan(...interface{}) error
+}
```
AlekSi, line 34: Let's try to use a simpler interface like the one being proposed at https://github.com/golang/go/discussions/54245

AlekSi (changes requested):

rumyantseva (commented):
Thanks for sending the PR! I agree with @AlekSi's comments.

AlekSi:
@nicolascb, we updated the #898 description to mention a new interface. Will you have time to help us with it?

PR #1091, opened 2022-08-28
Implement simple query pushdown for Tigris

AlekSi:
# Description

## Readiness checklist

### DIFF
--- a/internal/handlers/tigris/tigrisdb/query.go
+++ b/internal/handlers/tigris/tigrisdb/query.go
@@ -16,6 +16,7 @@ package tigrisdb
 
 import (
 	"context"
+	"encoding/json"
 
 	"github.com/tigrisdata/tigris-client-go/driver"
 	"go.uber.org/zap"
@@ -23,12 +24,18 @@ import (
 	"github.com/FerretDB/FerretDB/internal/tjson"
 	"github.com/FerretDB/FerretDB/internal/types"
 	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+	"github.com/FerretDB/FerretDB/internal/util/must"
 )
+
+const pushdownEnabled = true
 
 // FetchParam represents options/parameters used by the fetch/query.
 type FetchParam struct {
 	DB         string
 	Collection string
+
+	// Query filter for possible pushdown; may be ignored in part or entirely.
+	Filter *types.Document
 }
 
 // QueryDocuments fetches documents from the given collection.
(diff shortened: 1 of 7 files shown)


.github/workflows/integration.yml
```diff
@@ -31,10 +31,33 @@ jobs:
...
+          - name: "PostgreSQL (no pushdown)"
+            task: "pg"
+            buildtags: "ferretdb_tigris,ferretdb_nopushdown"
+          - name: "Tigris (no pushdown)"
+            task: "tigris"
+            buildtags: "ferretdb_tigris,ferretdb_nopushdown"
+          - name: "Tigris main (no pushdown)"
+            task: "tigris"
+            buildtags: "ferretdb_tigris,ferretdb_nopushdown"
+            tigris_dockerfile: "tigris_main"
```
noisersup, line 60: I don't understand the purpose of not using pushdown

internal/handlers/tigris/tigrisdb/query.go
```diff
@@ -79,3 +87,38 @@ func (tdb *TigrisDB) QueryDocuments(ctx context.Context, param FetchParam) ([]*t
...
+		}
+
+		// filter by the exact _id value
+		id := must.NotFail(tjson.Marshal(v))
+		f := map[string]any{"_id": json.RawMessage(id)}
+
+		return must.NotFail(json.Marshal(f))
+	}
+
+	return driver.Filter(`{}`)
```
w84thesun, line 123: I think we could simplify that `for` cycle a bit and make last return statement actually return the filter.
```suggestion

	filter := map[string]any{}

	for k, v := range filter.Map() {
		// filter only by _id for now
		if k != "_id" {
			continue
		}

		switch v.(type) {
		case string:
			// filtering by string values is complicated if the storage supports encodings, collations, etc,
			// but Tigris doesn't support any of these
		case types.ObjectID:
			// filtering by ObjectID is always safe
		default:
			// skip other types for now
			continue
		}

		// filter by the exact _id value
		id := must.NotFail(tjson.Marshal(v))
		filter["_id"] = json.RawMessage(id)
	}

	return must.NotFail(json.Marshal(filter))
```

w84thesun (changes requested):
Overall looks good, just a small suggestion for the `queryDocumentsFilter` method.

.github/workflows/integration.yml
AlekSi: The reason was to do extra cross-testing, but that's already covered by our testing against MongoDB

internal/handlers/tigris/tigrisdb/query.go
AlekSi: Done with a small tweak – `filter` is the argument name

w84thesun (approved):
LGTM

rumyantseva (approved):
I guess you checked (maybe manually) that we really go to the database :)

LGTM overall.

PR #1094, opened 2022-08-30
Check Tigris schema on updates and inserts

### LINKED ISSUE #1061: Tigris: check schema for insert and update operations
Currently, when we fetch data from Tigris, we check that the data correspond to the schema. We must do it as we want to unmarshal data to the right types.

On the other hand, when we insert or update, we don't have schema validation on our side.
We should add schema validation so it works the same way.

Motivation: it's not enough to rely on Tigris schema validation (there was already a case when we were able to insert but weren't able to fetch back).

```[tasklist]
### Relevant issues
- [ ] #1830
- [x] #1835
```

To Do: 
- [x] write a function in the tjson package to validate data by schema.
- [x] call this function before performing insert/update operations.
- [ ] add schema to shared data definitions.
- [x] write an integration (Tigris only) test to check how invalid data is validated.
- [ ] Check how we can use this validation to skip unsuitable compat tests, see https://github.com/noisersup/FerretDB/blob/35b050fd7ce29614d018b2f5fd091e546eb6db42/integration/update_compat_test.go#L128-L133
- [ ] Check #1835, it has a nice example of broken validation

Related: https://github.com/FerretDB/FerretDB/issues/1835

w84thesun:
# Description

This PR closes #1061.

## Readiness checklist

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -53,7 +53,7 @@ env:
 
 jobs:
   build:
-    name: Build
+    name: Build image
     runs-on: ubuntu-20.04
 
     if: >
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,5 +1,5 @@
 ---
-name: Packages
+name: Docs
 on:
   pull_request:
     types:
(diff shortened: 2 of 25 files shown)


w84thesun:
As we discussed with Elena the proper validation requires too much effort, so we close this one.

PR #1107, opened 2022-09-02
Record incoming data for fuzzing

### LINKED ISSUE #1102: Record incoming data for fuzzing
When FerretDB is started with `-test-record=<DIR>` flag, it should record all data coming from connected clients in files (one per connection) in the given directory. That directory should be created if it does not exist.

Recording should be done on the lowest possible layer, [right after we read bytes from the connection and before we pass them to the packet decoder](https://github.com/FerretDB/FerretDB/blob/11c1a243d474e9e33dfe714c38ef9b06697d4ba0/internal/clientconn/conn.go#L142-L162). [`io.MultiWriter`](https://pkg.go.dev/io#MultiWriter) or [MultiReader](https://pkg.go.dev/io#MultiReader) may help with that.

We should enable that mode in our integration tests. We will use that data later for fuzzing.

Logging should stay as it is now. This issue is not about logging.

noisersup:
# Description

This PR closes #1102 .

## Readiness checklist


cmd/ferretdb/main.go
```diff
@@ -151,7 +152,7 @@ func main() {
 
 	prometheus.DefaultRegisterer.MustRegister(l)
 
-	err = l.Run(ctx)
+	err = l.Run(ctx, *testRecordF)
```
AlekSi, line 155: That should be in `NewListenerOpts`

internal/clientconn/conn.go
```diff
@@ -139,8 +143,23 @@ func (c *conn) run(ctx context.Context) (err error) {
...
+	if testRecordPath != "" {
+		//nolint:nosnakecase// os package require to use underscore
+		f, err := os.OpenFile(testRecordPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o644)
+		if err != nil {
+			return err
+		}
+
+		r := io.TeeReader(c.netConn, f)
+		bufr = bufio.NewReader(r)
+	} else {
```
AlekSi, line 157: Please drop `else`

internal/clientconn/conn.go
```diff
@@ -139,8 +143,23 @@ func (c *conn) run(ctx context.Context) (err error) {
 		close(done)
 	}()
 
-	bufr := bufio.NewReader(c.netConn)
+	var bufr *bufio.Reader
+	// if test record path is set, split netConn reader to write to file and bufr
+	if testRecordPath != "" {
+		//nolint:nosnakecase// os package require to use underscore
+		f, err := os.OpenFile(testRecordPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o644)
```
AlekSi, line 150: We also should create directory if it doesn't exist

internal/clientconn/conn.go
```diff
@@ -140,7 +146,28 @@ func (c *conn) run(ctx context.Context) (err error) {
...
+
+		//nolint:nosnakecase// os package require to use underscore
+		f, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o744)
+		if err != nil {
+			return err
+		}
+
+		r := io.TeeReader(c.netConn, f)
+		bufr = bufio.NewReader(r)
+	}
```
noisersup, line 167: @AlekSi  in place of `else` statement - do we want to redeclare bufr variable like here, or use switch statement?

internal/clientconn/conn.go
```diff
@@ -140,7 +146,27 @@ func (c *conn) run(ctx context.Context) (err error) {
...
 
 	bufr := bufio.NewReader(c.netConn)
+
+	// if test record path is set, split netConn reader to write to file and bufr
+	if c.testRecordPath != "" {
+		if err := os.MkdirAll(c.testRecordPath, 0o644); err != nil {
+			return err
+		}
+
+		path := filepath.Join(c.testRecordPath, "record.log")
```
w84thesun, line 156: We should create different file names for every client (one for connection). Right now we will save all data in `record.log`. We can use `c.netConn` to get client remote addr and add current time to create sort of unique file name.

w84thesun (changes requested):

internal/clientconn/conn.go
noisersup: Nice catch, forgot about that :smile:

w84thesun (approved):

internal/clientconn/conn.go
```diff
@@ -140,7 +146,29 @@ func (c *conn) run(ctx context.Context) (err error) {
...
 
 	bufr := bufio.NewReader(c.netConn)
+
+	// if test record path is set, split netConn reader to write to file and bufr
+	if c.testRecordPath != "" {
+		if err := os.MkdirAll(c.testRecordPath, 0o644); err != nil {
+			return err
+		}
+
+		filename := fmt.Sprintf("%s_%s.log", c.netConn.RemoteAddr().String(), time.Now().Format("2006-02-01_15:04:05"))
```
AlekSi, line 156:
```suggestion
		filename := fmt.Sprintf("%s_%s.bin", c.netConn.RemoteAddr().String(), time.Now().Format("2006-02-01_15:04:05"))
```

internal/clientconn/conn.go
```diff
@@ -140,7 +146,29 @@ func (c *conn) run(ctx context.Context) (err error) {
...
+
+		path := filepath.Join(c.testRecordPath, filename)
+
+		//nolint:nosnakecase// os package require to use underscore
```
AlekSi, line 160:
```suggestion
		//nolint:nosnakecase // os package require to use underscore
```

internal/clientconn/conn.go
```diff
@@ -140,7 +146,29 @@ func (c *conn) run(ctx context.Context) (err error) {
...
+		f, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o744)
+		if err != nil {
+			return err
+		}
+
+		r := io.TeeReader(c.netConn, f)
```
AlekSi, line 166: 👍

internal/clientconn/conn.go
```diff
@@ -140,7 +146,29 @@ func (c *conn) run(ctx context.Context) (err error) {
...
+		if err := os.MkdirAll(c.testRecordPath, 0o644); err != nil {
+			return err
+		}
+
+		filename := fmt.Sprintf("%s_%s.log", c.netConn.RemoteAddr().String(), time.Now().Format("2006-02-01_15:04:05"))
+
+		path := filepath.Join(c.testRecordPath, filename)
+
+		//nolint:nosnakecase// os package require to use underscore
+		f, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o744)
```
AlekSi, line 161: I think that should be just `os.Create`: we never append to existing file thanks to unique timestamp file name, we don't care about permissions mask beyond umask, etc.

internal/clientconn/conn.go
```diff
@@ -140,7 +146,29 @@ func (c *conn) run(ctx context.Context) (err error) {
...
+		if err != nil {
+			return err
+		}
+
```
AlekSi, line 165: ⚠️ We should close that file; otherwise, some data may be lost.

That looks like a right place to do it there with `defer f.Close()`, but please check it

AlekSi (changes requested):
We should explicitly close files open for writing

internal/clientconn/conn.go
AlekSi: I think the current code is fine

internal/clientconn/conn.go
noisersup: Nice catch, definitely something that shouldn't be forgotten. I checked and it looks like everything works fine with defer in this place.

internal/clientconn/conn.go
```diff
@@ -140,7 +146,31 @@ func (c *conn) run(ctx context.Context) (err error) {
...
 
 	bufr := bufio.NewReader(c.netConn)
+
+	// if test record path is set, split netConn reader to write to file and bufr
+	if c.testRecordPath != "" {
+		if err := os.MkdirAll(c.testRecordPath, 0o755); err != nil {
+			return err
+		}
+
+		filename := fmt.Sprintf("%s_%s.bin", time.Now().Format("2006-02-01_15:04:05"), c.netConn.RemoteAddr().String())
```
noisersup, line 156: I've changed filenaming to start with a time and not address to make it more sorted.

internal/clientconn/conn.go
```diff
@@ -140,7 +146,31 @@ func (c *conn) run(ctx context.Context) (err error) {
...
+
+		path := filepath.Join(c.testRecordPath, filename)
+
+		//nolint:nosnakecase // os package require to use underscore
```
w84thesun, line 160:
```suggestion
```

cmd/ferretdb/main.go
```diff
@@ -42,6 +42,7 @@ var (
 	proxyAddrF  = flag.String("proxy-addr", "127.0.0.1:37017", "proxy address")
 	debugAddrF  = flag.String("debug-addr", "127.0.0.1:8088", "debug address")
 	modeF       = flag.String("mode", string(clientconn.AllModes[0]), fmt.Sprintf("operation mode: %v", clientconn.AllModes))
+	testRecordF = flag.String("test-record", "", "location of record file with binary data coming from connected clients")
```
w84thesun, line 45: Let's be a bit more specific 😃 
```suggestion
	testRecordF = flag.String("test-record", "", "directory of record files with binary data coming from connected clients")
```

w84thesun (changes requested):

w84thesun (approved):

AlekSi (approved):
:shipit:

Issue #940, opened 2022-07-25
Support `ordered` `insert`s 


AlekSi:
After #1618 (It's needed to handle postgres IF NOT EXISTS issues).

When inserting documents with `ordered` set to `true` (which should be set as a default value), insert documents one-by-one in seperate transactions, skip invalid documents (for example, with `_id` values that are already present) and return errors just for them.

With `ordered` set to `false`, attempt to insert all documents in one transaction.

See https://docs.mongodb.com/manual/reference/method/db.collection.insertMany/

https://github.com/FerretDB/FerretDB/pull/1118

**Details:**

- Use `InTransactionRetry` if needed

**DoD:**

- [x] `ordered` param is supported
- [x] Check for both pg and Tigris. If Tigris doesn't work, create a separate issue.
- [x] Update supported commands doc - https://docs.ferretdb.io/reference/supported_commands/
- [x] Write compat tests. For example, we can attempt to insert a few docs with one invalid (e.g. with a regex as `_id` - see `InsertIDRegex` test case).

noisersup:
I believe that this one is pretty similar to #200 and don't understand why do we have 2 issues for that.

PR #1118, opened 2022-09-06
✨ feat: support ordered insert

si3nloong:
# Description

Previously, the insert statement doesn't respect `ordered` flag. After the change, it will construct a **INSERT Multiple Rows** SQL statement instead of multiple single **INSERT** statements.

This PR closes [#940](https://github.com/FerretDB/FerretDB/issues/940). 

## Readiness checklist


AlekSi (commented):
> After the change, it will construct a INSERT Multiple Rows SQL statement instead of multiple single INSERT statements.

I'm afraid that's not how MongoDB handles ordered inserts. But let's see if we can reproduce the problem with tests. Could you please add tests similar to https://github.com/FerretDB/FerretDB/blob/main/integration/delete_compat_test.go

si3nloong:
Screenshot 2022-09-09 at 1 38 44 PM

@AlekSi I think what you said is correct, I made some mistakes there, but I will amend it and add tests.

rumyantseva (commented):
I like the idea of having tests for ordered insert. I'll take another look later.

AlekSi:
#1673

Issue #1101, opened 2022-09-01
Document validation research


AlekSi:
We should investigate how we can rework and improve data validation.

We want to be able to validate incoming documents and reject invalid values (e.g., NaNs) and invalid type combinations (e.g., nested arrays).

We probably want to be able to validate incoming command documents and data documents separately. For example, we don't want to handle nested arrays in data, but they are perfectly fine in commands.

Validation probably should happen after unmarshalling. For example, FerretDB should be able to decode NaN values in incoming documents (using `bson` and `wire` packages) but reject them after that. That should simplify testing and improve error messages.

Similarly, `fjson` package should be able to represent NaN values while we continue to use it for logging.

All existing validation code probably should be reworked or removed.

See https://github.com/FerretDB/FerretDB/labels/validation

w84thesun:
Tasks to be done:

- [ ] #1000 
- [ ] #693 
- [ ]

w84thesun:
The only one thing that left is: #666.
We should either: 
* deprecate duplicate keys in command documents 
* rewrite Document type to support duplicate keys
* introduce new type to support duplicate keys

rumyantseva:
Added @w84thesun's tasks to the epic #871.

PR #1124, opened 2022-09-07
Support `$max` field update operator

### LINKED ISSUE #624: Support `$max` field update operator
https://www.mongodb.com/docs/manual/reference/operator/update/max/

noisersup:
# Description
- [x] functionality
  - [x] handle integers
  - [x] handle all scalar types
  - [x] handle dates
  - [x] handle other types
  - [x] make sure that mixed types are handled correctly
  - [x] implement lexicographic order if string
- [x] integration tests
- [x] compat tests

This PR closes #624 .

## Readiness checklist


integration/update_field_test.go
```diff
@@ -671,6 +671,178 @@ func TestUpdateFieldInc(t *testing.T) {
 	})
 }
 
+func TestUpdateFieldMax(t *testing.T) {
```
w84thesun, line 674: Can we move those cases to compat test and remove this one?

w84thesun (changes requested):
Overall good, please fix linter.

integration/update_field_test.go
noisersup: Then what's the point of other Tests in this file (like `TestUpdateFieldInc` for an example?

noisersup: I've updated compat tests, but I think that we still need integration tests because compat tests for both handlers are skipped for now because of document comparison.

integration/update_field_compat_test.go
```diff
@@ -54,6 +58,86 @@ func TestUpdateFieldCompatInc(t *testing.T) {
...
+			update: bson.D{{"$max", bson.D{{"v", -54.32}}}},
+		},
+
+		"MultipleQueries": {
+			update: bson.D{{"$max", bson.D{{"v", int32(39)}, {"a", int32(30)}}}},
+		},
+		"DuplicateQuery": {
+			update: bson.D{{"$max", bson.D{{"v", int32(39)}, {"v", int32(30)}}}},
+			skip:   "Handle duplicates correctly",
+		},
```
noisersup, line 96:
```
expected: mongo.WriteException(mongo.WriteException{WriteConcernError:(*mongo.WriteConcernError)(nil), WriteErrors:mongo.WriteErrors{mongo.WriteError{Index:0, Code:40, Message:"Updating the path 'v' would create a conflict at 'v'", Details:bson.Raw(nil), Raw:bson.Raw(nil)}}, Labels:[]string(nil), Raw:bson.Raw(nil)})
actual  : mongo.CommandError(mongo.CommandError{Code:0, Message:"connection(127.0.0.1:37525[-4]) socket was unexpectedly closed: EOF", Labels:[]string{"NetworkError", "RetryableWriteError"}, Name:"", Wrapped:topology.ConnectionError{ConnectionID:"127.0.0.1:37525[-4]", Wrapped:(*errors.errorString)(0xc00007a120), init:false, message:"socket was unexpectedly closed"}, Raw:bson.Raw(nil)})
```

Do we have any issue related to that? The closest one I found is #1101.

integration/update_field_test.go
noisersup: Nevermind, integration tests removed.

internal/handlers/common/update.go
```diff
@@ -328,6 +334,45 @@ func processIncFieldExpression(doc *types.Document, updateV any) (bool, error) {
...
+		// if the key is not found, set it
+		if val != nil {
+			res := types.CompareOrder(val, maxVal, types.Ascending)
+			switch res {
+			case types.Equal:
+				fallthrough
+			case types.Greater:
+				continue
+			case types.Less:
+			case types.Incomparable:
```
noisersup, line 362: Should I leave an empty `types.Less` case or make a nolint here?

internal/handlers/common/update.go
```diff
@@ -328,6 +334,45 @@ func processIncFieldExpression(doc *types.Document, updateV any) (bool, error) {
...
+	for _, field := range maxExpression.Keys() {
+		val, _ := doc.Get(field)
+
+		maxVal, err := maxExpression.Get(field)
+		if err != nil {
+			// if max field does not exist, don't change anything
+			continue
+		}
+
+		// if the key was not found, set it
```
w84thesun, line 353: The comment doesn’t describe what’s happening in that if statement. Inside that if we are calling Compare and analyse compare results.

internal/handlers/common/update.go
w84thesun: You could add a comment like “document value less than update one, set the new value”

w84thesun (changes requested):

internal/handlers/common/update.go
noisersup: But it describes what's happening when condition is not met and the whole motivation behind this if statement. But I can rephrase it to "if the document value was found, compare it with max value"

w84thesun (approved):

internal/types/compare_order.go
```diff
@@ -64,6 +64,8 @@ func detectDataType(value any) compareTypeOrderResult {
 		return booleanDataType
 	case time.Time:
 		return dateDataType
+	case *Document:
+		return documentDataType
```
AlekSi, line 68: It should be above array to follow our order

integration/update_field_compat_test.go
AlekSi: > socket was unexpectedly closed: EOF

The real error is in the log above

integration/update_field_compat_test.go
```diff
@@ -54,6 +58,86 @@ func TestUpdateFieldCompatInc(t *testing.T) {
 	testUpdateCompat(t, testCases)
 }
 
+func TestUpdateFieldCompatMax(t *testing.T) {
+	t.Parallel()
+	setup.SkipForTigrisWithReason(t, "tigris does not support dynamic types")
```
AlekSi, line 63: What does it mean and why we skip all tests for Tigris? We should not do that

AlekSi (changes requested):

integration/update_field_compat_test.go
noisersup: When you specify a max value in a type other than existing in database, it will compare them and overwrite based on type comparison order. In tigris, as far as I know we cannot set the value with different type than existing (i.e. types are not dynamic)

AlekSi: So we can't test that with Tigris at all and have to skip all tests?

noisersup: Definitely there should be some basic tigris testing. Although I'm not sure if there's any way to modify compat tests to exclude some collections for specified test

AlekSi: https://github.com/FerretDB/FerretDB/blob/be84709aa9d9363bf566485a25ea34006e9aa223/integration/update_compat_test.go#L128-L133

Those tests should already be skipped by that code

integration/update_field_compat_test.go
noisersup: > The real error is in the log above
```
[document.go:242 bson.(*Document).ReadFrom] bson.Document.ReadFrom: types.ConvertDocument: types.Document.validate: keys and values count mismatch: 1 != 2
```
I still believe that key duplicates are handled incorrectly

integration/update_field_compat_test.go
noisersup: Thanks, it works!

integration/update_field_compat_test.go
```diff
@@ -54,6 +56,85 @@ func TestUpdateFieldCompatInc(t *testing.T) {
...
+		},
+		"Double": {
+			update: bson.D{{"$max", bson.D{{"v", 54.32}}}},
+		},
+		"DoubleNegative": {
+			update: bson.D{{"$max", bson.D{{"v", -54.32}}}},
+		},
+
+		"MultipleQueries": {
+			update: bson.D{{"$max", bson.D{{"v", int32(39)}, {"a", int32(30)}}}},
```
rumyantseva, line 88: Do I understand correctly that a test case where we set `$max` to a field that doesn't exist is covered by this case too?

I mean something like
```
update: bson.D{{"$max", bson.D{{"a", int32(30)}}}},
```

rumyantseva (commented):
Looks good to me overall, asked one question.

integration/update_field_compat_test.go
noisersup: It looks like it is (we compare modifiedCount and collection values) but to be more detailed I can add another test case. It will be better for debugging. Thanks!

noisersup:
Tigris converts incoming data to underlaying type of existing value in database, for an example when we run [Int32Lower](https://github.com/FerretDB/FerretDB/pull/1124/files#diff-02beb83969c6feb8b48d15aebcd299647e0aa2e8b1e5c531e74eb2c4f636ad85R63-R65) on double-smallest collection, we send the int32 data to the tigris and tigris on the backend converts the int to the float type, which results in failing test:
```
--- expected
+++ actual
@@ -6,3 +6,5 @@
   "_id": "double-smallest",
-  "v": 30
+  "v": {
+    "$f": 30
+  }
```

@w84thesun said that there were some efforts to check the type before sending it to tigris but it wasn't that easy.

@AlekSi Should we skip tests for tigris here?

noisersup:
Test for tigris are not failing locally, there are issues with tigris build

integration/update_field_compat_test.go
```diff
@@ -54,6 +56,97 @@ func TestUpdateFieldCompatInc(t *testing.T) {
...
+			update:        bson.D{{"$max", bson.D{{"v", false}}}},
+			skipForTigris: "fixing tigris validation needed",
+		},
+		"EmptyOperand": {
+			update:     bson.D{{"$max", bson.D{}}},
+			resultType: emptyResult,
+		},
+		"DateTime": {
+			update:        bson.D{{"$max", bson.D{{"v", primitive.NewDateTimeFromTime(time.Date(2021, 11, 1, 12, 18, 42, 123000000, time.UTC))}}}},
+			skipForTigris: "fixing tigris validation needed",
```
rumyantseva, line 139: I don't think it could be fixed at all :) 

So, if you want this test to work for Tigris, you must work with a data set where v is defined as a date. Tigris doesn't allow to store multiple types in the same field. 

A possible solution would be to separate those test cases into a few different functions based on data (e.g. one function for bools, another function for doubles, another function for date etc). Each function will call `setup.Setup` for a particular data provider (separate for bools, separate for ints etc). This way it will work for Tigris.

However, such test cases will limit the number of situations we check for Postgres (e.g. for postgres we might want to set a date to the field that stored bools). I believe this should be a separate test case for postgres only (where we  can use "skip for Tigris" and set a particular reason that Tigris doesn't support such data types). 

Let's discuss if everyone agrees with this approach.

AlekSi: > So, if you want this test to work for Tigris, you must work with a data set where v is defined as a date.

Correct. We have that dataset (`DateTimes`).

> Tigris doesn't allow to store multiple types in the same field.

That's also correct.

> A possible solution

Sorry, but for what problem?

rumyantseva: > Sorry, but for what problem?

For the case when we want to check that storing various data types in the same field works correctly for Postgres.

rumyantseva: @AlekSi or I don't understand your question. This is why sometimes I prefer synchronous calls over such discussions.

noisersup: If I understand correctly you wan to move these test cases to integration tests?

AlekSi: That should be handled by `testUpdateCompat`. In fact, there is already a code for that:

https://github.com/noisersup/FerretDB/blob/35b050fd7ce29614d018b2f5fd091e546eb6db42/integration/update_compat_test.go#L128-L133

It might not work, but then we should just™ fix it

rumyantseva: So, to resume:
- if we want to test it with Tigris, we need to split those test cases into groups based on particular data types
- if after that we want to test that storing various data types still works for postgres, we need to add an additional postgres-only test for it.

noisersup: @AlekSi the point is that tigris doesn't return any error in that case, because if it's possible they parse your input to the current existing type

rumyantseva: @noisersup as agreed at the daily sync, please mention this issue https://github.com/FerretDB/FerretDB/issues/1061 in "skipForTigris", and that will be it for this task.

rumyantseva (approved):
LGTM

w84thesun (approved):
LGTM

rumyantseva:
@AlekSi in principle, we would like to merge it, but GitHub requires your approval as you had requested changes in this PR.

AlekSi (approved):

PR #1136, opened 2022-09-12
Add missing `$k` to the schema when creating collection in Tigris

rumyantseva:
# Description

In the `tjson` package when we encode documents (objects), we always add the extra field `$k` that stores the order of the fields. However, when we create collection with the given schema (`create_msg`), we don't add this `$k` field to the schema. As the result, we can create a collection but we can't insert into it later (as `$k` field is not allowed).

This PR fixes the bug.

See also [this community slack's thread](https://ferretdb.slack.com/archives/C02NWN1K3MJ/p1662989823378489) for more context.

## Readiness checklist


internal/tjson/schema.go
```diff
@@ -212,6 +212,25 @@ func (s *Schema) Unmarshal(b []byte) error {
 	return nil
 }
 
+// AddDocumentProperties adds missing $k properties to all the schema's objects (top-level and nested).
+func (s *Schema) AddDocumentProperties() {
```
rumyantseva, line 216: Alternatively, this could be unexproted and could be called from `Unmarshal`. I'm not sure which idea is better.

w84thesun: Is it possible to call that method from somewhere else?

rumyantseva: This is what I do currently - I call it from somewhere else :) 

The question is if we should allow it to be called from somewhere else, or we should add this "missing" JSON schema part right in the place where we unmarshal user-based schema.

To be honest, I'd prefer to add it to `Unmarshal`, but I'm not sure about it.

w84thesun: Let's add it to `Unmarshal`. If it would be necessary we always can make exported later on.

w84thesun (changes requested):
Unit test failed

rumyantseva:
@w84thesun I wonder why they started failing just now... (there are some pointers compared in them, so how did they pass before?..)

w84thesun (approved):

noisersup (approved):

internal/tjson/schema.go
```diff
@@ -212,9 +212,38 @@ func (s *Schema) Unmarshal(b []byte) error {
...
 
+	// Add $k properties that are necessary for documents.
+	s.addDocumentProperties()
+
 	return nil
 }
 
+// addDocumentProperties adds missing $k properties to all the schema's documents (top-level and nested).
+func (s *Schema) addDocumentProperties() {
+	if s.Type != Object && s.Type != "" {
```
AlekSi, line 223: That doesn't cover arrays of objects, right?

rumyantseva: This is only for either "objects" or high-level objects (when `Type` might be unset).

rumyantseva: Arrays might be needed too, yes.

AlekSi (approved):

PR #1152, opened 2022-09-15
Refactor `msg_delete` handlers

### LINKED ISSUE #1106: Remove incorrect type assertion in `delete` handlers
Our current `delete` command handlers work almost by accident. First, we create a `processQuery` closure that handles a single `delete` argument and returns either unwrapped (for arguments check) or wrapped (for actually changing data) `CommandError`. Then we check the error type using a regular type assertion that does not work with wrapped error.

It is even more problematic in the `tigris` handler, where we have this:

https://github.com/FerretDB/FerretDB/blob/edad1f885374ee2234798a46be216672807f04c4/internal/handlers/tigris/msg_delete.go#L237-L245

By wrapping an error, we make the type assertion work in the same way as `pg`'s `InTransaction` helper.

We should fix all that. One way to do it:
1. Extract `processQuery` closure into a method.
2. Split it into two (one would check parameters, the other would perform the actual delete) that will handle errors differently.

We should also enable `errorlint` linter for all code (#1105 enables it for the new code only) and fix all existing invalid error type assertions.

w84thesun:
# Description

This PR closes #1106.

## Readiness checklist


noisersup (approved):
LGTM

internal/handlers/pg/msg_delete.go
```diff
@@ -50,152 +50,62 @@ func (h *Handler) MsgDelete(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
 	}
 
-	var deleted int32
-	processQuery := func(i int) error {
-		// get document with filter
-		d, err := common.AssertType[*types.Document](must.NotFail(deletes.Get(i)))
-		if err != nil {
-			return err
-		}
+	sp := new(pgdb.SQLParam)
```
AlekSi, line 53: I think we agreed to standardise on
```go
var sp pgdb.SQLParam
// ...
foo(&sp)
```

https://dave.cheney.net/practical-go/presentations/gophercon-singapore-2019.html#_use_a_consistent_declaration_style

noisersup (approved):

AlekSi (approved):

PR #1166, opened 2022-09-20
fix: `docker-compose`

si3nloong:
# Description

Let **docker-compose** wait for database services to start using `depends_on` keywords.


AlekSi (commented):
Why we need that?

si3nloong:
> Why we need that?

Sorry, it was just my computer having some weird start-up, but I think the handling in `envtool/main.go` is good enough, I will close this.

PR #1168, opened 2022-09-21
Fuzz `wire` package with recorded data

### LINKED ISSUE #1140: Fuzz `wire` package with recorded data
When FerretDB is run with `task run` or similar commands, it stores recorded client sessions in the `records/` directory. We should use them in our `wire` package fuzzing. See `fuzzMessages` function in tests.

We should read all files in this directory, read wire messages in each file and use `f.Add` to add them to the seed corpus one-by-one.

No changes in CI configuration, etc.

noisersup:
# Description

This PR closes #1140.

## Readiness checklist


internal/wire/records_test.go
```diff
@@ -0,0 +1,72 @@
...
+	return resMsgs, nil
+}
+
+func FuzzRecords(f *testing.F) {
+	msgs, err := fetchRecords("./records/")
+	if err != nil {
+		f.Error(err)
+	}
+	f.Log(msgs)
+	fuzzMessages(f, msgs)
```
noisersup, line 71: Should we reuse `fuzzMessages` for that?

internal/wire/records_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func fetchRecords(recordsPath string) ([]testCase, error) {
+
+	// Fetch resursively every file path with ".bin" extension from recordsPath directory
```
noisersup, line 14: Do we prefer to fetch files recursively from directory or not? I think that's nice because in the future we could seperate logs from different sessions into seperate directories but it's up to you.

noisersup:
Should these tests be placed in seperate file like here, or I didn't catch something in issue?

internal/wire/records_test.go
AlekSi: Sure, let’s do it.

Not sure I would use a verb “fetch”, though. Read? Load?

internal/wire/records_test.go
```diff
@@ -0,0 +1,72 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Read every record file, parse their content to wire messages
+	// and store them in the testCase struct
+	var resMsgs []testCase
+	for _, path := range recordFiles {
+		f, err := os.OpenFile(path, os.O_RDONLY, 0o666)
```
AlekSi, line 35: Just `os.Open`. And `f ` is never closed.

internal/wire/records_test.go
AlekSi: I think we should do the opposite and use fetchRecords from inside fuzzMessages.

https://github.com/FerretDB/FerretDB/blob/33a94873e6c16324f7b1dfbf79be2489e30baf5e/internal/wire/wire_test.go#L118

After that loop, call `fetchRecrods `and `f.Add` more data to the seed corpus

internal/wire/records_test.go
```diff
@@ -0,0 +1,72 @@
...
+		for {
+			header, body, err := ReadMessage(r)
+			if err == io.EOF {
+				break
+			}
+			if err != nil {
+				return resMsgs, err
+			}
+
+			headBytes := []byte(header.String())
```
AlekSi, line 51: Why `.String()`? We need raw bytes

internal/wire/records_test.go
noisersup: I don't understand, I open reader for every file how `os.Open` would differ from that?

AlekSi: os.Open has less parameters :)

AlekSi (changes requested):
> I marked all done items in this checklist.

Please do that

internal/wire/records_test.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var resMsgs []testCase
+
+	// Read every record file, parse their content to wire messages
+	// and store them in the testCase struct
+	for _, path := range recordFiles {
+		f, err := os.Open(path)
```
AlekSi, line 53: We never call `f.Close()`, but we should

internal/wire/records_test.go
```diff
@@ -0,0 +1,92 @@
...
+				return nil, err
+			}
+
+			resMsgs = append(
+				resMsgs,
+				testCase{
+					headerB: headBytes,
+					bodyB:   bodyBytes,
+				},
+			)
```
AlekSi, line 87:
```suggestion
			resMsgs = append(resMsgs, testCase{
				headerB: headBytes,
				bodyB:   bodyBytes,
			})
```

internal/wire/wire_test.go
```diff
@@ -118,6 +120,20 @@ func fuzzMessages(f *testing.F, testCases []testCase) {
...
 	}
 
+	records, err := loadRecords(filepath.Join("..", "..", "records"))
+
+	// if err is not nil and not "no such file or directory" return it
+	switch {
+	case os.IsNotExist(err):
+		f.Logf("No records file will be added to the corpus: %s", err.Error())
+	case err != nil:
+		f.Error(err)
```
AlekSi, line 130: That should be fatal error; `f.Error` does not terminate the test

internal/wire/wire_test.go
```diff
@@ -118,6 +120,20 @@ func fuzzMessages(f *testing.F, testCases []testCase) {
 		f.Add(tc.expectedB)
 	}
 
+	records, err := loadRecords(filepath.Join("..", "..", "records"))
+
+	// if err is not nil and not "no such file or directory" return it
+	switch {
+	case os.IsNotExist(err):
```
AlekSi, line 127: I think that's not possible, or should not be possible. `loadRecords` should return empty or nil slice and nil error if there are not files. Any other error should be fatal

AlekSi (changes requested):

internal/wire/records_test.go
```diff
@@ -0,0 +1,97 @@
...
+// loadRecords gets recursively all .bin files from the recordsPath directory,
+// parses their content to wire Msgs and returns them as an array of testCase
+// structs with headerB and bodyB fields set.
+// If no records are found, it returns an empty slice and no error.
+func loadRecords(recordsPath string) ([]testCase, error) {
+	// Load recursively every file path with ".bin" extension from recordsPath directory
+	var recordFiles []string
+
+	err := filepath.WalkDir(recordsPath, func(path string, entry fs.DirEntry, err error) error {
+		if err != nil {
```
w84thesun, line 35: Do we need that error check here?

internal/wire/records_test.go
```diff
@@ -0,0 +1,97 @@
...
+			bodyBytes, err := body.MarshalBinary()
+			if err != nil {
+				return nil, err
+			}
+
+			resMsgs = append(resMsgs, testCase{
+				headerB: headBytes,
+				bodyB:   bodyBytes,
+			},
+			)
```
w84thesun, line 92:
```suggestion
			})
```

w84thesun (changes requested):
Overall good, small changes needed.

internal/wire/records_test.go
noisersup:
```go
func WalkDir(root string, fn fs.WalkDirFunc) error {
	info, err := os.Lstat(root)
	if err != nil {
		err = fn(root, nil, err)
	} else {
		err = walkDir(root, &statDirEntry{info}, fn)
	}
	if err == SkipDir {
		return nil
	}
	return err
}

```
It pass the error to the function if Lstat failed. So, yeah I think that's important to check for it

w84thesun: Yes, that's right but I can't see `Lstat` call 😃

noisersup: `Walkdir` does it in the first line. If it fails, it pass an error to the function provided as an argument

AlekSi: `err` is passed to the walk function for advanced error handling, but `if err != nil { return err }` is the most common implementation. In fact, I don't think I ever had any other implementation :D

internal/wire/wire_test.go
```diff
@@ -118,6 +119,19 @@ func fuzzMessages(f *testing.F, testCases []testCase) {
 		f.Add(tc.expectedB)
 	}
 
+	records, err := loadRecords(filepath.Join("..", "..", "records"))
+	if err != nil {
+		f.Fatal(err)
+	}
```
AlekSi, line 125:
```suggestion
	require.NoError(t, err)
```

I think we should not mix stdlib-style and testify-style assertions

internal/wire/wire_test.go
```diff
@@ -118,6 +119,19 @@ func fuzzMessages(f *testing.F, testCases []testCase) {
...
+
+	if len(records) == 0 {
+		f.Log("No records found so none of them will be added to the corpus")
+	}
```
AlekSi, line 129:
```suggestion
	f.Logf("%d recorded messages were added to the seed corpus", len(records))
```

So we always log the number

internal/wire/records_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Read every record file, parse their content to wire messages
+	// and store them in the testCase struct
+	for _, path := range recordFiles {
+		f, err := os.Open(path)
+		if err != nil {
+			return nil, err
+		}
+
+		defer f.Close()
```
AlekSi, line 63: I checked if there is a linter for that and found nothing 😑

internal/wire/records_test.go
```diff
@@ -0,0 +1,96 @@
...
+		if filepath.Ext(entry.Name()) == ".bin" {
+			recordFiles = append(recordFiles, path)
+		}
+
+		return nil
+	})
+
+	switch {
+	case os.IsNotExist(err):
+		return []testCase{}, nil
```
AlekSi, line 48: An empty non-nil slice suggests there is something complicated going on, that we need to differentiate between it and the nil slice, but that's not the case there. Let's just return nil.

internal/wire/records_test.go
AlekSi: Alternatively, we might want to move `os.IsNotExist` check there

AlekSi (commented):
Almost there, just a few stye suggestions

internal/wire/records_test.go
AlekSi: Sounds good

internal/wire/wire_test.go
noisersup: Nice idea!

internal/wire/records_test.go
noisersup: There's [bodyclose](https://github.com/timakin/bodyclose) But I believe it's only for net/http response bodies

AlekSi: Yes, that's correct, I checked it

internal/wire/records_test.go
noisersup: We return nil when there's an error. When file does not exist we don't return an error, so I wanted to differentiate from that by using empty slice. 
It's also more adapted to the usage. If we run the function and check if err != nil, we expect from the result to be useful:

```go
arr, err := fun()
if err != nil { return err }

// here we assume that arr will work with operators, `range`'s, etc.
for _, n := range arr { ... }
```

If we return nil without any error we would need to make another `if` to check if value is correct and (correct me if I'm wrong) it doesn't look like a good practice:

```go
arr, err := fun()
if err != nil { return err }

// if we don't validate the arr it will panic in the next statement
if arr != nil {...}

// here we assume that arr will work with operators, `range`'s, etc.
for _, n := range arr { ... }
```

noisersup: >  suggests there is something complicated going on, that we need to differentiate between it and the nil slice

I understand the point, so I mentioned about this behavior in the comment above the function.

Please let me know what's the final decision :)

AlekSi: > If we return nil without any error we would need to make another if to check if value is correct 

That's incorrect; it is perfectly fine to iterate over nil slice: https://go.dev/play/p/CWFD4PXSsfq

AlekSi (approved):

internal/wire/records_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"bufio"
+	"errors"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+)
+
+// loadRecords gets recursively all .bin files from the recordsPath directory,
+// parses their content to wire Msgs and returns them as an array of testCase
```
AlekSi, line 27:
```suggestion
// parses their content to wire Msgs and returns them as a slice of testCase
```

AlekSi (approved):

PR #1177, opened 2022-09-23
Add support for additional Tigris parameters

AlekSi:
# Description

## Readiness checklist

### DIFF
--- a/cmd/ferretdb/main.go
+++ b/cmd/ferretdb/main.go
@@ -53,8 +53,13 @@ var (
 	testConnTimeoutF = flag.Duration("test-conn-timeout", 0, "test: set connection timeout")
 )
 
-// tigrisURL is a Tigris URL. It is set in the main_tigris.go.
-var tigrisURL string
+// Tigris parameters that are set at main_tigris.go.
+var (
+	tigrisClientID     string
+	tigrisClientSecret string
+	tigrisToken        string
+	tigrisURL          string
+)
 
 // initFlags improves flags settings after all global flags are initialized
 // and all handler constructors are registered.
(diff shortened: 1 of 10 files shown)


cmd/ferretdb/main.go
```diff
@@ -53,8 +53,13 @@ var (
 	testConnTimeoutF = flag.Duration("test-conn-timeout", 0, "test: set connection timeout")
 )
 
-// tigrisURL is a Tigris URL. It is set in the main_tigris.go.
-var tigrisURL string
+// Tigris parameters that are set at main_tigris.go.
```
rumyantseva, line 56: With more and more handler-specific params, I wonder if we need to store them somewhere else but `main.go` file. What if in the future we have SQLite, Cockroach or some other DB-specific params? Our main.go will be full of them and yet for most of the users only a tiny subset will be needed.

AlekSi: Well, that's internal detail that we could refactor at any time. I just don't think that time is now.

rumyantseva: That sounds like something we could say about any tech debt :P
But I agree, it's not for now.

internal/handlers/tigris/tigris.go
```diff
@@ -34,8 +34,11 @@ func notImplemented(command string) error {
 
 // NewOpts represents handler configuration.
```
rumyantseva, line 35: Can we at least add more docs? E.g. are all the params required and where one can find what is each param needed for (maybe we can link some Tigris docs)?

rumyantseva (commented):
The code looks ok, but maybe we can add more docs somewhere? (E.g. some links where to read about those params as Tigris is quite a specific thing)

noisersup (dismissed):

AlekSi:
@rumyantseva I added comments for our embedded usage, PTAL

rumyantseva (approved):
LGTM

PR #1179, opened 2022-09-26
Add basic validation for queries.

### LINKED ISSUE #693: Explicitly disallow field names with `$`
DoD:

- [x] Add basic validation functions and calls mentioned in #871 
- [x] Disallow usage of `$` in data documents (but it is still allowed to use this symbol in command documents)
- [x] Add tests to check validation functions
- [ ] Finish FerretDB/dance#211
- [x] Create a separate issue for Tigris to have `tigrisdb` similar to `pgdb` 
       - #1248

rumyantseva:
# Description

This PR closes #693.

## Readiness checklist

### DIFF
--- a/internal/wire/op_query.go
+++ b/internal/wire/op_query.go
@@ -40,6 +40,7 @@ type OpQuery struct {
 
 func (query *OpQuery) msgbody() {}
 
+// readFrom composes an OpQuery from a buffered reader.
 func (query *OpQuery) readFrom(bufr *bufio.Reader) error {
 	if err := binary.Read(bufr, binary.LittleEndian, &query.Flags); err != nil {
 		return lazyerrors.Errorf("wire.OpQuery.ReadFrom (binary.Read): %w", err)


internal/types/types.go
```diff
@@ -97,7 +97,10 @@ var Null = NullType{}
 func validateValue(value any) error {
 	switch value := value.(type) {
 	case *Document:
-		return value.validate()
+		// TODO Call new validation here: https://github.com/FerretDB/FerretDB/issues/693
+		// Attention! This place require us to know whether the doc is a data doc or command doc!
```
rumyantseva, line 101: @AlekSi, an example where we need to know if it's a data or command document.

internal/types/document.go
```diff
@@ -216,17 +154,14 @@ func (d *Document) add(key string, value any) error {
 		return fmt.Errorf("types.Document.add: key already present: %q", key)
 	}
 
-	if !isValidKey(key) {
-		return fmt.Errorf("types.Document.add: invalid key: %q", key)
-	}
-
+	// TODO Should we still validate value if we don't validate key anymore?
```
rumyantseva, line 157: @AlekSi, this place is controversial to me now. I like the fact that we validated both `key` and `value` here. I think, for consistency, `isValidKey` should be called `validateKey` and should return the error, but overall the idea looked reasonable. Now, if we remove this `key` validation, we lose a part of the logic.

internal/types/types.go
rumyantseva, line 101: @AlekSi, an example where we need to know if it's a data or command document.

w84thesun (commented):
I think the `validateValue` function should be removed or renamed to either `ValidateDocument` or `ValidateCommand`.

rumyantseva:
@w84thesun, it looks to me that `validateDocument` and `validateCommand` are two different functions with different logic. E.g. the logic about keys validation will be different for them. `validateValue` looked quite reasonable to me, but in principle, we can move its logic to `validateDocument` and `validateCommand`, yes... What worries me is that after removing `validateValue`, we stop validating methods like `Set` completely...

w84thesun:
> What worries me is that after removing `validateValue`, we stop validating methods like `Set` completely...

I think that's ok. We can add correct validation later on if we will add TODO to those methods.

internal/types/types.go
AlekSi: Let's remove all validation code, not just comment it out. Half of it is wrong, and another is in the wrong place. We could always get old code from the git history if we need it

AlekSi (changes requested):

internal/types/array_test.go
```diff
@@ -51,20 +51,23 @@ func TestArray(t *testing.T) {
...
 	})
 
-	t.Run("NewArray", func(t *testing.T) {
-		t.Parallel()
-
-		a, err := NewArray(int32(42), 42)
-		assert.Nil(t, a)
-		assert.EqualError(t, err, `types.NewArray: index 1: types.validateValue: unsupported type: int (42)`)
-	})
+	// TODO add new validation solution: https://github.com/FerretDB/FerretDB/issues/693
```
rumyantseva, line 63: @AlekSi old validation (`validateValue`) used to check that in cases like `NewArray`, `NewDocument`, `Array.Append` and `Document.Set` attempt to set an invalid data type (e.g. just `int` without specifying `32` or `64`, or `nil`, or some other type lie this) is not possible. As I removed this validation, now it's possible, and all those tests fail. Should the new validation be called from such places? Or do we consider such cases impossible and should delete tests?

internal/wire/op_msg_test.go
```diff
@@ -22,136 +22,134 @@ import (
...
-		sections: []OpMsgSection{{
-			Documents: []*types.Document{must.NotFail(types.NewDocument(
-				"buildInfo", int32(1),
-				"lsid", must.NotFail(types.NewDocument(
-					"id", types.Binary{
-						Subtype: types.BinaryUUID,
-						B: []byte{
-							0xa3, 0x19, 0xf2, 0xb4, 0xa1, 0x75, 0x40, 0xc7,
-							0xb8, 0xe7, 0xa3, 0xa3, 0x2e, 0xc2, 0x56, 0xbe,
+var msgTestCases = []testCase{
```
rumyantseva, line 25: Somehow, `task fmt` reformatted this.

AlekSi (changes requested):
Integration test still fails

w84thesun (approved):
LGTM

noisersup (approved):

AlekSi (approved):

PR #1207, opened 2022-10-04
Add simple query pushdown for PostgreSQL

### LINKED ISSUE #896: Support simple query pushdown for Postgres
## Support simple query pushdown

The goal of this task is to enable simple query pushdown for querier that look like `{_id: <ObjectID>}`.
Only that field name (`_id`) is supported, and only that value type (`ObjectID`).

In the future, we will add support for other fields (starting with simple scalar fields),
other values (starting with strings, numbers, and other simple scalar values),
other conditions like `{_id: <ObjectID>, field: value}` (so only the first part is pushed down).

## Postgres

If a field name is different, value type is not `ObjectID`, or some other condition is present,
we should use the previous version of the code that SELECTs the whole table without any WHERE condition.

If the value type is `ObjectID` but the data in the value is somehow corrupted, raise an error (`fjson` unmarshal).

If those conditions are met, we send a SELECT query with a WHERE condition.

Proof of concept for a `{_id: <ObjectID>}` pushdown query, PostgreSQL:

GIN index is not used by PostgreSQL in any queries below.
So let's use the first one just because it is first.

```sql
-- this will not use these indexes:
-- CREATE INDEX values_id_idx ON public."values" USING gin ((((_jsonb -> '_id'::text) -> '$o'::text)))
-- CREATE INDEX values_id_idx ON values (((_jsonb ->> '_id')::text));
select * from values where (_jsonb->'_id')::jsonb->>'$o' = '507f1f77bcf86cd799439011'; -- no, seq scan will it use index? not that indexes
```

## Test

Integration tests: Provide a new test flag that enables a query pushdown.
Run with and without flag: results must be the same.
When explain feature will be ready, compare the query plan.

## Links

Analysis task: #847

w84thesun:
# Description

This PR closes #896.

## Readiness checklist

### DIFF
--- a/internal/handlers/pg/msg_count.go
+++ b/internal/handlers/pg/msg_count.go
@@ -65,10 +65,12 @@ func (h *Handler) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
 	if sp.DB, err = common.GetRequiredParam[string](document, "$db"); err != nil {
 		return nil, err
 	}
+
 	collectionParam, err := document.Get(document.Command())
 	if err != nil {
 		return nil, err
 	}
+
 	var ok bool
 	if sp.Collection, ok = collectionParam.(string); !ok {
 		return nil, common.NewErrorMsg(
@@ -77,9 +79,7 @@ func (h *Handler) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
 		)
 	}
 
-	if filter != nil && filter.Has("_id") {
-		sp.SqlFilters = must.NotFail(types.NewDocument("_id", must.NotFail(filter.Get("_id"))))
-	}
+	sp.Filters = filter
 
 	resDocs := make([]*types.Document, 0, 16)
 	err = h.pgPool.InTransaction(ctx, func(tx pgx.Tx) error {
--- a/internal/handlers/pg/msg_count.go
+++ b/internal/handlers/pg/msg_count.go
@@ -78,7 +78,15 @@ func (h *Handler) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
 	}
 
 	if filter.Has("_id") {
-		sp.SqlFilters = must.NotFail(types.NewDocument("_id", must.NotFail(filter.Get("_id"))))
+		id, err := filter.Get("_id")
+		if err != nil {
+			panic("can't get _id for some reason")
+		}
+
+		sp.SqlFilters, err = types.NewDocument("_id", id)
+		if err != nil {
+			panic("can't create document for some reason")
+		}
 	}
 
 	resDocs := make([]*types.Document, 0, 16)
--- a/internal/handlers/pg/msg_count.go
+++ b/internal/handlers/pg/msg_count.go
@@ -77,11 +77,10 @@ func (h *Handler) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
 		)
 	}
 
-	var sqlFilter *types.Document
+	sp.SqlFilters = nil
 	if filter.Has("_id") {
-		sqlFilter = must.NotFail(types.NewDocument("_id", must.NotFail(filter.Get("_id"))))
+		sp.SqlFilters = must.NotFail(types.NewDocument("_id", must.NotFail(filter.Get("_id"))))
 	}
-	sp.SqlFilters = sqlFilter
 
 	resDocs := make([]*types.Document, 0, 16)
 	err = h.pgPool.InTransaction(ctx, func(tx pgx.Tx) error {
--- a/internal/handlers/pg/msg_count.go
+++ b/internal/handlers/pg/msg_count.go
@@ -62,6 +62,7 @@ func (h *Handler) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
 	}
 
 	var sp pgdb.SQLParam
+
 	if sp.DB, err = common.GetRequiredParam[string](document, "$db"); err != nil {
 		return nil, err
 	}
@@ -79,7 +80,7 @@ func (h *Handler) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
 		)
 	}
 
-	sp.Filters = filter
+	sp.Filter = filter
 
 	resDocs := make([]*types.Document, 0, 16)
 	err = h.pgPool.InTransaction(ctx, func(tx pgx.Tx) error {
--- a/internal/handlers/pg/msg_count.go
+++ b/internal/handlers/pg/msg_count.go
@@ -77,9 +77,15 @@ func (h *Handler) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
 		)
 	}
 
+	var sqlFilter *types.Document
+	if filter.Has("_id") {
+		sqlFilter = must.NotFail(types.NewDocument("_id", must.NotFail(filter.Get("_id"))))
+	}
+	sp.SqlFilters = sqlFilter
+
 	resDocs := make([]*types.Document, 0, 16)
 	err = h.pgPool.InTransaction(ctx, func(tx pgx.Tx) error {
-		fetchedChan, err := h.pgPool.QueryDocuments(ctx, tx, &sp, filter)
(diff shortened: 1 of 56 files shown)


internal/handlers/pg/pgdb/query.go
```diff
@@ -173,13 +174,41 @@ func buildQuery(ctx context.Context, tx pgx.Tx, sp *SQLParam) (string, error) {
...
+// appendSqlFilters adds WHERE clause with given filters to the query.
+func appendSqlFilters(q string, sqlFilter *types.Document) string {
+	var filters []string
+
+	for k, v := range sqlFilter.Map() {
+		switch k {
+		case "_id":
+			switch v := v.(type) {
+			case *types.ObjectID:
+				filters = append(filters, fmt.Sprintf(`(_jsonb->'_id')::jsonb->>'$o' = '%s'`, must.NotFail(fjson.Marshal(v))))
```
AlekSi, line 197: We should use placeholders instead of `%s`

internal/handlers/pg/msg_count.go
```diff
@@ -77,6 +77,10 @@ func (h *Handler) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
 		)
 	}
 
+	if filter != nil && filter.Has("_id") {
+		sp.SqlFilters = must.NotFail(types.NewDocument("_id", must.NotFail(filter.Get("_id"))))
```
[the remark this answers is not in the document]
w84thesun: It would return 1 as the count result but it is still possible to do so.

internal/handlers/pg/pgdb/query.go
w84thesun: Done.

internal/handlers/pg/msg_find.go
```diff
@@ -119,6 +119,10 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
 		}
 	}
 
+	if filter != nil && filter.Has("_id") {
+		sp.SqlFilters = must.NotFail(types.NewDocument("_id", must.NotFail(filter.Get("_id"))))
```
AlekSi, line 123: Why we don't move that copy&pasted logic inside `QueryDocuments`? That's what we do in the Tigris handler

rumyantseva (changes requested):
I agree with @AlekSi's idea. I also asked a question in the place where I don't understand the query.

internal/handlers/pg/pgdb/query.go
```diff
@@ -149,35 +150,79 @@ func Explain(ctx context.Context, tx pgx.Tx, sp SQLParam) (*types.Array, error)
...
+	}
+
+	var query string
+
+	if sp.Explain {
+		query += `EXPLAIN (VERBOSE true, FORMAT JSON) `
 	}
 
-	q := `SELECT _jsonb `
+	query += ` SELECT _jsonb `
```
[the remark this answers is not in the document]
w84thesun: It could be EXPLAIN which is appended earlier if flag provided.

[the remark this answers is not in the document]
w84thesun: It could be EXPLAIN which is appended earlier if flag provided.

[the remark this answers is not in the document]
rumyantseva: @w84thesun, but in the example above you have `query += EXPLAIN (VERBOSE true, FORMAT JSON) `. Doesn't it cover the `EXPLAIN` case? And what could be even before `EXPLAIN` in that case above?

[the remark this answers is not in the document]
w84thesun: Changed the statement above which adds EXPLAIN.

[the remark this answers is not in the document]
rumyantseva: ok, got it now! :)

internal/handlers/pg/pgdb/query.go
```diff
@@ -149,35 +150,79 @@ func Explain(ctx context.Context, tx pgx.Tx, sp SQLParam) (*types.Array, error)
 //
 // It returns (possibly wrapped) ErrSchemaNotExist or ErrTableNotExist
 // if schema/database or table/collection does not exist.
-func buildQuery(ctx context.Context, tx pgx.Tx, sp *SQLParam) (string, error) {
+func buildQuery(ctx context.Context, tx pgx.Tx, sp *SQLParam) (string, []any, error) {
```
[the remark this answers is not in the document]
w84thesun: I just specified that we return arguments.

internal/handlers/pg/msg_find.go
w84thesun: Moved and simplified the code a bit for now.

internal/handlers/pg/pgdb/query.go
```diff
@@ -147,37 +148,82 @@ func Explain(ctx context.Context, tx pgx.Tx, sp SQLParam) (*types.Array, error)
...
+		default:
+			continue
+		}
+	}
+
+	var query string
+
+	if len(filters) > 0 {
+		query = ` WHERE ` + strings.Join(filters, " AND ")
+		query = query[:len(query)-5]
```
rumyantseva, line 223: Hmm, this part is unclear to me now :) why `-5`? let's leave a comment about it?

rumyantseva (changes requested):
One more question!

internal/handlers/pg/pgdb/query.go
w84thesun: Moved AND to const and added comment.

internal/handlers/pg/pgdb/query.go
```diff
@@ -147,37 +153,82 @@ func Explain(ctx context.Context, tx pgx.Tx, sp SQLParam) (*types.Array, error)
...
+		default:
+			continue
+		}
+	}
+
+	var query string
+
+	if len(filters) > 0 {
+		query = ` WHERE ` + strings.Join(filters, andClause)
+		query = query[:len(query)-len(andClause)]
```
rumyantseva, line 228: I still don't understand what we want to trim here :) Do you have an example?

w84thesun: When we made strings.Join call we and up with `" AND “` in the and of the query which we are trimming here.

rumyantseva: Yes, this is the exact part I don't understand! I believe, this is not how Join works :)
Proof me wrong: https://go.dev/play/p/7qExnl4j76p

w84thesun: Yeah, you right.

rumyantseva (dismissed):
I don't see anything odd.
That would be nice to add tests, but we have a separate task for it.

internal/handlers/pg/pgdb/query.go
```diff
@@ -147,37 +148,81 @@ func Explain(ctx context.Context, tx pgx.Tx, sp SQLParam) (*types.Array, error)
...
+			}
+		default:
+			continue
+		}
+	}
+
+	var query string
+
+	if len(filters) > 0 {
+		query = ` WHERE ` + strings.Join(filters, " AND ")
```
AlekSi, line 222: Let's add `(` and `)` so conditions would look like `WHERE (_id = 1) AND (_id = 2)`

w84thesun: I added parentheses to filter generation part.

internal/handlers/pg/pgdb/query.go
```diff
@@ -147,37 +148,81 @@ func Explain(ctx context.Context, tx pgx.Tx, sp SQLParam) (*types.Array, error)
...
+	var p Placeholder
+
+	for k, v := range sqlFilters.Map() {
+		switch k {
+		case "_id":
+			switch v := v.(type) {
+			case types.ObjectID:
+				filters = append(filters, fmt.Sprintf(`((_jsonb->'_id')::jsonb->>'$o' = %s)`, p.Next()))
+
+				args = append(args, pgx.Identifier{string(must.NotFail(fjson.Marshal(v)))}.Sanitize())
```
AlekSi, line 212: `pgx.Identifier` should be used only for identifiers, like table or column names. It should not be used for values. Why we need it there?

w84thesun: Removed.

AlekSi:
Tests fail

AlekSi (approved):

PR #1231, opened 2022-10-06
Add “metrics” section to `serverStatus` response

### LINKED ISSUE #1201: Add “metrics” to `serverStatus` response
We should return FerretDB metrics in the `metrics` field of `serverStatus` command response. See https://www.mongodb.com/docs/manual/reference/command/serverStatus/#metrics

We already have `ConnMetrics` that could provide information for `metrics.commands` field. Let's return them.

Let's also check what else we could add quickly.

We should write an integration test. For example:
* send some command like `ping`;
* send `serverStatus` command;
* check that `metrics.ping.total` is not zero.

noisersup:
# Description

This PR closes #1201 .

## Readiness checklist


internal/handlers/common/serverstatus.go
```diff
@@ -19,14 +19,15 @@ import (
...
+	"github.com/FerretDB/FerretDB/internal/clientconn"
 	"github.com/FerretDB/FerretDB/internal/types"
 	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
 	"github.com/FerretDB/FerretDB/internal/util/must"
 	"github.com/FerretDB/FerretDB/internal/util/version"
 )
 
 // ServerStatus returns a common part of serverStatus command response.
-func ServerStatus(startTime time.Time) (*types.Document, error) {
+func ServerStatus(startTime time.Time, cm *clientconn.ConnMetrics) (*types.Document, error) {
```
noisersup, line 30: Here (`common` package) we need to import the `clientconn.ConnMetrics`, because it's required to provide the metrics to the `serverStatus` cmd

internal/clientconn/conn_metrics.go
```diff
@@ -70,6 +93,83 @@ func (cm *ConnMetrics) Collect(ch chan<- prometheus.Metric) {
 	cm.aggregationStages.Collect(ch)
 }
 
+func (cm *ConnMetrics) Responses() map[string]CommandMetrics {
+	res := make(map[string]CommandMetrics)
+
+	// initialize commands in the map to show zero values in the metrics output
+	for k := range common.Commands {
```
noisersup, line 100: Here (`clientconn` package) we need to import `common.Commands` to fill the map with all of the listed commands

noisersup (commented):
After merging with main the severstatus code was splitted from `pg`/`tigris` packages to `common` package. 

Trouble is that I import `common` from `clientconn` which is used by serverStatus code.

Moving the `common.Commands` to the seperate package sounds reasonable to me, because now it's a code used by different packages for completely different purposes.

We could also pass `[]string` with command names inside the `NewListenerMetrics` and `NewCommandMetrics`, although it requires to fetch only keys from `common.Commands` and we use `NewListenerMetrics` in a couple of places so it would require some helper function.

There are probably some other solutions so let's maybe discuss it

internal/handlers/common/serverstatus.go
AlekSi: I don't think we should import `common` from `clientconn`, yeah.

Let's try passing commands as `[]string`

cmd/ferretdb/main.go
```diff
@@ -212,9 +214,17 @@ func run() {
 
 	go debug.RunHandler(ctx, cli.DebugAddr, metricsRegisterer, logger.Named("debug"))
 
+	cmdsList := make([]string, len(common.Commands))
+	for k := range common.Commands {
+		cmdsList = append(cmdsList, k)
+	}
+
+	metrics := connmetrics.NewListenerMetrics(cmdsList)
```
AlekSi, line 222:
```suggestion
	metrics := connmetrics.NewListenerMetrics(maps.Keys(common.Commands))
```

https://pkg.go.dev/golang.org/x/exp/maps#Keys

But we also might want to sort them to make results stable

internal/clientconn/connmetrics/conn_metrics.go
```diff
@@ -0,0 +1,183 @@
...
+
+		cm := res[command]
+		if cm == nil {
+			cm = BasicCommandMetrics{}
+		}
+
+		switch cm := cm.(type) {
+		case UpdateCommandMetrics:
+			cm.Total += value
+			if result != "ok" && result != "Unset" {
```
AlekSi, line 160: Where that "Unset" comes from? We should prevent it from appearing, not try to handle it there

noisersup: Isn't Unset just error status?

internal/clientconn/connmetrics/conn_metrics.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+type UpdateCommandMetrics struct {
+	ArrayFilters int64
+	Failed       int64
+	Pipeline     int64
+	Total        int64
+}
+
+// newConnMetrics creates new conn metrics.
```
w84thesun, line 50:
```suggestion
// newConnMetrics creates connection metrics.
```

internal/clientconn/connmetrics/conn_metrics.go
AlekSi: I don't know. Where we set that value to that label?

internal/clientconn/connmetrics/conn_metrics.go
```diff
@@ -0,0 +1,183 @@
...
+//
+// The cmds is the list of all expected commands that could be measured.
+// After providing them, they will be set with the zero values.
```
w84thesun, line 53:
```suggestion
// Commands provided in the list will be set with zero values.
```

internal/clientconn/connmetrics/conn_metrics.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// Collect implements prometheus.Collector.
+func (cm *ConnMetrics) Collect(ch chan<- prometheus.Metric) {
+	cm.Requests.Collect(ch)
+	cm.Responses.Collect(ch)
+	cm.AggregationStages.Collect(ch)
+}
+
+func (cm *ConnMetrics) GetResponses() map[string]CommandMetrics {
```
w84thesun, line 101: Please add the function comment.

internal/clientconn/connmetrics/conn_metrics.go
```diff
@@ -0,0 +1,183 @@
...
+		close(ch)
+	}()
+
+	for m := range ch {
+		var content dto.Metric
+		must.NoError(m.Write(&content))
+
+		var stage int
+
+		var command, opcode, result string
```
w84thesun, line 129:
```suggestion
		var stage int
		var command, opcode, result string
```

internal/clientconn/connmetrics/conn_metrics.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+
+		switch cm := cm.(type) {
+		case UpdateCommandMetrics:
+			cm.Total += value
+			if result != "ok" && result != "Unset" {
+				cm.Failed += value
+			}
+			cm.Pipeline += int64(stage)
+			// TODO: add metrics for arrayFilters
```
w84thesun, line 164: Should we create an issue for `arrayFilters` metrics?

internal/clientconn/connmetrics/conn_metrics.go
```diff
@@ -0,0 +1,183 @@
...
+		// update related operators have more fields in the output
+		switch cmd {
+		case "update", "clusterUpdate", "findAndModify":
+			res[cmd] = UpdateCommandMetrics{}
+		default:
+			res[cmd] = BasicCommandMetrics{}
+		}
+	}
+
+	ch := make(chan prometheus.Metric)
```
w84thesun, line 115: If it won't collide with any imports, `metrics` would be a better name for that channel.

internal/clientconn/connmetrics/conn_metrics.go
```diff
@@ -0,0 +1,183 @@
...
+			}
+		}
+
+		if opcode != "OP_MSG" {
+			continue
+		}
+
+		value := int64(content.GetCounter().GetValue())
+
+		cm := res[command]
```
w84thesun, line 152: You use `cm` as the shorthand for `ConnMetrics` below. If this `cm` stands for `command` better replace it with `cmd`.

w84thesun (changes requested):

internal/clientconn/connmetrics/conn_metrics.go
noisersup: It comes from `getParameter` command

noisersup: https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/pg/msg_getparameter.go#L85

noisersup: Yes, it returns 2 values (one of them is UnsetErr) and sets result based on Unset protoerror [here](https://github.com/FerretDB/FerretDB/blob/main/internal/clientconn/conn.go#L368)

internal/clientconn/connmetrics/conn_metrics.go
noisersup: you right

internal/clientconn/connmetrics/conn_metrics.go
noisersup: `cmdMetric` will be good?

internal/clientconn/connmetrics/conn_metrics.go
noisersup: They are not implemented in code and we don't have any issue for them. @AlekSi wdyt?

rumyantseva (changes requested):
Some linters fail

internal/clientconn/connmetrics/conn_metrics.go
w84thesun: Here, I created one https://github.com/FerretDB/FerretDB/issues/1259

integration/setup/common.go
```diff
@@ -91,9 +95,15 @@ func SkipForPostgresWithReason(tb testing.TB, reason string) {
 func setupListener(tb testing.TB, ctx context.Context, logger *zap.Logger) int {
 	tb.Helper()
 
+	cmdsList := maps.Keys(common.Commands)
+	sort.Sort(sort.StringSlice(cmdsList))
```
w84thesun, line 99: Linter missed that one. We can use `sort.Strings(...)` here too.

rumyantseva (commented):
It seems that Tigris panics, and some Tigris tests don't pass (or something like that). Does it pass on your local env?

internal/clientconn/connmetrics/conn_metrics.go
```diff
@@ -0,0 +1,185 @@
...
+		// update related operators have more fields in the output
+		switch cmd {
+		case "update", "clusterUpdate", "findAndModify":
+			res[cmd] = UpdateCommandMetrics{}
+		default:
+			res[cmd] = BasicCommandMetrics{}
+		}
+	}
+
+	metrics := make(chan prometheus.Metric)
```
rumyantseva, line 118: I don't understand the need for that channel. Does it simplify/speed up something?

noisersup: Prometheus package is using channels

rumyantseva: What is the logic behind it? This is an unbuffered channel, and in the code below range through it, why? How do we expect this range to work? Is it a one-time read?

rumyantseva:
@noisersup ok, then that looks like the same concurrency problem they've fixed... I hope, we will update to a newer Tigris version soon.

internal/clientconn/connmetrics/conn_metrics.go
noisersup: https://github.com/FerretDB/FerretDB/pull/1231/files/1a4633e6dbf3ea84668460ab69792c51169c3efc#diff-66d1358cff6f92642975f6daa16960a7c7bfc30bdefffc90193b55dce2bd426fR123

Channel will be closed after sending all data, so range will stop executing

noisersup: Using buffer here would create a deadlock, because prometheus output will be longer with every command execution

rumyantseva (approved):
I don't have concerns anymore, we might need some Tigris-related action items though.

w84thesun (dismissed):
LGTM

AlekSi (changes requested):
> I added/updated comments for both exported and unexported top-level declarations (functions, types, etc).

That's not done

internal/clientconn/connmetrics/conn_metrics.go
```diff
@@ -0,0 +1,185 @@
...
+	Failed int64
+	Total  int64
+}
+
+type UpdateCommandMetrics struct {
+	ArrayFilters int64
+	Failed       int64
+	Pipeline     int64
+	Total        int64
+}
```
w84thesun, line 48: Please add comments to those types declarations.

rumyantseva (commented):
Comments are needed (see @w84thesun's review), the rest looks good to me.

w84thesun (approved):

AlekSi (approved):

PR #1229, opened 2022-10-06
Bump Tigris

AlekSi:
# Description

This PR closes #{issue_number}.

## Readiness checklist

* [ ] I added tests for new functionality or bugfixes.
* [ ] I ran `task all`, and it passed.
* [ ] I added/updated comments for both exported and unexported top-level declarations (functions, types, etc).
* [ ] I checked comments rendering with `task godocs`.
* [ ] I ensured that the title is good enough for the changelog.
* [ ] (for maintainers only) I set Reviewers ([`@FerretDB/core`](https://github.com/orgs/FerretDB/teams/core)), Assignee, Labels, Project and project's Sprint fields.
* [ ] I marked all done items in this checklist.

PR #1239, opened 2022-10-11
Migrate to a newer Tigris version and fix relevant tests

rumyantseva:
# Description

This PR fixes the problems related to the newer Tigris version migration (see `engineering #10`).

Identified cases:
- Sometimes, `\n` is returned from Tigris at the end of a valid document. In such cases `tjson.checkConsumed` shouldn't return an error as we consider such a string valid.
- If a field is described as a number in the schema, it's possible to set a string containing a valid number (e.g. `"32"` will be stored as `32` for int field, or `"32.32"` will be stored as `32.32` for float fields), see `diff.md`.

## Readiness checklist


AlekSi:
I think that's the same as #1229

website/docs/diff.md
```diff
@@ -15,6 +15,8 @@ sidebar_position: 6
    * database name cannot contain capital letters.
 5. For Tigris, FerretDB requires Tigris schema validation for `msg_create`: validator must be set as `$tigrisSchemaString`.
    The value must be a JSON string representing JSON schema in [Tigris format](https://docs.tigrisdata.com/overview/schema).
+6. For Tigris, if field represents a number, an attempt to set a string that represents a suitable number
```
rumyantseva, line 18: @Fashander, I need your help to phrase it better :) Do you have a moment to help me with the wording? I can show you this case so that we can describe it better.

Fashander:
```suggestion
For Tigris, if a field is represented as a number, any attempt to assign a string containing a number will pass as a number (e.g. "30" for int32 field)
```

w84thesun (approved):
Nice!

website/docs/diff.md
```diff
@@ -15,6 +15,8 @@ sidebar_position: 6
    * database name cannot contain capital letters.
 5. For Tigris, FerretDB requires Tigris schema validation for `msg_create`: validator must be set as `$tigrisSchemaString`.
    The value must be a JSON string representing JSON schema in [Tigris format](https://docs.tigrisdata.com/overview/schema).
+6. For Tigris, if a field is represented as a number, any attempt to assign a string containing a number will pass as a number
```
AlekSi, line 18: ⚠️

The idea of that list of user-visible differences was to have a corresponding dance test for each item. Are we going to add that?

I would say we should not bother with it until we add MongoDB-style schema validation, so let's remove that item for now.

website/docs/diff.md
```diff
@@ -15,6 +15,8 @@ sidebar_position: 6
    * database name cannot contain capital letters.
 5. For Tigris, FerretDB requires Tigris schema validation for `msg_create`: validator must be set as `$tigrisSchemaString`.
```
AlekSi, line 16: While you are on it…

```suggestion
5. For Tigris, FerretDB requires Tigris schema validation for `create` command: validator must be set as `$tigrisSchemaString`.
```

`msg_create` is not something our users know about

internal/tjson/tjson.go
```diff
@@ -66,12 +66,16 @@ type tjsontype interface {
...
+		b = bytes.TrimSpace(b)
+		if l := len(b); l != 0 {
+			return lazyerrors.Errorf("%d bytes remains in the decoder: `%s` (%b)", l, b, b)
+		}
 	}
 
 	if l := r.Len(); l != 0 {
 		b, _ := io.ReadAll(r)
-		return lazyerrors.Errorf("%d bytes remains in the reader: %s", l, b)
+		return lazyerrors.Errorf("%d bytes remains in the reader: `%s` (%b)", l, b, b)
```
AlekSi, line 78:
```suggestion
		return lazyerrors.Errorf("%[1]d bytes remains in the reader: `%[2]s` (%[2]#02x)", l, b)
```

internal/tjson/tjson.go
```diff
@@ -66,12 +66,16 @@ type tjsontype interface {
 func checkConsumed(dec *json.Decoder, r *bytes.Reader) error {
 	if dr := dec.Buffered().(*bytes.Reader); dr.Len() != 0 {
 		b, _ := io.ReadAll(dr)
-		return lazyerrors.Errorf("%d bytes remains in the decoded: %s", dr.Len(), b)
+
+		b = bytes.TrimSpace(b)
```
AlekSi, line 70: Let's add a comment explaining why. We don't do that in pjson, for example.

integration/update_field_compat_test.go
```diff
@@ -103,10 +103,12 @@ func TestUpdateFieldCompatMax(t *testing.T) {
 
 		// Strings are not converted to numbers
```
AlekSi, line 104:
```suggestion
		// Strings are not converted to numbers (except for Tigris with int64 fields)
```

website/docs/diff.md
AlekSi: > The idea of that list of user-visible differences was to have a corresponding dance test for each item

Actually, please add a markdown comment above that list mentioning that

rumyantseva: @AlekSi sorry, I don't understand what should be mentioned where. Could you suggest a change, please?
If `dance` supports Tigris, let's add point 6, yes.

internal/tjson/tjson.go
AlekSi: btw, the explanation in the PR description is great, but it is not visible when someone reads this code after the PR is merged

rumyantseva: fixed

website/docs/diff.md
rumyantseva: fixed

AlekSi (approved):

w84thesun (approved):

PR #1261, opened 2022-10-14
Tweak supported wire protocol versions

AlekSi:
We don't actively test with 5.1, but that allows us to support some software that wasn't updated for 6.0 yet.


chilagrow (approved):

noisersup (changes requested):
It seems that `TestCommandsReplicationIsMaster` tests are failing because of different wire versions

noisersup (approved):

chilagrow (approved):

PR #1265, opened 2022-10-16
Add ability to subscribe to state changes

AlekSi:
# Description

This PR:
* adds the ability to subscribe to state changes, so when telemetry is enabled or disabled in runtime, we can start or stop the reporter;
* moves `startTime` to the state, so we don't have multiple copies of it;
* fixes passing metrics to the Tigris handler.

## Readiness checklist


internal/handlers/registry/tigris.go
```diff
@@ -25,11 +25,13 @@ import (
...
-			ClientSecret: opts.TigrisClientSecret,
-			Token:        opts.TigrisToken,
-			URL:          opts.TigrisURL,
-			L:            opts.Logger,
+			ClientID:      opts.TigrisClientID,
+			ClientSecret:  opts.TigrisClientSecret,
+			Token:         opts.TigrisToken,
+			URL:           opts.TigrisURL,
+			L:             opts.Logger,
+			Metrics:       opts.Metrics,
```
AlekSi, line 33: Metrics passing was lost

internal/util/state/provider.go
```diff
@@ -22,88 +21,127 @@ import (
...
+	// Simply overwrite state to handle all errors and edge cases
+	// like missing directory, corrupted file, invalid UUID, etc.
+	if err := persistState(p.s, p.filename); err != nil {
+		return p, fmt.Errorf("failed to persist state: %w", err)
 	}
 
 	return p, nil
 }
 
+// MetricsCollector return Prometheus metrics collector for that provider.
```
rumyantseva, line 64:
```suggestion
// MetricsCollector returns Prometheus metrics collector for that provider.
```

internal/util/state/provider.go
```diff
@@ -22,88 +21,127 @@ import (
...
+	p.s.fill()
+
+	err := persistState(p.s, p.filename)
+	if err != nil {
+		err = fmt.Errorf("failed to persist state: %w", err)
 	}
 
-	if err := os.WriteFile(p.filename, b, 0o666); err != nil {
-		return nil, fmt.Errorf("failed to write state file: %w", err)
+	// skip subscribers that already have notification waiting for them
```
rumyantseva, line 120: If I understand correctly, this comment describes the empty `default:` case, right? should it be mentioned after `default` then?

internal/util/state/provider.go
```diff
@@ -22,88 +21,127 @@ import (
...
+	for ch := range p.subs {
+		select {
+		case ch <- struct{}{}:
```
rumyantseva, line 123: This implementation looks safe to me (e.g. if we called `Unsubscribe` before, we don't have a chance to write into `ch` again as it will be out of `p.subs`). Still, I wonder if we need some tests, especially the ones that cover situations when `ch` is already full and that we don't wait for it.

rumyantseva (commented):
In principle, I like the implementation and the fact that there are some simple tests to check channel-related logic.

Personally, I'd try to test some corner cases (like calling `Update`, or `Subscribe` or `Unsubscribe` multiple times or in a wrong order, but on the other hand, I'm not sure if such tests make sense.

internal/util/state/provider.go
AlekSi: Well, it describes the whole construct. Also, the `default` case is only for a single subscriber

w84thesun (changes requested):
PG tests failed

internal/util/state/provider.go
AlekSi: Done by removing `Unsubscribe` since we don't actually need it

w84thesun (approved):
LGTM

AlekSi:
@rumyantseva PTAL

rumyantseva (approved):
LGTM

PR #1277, opened 2022-10-18
Add support for empty command documents

### LINKED ISSUE #1000: Add support for empty documents
It is possible to pass empty document to command document. One of it's usages as replacement document in update operation.

- [ ] Check for this issue number being mentioned in source code
- [ ] Check that validation function passes such documents
- [ ] Check `findAndModify` command
- [ ] Enable tests

Example: when you pass an empty `bson.D` for update with replacement document, the result we store should be the doc with only `_id` field being set.

w84thesun:
# Description

Closes #1000.

## Readiness checklist


noisersup:
Tests are still failing, because it's needed to change some fields there, but please check if it's suitable for our requirements

integration/update_test.go
```diff
@@ -228,15 +228,6 @@ func TestUpdateReplaceDocuments(t *testing.T) {
 			expectedFilter: bson.D{{"_id", "document-composite"}},
 			expected:       bson.D{{"_id", "document-composite"}, {"replacement-value", int32(1)}},
 		},
-		// TODO: https://github.com/FerretDB/FerretDB/issues/1000
```
w84thesun, line 231: This test is basically the same as "ReplaceEmptyDocument" in update_compat_test.go

internal/handlers/pg/msg_update.go
```diff
@@ -68,6 +68,7 @@ func (h *Handler) MsgUpdate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 	}
 
 	var created bool
+
```
AlekSi, line 71: All changes in this file are whitespace changes that also make this file different from Tigris version. Why we need them?

w84thesun: Restored as it was. `MsgUpdate` and `MsgFindAndModify` are a huge mess. I think we should refactor them.

noisersup (approved):

rumyantseva (approved):
LGTM

PR #1281, opened 2022-10-19
Add experimental mergify configuration

AlekSi:
# Description

See https://mergify.com and https://docs.mergify.com/configuration/

That configuration should remove the need to constantly click the "Update branch" button.

It probably will be replaced by a GitHub-native merge queue (https://github.com/features/merge-queue/signup) when we get access to it.

## Readiness checklist


.github/mergify.yml
```diff
@@ -0,0 +1,16 @@
+---
+# see https://docs.mergify.com/configuration/
+pull_request_rules:
+  - name: "Update out-of-date ready PRs"
+    conditions:
+      - "-closed"
+      - "-conflict"
```
rumyantseva, line 7: I looked at https://docs.mergify.com/conditions/#conditions, and from docs, I'd read this condition as "and the pull request has a conflict with the base branch". Did I misunderstand it?

rumyantseva (commented):
The configuration looks strange to me - are the conditions "or" conditions or "and" conditions? It's not straightforward from the docs, but if you checked it, I'm ok with the configuration.

.github/mergify.yml
```diff
@@ -0,0 +1,16 @@
+---
+# see https://docs.mergify.com/configuration/
+pull_request_rules:
+  - name: "Update out-of-date ready PRs"
+    conditions:
```
rumyantseva, line 5: There is `-or` operator in the docs (https://docs.mergify.com/conditions/#combining-conditions-with-operators), I'd suggest using it to be explicit.

.github/mergify.yml
AlekSi: From https://docs.mergify.com/conditions/#grammar

> The optional `-` prefix is equivalent to the not operator.

rumyantseva: Ah! I actually misread this and didn't notice `-`. And I guess by default, the conditions use `and`?

.github/mergify.yml
rumyantseva: Or in our case, it's rather `and`...

.github/mergify.yml
AlekSi: Correct

AlekSi:
@rumyantseva PTAL

rumyantseva (approved):
I'd prefer a longer syntax (`not` and `and` being mentioned explicitly) to improve readability and not confuse people, but it's not a strong opinion, so I'm ready to approve it.

AlekSi:
All their examples use implicit `and`: https://docs.mergify.com/examples/ Just like MongoDB with filters without `$and` :)

w84thesun (approved):

chilagrow (approved):
Let's try it out!

PR #1293, opened 2022-10-20
Explicitly disallow duplicate keys in data documents

### LINKED ISSUE #364: Explicitly disallow duplicate keys in data documents

rumyantseva:
# Description

Closes #364.

As agreed with the team, quite a big refactoring was made. Now `types.Document`'s internal representation consists of a slice of fields where each field has a `string` field and `any` value.
That approach allows having duplicated keys in documents in principle.

However, `Document.ValidateData` checks that there are no duplicates.

Minor refactoring: `Document.Set`'s signature doesn't return `error `anymore, as, in fact, the implementation always returns `nil`. With that, there is no need to make the code more complicated (with all the checks for errors).

## Readiness checklist


internal/types/document_validation.go
```diff
@@ -50,6 +53,11 @@ func (d *Document) ValidateData() error {
 		if strings.Contains(key, "$") {
 			return newValidationError(fmt.Errorf("invalid key: %q (key must not contain $)", key))
 		}
+
+		if _, ok := duplicateChecker[key]; ok {
+			return newValidationError(fmt.Errorf("invalid key: %q (duplicate keys are not allowed)", key))
```
chilagrow, line 58: Shall we add validation test for this scenario.

internal/types/document.go
```diff
@@ -146,92 +179,85 @@ func (d *Document) Command() string {
...
-		return nil
 	}
 
-	d.m[key] = value
-	return nil
+	d.fields = append(d.fields, field{key, value})
 }
 
 // Remove the given key and return its value, or nil if the key does not exist.
+// If there are duplicate keys, only the first value is deleted and returned.
```
chilagrow, line 251: Was this what we decided or what MongoDB does?

chilagrow (commented):
Thanks for great refactor too :muscle:

internal/types/document.go
rumyantseva: This is a very internal thing, so it's up to us to decide.

chilagrow: I'm happy with what we have, particular the documented comment :hugs:

internal/types/document.go
```diff
@@ -146,92 +179,85 @@ func (d *Document) Command() string {
...
+		if field.key == key {
+			return field.value, nil
+		}
 	}
 
 	return nil, fmt.Errorf("types.Document.Get: key not found: %q", key)
 }
 
 // Set sets the value for the given key, replacing any existing value.
+// If the key is duplicated, only first value will be replaced.
```
AlekSi, line 223: Why not panic in that case?

internal/types/document_sort.go
```diff
@@ -0,0 +1,32 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+import "sort"
+
+// SortByKeys sorts the document by its keys. It modifies the document in place.
```
AlekSi, line 19: ⚠️ I don't think we should have that method or implement sort.Interface. There are no uses for modifying a document like this in place

internal/types/document.go
```diff
@@ -35,8 +35,13 @@ type document interface {
 //
 // Duplicate field names are not supported yet.
```
AlekSi, line 36: They are supported now

internal/types/document.go
AlekSi: I think we should panic in that case

internal/types/document.go
```diff
@@ -250,9 +276,11 @@ func (d *Document) GetByPath(path Path) (any, error) {
 // SetByPath sets value by given path. If the Path has only one element, it sets the value for the given key.
 // If some parts of the path are missing, they will be created.
 // The Document type will be used to create these parts.
+// If multiple fields match the path, only the first one will be set.
```
AlekSi, line 279: Let's panic

AlekSi (changes requested):
~Don't~ do panic

internal/types/document.go
```diff
@@ -113,27 +127,46 @@ func (d *Document) Len() int {
...
+	return len(d.fields)
 }
 
 // Map returns this document as a map. Do not modify it.
 //
+// If there are duplicate keys in the document, the last value is set in the corresponding field.
+//
 // It returns nil for nil Document.
+//
+// Deprecated: as Document might have duplicate keys, map is not a good representation of it.
```
noisersup, line 140: Don't we want to add some function returning a custom struct []{key,value} for that? It would be helpful when you want to access every value inside the document

noisersup (commented):
Overall LGTM

internal/types/document.go
rumyantseva: We have other tickets to refactor map-related things. So, probably, yes, but now now :)

internal/types/document.go
```diff
@@ -113,27 +127,46 @@ func (d *Document) Len() int {
...
+
+	m := make(map[string]any, len(d.fields))
+	for _, field := range d.fields {
+		m[field.key] = field.value
+	}
+
+	return m
 }
 
-// Keys returns document's keys. Do not modify it.
```
AlekSi, line 129: Well, yeah, someone ignored that warning :)

w84thesun (commented):
I do like how it turned out. Especially the updated `Set` method 😄
I don't quite understand how it would work with the data documents on `insert` which could contain duplicate keys. Do we have a test for that case?

internal/types/document.go
rumyantseva: fixed

internal/types/document_validation.go
rumyantseva: added a unit test and integration test (skipped for another reason for now)

internal/types/document.go
rumyantseva: fixed

internal/types/document.go
rumyantseva: fixed

internal/types/document_sort.go
rumyantseva: fixed

internal/types/document.go
rumyantseva: fixed

chilagrow (approved):

w84thesun (approved):

noisersup (approved):

Issue #1151, opened 2022-09-15
Validate command documents after reading them from the wire


w84thesun:
We should disallow NaN and negative zero values globally for all data and "command" (that probably do not exist) documents.

Let's reject them in the `wire` package after reading them from the connection. `bson` package should still support them because we need to read the full documents first. `fjson` package should still support them because we need it for logging. But let's remove all checks for NaNs and negative zeros from other packages.

The implementation probably should be an unexported function in the `wire` package, not a method in the `types` package.

References to this issue in the source code, diff documentation, integration tests, add unit tests and dance tests are in the scope (as usual).

Add check in both [op_msg,go ](https://github.com/FerretDB/FerretDB/blob/main/internal/wire/op_msg.go#L58) and op_query.go 

Dance tests scope:
- [x] NaN, negative zero in data documents - copy diff tests to validate documents and create a separate file

rumyantseva:
We need to decide about command documents validation - where to call from etc

PR #1296, opened 2022-10-20
Deprecate infinity values in data documents

noisersup:
# Description

This PR is a part of the #1151.

## Readiness checklist


internal/types/document_validation.go
```diff
@@ -38,9 +39,9 @@ func (e *ValidationError) Error() string {
 // ValidateData checks if the document represents a valid "data document".
 // If the document is not valid it returns *ValidationError.
 func (d *Document) ValidateData() error {
-	// The following block should be used to checks that keys are valid.
-	// All further key related validation rules should be added here.
```
noisersup, line 42: @rumyantseva is there any background behind making this loop only for keys? Can I leave the change or should I make another loop for values?

internal/types/document.go
```diff
@@ -93,6 +93,10 @@ func NewDocument(pairs ...any) (*Document, error) {
 		}
 	}
 
+	if err := doc.ValidateCmd(); err != nil {
+		return nil, fmt.Errorf("types.NewDocument: %w", err)
+	}
+
```
noisersup, line 99: Not sure if my reasoning behind command documents is correct

internal/types/document_validation.go
rumyantseva: @noisersup up to you. Please take a looks at https://github.com/FerretDB/FerretDB/pull/1293, there won't be separate way to store values soon, so ost likely, both keys and values will be in the same loop.

internal/types/document.go
rumyantseva: This is a good question! Not every document is a command document. I think `ValidateCmd` should be called from handlers. @AlekSi, please correct me if I'm wrong.

w84thesun (approved):
LGTM 👍

integration/commands_administration_test.go
```diff
@@ -876,8 +876,6 @@ func TestCommandsAdministrationServerStatusMetrics(t *testing.T) {
 	} {
 		name, tc := name, tc
 		t.Run(name, func(t *testing.T) {
-			t.Parallel()
```
rumyantseva, line 879: Why did you remove it?

internal/types/document_validation_test.go
```diff
@@ -46,6 +47,14 @@ func TestDocumentValidateData(t *testing.T) {
 			doc:    must.NotFail(NewDocument("$v", "bar")),
 			reason: errors.New(`invalid key: "$v" (key must not contain $)`),
 		},
+		"Inf+": {
+			doc:    must.NotFail(NewDocument("v", math.Inf(1))),
+			reason: errors.New(`invalid value: +Inf (infinity values are not allowed)`),
```
rumyantseva, line 52: Let's also log the field that gave an invalid value (so, it'll be easier to debug and just more user-friendly in general).

rumyantseva (changes requested):
Overal looks good to me.
I have a few minor questions, plus we need to document the difference in `diff.md`. 
Also, from my perspective, the tests you removed should be moved to `dance` (maybe you have already considered it).

integration/query_comparison_test.go
```diff
@@ -588,19 +572,11 @@ func TestQueryComparisonGt(t *testing.T) {
...
-				"double", "double-big", "double-max", "double-positive-infinity", "double-smallest", "double-whole",
+				"double", "double-big", "double-max", "double-smallest", "double-whole",
 				"int32", "int32-max",
 				"int64", "int64-big", "int64-max",
 			},
 		},
-		"DoubleMax": {
-			value:       math.MaxFloat64,
-			expectedIDs: []any{"double-positive-infinity"},
-		},
```
chilagrow, line 599: We don't need to remove this test case.

integration/commands_administration_test.go
noisersup: I believe that's the reason why pg tests fail sometimes. `serverStatus` is the command which is not related to specific collection, so running these test cases in parallel makes it possible to read values in "wrong" order

internal/types/document_validation_test.go
noisersup: Does it look ok?

integration/query_comparison_test.go
noisersup: Nice catch!

chilagrow (approved):
:rocket:

integration/commands_administration_test.go
AlekSi: But what part of the test expects a particular order?

internal/types/document_validation_test.go
rumyantseva: Yeah, that looks good to me.

integration/commands_administration_test.go
noisersup: Nevermind, now looking at tests it looks that they are order agnostic

AlekSi (approved):

rumyantseva (changes requested):
I didn't get an answer to the question from my past review:
- when will we update diff.md
- when will we add dance tests

AlekSi:
> when will we add dance tests

It is impossible to do that in this PR.

> when will we update diff.md

When we will be back working on this issue.

That PR is good enough otherwise. It does not close the issue.

rumyantseva (approved):
All the questions are answered, and the code looks good to me.

chilagrow (approved):

PR #1312, opened 2022-10-24
Allow dashes (`-`) in collection names

### LINKED ISSUE #1280: Allow dashes (`-`) in collection names
### What should be done?

Currently, we reject collection names with dashes (`-`). We should allow them because they are common in real-life apps.

We should update:
* [x] code;
* [x] tests;
* [ ] `testutil.CollectionName` helper;
* [x] diff tests in the dance repo;
* [x] documentation.

noisersup:
# Description

Closes #1280 .

## Readiness checklist


noisersup:
We don't check for dashes inside the dance tests so I'm not sure what should we change there

AlekSi (approved):

rumyantseva (commented):
The code looks good to me in general.
I remember, at planning, we were discussing if we should start using dashes instead of undescores in `CollectionName` testutil's handler - https://github.com/FerretDB/FerretDB/blob/74d6f833660083a9e28cebd60186ca1224f0e4d1/internal/util/testutil/db.go#L92. So, you decided not doing it?

noisersup:
@rumyantseva I was confused about this part and thought that discussion and mention about this function was only because it **would** not work correctly. Now, I can see that the point of this change is different. Do we finally prefer using underscores there or dashes?

rumyantseva:
@noisersup I think the idea was to have more test cases. Personally, I'm ok with either dashes or underscores.

rumyantseva (approved):
LGTM as everything is tested.

PR #1316, opened 2022-10-24
Add stress tests for settings table and fix simple issues with transactions

### LINKED ISSUE #1282: Add stress tests for settings table
* [x] Add a stress integration test that creates X collections in parallel in X goroutines.
* [ ] As the test will always fail and implementation is not in scope, either skip the test or not merge it.

rumyantseva:
# Description

Closes #1282. 

See also my comments in the original issue.
See also #1318 for `SchemaStats` related issues.

## Readiness checklist


integration/create_test.go
```diff
@@ -26,6 +27,43 @@ import (
...
+	wg.Wait()
+
+	// check that all collections were created, and we can query them
+	for i := 0; i < collNum; i++ {
+		t.Run(fmt.Sprintf("check_stress_%d", i), func(t *testing.T) {
+			t.Parallel()
+
+			collName := fmt.Sprintf("stress_%d", i)
+			res := db.Collection(collName).FindOne(ctx, bson.D{})
+			require.Equal(t, mongo.ErrNoDocuments, res.Err())
```
chilagrow, line 62: For a non-existing collection, we do not get `ErrNoDocuments`, correct? Just want to ensure that `ErrNoDocuments` is an error we get if the collection exists and no document in the collection.

internal/handlers/pg/msg_create.go
```diff
@@ -72,25 +72,44 @@ func (h *Handler) MsgCreate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+	// but it's a normal situation, and we should create both collections in such a case.
+
 	err = h.PgPool.InTransaction(ctx, func(tx pgx.Tx) error {
 		if err := pgdb.CreateDatabaseIfNotExists(ctx, tx, db); err != nil {
-			if errors.Is(pgdb.ErrInvalidDatabaseName, err) {
+			switch {
+			case errors.Is(err, pgdb.ErrAlreadyExist):
+				// If the DB was created from a parallel query, it's ok.
+				// However, in this case one of the transaction queries failed,
+				// so we need to rollback the transaction.
```
chilagrow, line 86: Nice catch!

integration/create_test.go
rumyantseva: Good point! If collection doesn't exist, it gives `ErrNoDocuments` 🙈

rumyantseva:
Ooops! Not ready for review yet!

integration/create_test.go
rumyantseva: That actually revealed some other problems, see the recent update :)

integration/create_test.go
```diff
@@ -26,6 +27,57 @@ import (
 	"github.com/FerretDB/FerretDB/integration/setup"
 )
 
+func TestCreateStressPostgres(t *testing.T) {
+	setup.SkipForPostgresWithReason(t, "https://github.com/FerretDB/FerretDB/issues/1206")
+	setup.SkipForTigrisWithReason(t, "Tigris needs a schema for collection creation")
```
AlekSi, line 32: Could we add a schema and make this test handler-independent?

integration/create_test.go
```diff
@@ -26,6 +27,57 @@ import (
...
+		wg.Add(1)
+
+		go func(i int) {
+			collName := fmt.Sprintf("stress_%d", i)
+			err := db.CreateCollection(ctx, collName)
+			_, errIns := db.Collection(collName).InsertOne(ctx, bson.D{{"_id", "foo"}, {"v", "bar"}})
+
+			wg.Done()
+
+			require.NoError(t, err)
```
AlekSi, line 52:
```suggestion
			assert.NoError(t, err)
```

`require` does not work correctly from non-main goroutine: https://pkg.go.dev/testing#T.FailNow

integration/create_test.go
```diff
@@ -26,6 +27,57 @@ import (
...
+	var wg sync.WaitGroup
+	for i := 0; i < collNum; i++ {
+		wg.Add(1)
+
+		go func(i int) {
+			collName := fmt.Sprintf("stress_%d", i)
+			err := db.CreateCollection(ctx, collName)
+			_, errIns := db.Collection(collName).InsertOne(ctx, bson.D{{"_id", "foo"}, {"v", "bar"}})
+
+			wg.Done()
```
AlekSi, line 50:
```suggestion
			defer wg.Done()
```

And move on top

integration/create_test.go
```diff
@@ -26,6 +27,57 @@ import (
...
+func TestCreateStressPostgres(t *testing.T) {
+	setup.SkipForPostgresWithReason(t, "https://github.com/FerretDB/FerretDB/issues/1206")
+	setup.SkipForTigrisWithReason(t, "Tigris needs a schema for collection creation")
+
+	t.Parallel()
+
+	ctx, collection := setup.Setup(t) // no providers there, we will create collections concurrently
+	db := collection.Database()
+
+	collNum := 10
```
AlekSi, line 39:
```suggestion
	const collNum = runtime.GOMAXPROCS() * 10
```

integration/create_test.go
```diff
@@ -26,6 +27,57 @@ import (
...
+
+	var wg sync.WaitGroup
+	for i := 0; i < collNum; i++ {
+		wg.Add(1)
+
+		go func(i int) {
+			collName := fmt.Sprintf("stress_%d", i)
```
AlekSi, line 46: I think we need an extra trigger to start (not just create) all goroutines at the same time. Something like:

```go
ready := make(chan struct{}, collNum)
start := make(chan struct{})

// ...
go func(i int) {
  ready <- struct{}
  <-start
  // ...
}

// ...

for i := 0; i < collNum; i++ {
  <- ready
}

close(start)
```

integration/create_test.go
```diff
@@ -26,6 +27,57 @@ import (
...
+		}(i)
+	}
+
+	wg.Wait()
+
+	colls, err := db.ListCollectionNames(ctx, bson.D{})
+	require.NoError(t, err)
+
+	// TODO https://github.com/FerretDB/FerretDB/issues/1206
+	// This check fails now. Even though all the collections are created as separate tables in the database,
```
AlekSi, line 63: Well, it doesn't, because we skip that test for both PostgreSQL and Tigris

integration/create_test.go
```diff
@@ -26,6 +27,57 @@ import (
...
+	// the settings table doesn't store all of them because of concurrency issues.
+	require.Len(t, colls, collNum)
+
+	// check that all collections were created, and we can query them
+	for i := 0; i < collNum; i++ {
+		t.Run(fmt.Sprintf("check_stress_%d", i), func(t *testing.T) {
```
AlekSi, line 69: `i := i` to make an iteration-specific copy of `i`

integration/create_test.go
```diff
@@ -26,6 +27,57 @@ import (
...
+			t.Parallel()
+			collName := fmt.Sprintf("stress_%d", i)
```
AlekSi, line 71: otherwise, that `i` could be incorrect in parallel subtest

AlekSi (commented):
tbh, I'm quite confused how things are split between this PR and #1318: both add tests, tweak tests, skip tests, and fix code. Let's make one PR with this stress test that is not skipped and fixes that make it pass

integration/create_test.go
rumyantseva: I understand the idea with `start`, but what is `ready` for? Is it to replace `wg`?

AlekSi: To start all created goroutines when they are ready, not just when they are created.

We could use a wait group for that, but a different variable.

integration/create_test.go
rumyantseva: fixed

integration/create_test.go
rumyantseva: fixed

integration/create_test.go
rumyantseva: fixed

integration/create_test.go
rumyantseva: fixed

integration/create_test.go
rumyantseva: This is unclear to me, could you explain the idea behind it?

chilagrow (dismissed):
Looks good!

integration/create_test.go
```diff
@@ -26,6 +29,94 @@ import (
...
 
+func TestCreateStress(t *testing.T) {
+	setup.SkipForPostgresWithReason(t, "https://github.com/FerretDB/FerretDB/issues/1206")
+
+	t.Parallel()
+
+	ctx, collection := setup.Setup(t) // no providers there, we will create collections concurrently
+	db := collection.Database()
+
+	collNum := runtime.NumCPU() * 10
```
AlekSi: That should be `runtime.GOMAXPROCS(-1)` as we are interested in the number of CPU cores ("procs") that the Go scheduler is allowed to use, not just in the number of CPU cores

integration/create_test.go
```diff
@@ -26,6 +29,94 @@ import (
...
+			}`, collName, i,
+			)
+			opts := options.CreateCollectionOptions{
+				Validator: bson.D{{"$tigrisSchemaString", schema}},
+			}
+
+			// Attempt to create a collection for Tigris with a schema.
+			// If we get an error, it's not Tigris, so we create collection without schema
+			err := db.CreateCollection(ctx, collName, &opts)
+			if err != nil {
```
AlekSi, line 75: Well, that could also be a concurrency issue that we just ignore, defeating the purpose of the test.

I think we should check that error. We already do that in one other test

internal/handlers/tigris/tigrisdb/databases.go
```diff
@@ -37,19 +39,32 @@ func (tdb *TigrisDB) createDatabaseIfNotExists(ctx context.Context, db string) (
...
 	// but keep in mind that it can be created in concurrent connection.
-	err = tdb.Driver.CreateDatabase(ctx, db)
-	switch err := err.(type) {
-	case nil:
-		return true, nil
-	case *driver.Error:
-		if IsAlreadyExists(err) {
-			return false, nil
-		}
+	// If we detect that other creation is in flight, we give up to three attempts to create the database.
```
AlekSi, line 42: Do we want to keep it, or we want Tigris to fix it?

internal/handlers/tigris/tigrisdb/errors.go
```diff
@@ -34,6 +37,22 @@ func IsAlreadyExists(err error) bool {
 	return pointer.Get(e).Code == api.Code_ALREADY_EXISTS //nolint:nosnakecase // Tigris named their const that way
 }
 
+// IsOtherCreationInFlight returns true if an attempt to create the database with the given name is already in progress.
+// This function is implemented to keep nolint in a single place.
+func IsOtherCreationInFlight(err error) bool {
```
AlekSi, line 42: Let's un-export it

integration/create_test.go
AlekSi: Doh, too late!

internal/handlers/tigris/tigrisdb/databases.go
rumyantseva: I'm afraid we have to keep it. Tigris said they would fix the error code (so it's not "unknown" anymore). But as far as I understand, they consider such a situation possible - the DB is in the creation process from some other request, and we don't know yet if it'd be created successfully. So we need another attempt to understand how it went.

integration/create_test.go
rumyantseva: fixed

internal/handlers/tigris/tigrisdb/errors.go
rumyantseva: fixed

internal/handlers/tigris/tigrisdb/databases.go
AlekSi: My understanding is that their client would do retrying transparently for us.

Let's create an issue for that in our repo and add TODO there

rumyantseva: added

integration/create_test.go
```diff
@@ -26,6 +31,100 @@ import (
...
+					"v": {"type": "string"}
+				}
+			}`, collName, i,
+			)
+			opts := options.CreateCollectionOptions{
+				Validator: bson.D{{"$tigrisSchemaString", schema}},
+			}
+
+			// Attempt to create a collection for Tigris with a schema.
+			// If we get an error, it's not Tigris, so we create collection without schema
```
AlekSi, line 75: > it's not Tigris

Well, it is never Tigris. It is either FerretDB or MongoDB. And I guess only MongoDB returns that error?

AlekSi (dismissed):

integration/create_test.go
rumyantseva: @AlekSi, that's true, but I'm not sure how to write such a comment, could you rephrase it, so it looks better?

AlekSi: > If we get an error, that's MongoDB (FerretDB ignores that argument for non-Tigris handlers), so we create collection without schema

w84thesun (approved):
LGTM

chilagrow (approved):

noisersup (approved):

integration/create_test.go
```diff
@@ -26,6 +31,101 @@ import (
...
+			}
+
+			// Attempt to create a collection for Tigris with a schema.
+			// If we get an error, that's MongoDB (FerretDB ignores that argument for non-Tigris handlers),
+			// so we create collection without schema.
+			err := db.CreateCollection(ctx, collName, &opts)
+			if err != nil {
+				var cmdErr *mongo.CommandError
+				if errors.As(err, &cmdErr) {
+					if strings.Contains(cmdErr.Message, `unknown top level operator: $tigrisSchemaString`) {
```
AlekSi, line 81: We have the same comment there and at https://github.com/FerretDB/FerretDB/pull/1318#discussion_r1009398501, but the implementation of error checking is very different. Which one is correct?

PR #1326, opened 2022-10-26
forbid regex types and arrays in document's `_id` field

### LINKED ISSUE #1235: Check that `_id` is not regex or array
### What should be done?

This issue is based on old TODOs in the code that suggest checking for `_id` not being a regex or array.

TODO:
- [x] Check TODOs that mention the issue number (`https://github.com/FerretDB/FerretDB/issues/1235`) - let's remove them as the check for `_id` should be in validation function only
- [x] Implement the change for document validation
- [x] Add tests - set id as array, regex and expect to receive an error (unit tests for validation func, compat tests)

noisersup:
# Description

Closes #1235.

## Readiness checklist

### DIFF
--- a/internal/types/document.go
+++ b/internal/types/document.go
@@ -16,6 +16,7 @@ package types
 
 import (
 	"fmt"
+	"reflect"
 	"strconv"
 
 	"golang.org/x/exp/slices"
@@ -229,6 +230,16 @@ func (d *Document) Set(key string, value any) {
 
 	if key == "_id" {
 		// TODO check that value is not regex or array: https://github.com/FerretDB/FerretDB/issues/1235
+
+		switch value := value.(type) {
+		case Regex:
+			panic(fmt.Sprintf("types.Document.Set: _id is a regex: %v", value))
+
+		default:
+			if reflect.TypeOf(value).Kind() == reflect.Slice {
+				panic(fmt.Sprintf("types.Document.Set: _id is a slice: %v", value))
+			}
+		}
 
 		// ensure that _id is the first field
 		if i := slices.Index(d.Keys(), key); i >= 0 {
--- a/internal/types/document_test.go
+++ b/internal/types/document_test.go
@@ -54,6 +54,18 @@ func TestDocument(t *testing.T) {
 		assert.Equal(t, "foo", doc.Command())
 	})
 
+	t.Run("ArrayValues", func(t *testing.T) {
+		t.Parallel()
+		var doc Document
+		doc.Set("_id", []string{})
+	})
+
+	t.Run("RegexValues", func(t *testing.T) {
+		t.Parallel()
+		var doc Document
+		doc.Set("_id", Regex{Pattern: "Boo! I'm Regex placed in a wrong place!!"})
+	})
+
 	t.Run("NewDocument", func(t *testing.T) {
 		t.Parallel()
 
--- a/internal/types/document_test.go
+++ b/internal/types/document_test.go
@@ -57,7 +57,9 @@ func TestDocument(t *testing.T) {
 	t.Run("ArrayValues", func(t *testing.T) {
 		t.Parallel()
 		var doc Document
-		doc.Set("_id", []string{})
+		assert.Panicsf(t, func() {
+			doc.Set("_id", []string{"foo", "bar"})
+		}, `types.Document.Set: _id is a slice: []string{"foo","bar"}`)
 	})
 
 	t.Run("RegexValues", func(t *testing.T) {


internal/handlers/pg/msg_insert.go
```diff
@@ -123,5 +123,5 @@ func (h *Handler) insert(ctx context.Context, tx pgx.Tx, sp *pgdb.SQLParam, doc
 		return common.NewErrorMsg(common.ErrInvalidNamespace, msg)
 	}
 
-	return lazyerrors.Error(err)
+	return common.NewWriteErrorMsg(53, err.Error())
```
noisersup, line 126: Not sure how to handle error code here. Technically no test is failing so I assume that's only one code used here. However, I assume that we want to make this code a variable inside ErrorCodes

noisersup (commented):
It's not ready to merge but please take a look

integration/insert_compat_test.go
```diff
@@ -0,0 +1,147 @@
...
+		"InsertEmptyDocument": {
+			insert:     bson.D{},
+			resultType: nonEmptyResult,
+		},
+		"InsertIDArray": {
+			insert:     bson.D{{"_id", bson.A{"foo", "bar"}}},
+			resultType: emptyResult,
+		},
+		"InsertIDRegex": {
+			insert:     bson.D{{"_id", "^regex$"}},
```
noisersup, line 44: I understand thats not a valid regex type syntax and that's just a basic string, but I noticed unexpected behavior. This operation is correct so the change is applied to collection, but every subtest inside this test cases says that this id is duplicated, which from my understanding shouldn't be true because  `SetupCompat` creates different collections for every subtest

internal/handlers/pg/msg_insert.go
rumyantseva: Yeah, this `53` should be a constant in the `internal/handlers/common/error.go` file.

internal/types/document_validation.go
```diff
@@ -79,5 +79,14 @@ func (d *Document) ValidateData() error {
 		return newValidationError(fmt.Errorf("invalid document: document must contain '_id' field"))
 	}
 
+	v := must.NotFail(d.Get("_id"))
+
+	switch v.(type) {
+	case Array, *Array:
+		return fmt.Errorf("The '_id' value cannot be of type array")
```
rumyantseva, line 86: Ok, after reading the whole PR, I got what you asked yesterday :) 

I suggest adding the `code common.ErrorCode` field to `ValidationError` struct, so we can either set it as `ErrFailedToParse` or that new const for `53` for those new cases. 

Then we can still use `ValidationError` for all the cases here, and we can extend it if we need more error codes in the future.

@noisersup @chilagrow @w84thesun @AlekSi, what do you think?

noisersup: so the ValidationErrors's `Error()` function will return different output when the code is equal to `ErrFailedToParse` and different if it's set to `53`?

rumyantseva:
```
if errors.As(err, &valErr) {
	return common.NewErrorMsg(valErr.Code(), err.Error())
}
```

where `valErr.Code()` calls a new method `Code` that returns the `code` field from the structure.

rumyantseva: And `Error` func can be rewritten, yes (e.g. we can remove `Invalid document, reason` part).

rumyantseva: No need to handle different error codes differently, let's rather make`ValidateData` more universal (and remove everything that doesn't let it be universal).

integration/update_compat_test.go
```diff
@@ -125,7 +125,7 @@ func testUpdateCompat(t *testing.T, testCases map[string]updateCompatTestCase) {
 								targetErr = UnsetRaw(t, targetErr)
 								compatErr = UnsetRaw(t, compatErr)
 
-								// Skip updates that could not be performed due to Tigris schema validation.
+								// Skip inserts that could not be performed due to Tigris schema validation.
```
rumyantseva, line 128: I think in this case, we still skip updates, not inserts (I guess it was copy-pasted after an attempt to rewrite something 😉)

rumyantseva, line 128:
```suggestion
								// Skip updates that could not be performed due to Tigris schema validation.
```

noisersup: Yeah this part of code was rewritten due to linters though I needed to "rollback" for now and I probably appended the comment by accident, nice catch :)

internal/types/document_validation.go
noisersup: Yay!

rumyantseva: @noisersup, and yesterday I didn't get what you meant, sorry for that.

noisersup: No worries!

integration/insert_compat_test.go
rumyantseva: I think I understand :)

You have a loop `for _, doc := range allDocs {`, and it seems that you try to insert multiple times there. It was reasonable for update compat test where we might want to try updating every single value, but for `insert` it doesn't make sense. I'll leave a comment there.

integration/insert_compat_test.go
```diff
@@ -0,0 +1,147 @@
...
+			require.NotNil(t, insert)
+
+			var nonEmptyResults bool
+			for i := range targetCollections {
+				targetCollection := targetCollections[i]
+				compatCollection := compatCollections[i]
+				t.Run(targetCollection.Name(), func(t *testing.T) {
+					t.Helper()
+
+					allDocs := FindAll(t, ctx, targetCollection)
```
rumyantseva, line 89: I think this is unnecessary (nor is the loop with `allDocs`). We can insert the doc once, and it's enough.

integration/insert_compat_test.go
```diff
@@ -0,0 +1,147 @@
...
+						id, ok := doc.Map()["_id"]
+						require.True(t, ok)
+
+						t.Run(fmt.Sprint(id), func(t *testing.T) {
+							t.Helper()
+
+							var targetInsertRes, compatInsertRes *mongo.InsertOneResult
+							var targetErr, compatErr error
+
+							targetInsertRes, targetErr = targetCollection.InsertOne(ctx, insert)
```
rumyantseva, line 101: So, currently, you are trying to insert as many times as the number of docs in the collection (as you iterate through them). If insertion was successful, the second one would give you "duplicate key".

rumyantseva (changes requested):
I left some comments where changes might be needed.

integration/insert_compat_test.go
noisersup: Oh that makes sense right now! Thank you:)

integration/insert_compat_test.go
```diff
@@ -0,0 +1,143 @@
...
+			require.NotNil(t, insert)
+
+			var nonEmptyResults bool
+			for i := range targetCollections {
+				targetCollection := targetCollections[i]
+				compatCollection := compatCollections[i]
+				t.Run(targetCollection.Name(), func(t *testing.T) {
+					t.Helper()
+
+					t.Helper()
```
AlekSi, line 88: Very helper 😆

internal/types/document_validation.go
```diff
@@ -79,5 +79,14 @@ func (d *Document) ValidateData() error {
 		return newValidationError(fmt.Errorf("invalid document: document must contain '_id' field"))
 	}
 
+	v := must.NotFail(d.Get("_id"))
+
+	switch v.(type) {
+	case Array, *Array:
```
AlekSi, line 85:
```suggestion
	case *Array:
```

Array should never be passed by value

internal/types/document_validation.go
```diff
@@ -79,5 +79,14 @@ func (d *Document) ValidateData() error {
...
+		return fmt.Errorf("The '_id' value cannot be of type array")
+	case Regex, *Regex:
```
AlekSi, line 87:
```suggestion
	case Regex:
```

Scalar value should never be passed by pointer

internal/types/document_validation.go
noisersup: Using `common.ErrorCode` here creates import cycle. I suggest moving ValidationError into `common` package

AlekSi: Let's do the simplest thing now. We could always replace it with method/interface/types/etc later.

There:
```go
type ValidationErrorCode int

var (
  ErrBadID ValidationErrorCode = iota
)

type ValidationError struct {
  Code ValidationErrorCode
  reason error
}
```

Somewhere in handler layer:

```go
func checkError(err error) error {
switch err := err.(type) {
  case *ValidationError:
    switch err.Code {
      case types.ErrBadID:
        return ErrInvalidID
    }
}
```

noisersup:
Need a help with a linter. It screams about deprecation because of generating files, though we don't have any quick alternatives to the `Map` function and it's probably a topic for another issue. I tried to place //nolint:statickcheck but it said that the nolint is unused

rumyantseva (changes requested):
I like the current implementation idea! Added a couple of comments (not super critical).

internal/handlers/common/error.go
```diff
@@ -183,3 +187,20 @@ func formatBitwiseOperatorErr(err error, operator string, maskValue any) error {
...
+// CheckError checks error type and returns properly translated error.
+func CheckError(err error) error {
+	var ve *types.ValidationError
+	if errors.As(err, &ve) {
+		switch ve.Code() {
+		case types.ErrBadID:
+			return NewWriteErrorMsg(ErrInvalidID, err.Error())
+		case types.ErrValidation:
+			fallthrough
+		default:
```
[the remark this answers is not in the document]
AlekSi: We also have linters that check that switches are exhaustive, and that `default` would prevent them from working

integration/insert_compat_test.go
```diff
@@ -0,0 +1,143 @@
...
+	for name, tc := range testCases {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Helper()
+
+			if tc.skip != "" {
+				t.Skip(tc.skip)
+			}
+			if tc.skipForTigris != "" {
+				setup.SkipForTigrisWithReason(t, tc.skipForTigris)
```
chilagrow, line 72: There is skip for tigris due to Tigris schema validation later on. Do we need this skip?

integration/insert_compat_test.go
```diff
@@ -0,0 +1,143 @@
...
+					var targetFindRes, compatFindRes bson.D
+					targetCur, err := targetCollection.Find(ctx, bson.D{{}})
+					require.NoError(t, err)
+					targetCur.Decode(&targetFindRes)
+
+					compatCur, err := compatCollection.Find(ctx, bson.D{{}})
+					require.NoError(t, err)
+					compatCur.Decode(&targetFindRes)
+
+					AssertEqualDocuments(t, compatFindRes, targetFindRes)
```
rumyantseva, line 129: Should we find by `_id` to compare inserted documents here? (The idea is to check that what we inserted is really stored equally).

rumyantseva (approved):
Looks good to me overall.

I think, one tiny thing could be improved in insert compat tests (see my comment), but as this task blocks some other things, I consider this request for changes minor.

chilagrow (commented):
Seems good. I just have questions about skipped tests.

integration/insert_compat_test.go
noisersup: There are a couple of things that are not implemented for tigris handler yet which is a reason for this skip

integration/insert_compat_test.go
```diff
@@ -0,0 +1,143 @@
...
+		},
+	}
+
+	testInsertCompat(t, testCases)
+}
+
+type insertCompatTestCase struct {
+	insert        bson.D                   // required
+	resultType    compatTestCaseResultType // defaults to nonEmptyResult
+	skip          string                   // skips test if non-empty
```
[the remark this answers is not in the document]
noisersup: It's commonly used inside compat tests and as this code will be probably shared between some tests I assume that skip functionality will be helpful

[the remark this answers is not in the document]
chilagrow: Ah I see, ok let's keep it for the consistency.

integration/insert_compat_test.go
noisersup: I think that checking whole collection content is much more full-bodied. What if the insert function misbehaves and creates additional entry in collection? Checking only for specific `_id` would not cover this case

chilagrow (approved):

PR #1334, opened 2022-10-27
Deprecate nested arrays

### LINKED ISSUE #1305: Remove support for nested arrays
Out data validation should reject documents that contain arrays immediately within arrays.
* `{"foo": [1, [2]]}` – should be rejected;
* `{"foo": [ 1, {"bar": [2]} ]}` – should be allowed.

DoD:

- [x] Implement the check in data validation function
- [x] Write unit tests
- [x] Some integration tests in ferrretdb will start failing because there are nested arrays there, such tests need to be removed
- [ ] If it's easy to remove some nested-arrays-related logic (e.g. some code is not in use), feel free to remove it. 
- [x] Update `diff.md`
- [ ] Write dance diff tests

w84thesun:
# Description

Nested array logic is still in place. Will try to remove it in a separate PR after the dance tests would be done.

Closes #1305.

## Readiness checklist


rumyantseva (approved):
LGTM! Also thanks for validation of nested documents in `ValidateData`, we sort of forgot about it :)

integration/query_array_test.go
```diff
@@ -543,7 +508,7 @@ func TestQueryArrayAll(t *testing.T) {
 		},
 		"EmptyNested": {
 			filter:      bson.D{{"v", bson.D{{"$all", bson.A{bson.A{}}}}}},
```
chilagrow, line 510: Probably outside of the scope, when we query with nested array filter, what is our expectation? maybe @AlekSi

w84thesun: Nice catch, removed that test.

rumyantseva (approved):
LGTM

internal/types/document_validation.go
```diff
@@ -38,6 +41,11 @@ func (e *ValidationError) Error() string {
 	return fmt.Sprintf("Invalid document, reason: %s.", e.reason)
 }
 
+// Unwrap returns unwrapped error.
+func (e *ValidationError) Unwrap() error {
```
AlekSi, line 45: With `reason` not being exported and `Error() error` using `%s` instead of `%w`, `Unwrap() error` exposes internal details that were hidden before. Let's keep them hidden

AlekSi (changes requested):

integration/query_array_test.go
```diff
@@ -525,11 +495,6 @@ func TestQueryArrayAll(t *testing.T) {
 			expectedIDs: []any{"array-three", "array-three-reverse"},
 			expectedErr: nil,
 		},
-		"ArrayEmbeddedEqual": {
-			filter:      bson.D{{"v", bson.D{{"$all", bson.A{bson.A{int32(42), "foo"}}}}}},
-			expectedIDs: []any{"array-first-embedded", "array-last-embedded", "array-middle-embedded"},
-			expectedErr: nil,
-		},
```
chilagrow, line 532: ArrayEmbeddedReverseOrder also has nested array filter.

chilagrow (commented):
One tiny comment

noisersup (dismissed):
Looks good to me!

internal/types/document.go
```diff
@@ -23,6 +23,11 @@ import (
 	"github.com/FerretDB/FerretDB/internal/util/must"
 )
 
+const (
+	// IdKey is the key for _id field.
+	IdKey = "_id"
```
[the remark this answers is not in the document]
w84thesun: It's because of linter warnings about that constant.

internal/types/document_validation.go
w84thesun: Removed that method.

internal/types/document.go
rumyantseva, line 28: In the file `document_validation.go` ids in consts are called ID (capital letters), so I guess here those should be capital letters too?

```suggestion
	IDKey = "_id"
```

rumyantseva, line 28: Actually, do we need this const to be exported? Or can we leave it unexported?

```suggestion
	idKey = "_id"
```

rumyantseva (commented):
`ValidateData` logic looks good to me, one tiny suggestion to make unexported a const that doesn't need to be exported.

internal/types/document_validation.go
```diff
@@ -85,31 +89,60 @@ func (d *Document) ValidateData() error {
...
+		var vErr *ValidationError
+
+		switch v := value.(type) {
+		case *Document:
+			err := v.ValidateData()
+			if err != nil && errors.As(err, &vErr) {
+				if vErr.code != ErrIDNotFound {
+					return err
+				}
+			}
```
chilagrow, line 107: This will silently ignore error if `ValidateData()` ever in the future returns other than ValidationError. I know it's not the case now but I think we could be a bit more defensive here, WDYT?

chilagrow (commented):
Just a small comment :hugs:

internal/types/document_validation.go
w84thesun: Nice catch! Fixed it.

internal/types/document.go
w84thesun: We probably don't need it, unexported it.

chilagrow (dismissed):

internal/types/document.go
```diff
@@ -23,6 +23,11 @@ import (
 	"github.com/FerretDB/FerretDB/internal/util/must"
 )
 
+const (
+	// The idKey is key for _id field.
+	idKey = "_id"
```
AlekSi, line 28: tbh, I never understood constants like this one, especially given the fact that we have "_id" string across many files in many packages

internal/types/document_validation.go
```diff
@@ -85,31 +89,66 @@ func (d *Document) ValidateData() error {
...
+		case float64:
+			// TODO Add dance tests for infinity: https://github.com/FerretDB/FerretDB/issues/1151
+			if math.IsInf(v, 0) {
+				return newValidationError(
+					ErrValidation, fmt.Errorf("invalid value: { %q: %f } (infinity values are not allowed)", key, v),
+				)
+			}
+		case Regex:
+			if key == idKey {
+				return newValidationError(ErrWrongIDType, fmt.Errorf("The '_id' value cannot be of type regex"))
```
AlekSi, line 145: For example, should that error text use `idKey`?

internal/types/document.go
w84thesun: I introduce that constant just because of the linter warning.

internal/types/document_validation.go
w84thesun: Not sure. I think no, because this is the error message and there is no other possible options.

chilagrow (approved):

noisersup (approved):

rumyantseva (approved):
LGTM

internal/types/document.go
AlekSi: In the next PR, let's remove that constant and tweak linter settings instead. We could add "_id" string to exceptions, for example

AlekSi (approved):

internal/types/document.go
AlekSi: @w84thesun Please ack, resolve conversation, and merge PR

PR #1380, opened 2022-11-08
Implement `enableFreeMonitoring`, `disableFreeMonitoring` and `getFreeMonitoringStatus` commands

### LINKED ISSUE #1323: Implement `enableFreeMonitoring` / `disableFreeMonitoring`
### What should be done?

When the telemetry state is undecided, it should be possible to enable and disable it at runtime.

Let's use `enableFreeMonitoring` and `disableFreeMonitoring` commands for that. It should not be possible to change the state of telemetry is enabled or disabled via a command-line flag, an environment variable, or a file name.

Let's also implement `getFreeMonitoringStatus` command and tweak `serverStatus` command to return the current status.

noisersup:
# Description

Closes #1323.

## Readiness checklist

### DIFF
--- a/internal/handlers/dummy/msg_getfreemonitoringstatus.go
+++ b/internal/handlers/dummy/msg_getfreemonitoringstatus.go
@@ -23,5 +23,5 @@ import (
 
 // MsgGetFreeMonitoringStatus implements HandlerInterface.
 func (h *Handler) MsgGetFreeMonitoringStatus(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
-	return common.MsgGetFreeMonitoringStatus(ctx, msg)
+	return common.MsgGetFreeMonitoringStatus(ctx, msg, nil)
 }
--- a/internal/handlers/common/msg_getfreemonitoringstatus.go
+++ b/internal/handlers/common/msg_getfreemonitoringstatus.go
@@ -17,18 +17,30 @@ package common
 import (
 	"context"
 
+	"github.com/AlekSi/pointer"
 	"github.com/FerretDB/FerretDB/internal/types"
 	"github.com/FerretDB/FerretDB/internal/util/must"
+	"github.com/FerretDB/FerretDB/internal/util/state"
 	"github.com/FerretDB/FerretDB/internal/wire"
 )
 
 // MsgGetFreeMonitoringStatus is a common implementation of the getFreeMonitoringStatus command.
-func MsgGetFreeMonitoringStatus(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
+func MsgGetFreeMonitoringStatus(ctx context.Context, msg *wire.OpMsg, state *state.State) (*wire.OpMsg, error) {
+
+	telemetryState := "disabled"
+	telemetryMsg := "monitoring is not enabled"
+
+	if state != nil && pointer.Get(state.Telemetry) {
+		telemetryState = "enabled"
+		telemetryMsg = "monitoring is enabled"
+	}
+
 	var reply wire.OpMsg
+
 	must.NoError(reply.SetSections(wire.OpMsgSection{
 		Documents: []*types.Document{must.NotFail(types.NewDocument(
-			"state", "disabled",
-			"message", "monitoring is not enabled",
+			"state", telemetryState,
+			"message", telemetryMsg,
 			"ok", float64(1),
 		))},
 	}))
--- a/internal/handlers/pg/msg_getfreemonitoringstatus.go
+++ b/internal/handlers/pg/msg_getfreemonitoringstatus.go
@@ -23,5 +23,5 @@ import (
 
 // MsgGetFreeMonitoringStatus implements HandlerInterface.
 func (h *Handler) MsgGetFreeMonitoringStatus(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
-	return common.MsgGetFreeMonitoringStatus(ctx, msg)
+	return common.MsgGetFreeMonitoringStatus(ctx, msg, h.StateProvider.Get())
 }
--- a/internal/handlers/tigris/msg_getfreemonitoringstatus.go
+++ b/internal/handlers/tigris/msg_getfreemonitoringstatus.go
@@ -23,5 +23,5 @@ import (
 
 // MsgGetFreeMonitoringStatus implements HandlerInterface.
 func (h *Handler) MsgGetFreeMonitoringStatus(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
-	return common.MsgGetFreeMonitoringStatus(ctx, msg)
+	return common.MsgGetFreeMonitoringStatus(ctx, msg, h.StateProvider.Get())
 }


Fashander (approved):

website/docs/Reference/telemetry.md
```diff
@@ -80,6 +80,18 @@ Telemetry can be disabled using any of the following options:
    until the `donottrack` string is removed.
    :::
 
+5. Use the `db.disableFreeMonitoring()` command on runtime.
+
+   ```sh
+   test> db.disableFreeMonitoring()
+   { ok: 1 }
+   ```
```
rumyantseva, line 88: I look at the other examples and it seems that for consistency it should be something like
```suggestion
   ```js
db.disableFreeMonitoring()
   ```
```

(with no response example)

integration/commands_administration_test.go
```diff
@@ -915,6 +915,52 @@ func TestCommandsAdministrationServerStatusMetrics(t *testing.T) {
...
+			command:        bson.D{{"setFreeMonitoring", 1}, {"action", "disable"}},
+			expectedRes:    bson.D{{"ok", float64(1)}},
+			expectedStatus: "disabled",
+		},
+	} {
+		name, tc := name, tc
+
+		t.Run(name, func(t *testing.T) {
+			res := s.Collection.Database().RunCommand(s.Ctx, tc.command)
+			require.NoError(t, res.Err())
```
chilagrow, line 945: You'd want to check `expectedRes`.

integration/commands_freemonitoring_test.go
```diff
@@ -32,28 +31,22 @@ func TestCommandsFreeMonitoringGetFreeMonitoringStatus(t *testing.T) {
...
 	var actual bson.D
 	err := s.Collection.Database().RunCommand(s.Ctx, bson.D{{"getFreeMonitoringStatus", 1}}).Decode(&actual)
 	require.NoError(t, err)
 
-	m := actual.Map()
 	keys := CollectKeys(t, actual)
 
-	for k, item := range expected {
+	for _, k := range allowedKeys {
+		assert.Contains(t, allowedKeys, k)
```
chilagrow, line 45: Any reason why we don't check the value of each key anymore? Perhaps it's covered by other tests :thinking:

integration/commands_freemonitoring_test.go
```diff
@@ -64,24 +57,20 @@ func TestCommandsFreeMonitoringSetFreeMonitoring(t *testing.T) {
...
 	for name, tc := range map[string]struct {
-		command bson.D
-		err     *mongo.CommandError
+		command        bson.D
+		err            *mongo.CommandError
+		expectedRes    bson.D
+		expectedStatus string
 	}{
 		"Enable": {
-			command: bson.D{{"setFreeMonitoring", 1}, {"action", "enable"}},
```
chilagrow, line 71: Where do we plan to cover these tests where using command is disabled?

internal/util/telemetry/telemetry.go
```diff
@@ -60,15 +60,18 @@ func (s *Flag) UnmarshalText(text []byte) error {
 //   - common DO_NOT_TRACK environment variable;
 //   - executable name;
 //   - and the previously saved state.
-func initialState(f *Flag, dnt string, execName string, prev *bool, l *zap.Logger) (*bool, error) {
+//
+// The second returned value is true if the telemetry state should be locked, because of
+// setting telemetry via a command-line flag, an environment variable, or a filename.
+func initialState(f *Flag, dnt string, execName string, prev *bool, l *zap.Logger) (*bool, bool, error) {
```
chilagrow, line 66: It's quite difficult to follow each return value. Possibly using `var locked bool` and returning its value would improve.

internal/util/telemetry/telemetry.go
```diff
@@ -60,15 +60,18 @@ func (s *Flag) UnmarshalText(text []byte) error {
...
 	var disable bool
 
 	// https://consoledonottrack.com is not entirely clear about accepted values.
 	// Assume that "1", "t", "true", etc. mean that telemetry should be disabled,
 	// and other valid values, including "0" and empty string, mean undecided.
 	v, err := parseValue(dnt)
 	if err != nil {
-		return nil, err
+		return nil, false, err
```
chilagrow, line 74: missing test?

chilagrow (commented):
A few comments here. Great work so far!

integration/commands_freemonitoring_test.go
noisersup: MongoDB will respond with a different output. They will use the userReminder and url fields to notify about some FreeMonitoring specific things. The expected output for both handlers is fully covered inside the dance tests PR

integration/commands_freemonitoring_test.go
noisersup: I don't think there's easy way to make a check for that. We would need to initialize another FerretDB instance for that and we don't do anytthing similar in any test. I took the advice to make tests simple and get some feedback so if needed we would make some changes in tests

integration/commands_freemonitoring_test.go
chilagrow: Great thanks for the explanation!

internal/util/telemetry/telemetry.go
noisersup: It was already there so I missed that one, but sure this could be covered with a test, nice catch!

website/docs/Reference/telemetry.md
```diff
@@ -80,6 +80,18 @@ Telemetry can be disabled using any of the following options:
...
+
+   :::caution
+   If the telemetry is set via previously listed methods, it's not possible
```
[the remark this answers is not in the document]
noisersup: Okay then!:)

integration/commands_administration_test.go
noisersup: Nice catch!

integration/commands_freemonitoring_test.go
chilagrow: That's right. We run integration tests against an already running ferretdb, so it won't be simple. For now, let's go with what we have :)

internal/util/telemetry/telemetry.go
noisersup: Just to clarify, you mean that we could replace every `return ..., false, ...`  with `return ..., locked, ...`, right?

chilagrow: Yeah, for me that would make it easier to follow :hugs:

chilagrow (dismissed):

internal/handlers/dummy/msg_getfreemonitoringstatus.go
```diff
@@ -23,5 +23,5 @@ import (
 
 // MsgGetFreeMonitoringStatus implements HandlerInterface.
 func (h *Handler) MsgGetFreeMonitoringStatus(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
-	return common.MsgGetFreeMonitoringStatus(ctx, msg)
+	return common.MsgGetFreeMonitoringStatus(ctx, msg, nil)
```
rumyantseva, line 26: It looks a bit controversial to me that we check for state not being `nil` in `MsgGetFreeMonitoringStatus` but here we set it to `nil` on purpose. I understand the idea - this is just a stub, but I wonder if we should leave a comment or something.

website/docs/Reference/telemetry.md
```diff
@@ -80,6 +80,18 @@ Telemetry can be disabled using any of the following options:
    until the `donottrack` string is removed.
    :::
 
+5. Use the `db.disableFreeMonitoring()` command on runtime.
+
+   ```sh
```
rumyantseva, line 85:
```suggestion
   ```js
```

website/docs/Reference/telemetry.md
```diff
@@ -80,6 +80,18 @@ Telemetry can be disabled using any of the following options:
...
+   test> db.disableFreeMonitoring()
```
rumyantseva, line 86:
```suggestion
db.disableFreeMonitoring()
```

website/docs/Reference/telemetry.md
```diff
@@ -80,6 +80,18 @@ Telemetry can be disabled using any of the following options:
...
+   { ok: 1 }
```
rumyantseva, line 87:
```suggestion
```

internal/handlers/common/serverstatus.go
```diff
@@ -59,6 +64,15 @@ func ServerStatus(startTime time.Time, cm *connmetrics.ConnMetrics) (*types.Docu
...
 	}
 
+	telemetryState := "disabled"
+
+	switch {
+	case state.Telemetry == nil:
+		telemetryState = "undecided"
+	case pointer.GetBool(state.Telemetry):
+		telemetryState = "enabled"
+	}
```
rumyantseva, line 74: Don't we want to define `disabled`, `undecided`, `enabled` as constants in the telemetry related package? It seems that those words are met a few times in the source code.

rumyantseva (commented):
The logic looks quite clear to me. I left some relatively minor ideas. 
I think this PR needs an ultimate review from @AlekSi to make sure that the requirements around telemetry are met.

internal/handlers/common/serverstatus.go
noisersup: https://github.com/FerretDB/FerretDB/pull/1380/commits/4d18588abe0fd2bf50932edc0c4d9f7cb755dd54 something like that?

internal/handlers/dummy/msg_getfreemonitoringstatus.go
noisersup: Is it good now? :)

internal/handlers/common/serverstatus.go
rumyantseva: @noisersup yeah, looks good to me! (Linter complains as the comment should be either out of the `const` block or there should be dedicated comments for each cont).

rumyantseva (commented):
Looks good to me in general (only linter complaints are left)

rumyantseva (dismissed):
LGTM

chilagrow (dismissed):

AlekSi (commented):
There are merge conflicts now

internal/util/telemetry/telemetry.go
```diff
@@ -24,6 +24,13 @@ import (
 	"go.uber.org/zap"
 )
 
+// Telemetry states.
+const (
+	Enabled   string = "enabled"
+	Disabled  string = "disabled"
+	Undecided string = "undecided"
+)
```
AlekSi, line 32: We already have a representation for those: true, false, and nil. No need to add another one

internal/util/telemetry/reporter.go
```diff
@@ -78,11 +78,17 @@ type NewReporterOpts struct {
...
-	t, err := initialState(opts.F, opts.DNT, opts.ExecName, opts.P.Get().Telemetry, opts.L)
+	t, lock, err := initialState(opts.F, opts.DNT, opts.ExecName, opts.P.Get().Telemetry, opts.L)
 	if err != nil {
 		return nil, err
 	}
 
+	if t != nil && lock {
+		if err = opts.P.Update(func(s *state.State) { s.TelemetryLocked = true }); err != nil {
+			return nil, err
+		}
```
AlekSi, line 89: Those lines should be covered by test

integration/commands_administration_test.go
```diff
@@ -915,6 +915,49 @@ func TestCommandsAdministrationServerStatusMetrics(t *testing.T) {
 	}
 }
 
+func TestCommandsAdministrationServerStatusFreeMonitoring(t *testing.T) {
+	// this test shouldn't be run in parallel, because it requires specific state of the field which would be modify by the other tests.
```
w84thesun, line 919:
```suggestion
	// this test shouldn't be run in parallel, because it requires a specific state of the field which would be modified by the other tests.
```

internal/util/telemetry/telemetry.go
noisersup: @rumyantseva WDYT?

rumyantseva: @AlekSi I don't understand your comment. We don't use true, false, or nil in the places where we use these particular constants.

w84thesun: Those constants are used twice in the code, so maybe it's ok to leave them? Also, we can use those constants to compose the message like  "monitoring is " + state.

w84thesun (changes requested):
Overall good, small fix suggested.

internal/util/telemetry/telemetry.go
rumyantseva: (overall, I'm ok with whatever can help us finish this task asap)

internal/util/telemetry/reporter.go
noisersup: https://github.com/FerretDB/FerretDB/pull/1380/commits/3d6a9d7bc1ee0fd19e0fa7fcead28f7334d87157 done

internal/util/telemetry/telemetry.go
AlekSi: > We don't use true, false, or nil in the places where we use these particular constants.

Those constants represent things that already have a different representation. We don't use them in the state, for example.

> Those constants are used twice in the code, so maybe it's ok to leave them?

By that logic we should have a constant for "ok" string that is used everywhere.

noisersup: I reverted the change. All state's string representations are called without any const now

noisersup:
linters failing, let me clean my local environment

internal/util/telemetry/telemetry.go
rumyantseva: @AlekSi as someone who uses things like `http.MethodGet` instead of `"GET"` I don't feel it's bad to have a constant for `ok` :D But I got the point, yeah.

@noisersup thanks for reverting the change!

w84thesun (approved):

chilagrow (approved):

rumyantseva (approved):
LGTM

Fashander (approved):

AlekSi (approved):

Issue #593, opened 2022-05-11
Document code style for function calls


AlekSi:
In short:

# Good 1

Short line, `lll` linter does not complain, all other similar lines are short too:

```go
return false, NewError(ErrRegexMissingParen, err) 
```

# Good 2

That's just a standard Go.

```go
for {
  someCode()
  someOtherCode()
}
```

# Good 3

Long line, `lll` linter complains, or some similar lines are long.

This style is the same as in "Good 2".

```go
err := NewError( // no arguments there
  ErrRegexMissingParen,
  fmt.Sprintf("let's imaging that error mesasge is very long"),
) // no arguments there
```

# Bad 1

```go
for { someCode()
  someOtherCode()
}
```

No-one does that.

# Bad 2

```go
err := NewError(ErrRegexMissingParen,
  fmt.Sprintf("let's imaging that error mesasge is very long"),
)
```

That is as bad as "Bad 1".

# Bad 3

```go
err := NewError(
  ErrRegexMissingParen,
  fmt.Sprintf("a very long error message"))
```

The brackets should be symmetrical.

w84thesun:
Maybe we can add this to `Code style and conventions` section of CONTRIBUTING.md or to some REVIEW.md?

AlekSi:
Yes, that's exactly what this ticket is about

PR #1381, opened 2022-11-08
Add missing locks when update settings table

### LINKED ISSUE #1206: Fix the problem with settings table
### What should be done?

In pg, we store the list of collections in a single row in the settings table.
Because of that, there are some problems with accessing this data.

Example: when we create 10 collections concurrently, all the tables in pg are created correctly, but only a few of them are stored in the settings table by the end of collection creation. See `TestCreateStressPostgres` test 
 from https://github.com/FerretDB/FerretDB/pull/1316/files.

We must lock the table properly to ensure we don't lose data. Such work might require an approach like `select ... for update` that we currently don't use.

This approach might slow dow FerretDB, we have a dedicated issue #1347 to fix the problem of performance.

DoD:
- [x]  no TODOs / skipped tests mentioning the current issue number are in the code
- [x]  fix the comment about collection creation (Tigris vs Postgres)

rumyantseva:
# Description

Closes #1206.

Note: I didn't refactor how the settings table is used (as it's not in the scope of the task), I only fixed the exact problem with missing locks.

## Readiness checklist

### DIFF
--- a/integration/create_test.go
+++ b/integration/create_test.go
@@ -31,8 +31,6 @@ import (
 )
 
 func TestCreateStress(t *testing.T) {
-	t.Parallel()
-
 	ctx, collection := setup.Setup(t) // no providers there, we will create collections concurrently
 	db := collection.Database()
 
@@ -121,8 +119,6 @@ func TestCreateStress(t *testing.T) {
 }
 
 func TestCreateStressSameCollection(t *testing.T) {
-	t.Parallel()
-
 	ctx, collection := setup.Setup(t) // no providers there, we will create collection from the test
 	db := collection.Database()
 
--- a/integration/create_test.go
+++ b/integration/create_test.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"runtime"
 	"sync"
+	"sync/atomic"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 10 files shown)


internal/handlers/pg/msg_listdatabases.go
```diff
@@ -73,10 +76,21 @@ func (h *Handler) MsgListDatabases(ctx context.Context, msg *wire.OpMsg) (*wire.
...
+				if err == nil {
+					sizeOnDisk += tableSize
+					continue
+				}
+
+				var pgErr *pgconn.PgError
+				if errors.As(err, &pgErr) {
+					switch pgErr.Code {
+					case pgerrcode.UndefinedTable:
+						// Table was deleted after we got the list of tables, just ignore it
```
chilagrow, line 88: We can ignore this because all query or operation can handle the table not existing correct?

integration/create_test.go
```diff
@@ -72,15 +69,14 @@ func TestCreateStress(t *testing.T) {
 			}
 
 			// Attempt to create a collection for Tigris with a schema.
-			// If we get an error, that's MongoDB (FerretDB ignores that argument for non-Tigris handlers),
-			// so we create collection without schema.
+			// If we get an error that support for "validator" is not implemented, that's Postgres.
+			// If we get an error that "$tigrisSchemaString" is unknown, that's MongoDB.
```
chilagrow, line 73: I think this comment should come right after `// Attempt to create a collection for Tigris with a schema.`

integration/create_test.go
```diff
@@ -72,15 +69,14 @@ func TestCreateStress(t *testing.T) {
...
+			// In both cases, we create a collection without a schema.
 			err := db.CreateCollection(ctx, collName, &opts)
 			if err != nil {
-				var cmdErr *mongo.CommandError
-				if errors.As(err, &cmdErr) {
-					if strings.Contains(cmdErr.Message, `unknown top level operator: $tigrisSchemaString`) {
-						err = db.CreateCollection(ctx, collName)
-					}
+				if strings.Contains(err.Error(), `support for field "validator" is not implemented yet`) ||
```
AlekSi, line 77: I think we should create a helper method like `errorTextContains(error, string)` and use it everywhere in tests where needed instead of `strings.Contains`. That would help us understand where we don’t return proper errors yet.

integration/create_test.go
AlekSi: That comment is copy&pasted in several places, I think it should have exactly the same text to make it easier to spot that repeated problem

internal/handlers/pg/pgdb/settings.go
```diff
@@ -174,16 +173,44 @@ func getSettingsTable(ctx context.Context, tx pgx.Tx, db string) (*types.Documen
...
+// If the collection is already present in settings, ErrAlreadyExist will be returned.
+func setTableInSettings(ctx context.Context, tx pgx.Tx, db, collection, table string) error {
+	settings, err := getSettingsTable(ctx, tx, db, true)
+	if err != nil {
+		return lazyerrors.Error(err)
+	}
+
+	collections, ok := must.NotFail(settings.Get("collections")).(*types.Document)
+	if !ok {
+		return lazyerrors.Errorf("invalid settings document")
```
AlekSi, line 188: I think we could use unchecked type assertion there

internal/handlers/pg/pgdb/settings.go
```diff
@@ -174,16 +173,44 @@ func getSettingsTable(ctx context.Context, tx pgx.Tx, db string) (*types.Documen
...
+	}
+
+	if collections.Has(collection) {
+		return ErrAlreadyExist
```
AlekSi, line 192: That should be covered by tests

AlekSi (commented):
Looks good overall, but please check tests coverage

internal/handlers/pg/pgdb/settings.go
rumyantseva: I'm not sure it's possible to do it. 
I added a test that attempts to create a collection with the same name from a few parallel requests, but I'm not sure if it can reach this code as this case is very specific. 

Overall, the way how we work with settings needs to be simplified, and then we can reach better coverage. For example, I don't think we need that many requests of the settings table for every single operation, but this is how it's written, and it's not in the scope of the current task to refactor it (at least, this is not how the task is defined).

Do you have a particular example of a test that could be added?

chilagrow (approved):

integration/create_test.go
```diff
@@ -126,6 +120,106 @@ func TestCreateStress(t *testing.T) {
...
+			if err == nil {
+				created.Add(1)
+			} else {
+				AssertEqualError(t,
+					mongo.CommandError{
+						Code:    48,
+						Name:    "NamespaceExists",
+						Message: `Collection testcreatestresssamecollection.stress_same_collection already exists.`,
+					},
+					err)
```
AlekSi, line 185:
```suggestion
				AssertEqualError(
					t,
					mongo.CommandError{
						Code:    48,
						Name:    "NamespaceExists",
						Message: `Collection testcreatestresssamecollection.stress_same_collection already exists.`,
					},
					err,
				)
```

#593

internal/handlers/pg/pgdb/settings.go
AlekSi: Let's panic then. We will know if that code can be reached

chilagrow (approved):

noisersup (approved):

w84thesun (approved):

PR #1391, opened 2022-11-10
Allow duplicates in `bson` documents

### LINKED ISSUE #1263: Allow duplicates in `bson` documents
### What should be done?

`bson` package should support documents with duplicate keys, even other layers reject them. That should make testing and fuzzing easier.

### Where?

In `bson` package (just like in `types` package). Handlers are not in the scope.

---

- [x] Find the current issue's TODOs in the source code and modify them.
- [x] Check skipped tests.
- [ ] Diff tests are written for `insert`.

### Implementation comments

- For `insert`: if there are duplicate keys, an error is returned (this should be documented in `diff.md`)
- For `update` operators: if there are duplicate keys in the update document, an error is returned (see examples in `TestUpdateFieldCompat*` where `*` is an operator).

rumyantseva:
# Description

Closes #1263.

Dance diff tests: https://github.com/FerretDB/dance/pull/242

## Readiness checklist


rumyantseva:
@AlekSi I'm quite struggling with unit tests being failed. Please take a look, maybe you can notice something wrong quickly. Also, I scheduled a pair programming session for Tuesday.

internal/bson/document_test.go
```diff
@@ -171,7 +171,7 @@ var (
 	fuzz1 = testCase{
 		name: "fuzz1",
 		b:    testutil.MustParseDumpFile("testdata", "fuzz1.hex"),
-		bErr: `duplicate key ""`,
+		//	bErr: `duplicate key ""`,
```
rumyantseva, line 174: @AlekSi this is the place I need help with

internal/bson/document.go
```diff
@@ -34,35 +34,56 @@ const (
...
 type Document struct {
-	m    map[string]any
-	keys []string
+	fields []field
+}
+
+// field represents a field in the document.
+type field struct {
+	value any
+	key   string
```
w84thesun, line 49: It's not necessary, just seems right to put the key field first.
```suggestion
	key   string
	value any
```

w84thesun (dismissed):
I like how it's turned out! Added a small change suggestion.

internal/bson/document.go
rumyantseva: Linter complains about it :/ 
The `fieldalignment` thing

internal/types/document.go
```diff
@@ -167,6 +172,55 @@ func (d *Document) Keys() []string {
...
+		}
+
+		seen[field.key] = struct{}{}
+	}
+
+	return "", false
+}
+
+// RemoveDuplicateKeys removes duplicate keys from the document. The last value of each duplicate key is kept.
+func (d *Document) RemoveDuplicateKeys() {
```
AlekSi, line 207: We should not have that method at all.

Instead, we should validate documents in handler and return error, not silently remove duplicates.

`types.Document.ValidateData()` should already do that

AlekSi (changes requested):

internal/types/document.go
AlekSi: (yes, even if MongoDB works this way… Doing the same feels like a violation of common sense)

internal/types/document.go
```diff
@@ -184,9 +220,9 @@ func (d *Document) Command() string {
 func (d *Document) add(key string, value any) error {
```
chilagrow, line 220: It doesn't error in this func, any reason we want to keep error return value?

AlekSi: We already have a tracking issue for refactoring of bson and related packages

chilagrow (approved):

internal/wire/op_msg.go
```diff
@@ -259,7 +260,8 @@ func (msg *OpMsg) MarshalBinary() ([]byte, error) {
 				if err != nil {
 					return nil, lazyerrors.Error(err)
 				}
-				if d.WriteTo(secw); err != nil {
+
+				if err := d.WriteTo(secw); err != nil {
```
AlekSi, line 264: Nice!

internal/bson/document.go
```diff
@@ -247,10 +298,7 @@ func (doc *Document) ReadFrom(r *bufio.Reader) error {
 		}
 	}
 
-	if _, err := types.ConvertDocument(doc); err != nil {
-		return lazyerrors.Errorf("bson.Document.ReadFrom: %w", err)
-	}
-
+	*doc = Document{fields: fields}
```
AlekSi, line 301: That's fine for now, but note that previous code reused doc.m and doc.keys' capacities, and this new code replaces the whole document

AlekSi (approved):

w84thesun (approved):

PR #1397, opened 2022-11-14
Fix issues for the Unix listener

### LINKED ISSUE #1295: Fix issues for the Unix listener
### Versions

`main` on October 21

### What did you do?

Run FerretDB with a Unix socket listener. Run a few commands.

Those commands should work.

### What did you see instead?

They return internal errors or panic.

---

The fix itself should be easy: there are a few places where we do type assertion on `net.TCPAddr` that panic for Unix sockets.

But before fixing those places, we should make our integration tests use Unix sockets for in-process FerretDB. And that would require some setup refactoring. The simplest way to do it might be switching from passing port numbers to passing MongoDB URIs.

chilagrow:
# Description

Closes #1295. 

* `test-integration-pg` runs on unix socket, ` -target-unix-socket` is already specified in `Taskfile.yml`. 
* We don't support unix socket on windows, windows always uses TCP.

## Readiness checklist


internal/handlers/pg/msg_explain.go
```diff
@@ -84,7 +84,9 @@ func (h *Handler) MsgExplain(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 	var port int32
 	connInfo := conninfo.GetConnInfo(ctx)
 	if connInfo.PeerAddr != nil {
-		port = int32(connInfo.PeerAddr.(*net.TCPAddr).Port)
+		if tcpAddr, ok := connInfo.PeerAddr.(*net.TCPAddr); ok {
```
rumyantseva, line 87: In the case when `ok == false` the port is `0`, and then we consider the connection uses a socket, right?

internal/handlers/pg/msg_explain.go
```diff
@@ -84,7 +84,9 @@ func (h *Handler) MsgExplain(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+			port = int32(tcpAddr.Port)
```
rumyantseva, line 88: Do I understand correctly that if we want to test it, we need to run tests without `-target-unix-socket`? 
That looks a bit tricky as in `Taskfile.yml` we don't set  `-target-unix-socket` only for Tigris, and this case will not be covered (because it's the one for `pg`).

rumyantseva (commented):
Not a full review yet, asked a couple of questions to understand the code.

rumyantseva (changes requested):
Hmmm, it seems that something doesn't work for Mac.
I checked out the code locally and when I try to run tests on my env I see a lot of errors like

```
 task test-integration-tigris
task: [test-integration-tigris] go test -count=1 -race=true -tags=ferretdb_tigris -shuffle=on -coverpkg=../... -coverprofile=integration-tigris.txt . -handler=tigris

-test.shuffle 1668675367702040000
2022-11-17T09:56:07.704+0100    INFO    setup/common.go:233     Target system: in-process FerretDB with "tigris" handler.
2022-11-17T09:56:07.704+0100    INFO    setup/common.go:241     Compat system: port 37017.
2022-11-17T09:56:07.705+0100    INFO    debug   debug/debug.go:61       Starting debug server on http://127.0.0.1:58294/
--- FAIL: TestFindAndModifyCompatErrors (0.00s)
    --- FAIL: TestFindAndModifyCompatErrors/BadUpdateType (0.01s)
        common_unix.go:42: listen unix socket path too long len: 119, path: /var/folders/9p/cc9b8krs2zd1x9qx89fs1sjw0000gn/t/testfindandmodifycompaterrorsbadupdatetype2783407907/001/ferretdb.sock
```

internal/handlers/pg/msg_explain.go
chilagrow: Yes, without `-target-unix-socket` uses TCP. With current Taskfile.yml setup we won't be testing port related thing on CI.

internal/handlers/pg/msg_explain.go
chilagrow: Yeah. Also I didn't change returned document structure. The returned document will have "port":0 for unix socket connection.

chilagrow:
> Hmmm, it seems that something doesn't work for Mac. I checked out the code locally and when I try to run tests on my env I see a lot of errors like
> 
> ```
>  task test-integration-tigris
> task: [test-integration-tigris] go test -count=1 -race=true -tags=ferretdb_tigris -shuffle=on -coverpkg=../... -coverprofile=integration-tigris.txt . -handler=tigris
> 
> -test.shuffle 1668675367702040000
> 2022-11-17T09:56:07.704+0100    INFO    setup/common.go:233     Target system: in-process FerretDB with "tigris" handler.
> 2022-11-17T09:56:07.704+0100    INFO    setup/common.go:241     Compat system: port 37017.
> 2022-11-17T09:56:07.705+0100    INFO    debug   debug/debug.go:61       Starting debug server on http://127.0.0.1:58294/
> --- FAIL: TestFindAndModifyCompatErrors (0.00s)
>     --- FAIL: TestFindAndModifyCompatErrors/BadUpdateType (0.01s)
>         common_unix.go:42: listen unix socket path too long len: 119, path: /var/folders/9p/cc9b8krs2zd1x9qx89fs1sjw0000gn/t/testfindandmodifycompaterrorsbadupdatetype2783407907/001/ferretdb.sock
> ```

I attempted using `tmp/sockets/` dir but this didn't work for me. My contains upper case chars `FerretDB` in path `/home/chi/go/src/github.com/chilagrow/FerretDB/tmp/sockets` and unix socket path with upper case letter seems to have problem even after escaping it.

The alternative I chose was to use os.TempDir() and use shorter generated dir name and cleaning it up.

integration/setup/setup.go
```diff
@@ -49,6 +49,7 @@ type SetupOpts struct {
 type SetupResult struct {
 	Ctx           context.Context
 	Collection    *mongo.Collection
+	MongoURI      string
 	Port          uint16
```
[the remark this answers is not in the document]
chilagrow: Thanks good point!

integration/commands_administration_test.go
```diff
@@ -1046,9 +1046,11 @@ func TestCommandsAdministrationWhatsMyURI(t *testing.T) {
 	databaseName := s.Collection.Database().Name()
 	collectionName := s.Collection.Name()
 
+	// only check port number on TCP connection, no need to check on unix socket
+	isTCP := s.IsTPC(t)
```
w84thesun, line 1050: I believe there is a typo in `TCP`.

w84thesun (changes requested):
Checked it on Windows, looks like everything is working fine. A typo fix is needed.

integration/commands_administration_test.go
chilagrow: Yes :sweat_smile:  thanks for catching this :pray:

rumyantseva (approved):
LGTM.

Tests pass on my env.

rumyantseva (approved):
LGTM

w84thesun (approved):

PR #1400, opened 2022-11-15
Bump deps

AlekSi:
## Readiness checklist


tools/go.mod
```diff
@@ -4,14 +4,14 @@ go 1.19
 
 require (
 	github.com/BurntSushi/go-sumtype v0.0.0-20221020234012-480526a59796
-	github.com/go-task/task/v3 v3.17.0
-	github.com/golangci/golangci-lint v1.50.1 // sync with GitHub Actions workflow
```
chilagrow, line 8: do we want to keep this comment?

chilagrow (commented):
There is linter warning :thinking: interesting as the linter version wasn't updated

tools/go.mod
AlekSi: No, it was about https://github.com/golangci/golangci-lint-action, and it is very unlikely we would use it – it does not provide enough configurability

AlekSi:
That's because `golang.org/x/tools` was updated, that includes `vet` updates. PTAL

w84thesun (approved):

noisersup (approved):

PR #1401, opened 2022-11-15
Add documents validation to `wire` package

### LINKED ISSUE #1151: Validate command documents after reading them from the wire

w84thesun:
# Description

Closes #1151.

## Readiness checklist


internal/clientconn/conn.go
```diff
@@ -198,7 +198,43 @@ func (c *conn) run(ctx context.Context) (err error) {
...
+
+			resHeader = &wire.MsgHeader{
+				OpCode:        reqHeader.OpCode,
+				RequestID:     atomic.AddInt32(&c.lastRequestID, 1),
+				ResponseTo:    reqHeader.RequestID,
+				MessageLength: int32(wire.MsgHeaderLen + len(b)),
+			}
+
+			if err = wire.WriteMessage(bufw, resHeader, res); err != nil {
+				return
```
rumyantseva, line 228: So, we don't do anything with the error? We don't log it etc? Should we panic or something?

internal/clientconn/conn.go
```diff
@@ -198,7 +198,43 @@ func (c *conn) run(ctx context.Context) (err error) {
...
+			if err != nil {
+				panic(err)
+			}
+
+			resHeader = &wire.MsgHeader{
+				OpCode:        reqHeader.OpCode,
+				RequestID:     atomic.AddInt32(&c.lastRequestID, 1),
+				ResponseTo:    reqHeader.RequestID,
+				MessageLength: int32(wire.MsgHeaderLen + len(b)),
+			}
```
rumyantseva, line 225: Now we have two places where we make `resHeader`: here and in `route`. I'm not sure if it's a good solution, especially, having in mind that there are some `TODOs` around resHeader in `route`. Should we move both of them to a separate function that would make `resHeader`, or is it too much? (Maybe someone else has a better idea...)

internal/handlers/pg/msg_findandmodify.go
```diff
@@ -271,6 +271,8 @@ func (h *Handler) upsert(ctx context.Context, tx pgx.Tx, docs []*types.Document,
 			}
 		}
 
+		upsert.SetID()
```
rumyantseva, line 274: Hmm, I don't really understand the logic here. The code above has that `if !upsert.Has("_id")` check. Why do we need to set id again?

internal/handlers/pg/pjson/document.go
```diff
@@ -87,6 +87,8 @@ func (doc *documentType) UnmarshalJSON(data []byte) error {
 func (doc *documentType) MarshalJSON() ([]byte, error) {
 	td := types.Document(*doc)
 
+	td.SetID()
```
rumyantseva, line 90: Same here: why should we call `SetID` here?

internal/handlers/tigris/msg_findandmodify.go
```diff
@@ -249,6 +249,8 @@ func (h *Handler) upsert(ctx context.Context, docs []*types.Document, params *up
 			}
 		}
 
+		upsert.SetID()
```
rumyantseva, line 252: Same as upsert: the code above seems to be already setting id.

internal/wire/op_query.go
```diff
@@ -60,10 +61,18 @@ func (query *OpQuery) readFrom(bufr *bufio.Reader) error {
...
+
 	if err := q.ReadFrom(bufr); err != nil {
 		return err
 	}
-	query.Query = must.NotFail(types.ConvertDocument(&q))
+
+	doc := must.NotFail(types.ConvertDocument(&q))
+
+	if err := validateValue(doc); err != nil {
+		return NewValidationError(fmt.Errorf("wire.OpQuery.ReadFrom: validation failed for %v with: %v", doc, err))
```
rumyantseva, line 72: For clarity the fucntion's signature could be updated to highlight that a `ValidationError` could be returned.

internal/wire/validation.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+	case float64:
+		if math.IsNaN(v) {
+			return errors.New("NaN is not supported")
+		}
+
+		if v == 0 && math.Signbit(v) {
+			return errors.New("-0 is not supported")
+		}
+	}
```
rumyantseva, line 73: Let's add something like

```
default:
      panic("wire.validateValue: unknown type " + v)
```

to make sure we catch it if something goes wrong.

rumyantseva (commented):
I left some comments/questions.

A bit more extended comment for `SetID`: I like that with that we can get rid of the special `if` condition in the `Add` logic. I think `SetID` logic should be in a separate PR (if possible), so it'll be easier to review it. I partly understand the idea of `SetID`, but I'm not sure about the cases when id is already set - will the id key be sorted correctly, do we have tests that cover it?

internal/types/document.go
```diff
@@ -387,6 +363,34 @@ func (d *Document) isKeyDuplicate(targetKey string) bool {
 	return false
 }
 
+// SetID sets the _id field of the document at the first position.
+func (d *Document) SetID() {
+	if !d.Has("_id") {
```
chilagrow, line 368: I thought this function sets ID for any document, but I think what it does is it moves ID to the first index. And if a document doesn't have ID, it just does nothing. I suggest renaming it. ReorderIDIfExists or something that is more explicit

internal/wire/op_msg_test.go
```diff
@@ -147,6 +148,115 @@ var msgTestCases = []testCase{{
...
+		0x01, 0x76, 0x00, // double "v"
+		0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0xf8, 0x7f, // NaN
+		0x07, 0x5f, 0x69, 0x64, 0x00, // ObjectID "_id"
+		0x63, 0x77, 0xf2, 0x13, 0x75, 0x7c, 0x0b, 0xab, 0xde, 0xbc, 0x2f, 0x6a, // ObjectID value
+		0x00,                                                       // end of element 0 (document)
+		0x00,                                                       // end of document (array)
+		0x08, 0x6f, 0x72, 0x64, 0x65, 0x72, 0x65, 0x64, 0x00, 0x01, // "ordered" true
+		0x02, 0x24, 0x64, 0x62, 0x00, // "$db"
+		0x05, 0x00, 0x00, 0x00, // "test" length
+		0x74, 0x65, 0x73, 0x74, 0x00, // "test"
```
AlekSi, line 177: All that is not really needed since the only thing we need is the simplest document: `{"": NaN}`

w84thesun: The simplest document won't find the lack of recursion inside the validation function though.

internal/clientconn/conn.go
```diff
@@ -198,7 +198,43 @@ func (c *conn) run(ctx context.Context) (err error) {
 	for {
 		var reqHeader *wire.MsgHeader
 		var reqBody wire.MsgBody
+		var resHeader *wire.MsgHeader
+		var resBody wire.MsgBody
+		var validationErr *wire.ValidationError
+
 		reqHeader, reqBody, err = wire.ReadMessage(bufr)
+		if err != nil && errors.As(err, &validationErr) {
+			res := new(wire.OpMsg)
```
noisersup, line 207: I believe that we have a couple of lines in the code with this syntax:
```suggestion
			var res wire.OpMsg
```

internal/handlers/common/error.go
```diff
@@ -149,6 +150,11 @@
 		return writeErr, true
 	}
 
+	var validationErr *wire.ValidationError
+	if errors.As(err, &validationErr) {
+		return NewCommandError(ErrBadValue, err).(*CommandError), true //nolint:errorlint // false positive
```
noisersup, line 155: I know that there's a similar line in this function, but why does it exist? Isn't it possible to just return the value from `NewCommandError`?

noisersup (changes requested):
There's a lot of work done here! :) I left a couple of comments

internal/wire/validation.go
```diff
@@ -0,0 +1,76 @@
...
+
+	return d
+}
+
+// NewValidationError returns new ValidationError.
+func NewValidationError(err error) error {
+	return &ValidationError{msg: err.Error()}
+}
+
+// validateValue checks given value and return error if not supported value was encountered.
```
noisersup, line 50:
```suggestion
// validateValue checks given value and returns error if not supported value was encountered.
```

internal/handlers/common/error.go
w84thesun: `NewCommandError` return type is `error`, `ProtoErr` interface requires 3 missing methods.

internal/clientconn/conn.go
w84thesun: Fixed.

internal/clientconn/conn.go
```diff
@@ -328,8 +362,10 @@ func (c *conn) route(ctx context.Context, reqHeader *wire.MsgHeader, reqBody wir
 		document, err = msg.Document()
 
 		command = document.Command()
+
+		resHeader.OpCode = wire.OpCodeMsg
```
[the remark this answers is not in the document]
w84thesun: Yeah, I moved it here specifically to get `OpCode` value even we have error.

internal/wire/op_msg_test.go
```diff
@@ -147,6 +148,115 @@ var msgTestCases = []testCase{{
...
+		0x00, // end of document
```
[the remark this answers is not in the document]
w84thesun: Yeah, a bit.

internal/wire/validation.go
w84thesun: Added.

internal/types/document.go
w84thesun: Renamed it in #1544.

rumyantseva (dismissed):
Yesterday I left one comment where I wasn't sure about the current solution, but as nobody offered a better one, I'm ready to approve.

internal/clientconn/conn.go
w84thesun: Replaced return with panic.

internal/clientconn/conn.go
w84thesun: I tried to move it somewhere but the result wasn’t looking good. I suggest work on that later.

w84thesun:
> Yesterday I left one comment where I wasn't sure about the current solution, but as nobody offered a better one, I'm ready to approve.

Answered to that thread.

internal/bson/double_test.go
```diff
@@ -49,10 +45,6 @@ var doubleTestCases = []testCase{{
 	name: "-Infinity",
 	v:    pointer.To(doubleType(math.Inf(-1))),
 	b:    []byte{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xf0, 0xff},
-}, {
-	name: "NaN",
```
AlekSi, line 53: Issue states:

> bson package should still support them because we need to read the full documents first.

internal/types/fjson/double_test.go
```diff
@@ -49,10 +45,6 @@ var doubleTestCases = []testCase{{
 	name: "-Infinity",
 	v:    pointer.To(doubleType(math.Inf(-1))),
 	j:    `{"$f":"-Infinity"}`,
-}, {
-	name: "NaN",
```
AlekSi, line 53: Issues states:
> fjson package should still support them because we need it for logging.

chilagrow (commented):
Two highlights about not removing from fjson and bson package, it's good :+1:

noisersup (changes requested):
There are compilation errors (after automerge I guess)

noisersup (dismissed):
:rocket:

noisersup (approved):

AlekSi (approved):

PR #1402, opened 2022-11-15
Implement `debugError` command

### LINKED ISSUE #640: Implement `debugError` command
We have a stub for the special `debugError` command. We should implement it for testing FerretDB error handling.

Its command handler should accept a single string parameter that would change its behavior:
* for `ok`, it should return a successful response;
* for a numerical value like `2` (still as string), it should return an error with that code;
* for `panic` it should panic;
* for any other value, it should return `errors.New(that_value)`.

The integration test should be in the dance repo (diff test) and cover the above cases plus error codes 1 (internal error) and 0 (that seems to be possible for `getParameter`).

We don't document that difference from MongoDB.

noisersup:
# Description

Closes #640.

## Readiness checklist


internal/handlers/common/msg_debugerror.go
```diff
@@ -34,10 +37,29 @@ func MsgDebugError(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
 		return nil, err
 	}
 
-	// TODO https://github.com/FerretDB/FerretDB/issues/640
+	// check if parameter is an error code
+	if n, err := strconv.Atoi(expected); err != nil {
```
w84thesun, line 41: Shouldn't it be `err == nil`?

noisersup: It should... nice catch!

w84thesun (approved):

noisersup:
I have set donotmerge label because of the lack of dance tests

internal/handlers/common/msg_debugerror.go
```diff
@@ -34,10 +37,29 @@ func MsgDebugError(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
 		return nil, err
 	}
 
-	// TODO https://github.com/FerretDB/FerretDB/issues/640
+	// check if parameter is an error code
+	if n, err := strconv.Atoi(expected); err == nil {
+		errCode := ErrorCode(n)
```
rumyantseva, line 42: What if `n` is something that is not present in the list of `ErrorCode`s? 
For example, if `n == 123456`, how would this work?

rumyantseva (commented):
Asked a question.

internal/handlers/common/msg_debugerror.go
AlekSi: Good thing to cover with a dance test

noisersup: @rumyantseva The ErrorCode.String() method will return it as `"ErrorCode(123456)"`.

AlekSi (approved):

chilagrow (approved):

noisersup:
Dance tests are failing due to issues that are already on the main branch. When they are run individually, they pass. As agreed on the call - let's merge this PR and fix dance tests.

PR #1487, opened 2022-11-17
Simplify documents fetching for the `count` operator

### LINKED ISSUE #898: Offer a better design for fetching documents (iterator)

rumyantseva:
# Description

Closes #898.

This is one of PRs to complete #898. It replaces `h.PgPool.QueryDocuments` and a channel with `iterator` for `msg_count`.
Please take a look at this example (iterator interface, its place, iterator implementation to query documents, how it's used).

If it looks good, I'll continue replacing channel-based solution with the iterator in separate PRs.

My questions so far:

- [x] Would the iterator-based solution be slower? with channel we had a bit of concurrency to speed up the fetching process
- [x] In the initial task it's proposed to describe iterator as `Next() (E1, E2, error)`, `E1` for keys, `E2` for values, but I don't see a use case for it if we fetch documents from the database. 
- [x] I don't see places where we actually do need this interface passed as an interface (e.g. as a function parameter). So, why do we need it at all, and it's not enough to just implement it? 

## Readiness checklist


internal/util/iterator/iterator.go
```diff
@@ -0,0 +1,29 @@
...
+import "errors"
+
+// ErrIteratorDone  is returned when the iterator is read to the end.
+var ErrIteratorDone = errors.New("iterator is read to the end")
+
+// Interface is an iterator interface.
+type Interface[E1, E2 any] interface {
+	// Next returns a pair of values for containers, such as maps,
+	// for which elements have two values.
+	// If the iterator is at the end, it returns ErrEndOfIterator.
```
AlekSi, line 27: Can other errors be returned? Can they be wrapped?

AlekSi:
> Would the iterator-based solution be slower?

Nothing prevents the _implementation_ from prefetch rows/documents in the background. But I don't think we should focus on that now.

> but I don't see a use case for it if we fetch documents from the database

That interface is not only for fetching documents from the database. It could also be used instead of `for range d.Keys()` loop to iterate over document fields.

> So, why do we need it at all, and it's not enough to just implement it?

It would be easier for us when Go support that or similar interface natively in `for range` loops.

internal/handlers/pg/pgdb/query_iterator.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+
+		return nil, lazyerrors.Error(err)
+	}
+
+	rows, err := tx.Query(ctx, q, args...)
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+
```
AlekSi, line 52: We should set a finalizer to call `rows.Close`

internal/handlers/pg/pgdb/query_iterator.go
```diff
@@ -0,0 +1,80 @@
...
+	return &Iterator{
+		ctx:  ctx,
+		rows: rows,
+	}, nil
+}
+
+// Next implements iterator.Interface.
```
AlekSi, line 59: It doesn't

internal/handlers/pg/pgdb/query_iterator.go
rumyantseva: fixed

internal/handlers/pg/pgdb/query_iterator.go
rumyantseva: fixed

internal/util/iterator/iterator.go
rumyantseva: fixed

internal/util/iterator/iterator.go
```diff
@@ -0,0 +1,31 @@
...
+
+import "errors"
+
+// ErrIteratorDone  is returned when the iterator is read to the end.
+var ErrIteratorDone = errors.New("iterator is read to the end")
+
+// Interface is an iterator interface.
+type Interface[E1, E2 any] interface {
+	// Next returns a pair of values for containers, such as maps,
+	// for which elements have two values.
```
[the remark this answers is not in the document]
AlekSi: It is slice index, map key, etc. Just like in `for e1, e2 := range i`

chilagrow (approved):

w84thesun (dismissed):
I liked the implementation!

internal/handlers/pg/pgdb/query_iterator.go
```diff
@@ -0,0 +1,90 @@
...
+	currentIter atomic.Uint32
+}
+
+// NewIterator returns a new iterator for the given SQL param.
+// If the table for the given SQL params doesn't exist, it returns nil and no error.
+// If an error occurs, it returns nil and "lazy error".
+func NewIterator(ctx context.Context, tx pgx.Tx, sp *SQLParam) (*Iterator, error) {
+	q, args, err := buildQuery(ctx, tx, sp)
+	if err != nil {
+		if errors.Is(err, ErrTableNotExist) {
```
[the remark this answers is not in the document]
rumyantseva: @AlekSi could you propose the code change, so I understand what you mean? I'm not sure I understand your idea.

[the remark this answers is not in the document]
AlekSi: Let's split the proposed `NewIterator` into `NewIterator(ctx context.Context, rows pgx.Rows)` that would only construct an iterator, and `QueryDocuments` that would apply some additional logic and then would use that function

internal/util/iterator/iterator.go
```diff
@@ -0,0 +1,31 @@
...
+	// If the iterator is at the end, it returns ErrEndOfIterator as error.
```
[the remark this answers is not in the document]
rumyantseva: I expect there might be cases when that could be wrapped, updated the docs.

internal/handlers/pg/msg_count.go
```diff
@@ -85,37 +88,46 @@ func (h *Handler) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
...
-				if err != nil {
-					return err
-				}
+		for {
+			_, doc, ierr := it.Next()
+
+			switch {
+			case ierr == nil:
+				// do nothing
+			case errors.Is(ierr, iterator.ErrIteratorDone):
```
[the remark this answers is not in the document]
rumyantseva: Docs fixed

internal/util/iterator/iterator.go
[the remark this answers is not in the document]
rumyantseva: @AlekSi please propose the exact change, I resued the exact signature and docs from Go's proposal.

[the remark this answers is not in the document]
AlekSi: Code comment in the Go's proposal doesn't say anything about maps. I think the result is confusing.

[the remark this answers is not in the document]
AlekSi: (because a pair of values is tuple 😄)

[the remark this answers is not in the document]
rumyantseva: true... true... :)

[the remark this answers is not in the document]
rumyantseva: @AlekSi if you have better names for `E1`, `E2` please suggest them. I don't like names like `Key`, `Value` because we don't always talk about keys.

internal/util/iterator/iterator.go
```diff
@@ -0,0 +1,32 @@
...
+var ErrIteratorDone = errors.New("iterator is read to the end")
+
+// Interface is an iterator interface.
+type Interface[E1, E2 any] interface {
+	// Next returns an ordered pair.
+	// The meaning of that ordered pair depends on the type of the iterator.
+	// For example, for maps E1 is key and E2 is value, for slices E1 is index and E2 is value.
+	// If the iterator is at the end, it returns possibly wrapped ErrEndOfIterator as error.
+	// Other errors may be returned as well, they depend on the implementation and could be wrapped.
+	// Such errors should be documented in the implementation.
```
AlekSi, line 30:
```suggestion
	// Next returns the next key/value pair, where key is slice index, map key, document number, etc, and the value is the slice or map value, next document, etc.
	// Returned error could be (possibly wrapped) ErrEndOfIterator or some fatal error.
```

internal/handlers/pg/pgdb/query.go
```diff
@@ -106,6 +109,35 @@ func (pgPool *Pool) QueryDocuments(ctx context.Context, tx pgx.Tx, sp *SQLParam)
...
+	}
+
+	rows, err := tx.Query(ctx, q, args...)
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+
+	runtime.SetFinalizer(rows, func(rows pgx.Rows) {
+		rows.Close()
+		pgPool.Config().ConnConfig.Logger.Log(
```
rumyantseva, line 132: @AlekSi finalizer check (to make it visible in tests and in logs in general)

internal/handlers/pg/pgdb/query.go
```diff
@@ -64,6 +65,8 @@ type SQLParam struct {
 // Context cancellation is not considered an error.
 //
 // If the collection doesn't exist, fetch returns a closed channel and no error.
+//
+// Deprecated in favor of GetDocuments, TODO remove in https://github.com/FerretDB/FerretDB/issues/898.
```
AlekSi, line 69: It should be `Deprecated:`, with a colon

https://github.com/golang/go/wiki/Deprecated

internal/handlers/pg/pgdb/query_iterator.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	doc, err := pjson.Unmarshal(b)
+	if err != nil {
+		return 0, nil, lazyerrors.Error(err)
+	}
+
+	defer it.currentIter.Add(1)
+
+	return it.currentIter.Load(), doc.(*types.Document), nil
```
AlekSi, line 71:
```suggestion
	n := it.currentIter.Add(1)

	return n-1, doc.(*types.Document), nil
```

No real reason to use `defer`

internal/handlers/pg/pgdb/query_test.go
```diff
@@ -164,3 +167,112 @@ func TestQueryDocuments(t *testing.T) {
...
+		name:       "three",
+		collection: collectionName + "_three",
+		documents: []*types.Document{
+			must.NotFail(types.NewDocument("_id", "foo", "id", "1")),
+			must.NotFail(types.NewDocument("_id", "foo", "id", "2")),
+			must.NotFail(types.NewDocument("_id", "foo", "id", "3")),
+		},
+	}}
+
+	for _, tc := range cases {
```
AlekSi, line 215: I'm not sure it is worth it to make this test table-driven: there is no difference between cases "two" and "three", not much difference with case "one", the test has a lot of branches, and we don't even test the case when we break from the iteration with not all rows consumed.

I think, each test should be a separate special (not really) case like "non-existing-collection" below

internal/handlers/pg/pgdb/query_test.go
```diff
@@ -164,3 +167,112 @@ func TestQueryDocuments(t *testing.T) {
...
+				iterPrev = iter
+			}
+
+			require.NoError(t, tx.Commit(ctx))
+		})
+	}
+
+	// Special case: querying a non-existing collection.
+	t.Run("non-existing-collection", func(t *testing.T) {
+		tx, err := pool.Begin(ctx)
```
AlekSi, line 267: t.Parallel()

internal/handlers/pg/msg_count.go
```diff
@@ -85,37 +88,46 @@ func (h *Handler) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
 
 	resDocs := make([]*types.Document, 0, 16)
 	err = h.PgPool.InTransaction(ctx, func(tx pgx.Tx) error {
-		fetchedChan, err := h.PgPool.QueryDocuments(ctx, tx, &sp)
-		if err != nil {
-			return err
+		it, ierr := h.PgPool.GetDocuments(ctx, tx, &sp)
```
AlekSi, line 91:
```suggestion
		it, err := h.PgPool.GetDocuments(ctx, tx, &sp)
```

Otherwise, we will have to very careful with both `err` and `ierr` around

internal/handlers/pg/msg_count.go
```diff
@@ -85,37 +88,46 @@ func (h *Handler) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
...
+		for {
+			_, doc, ierr := it.Next()
+
+			switch {
+			case ierr == nil:
+				// do nothing
+			case errors.Is(ierr, iterator.ErrIteratorDone):
+				// no more documents
+				return nil
+			case errors.Is(ierr, context.Canceled), errors.Is(ierr, context.DeadlineExceeded):
```
AlekSi, line 110: I think checking `ctx.Err()` above would be easier

integration/setup/setup_compat.go
```diff
@@ -43,6 +43,10 @@ type SetupCompatOpts struct {
 	// Data providers.
 	Providers []shareddata.Provider
 
+	// If true, a non-existed collection will be added to the list of collections.
+	// This is useful to test the behavior when a collection is not found.
```
AlekSi, line 47: When it is _not_ useful?

If it is always useful, why make it configurable?

integration/count_compat_test.go
```diff
@@ -0,0 +1,124 @@
...
+				t.Run(targetCollection.Name(), func(t *testing.T) {
+					t.Helper()
+
+					var targetRes, compatRes bson.D
+					targetErr := targetCollection.Database().RunCommand(ctx, bson.D{
+						{"count", targetCollection.Name()},
+						{"query", filter},
+					}).Decode(&targetRes)
+					compatErr := compatCollection.Database().RunCommand(ctx, bson.D{
+						{"count", compatCollection.Name()},
```
AlekSi, line 70: Why not CountDocuments?

rumyantseva: When I did something like that somewhere else, you asked me, "Why not `RunCommand`?" :) 

My answer is that `RunCommand` is more universal, and it's easy to extend it if needed. But I don't mind using `CountDocuments` if there is a reason why it's better.

internal/handlers/pg/pgdb/query.go
rumyantseva: @AlekSi but iterator is not responsible for `rows` anymore (as you told me to do). Imagine, we had a channel instead of `rows`, and we close this channel in the finalizer. I can add a comment to the `newIterator` that whoever passes rows must close them.

integration/setup/setup_compat.go
rumyantseva: I'm glad you like it! Unfortunately, I can't enable it for all the tests right now, as some tests will fail (and the scope of this task will be increased). So, I created a separate issue: https://github.com/FerretDB/FerretDB/issues/1545

internal/handlers/pg/msg_count.go
rumyantseva: @AlekSi that's true, but govet complains about shadowing if I use `err`. Should we adjust something for `govet`?

internal/handlers/pg/msg_count.go
rumyantseva: @AlekSi hmmm... where above? I probably don't understand something. The idea is that we pass `ctx` to new iterator, and at the beginning of each `Next` we check if the context is not closed. It has some point, but it's not super critical, at least with the current implementation. So, should I remove it or something?

internal/handlers/pg/pgdb/query_iterator.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Iterator implements iterator.Interface to fetch documents from the database.
+type Iterator struct {
+	ctx         context.Context
+	rows        pgx.Rows
+	currentIter atomic.Uint32
+}
+
+// NewIterator returns a new iterator for the given pgx.Rows.
+func NewIterator(ctx context.Context, rows pgx.Rows) *Iterator {
```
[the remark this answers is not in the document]
rumyantseva: Hmm... I pressed "Commit suggestion", and nothing happened 🤔

internal/handlers/pg/pgdb/query_test.go
rumyantseva: fixed

internal/handlers/pg/msg_count.go
w84thesun: The other option would be something like this, I think:
```suggestion
		var it iterator.Interface[uint32, *types.Document]
		it, err = h.PgPool.GetDocuments(ctx, tx, &sp)
```

internal/util/iterator/iterator.go
```diff
@@ -0,0 +1,29 @@
...
+import "errors"
+
+// ErrIteratorDone  is returned when the iterator is read to the end.
+var ErrIteratorDone = errors.New("iterator is read to the end")
+
+// Interface is an iterator interface.
+type Interface[K, V any] interface {
+	// Next returns the next key/value pair, where the key is a slice index, map key, document number, etc,
+	// and the value is the slice or map value, next document, etc.
+	// Returned error could be (possibly wrapped) ErrEndOfIterator or some fatal error.
```
w84thesun, line 27:
```suggestion
	// Returned error could be (possibly wrapped) ErrIteratorDone or some fatal error.
```

w84thesun (commented):
Added small comment change suggestion.

integration/count_compat_test.go
AlekSi: By the same logic we should use RunCommand with "find" instead of Find and FindOne methods. This way, we may ensure that out implementation matches our tests… and maybe does not actually driver's Find and FindOne implementation. But we care about real applications first, and they use drivers, so we should test with drivers where possible, not with raw commands.

AlekSi: > When I did something like that somewhere else, you asked me, "Why not RunCommand?" :)

IIRC, that was a question for command without driver's method

internal/handlers/pg/msg_count.go
AlekSi: We should replace `case errors.Is(ierr, context.Canceled), errors.Is(ierr, context.DeadlineExceeded)` with `if ctx.Err() != nil` before switch. For example:

https://github.com/FerretDB/FerretDB/blob/8bc6b5dd2264d4c45ba8761ad98b1faa325f7bee/internal/clientconn/listener.go#L144-L156

We don't even inspect `err` if context was canceled

internal/util/iterator/iterator.go
rumyantseva: Nice catch!

integration/count_compat_test.go
rumyantseva: @AlekSi func fact: `CountDocuments` wraps aggregation and calls aggregation, not the `count` command. So we can't use it in this test for the `count` command.

internal/handlers/pg/msg_count.go
rumyantseva: fixed

internal/util/iterator/iterator.go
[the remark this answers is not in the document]
rumyantseva: fixed

internal/handlers/pg/pgdb/query_iterator.go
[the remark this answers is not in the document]
rumyantseva: fixed

internal/handlers/pg/pgdb/query.go
rumyantseva: fixed (tested)

internal/handlers/pg/pgdb/query_iterator.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/jackc/pgx/v4"
+
+	"github.com/FerretDB/FerretDB/internal/handlers/pg/pjson"
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/iterator"
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+)
+
+// Iterator implements iterator.Interface to fetch documents from the database.
+type Iterator struct {
```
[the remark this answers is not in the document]
rumyantseva: fixed

internal/handlers/pg/pgdb/query.go
rumyantseva: fixed

internal/handlers/pg/pgdb/query_test.go
rumyantseva: fixed

w84thesun (dismissed):

integration/count_compat_test.go
AlekSi: Great, please add a comment about that in the next PR

internal/handlers/pg/pgdb/query_test.go
```diff
@@ -168,3 +172,129 @@ func TestQueryDocuments(t *testing.T) {
...
+		cancel()
+		iter, doc, err = it.Next()
+		assert.EqualError(t, context.Canceled, err.Error())
+		assert.Equal(t, uint32(0), iter)
+		assert.Nil(t, doc)
+
+		// To be able to commit tx, we need iterator's finalizer to be called (to close pgx.Rows).
+		// Otherwise, tx.Commit will fail with "conn busy".
+		runtime.GC()
+		time.Sleep(10 * time.Millisecond)
```
AlekSi, line 257: I _think_ that might be `runtime.Gosched()` which is a bit more clear. Please try in the next PR

internal/handlers/pg/msg_count.go
```diff
@@ -85,37 +87,48 @@ func (h *Handler) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
...
-			// Drain the channel to prevent leaking goroutines.
-			// TODO Offer a better design instead of channels: https://github.com/FerretDB/FerretDB/issues/898.
-			for range fetchedChan {
-			}
-		}()
 
-		for fetchedItem := range fetchedChan {
-			if fetchedItem.Err != nil {
-				return fetchedItem.Err
+		if it == nil {
```
AlekSi, line 96: We can simplify that (in the next PR) by returning empty iterator in that case

AlekSi (approved):

chilagrow (dismissed):

integration/count_compat_test.go
rumyantseva: added todo

internal/handlers/pg/pgdb/query_test.go
```diff
@@ -168,3 +172,129 @@ func TestQueryDocuments(t *testing.T) {
...
+		require.NotNil(t, it)
+
+		iter, doc, err := it.Next()
+		assert.NoError(t, err)
+		assert.Equal(t, uint32(0), iter)
+		assert.Equal(t, expectedDocs[0], doc)
+
+		cancel()
+		iter, doc, err = it.Next()
+		assert.EqualError(t, context.Canceled, err.Error())
```
AlekSi, line 250:
```suggestion
		assert.Equal(t, context.Canceled, err)
```

No need to compare strings

AlekSi (commented):
Given that `pgx.Conn` is not thread-safe, I don't think we have other options other than introduce required (and thread-safe) Close method

noisersup (approved):

internal/handlers/pg/pgdb/query_iterator.go
```diff
@@ -0,0 +1,104 @@
...
+		closed:   false,
+		mxClosed: new(sync.Mutex),
+	}
+
+	runtime.SetFinalizer(qi, func(qi *queryIterator) {
+		qi.mxClosed.Lock()
+		defer qi.mxClosed.Unlock()
+		if !qi.closed {
+			// TODO: ask about fatalpanic
+			panic("queryIterator.Close() has not been called")
```
rumyantseva, line 56: Important: this panic is `fatalpanic` (unrecoverable) as it's inside the finalizer, and there is no way to catch the function that would execute it. So, if it happens, FerretDB stops, `recover` will be ignored.

AlekSi: > Are we ok with keeping this panic?

Yes

integration/count_compat_test.go
rumyantseva: fixed

w84thesun (approved):

noisersup (approved):
I like the solution! :slightly_smiling_face:

Issue #1341, opened 2022-10-28
Improve concurrent Tigris database creation handling


rumyantseva:
### What should be done?

When we call DB creation concurrently, if an attempt to create a database with the given name is already in progress, Tigris returns an unknown error with "duplicate key value, violates key constraint" message.

Tigris is going to fix it, so when they do it, we need to fix our part too. See [isOtherCreationInFlight](https://github.com/FerretDB/FerretDB/pull/1316/files) function and how it's used.

We need to clarify how Tigris solved the problem and make the changes accordingly. E.g. do we still need retries on our side or are they on Tigris side? 

### Where?

https://github.com/FerretDB/FerretDB/pull/1316/files#diff-29cf2f7e97f581987a3d3ee5a8a0d7d0b3b22987fc50e73778917e662e6d04d4

PR #1513, opened 2022-11-21
Fix parallel collection inserts for PostgreSQL

### LINKED ISSUE #1488: Concurrent creation of database for parallel inserts doesn't work when database doesn't exist
### Versions

Nov 17, `main`

### What did you do?

Run dance tests and got an internal error

```
document_validation_test.go:112: 
        	Error Trace:	/home/user/dance/tests/diff/document_validation_test.go:112
        	            	(InternalError) [pool.go:213 pgdb.(*Pool).InTransaction] [error.go:200 common.CheckError] [insert.go:44 pgdb.InsertDocument] database/schema or collection/table already exist
        	Test:       	TestDocumentValidation/Update/DollarSign
        --- FAIL: TestDocumentValidation/Update/DollarSign (0.03s)
```

No errors

### What did you see instead?

DB already exists error.

### Reasoning

This is what `CreateDatabaseIfNotExists` promises (` If the schema already exists, no error is returned`):

```go
// CreateDatabaseIfNotExists creates a new FerretDB database (PostgreSQL schema).
// If the schema already exists, no error is returned, and transaction is not aborted.
func CreateDatabaseIfNotExists(ctx context.Context, tx pgx.Tx, db string) error {
```

This is what this function could return in fact:

```go
	case pgerrcode.DuplicateSchema:
		return ErrAlreadyExist
	case pgerrcode.UniqueViolation, pgerrcode.DuplicateObject:
		// https://www.postgresql.org/message-id/CA+TgmoZAdYVtwBfp1FL2sMZbiHCWT4UPrzRLNnX1Nb30Ku3-gg@mail.gmail.com
		// The same thing for schemas. Reproducible by integration tests.
		return ErrAlreadyExist
```

In addition to that, the statement that  `transaction is not aborted` is false. If `PgError` is caught, the transaction *will be* aborted (`current transaction is aborted, commands ignored until end of transaction block`).

### TODO

- [x] If possible, add a stress test to reproduce the problem on FerrretDB itself (e.g. parallel insert into the same database that doesn't exist)
- [x]  Fix the problem (e.g. fix godoc, check for `ErrAlreadyExists` in callers, use multiple transactions when needed - like [here](https://github.com/FerretDB/FerretDB/blob/af2554ad143a40afb110b4277995507af7873f73/internal/handlers/pg/msg_create.go#L75))
- [ ] Check dance for the current issue being mentioned and fix relevant TODOs

noisersup:
# Description

Closes #1488.

## Readiness checklist


rumyantseva (changes requested):
Tigris tests fail because Tigris needs some special creation-related handling. It might be relevant to this issue: #1341. Please check why Tigris fails, and if it's the same as #1341 and not easy to fix, add a todo with the issue number.

rumyantseva:
@noisersup yeah, they started doing it, and we need to implement #1341 to support it correctly. So, for the current issue, I suggest to skip/todo the places about Tigris and concurrent collection creation and mentioning #1341.

w84thesun (approved):
LGTM

internal/handlers/pg/pgdb/collections.go
```diff
@@ -144,8 +144,13 @@ func CreateCollection(ctx context.Context, tx pgx.Tx, db, collection string) err
 // If needed, it creates both database and collection.
 //
 // True is returned if collection was created.
-func CreateCollectionIfNotExist(ctx context.Context, tx pgx.Tx, db, collection string) (bool, error) {
-	exists, err := CollectionExists(ctx, tx, db, collection)
+func CreateCollectionIfNotExist(ctx context.Context, pgPool *Pool, db, collection string) (bool, error) {
```
chilagrow, line 147: This probably is outside the scope of this task, just noting that even if collection did exist when this func was called, nothing is preventing from collection to be deleted after calling this func.

chilagrow (changes requested):

noisersup:
As agreed on the call - error handling (rollbacking transaction) of the insert method will be improved on #940.

internal/handlers/pg/msg_insert.go
```diff
@@ -63,22 +61,22 @@ func (h *Handler) MsgInsert(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+		var doc any
+
+		doc, err = docs.Get(i)
+		if err != nil {
+			return nil, lazyerrors.Error(err)
 		}
-		return nil
-	})
+
+		err = h.insert(ctx, &sp, doc)
```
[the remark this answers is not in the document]
noisersup: Nice catch!
https://github.com/FerretDB/FerretDB/pull/1513#issuecomment-1326212210

internal/handlers/pg/pgdb/collections.go
noisersup: Yeah, things would be much easier if we could have collection creation and inserting in the same transaction

rumyantseva: As discussed in the call, an alternative approach could be using the same transaction and retrying it. The problem with it is with ordered inserts there are some errors that could be retried (from db and collection creation) and errors that couldn't (from inserts themselves). With such approach, we need to find a way to handle those retries correctly...

@noisersup let's wait for more review, but personally I'd vote for merging the current solutions with multiple transactions.

rumyantseva (approved):
LGTM

chilagrow (approved):

AlekSi:
Please don't forget to enable auto-merge

PR #1526, opened 2022-11-22
Move array query tests to compat tests

### LINKED ISSUE #1450: Move more query tests to compat tests
### What should be done?

### Where?

chilagrow:
# Description

Closes #1450.

This is one of many PR to move query integration tests to compat tests. This PR moves array query tests to compat test.
The tests that were not moved were
* Requires additional test Bson in shareddata, and currently adding that is too painful because too many tests are failing. This will be done once more tests were moved to compat, that will make it easier.
* Nan and Negative Zero because they will be removed by https://github.com/FerretDB/FerretDB/pull/1401
* Others that were not moved have TODO with issue link attached.
## Readiness checklist

### DIFF
--- a/integration/basic_test.go
+++ b/integration/basic_test.go
@@ -61,6 +61,8 @@ func TestFindNothing(t *testing.T) {
 	ctx, collection := setup.Setup(t)
 
 	var doc bson.D
+
+	// FindOne sets limit parameter to 1, Find leaves it unset.
 	err := collection.FindOne(ctx, bson.D{}).Decode(&doc)
 	require.Equal(t, mongo.ErrNoDocuments, err)
 	assert.Equal(t, bson.D(nil), doc)
--- a/integration/basic_test.go
+++ b/integration/basic_test.go
@@ -60,16 +60,8 @@ func TestFindNothing(t *testing.T) {
 	t.Parallel()
 	ctx, collection := setup.Setup(t)
 
-	cursor, err := collection.Find(ctx, bson.D{})
-	require.NoError(t, err)
-
-	var docs []bson.D
-	err = cursor.All(ctx, &docs)
-	require.NoError(t, err)
-	assert.Equal(t, []bson.D(nil), docs)
-
 	var doc bson.D
-	err = collection.FindOne(ctx, bson.D{}).Decode(&doc)
+	err := collection.FindOne(ctx, bson.D{}).Decode(&doc)
 	require.Equal(t, mongo.ErrNoDocuments, err)
 	assert.Equal(t, bson.D(nil), doc)
 }
--- a/integration/query_array_compat_test.go
+++ b/integration/query_array_compat_test.go
@@ -22,122 +22,104 @@ import (
 	"go.mongodb.org/mongo-driver/bson/primitive"
 )
 
-func TestQueryArrayCompat(t *testing.T) {
+func TestQueryArrayCompatSize(t *testing.T) {
 	t.Parallel()
 
 	testCases := map[string]queryCompatTestCase{
-		"AllString": {
-			filter:        bson.D{{"v", bson.D{{"$all", bson.A{"foo"}}}}},
+		"float64": {
+			filter:        bson.D{{"v", bson.D{{"$size", float64(2)}}}},
 			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
 		},
-		"AllStringRepeated": {
-			filter:        bson.D{{"v", bson.D{{"$all", bson.A{"foo", "foo", "foo"}}}}},
+		"int32": {
+			filter:        bson.D{{"v", bson.D{{"$size", int32(2)}}}},
 			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
 		},
-		"AllStringEmpty": {
-			filter:        bson.D{{"v", bson.D{{"$all", bson.A{""}}}}},
+		"int64": {
+			filter:        bson.D{{"v", bson.D{{"$size", int64(2)}}}},
 			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
 		},
-		"AllWhole": {
-			filter:        bson.D{{"v", bson.D{{"$all", bson.A{int32(42)}}}}},
-			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
-		},
-		"AllWholeNotFound": {
-			filter:     bson.D{{"v", bson.D{{"$all", bson.A{int32(44)}}}}},
+		"InvalidUse": {
+			filter:     bson.D{{"$size", 2}},
 			resultType: emptyResult,
 		},
-		"AllZero": {
-			filter:        bson.D{{"v", bson.D{{"$all", bson.A{math.Copysign(0, +1)}}}}},
-			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
-		},
-		"AllDouble": {
-			filter:        bson.D{{"v", bson.D{{"$all", bson.A{42.13}}}}},
-			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
-		},
-		"AllDoubleMax": {
-			filter:        bson.D{{"v", bson.D{{"$all", bson.A{math.MaxFloat64}}}}},
-			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
-		},
-		"AllDoubleMin": {
-			filter:        bson.D{{"v", bson.D{{"$all", bson.A{math.SmallestNonzeroFloat64}}}}},
-			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
-		},
-		"AllMultiAll": {
-			filter:        bson.D{{"v", bson.D{{"$all", bson.A{"foo", 42}}}}},
-			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
-		},
-		"AllMultiAllWithNil": {
-			filter:        bson.D{{"v", bson.D{{"$all", bson.A{"foo", nil}}}}},
-			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
-		},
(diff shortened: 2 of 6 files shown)


integration/query_compat_test.go
```diff
@@ -113,6 +119,216 @@ func TestQueryCompat(t *testing.T) {
 	t.Parallel()
 
 	testCases := map[string]queryCompatTestCase{
+		"ArrayOne": {
```
AlekSi, line 122: I think we should move all test into the same file and function – it would make it harder to run only a subset of tests. Consider, for example, our existing `query_logical_compat_test.go` file. So, for example, all queries with arrays could be moved to `query_array_compat_test.go` file and `TestQueryArrayCompat` function

integration/query_array_test.go
```diff
@@ -50,31 +50,12 @@ func TestQueryArraySize(t *testing.T) {
...
 		"NegativeZero": {
 			filter:      bson.D{{"v", bson.D{{"$size", math.Copysign(0, -1)}}}},
 			expectedIDs: []any{"array-empty"},
 		},
-		"NotFound": {
-			filter:      bson.D{{"v", bson.D{{"$size", 4}}}},
-			expectedIDs: []any{},
-		},
 		"InvalidType": {
+			// TODO: move to compat https://github.com/FerretDB/FerretDB/issues/1539
```
AlekSi, line 58: Please mark this PR as ready for review soon, and then investigate that issue, as we clearly have code for that: https://github.com/FerretDB/FerretDB/blob/8029b86ae0a2a2b2f4a22adf8e9e79d960c131a2/internal/handlers/common/filter.go#L712-L721

integration/query_compat_test.go
chilagrow: Just noticed that myself as well. I created function for each operator like `TestQueryArrayCompatAll` `TestQueryArrayCompatSize`

integration/query_array_test.go
chilagrow: Yeah let me do that soon. I will only address array query compat test in this PR.

chilagrow:
```
            update_field_test.go:273: 
                	Error Trace:	/home/runner/work/FerretDB/FerretDB/integration/path.go:103
                	            				/home/runner/work/FerretDB/FerretDB/integration/update_field_test.go:273
                	Error:      	Max difference between 2022-11-23 07:03:05.894 +0000 UTC and 2022-11-23 07:06:18.972 +0000 UTC allowed is 3m0s, but difference was -3m13.078s
                	Test:       	TestUpdateFieldCurrentDate/currentDate/Date
```
Just saw CI failed because max difference was bigger 3m. Should I be concerned? Maybe it's transient, I've rerun the CI.

rumyantseva:
@chilagrow regarding 3min max difference for time, I think something was just too slow, so the execution of this test took longer than usual. Ideally, that would be great to find something better than this fixed 3 min difference, but at the moment when this test was written, we didn't have better ideas. I'd say it might make sense to take a look at this test for learning purposes, but most likely, there is not much we can do about it.

integration/query_compat_test.go
chilagrow: Having separate tests like `TestQueryArrayCompatAll` and `TestQueryArrayCompatSize` might be the reason why it slows down. I'll put them all in one func `TestQueryArrayCompat` as suggested.

integration/basic_test.go
```diff
@@ -60,16 +60,8 @@ func TestFindNothing(t *testing.T) {
 	t.Parallel()
 	ctx, collection := setup.Setup(t)
 
-	cursor, err := collection.Find(ctx, bson.D{})
```
rumyantseva, line 63: This one is the same as query compat test

```go
"Empty": {
  filter: bson.D{},
},
```

and this is why it's removed, right?

integration/query_array_compat_test.go
```diff
@@ -0,0 +1,224 @@
...
+			resultType: emptyResult,
+		},
+		"DotNotationPositionIndexAtTheEndOfArray": {
+			filter:        bson.D{{"v.1", bson.D{{"$type", "double"}}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
+		},
+		"DotNotationPositionTypeNull": {
+			filter:        bson.D{{"v.1", bson.D{{"$type", "double"}}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
+		},
```
rumyantseva, line 100: What is the difference between test cases `DotNotationPositionIndexAtTheEndOfArray` and `DotNotationPositionIndexAtTheEndOfArray`?

integration/query_array_test.go
```diff
@@ -155,50 +130,8 @@ func TestQueryArrayDotNotation(t *testing.T) {
...
-			filter:      bson.D{{"v.5", bson.D{{"$type", "double"}}}},
-			expectedIDs: []any{},
-		},
-		"PositionIndexAtTheEndOfArray": {
-			filter:      bson.D{{"v.1", bson.D{{"$type", "double"}}}},
-			expectedIDs: []any{"array-two"},
-		},
-
-		"PositionTypeNull": {
-			filter:      bson.D{{"v.0", bson.D{{"$type", "null"}}}},
```
rumyantseva, line 168: This one is removed but not added. I guess the case `DotNotationPositionTypeNull` should contain it.

rumyantseva (changes requested):
Partly reviewed and found one typo / wrong copy-paste. Continuing with reviewing :)

integration/query_array_compat_test.go
```diff
@@ -0,0 +1,224 @@
...
+		"EqualityEmpty": {
+			filter:        bson.D{{"v", bson.A{}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
+		},
+		"EqualityNull": {
+			filter:        bson.D{{"v", bson.A{nil}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
+		},
+		"SizeFloat64": {
```
rumyantseva, line 197: Structuring idea: for some compat cases, we define multiple functions in the same file for different operations/operators. I find it helpful from the readability perspective. I wonder if we could do the same here. For example, everything `$size`-related could be in `TestQueryArrayCompatSize` or everything `$all`-related in `TestQueryArrayCompatAll`. As a referene, this is what we did in `query_logical_compat_test.go`, maybe we can do something similar here?

rumyantseva:
@chilagrow my review is done (one tiny thing to fix and structuring idea, see above).

integration/query_array_compat_test.go
AlekSi: > `DotNotationPositionIndexAtTheEndOfArray` and `DotNotationPositionIndexAtTheEndOfArray`

That's the same name

rumyantseva: @AlekSi indeed! :D bad copy-pasting, fixed.

integration/query_array_compat_test.go
chilagrow: I'll separated them, initially I thought they would be faster so I put them in the same func. But I just learnt that on CI it's faster to run multiple tests than one big test. Interesting.

Using multiple tests like `TestQueryArrayCompatSize`  `TestQueryArrayCompatAll` https://github.com/FerretDB/FerretDB/actions/runs/3529897317/jobs/5921499750**
```
ok  	github.com/FerretDB/FerretDB/integration	168.466s	coverage: 61.0% of statements in ../...
```
using one big test
https://github.com/FerretDB/FerretDB/actions/runs/3530990337/jobs/5923627738
```
ok  	github.com/FerretDB/FerretDB/integration	206.855s	coverage: 61.0% of statements in ../...
```

rumyantseva: @chilagrow, well, can we compare the time of two different CI runs? I'm not sure how GitHub Actions distribute resources, but I expect it might be possible that even the same test might take a different time to run on CI depends on available resources.

chilagrow: true, never mind.

integration/query_array_test.go
chilagrow: I updated the issue with example https://github.com/FerretDB/FerretDB/issues/1539
This only fails on scalar values.

integration/basic_test.go
chilagrow: Yes I removed test for find because it's covered by compat.

I added comment in that test.

integration/query_array_compat_test.go
chilagrow: Fixed

integration/query_array_test.go
```diff
@@ -406,6 +247,8 @@ func TestQueryArrayAll(t *testing.T) {
 	// Insert additional data to check more complicated cases:
 	// - a longer array of ints;
 	// - a field is called differently and needs to be found with the {$all: [null]} case.
+	// TODO Add "many-integers" to shareddata.Composites once more
```
rumyantseva, line 250: Do we need an issue number for this `TODO`, or is it in the scope of the current task?

rumyantseva (approved):
LGTM, one minor question

integration/query_array_test.go
chilagrow: It's in scope of the current task, it will be in upcoming PR :hugs:

integration/query_array_compat_test.go
```diff
@@ -0,0 +1,260 @@
...
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+func TestQueryArrayCompatSize(t *testing.T) {
+	t.Parallel()
+
+	testCases := map[string]queryCompatTestCase{
+		"float64": {
+			filter:        bson.D{{"v", bson.D{{"$size", float64(2)}}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
```
AlekSi, line 31: I think it might be easier to skip the whole test for Tigris, not all individual test cases.

chilagrow: Let's do it that way :+1:

PR #1544, opened 2022-11-23
Move `_id` field check to separate function

w84thesun:
# Description

Refs #1151.

I moved code related to the `_id` check to that PR as was requested.

## Readiness checklist

### DIFF
--- a/internal/handlers/pg/msg_findandmodify.go
+++ b/internal/handlers/pg/msg_findandmodify.go
@@ -271,7 +271,7 @@ func (h *Handler) upsert(ctx context.Context, tx pgx.Tx, docs []*types.Document,
 			}
 		}
 
-		err := h.insert(ctx, tx, params.sqlParam, upsert)
+		err := h.insert(ctx, params.sqlParam, upsert)
 		if err != nil {
 			return nil, false, err
 		}
--- a/internal/handlers/pg/msg_findandmodify.go
+++ b/internal/handlers/pg/msg_findandmodify.go
@@ -271,9 +271,6 @@ func (h *Handler) upsert(ctx context.Context, tx pgx.Tx, docs []*types.Document,
 			}
 		}
 
-		// we must place the _id to the first index of the upsert
-		upsert.MoveIDToTheFirstIndex()
-
 		err := h.insert(ctx, tx, params.sqlParam, upsert)
 		if err != nil {
 			return nil, false, err
--- a/internal/handlers/pg/msg_findandmodify.go
+++ b/internal/handlers/pg/msg_findandmodify.go
@@ -271,6 +271,9 @@ func (h *Handler) upsert(ctx context.Context, tx pgx.Tx, docs []*types.Document,
 			}
 		}
 
+		// we must place the _id to the first index of the upsert
+		upsert.MoveIDToTheFirstIndex()
+
 		err := h.insert(ctx, tx, params.sqlParam, upsert)
 		if err != nil {
 			return nil, false, err
--- a/internal/handlers/tigris/tjson/document.go
+++ b/internal/handlers/tigris/tjson/document.go
@@ -96,6 +96,8 @@ func (doc *documentType) UnmarshalJSONWithSchema(data []byte, schema *Schema) er
 func (doc *documentType) MarshalJSON() ([]byte, error) {
 	td := types.Document(*doc)
 
+	td.MoveIDToTheFirstIndex()
+
 	var buf bytes.Buffer
 
 	buf.WriteString(`{"$k":`)
--- a/internal/handlers/tigris/tjson/document.go
+++ b/internal/handlers/tigris/tjson/document.go
@@ -96,8 +96,6 @@ func (doc *documentType) UnmarshalJSONWithSchema(data []byte, schema *Schema) er
 func (doc *documentType) MarshalJSON() ([]byte, error) {
 	td := types.Document(*doc)
 
-	td.MoveIDToTheFirstIndex()
-
 	var buf bytes.Buffer
 
 	buf.WriteString(`{"$k":`)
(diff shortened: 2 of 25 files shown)


internal/types/document.go
```diff
@@ -387,6 +363,35 @@ func (d *Document) isKeyDuplicate(targetKey string) bool {
...
+
+	for i, key := range d.Keys() {
+		if key == "_id" {
+			idIdx = i
+			break
+		}
+	}
+
+	if idIdx == 0 {
+		return
```
rumyantseva, line 387: This place is reached if `MoveIDToTheFirstIndex` is called for a document that doesn't contain `_id` at all. Is it a possible (correct) case, or should we panic if that happened?

internal/handlers/pg/msg_findandmodify.go
```diff
@@ -271,6 +271,9 @@ func (h *Handler) upsert(ctx context.Context, tx pgx.Tx, docs []*types.Document,
 			}
 		}
 
+		// we must place the _id to the first index of the upsert
+		upsert.MoveIDToTheFirstIndex()
```
rumyantseva, line 275: I see you call it for upsert but not for `insert` or `update`. Why? Wouldn't it be enough to call it `types.Document.ValidateData` instead?

internal/types/document.go
w84thesun: That check should be removed as we have it before the for loop.

internal/handlers/pg/msg_findandmodify.go
w84thesun: For insert or update operations it would be called inside MarshallJSON. I don't get the part for `types.Document.ValidateData`.

rumyantseva: @w84thesun it's a different check. If your doc doesn't have `_id` at all, it will come to this check. If this `if` doesn't exist, and your test doesn't have `_id`, the code below will be executed.

rumyantseva: @w84thesun I mean, why do we need to have `MoveIDToTheFirstIndex` in a few different places, if for all of them `Document.ValidateData` is called before insert or update. Can we just move this logic of id sorting there? So we ensure that it's sorted before we send it to the db. Or it's not enough?

rumyantseva: Also, `h.insert` is called on the line 277, and it executes the same thing as for `insert`. You say, the logic for `insert` is already in `MarshalJSON`, why do we need this extra line then?

w84thesun: > Also, `h.insert` is called on the line 277, and it executes the same thing as for `insert`. You say, the logic for `insert` is already in `MarshalJSON`, why do we need this extra line then?

No need, removed it.

internal/handlers/tigris/tjson/document.go
```diff
@@ -96,6 +96,8 @@ func (doc *documentType) UnmarshalJSONWithSchema(data []byte, schema *Schema) er
 func (doc *documentType) MarshalJSON() ([]byte, error) {
 	td := types.Document(*doc)
 
+	td.MoveIDToTheFirstIndex()
```
rumyantseva, line 99: Why do we need it here?

internal/types/document_validation.go
```diff
@@ -67,7 +67,6 @@ func (d *Document) ValidateData() error {
 	duplicateChecker := make(map[string]struct{}, len(keys))
 	var idPresent bool
 
-	// TODO: make sure that `_id` is the first item in the map
```
rumyantseva, line 70: 👍

internal/handlers/tigris/tjson/document.go
w84thesun: Sorry, missed that one occurrence.

internal/types/document_validation.go
AlekSi: TBH, I don't understand why we did not do that there and instead extracted an exported function for something that should be an internal detail

rumyantseva: I had the same comment initially, but at some point I understand the idea of splitting the stage where we "normalize" data and where we "validate" data.

AlekSi: That might be a good idea, but that doesn't explain why we do that in handlers, of all things. That's too late for that. I still think that should be an internal detail of `type` package

rumyantseva: @AlekSi do you have a more exact suggestion where this could be moved? To which place/stage?

w84thesun: Should I move `MoveIDToTheFirstIndex` inside `ValidateData`?

internal/types/document_test.go
```diff
@@ -75,23 +75,31 @@ func TestDocument(t *testing.T) {
...
-		doc.Set("_id", "bar")
-		assert.Equal(t, []string{"_id", "foo"}, doc.Keys())
+		doc = must.NotFail(NewDocument(
+			"foo", "bar",
+			"baz", "qux",
+			"_id", int32(42),
+		))
+		doc.MoveIDToTheFirstIndex()
+		assert.Equal(t, []string{"_id", "foo", "baz"}, doc.Keys())
 	})
```
rumyantseva, line 103: I like that we have tests for different options! Let's also add one for the case when `_id` is not present in the doc at all? So then we expect nothing is changed in the doc (but nothing panics etc). Codecov currently shows this case uncovered.

w84thesun: Done.

chilagrow (approved):

rumyantseva (approved):
LGTM

AlekSi (changes requested):
Let's discuss on DS

AlekSi (approved):

rumyantseva (approved):
LGTM

PR #1560, opened 2022-11-24
Move more query tests to compat and change filter bad value check order

### LINKED ISSUE #1450: Move more query tests to compat tests

chilagrow:
# Description

Closes #1450.

## Readiness checklist


integration/query_compat_test.go
```diff
@@ -28,14 +28,14 @@ import (
...
-	skipForTigris string                   // skips test for Tigris if non-empty
+	filter     bson.D                   // required
+	sort       bson.D                   // defaults to `bson.D{{"_id", 1}}`
+	resultType compatTestCaseResultType // defaults to nonEmptyResult
 }
 
 // testQueryCompat tests query compatibility test cases.
-func testQueryCompat(t *testing.T, testCases map[string]queryCompatTestCase) {
+// If skipForTigris is not empty, tigris tests are skipped.
+func testQueryCompat(t *testing.T, skipForTigris string, testCases map[string]queryCompatTestCase) {
```
chilagrow, line 38: Skip whole test for tigris https://github.com/FerretDB/FerretDB/pull/1526#discussion_r1031076525

internal/handlers/common/filter.go
```diff
@@ -720,6 +715,11 @@ func filterFieldExprSize(fieldValue any, sizeValue any) (bool, error) {
 		)
 	}
 
+	arr, ok := fieldValue.(*types.Array)
+	if !ok {
+		return false, nil
+	}
```
chilagrow, line 721: We check array type last because compat test told us we should check values first.

w84thesun (approved):

internal/handlers/common/filter.go
rumyantseva: @chilagrow as codecov complains, do we have a test for `size` where we call it for a non-array? Or is it a separate issue?

rumyantseva (approved):
LGTM. Asked one minor question in comments.

integration/query_array_compat_test.go
```diff
@@ -27,32 +27,45 @@ func TestQueryArrayCompatSize(t *testing.T) {
...
+		},
 		"Zero": {
-			filter:        bson.D{{"v", bson.D{{"$size", 0}}}},
-			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/908",
+			filter: bson.D{{"v", bson.D{{"$size", 0}}}},
 		},
 	}
 
-	testQueryCompat(t, testCases)
+	skipForTigris := "https://github.com/FerretDB/FerretDB/issues/908"
```
AlekSi, line 67: See above

internal/handlers/common/filter.go
chilagrow: Yes we do, we call them from many non array. Is this a problem?

For `$size` and `$elemMatch` we now call from every provider from shareddata such as [Double](https://github.com/FerretDB/FerretDB/blob/main/integration/shareddata/scalars.go#L101).

integration/query_array_compat_test.go
```diff
@@ -27,32 +27,45 @@ func TestQueryArrayCompatSize(t *testing.T) {
```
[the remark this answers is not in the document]
chilagrow: Thank you!

integration/query_array_compat_test.go
chilagrow: :pray:

internal/handlers/common/filter.go
rumyantseva: Sounds reasonable! I wonder why codecov didn't catch it. Anyway, it's not in the scope of the current task.

integration/query_compat_test.go
```diff
@@ -115,6 +110,8 @@ func testQueryCompat(t *testing.T, testCases map[string]queryCompatTestCase) {
 }
 
 func TestQueryCompat(t *testing.T) {
+	setup.SkipForTigrisWithReason(t, "https://github.com/FerretDB/FerretDB/issues/908")
```
AlekSi, line 113: Ok, but why skip that one?

chilagrow: yes true...
thanks for catching this one.

integration/query_array_compat_test.go
```diff
@@ -20,42 +20,60 @@ import (
 
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/primitive"
+
+	"github.com/FerretDB/FerretDB/integration/setup"
 )
 
 func TestQueryArrayCompatSize(t *testing.T) {
+	setup.SkipForTigrisWithReason(t, "https://github.com/FerretDB/FerretDB/issues/908")
```
AlekSi, line 28: Now is the question – why we skip that test if we fixed it, and arrays are just skipped for both target and compat systems when target uses Tigris?

chilagrow: This isn't fixed. In shareddata we have `Composites` which contain mix types in the array. Because we [skip creating that collection](https://github.com/FerretDB/FerretDB/blob/main/integration/setup/setup.go#L151) for tigris in setup, when we query on compat tests we see mismatch. I'd say we leave tigris out for now.

AlekSi: > Because we skip creating that collection for tigris in setup, when we query on compat tests we see mismatch

You are linking to the wrong function. `TestQueryArrayCompatSize` calls `testQueryCompat` that calls `SetupCompat` that calls `setupCompatCollections`, not `setupCollection`.

https://github.com/chilagrow/FerretDB/blob/3331480fa0c2ca01f7839caa9508d78fd769537a/integration/setup/setup_compat.go#L166-L172

That code target's handler for both target and compat systems, skipping creating collection in both system if it is not compatible with target system. There is no mismatch.

chilagrow: This is interesting because, the test still continues to not work on tigris and my above assumption is wrong. I'll create an issue for investigating that.

AlekSi: Great, let's change `SkipForTigrisWithReason`'s argument then; that test is skipped not because Tigris does not support arrays yet

noisersup (approved):
LGTM!

w84thesun (approved):

PR #1579, opened 2022-11-28
Move query tests to compat by comparing the error code instead of the error message

### LINKED ISSUE #1450: Move more query tests to compat tests

chilagrow:
# Description

Closes #1450.

## Readiness checklist


integration/query_array_test.go
```diff
@@ -29,84 +28,6 @@ import (
 	"github.com/FerretDB/FerretDB/integration/shareddata"
 )
 
-func TestQueryArraySize(t *testing.T) {
```
chilagrow, line 32: This test was already moved to compat.

integration/query_compat_test.go
```diff
@@ -76,9 +81,8 @@ func testQueryCompat(t *testing.T, testCases map[string]queryCompatTestCase) {
 
 					if targetErr != nil {
 						t.Logf("Target error: %v", targetErr)
-						targetErr = UnsetRaw(t, targetErr)
-						compatErr = UnsetRaw(t, compatErr)
-						assert.Equal(t, compatErr, targetErr)
+						AssertEqualErrorCode(t, compatErr, targetErr)
```
chilagrow, line 84: We only compare error code in compat tests.

AlekSi (commented):
We should also compare errors's `Name` field, and maybe `Wrapped`

integration/helpers.go
```diff
@@ -166,6 +167,25 @@ func AssertEqualError(t testing.TB, expected mongo.CommandError, actual error) b
...
+
+	if ok := errors.As(expected, &eErr); !ok {
+		return assert.Equal(t, expected, actual)
+	}
+
+	if ok := errors.As(actual, &aErr); !ok {
+		return assert.Equal(t, expected, actual)
+	}
+
+	assert.Equal(t, eErr.Name, aErr.Code)
```
w84thesun, line 183: Should it be `assert.Equal(t, eErr.Name, aErr.Name)`?

w84thesun (commented):
Overall good, one comment about `AssertEqualErrorCode`.

integration/query_compat_test.go
rumyantseva: I understand why we started doing it (@AlekSi proposed that and everyone agreed), but I have a new question now. 

If we don't check the error message of our errors in compat tests, and we don't have non-compat tests anymore, where do we check the error message of our errors?..  I guess, for query tests it's ok'ish, especially, if error messages are just constants. But in principle, it might make sense to check some error messages (e.g. if they contain variable parts). This is a side note, just wanted to share my thoughts.

chilagrow: This is a valid point, I would like to know how exactly we need to be for the variable parts.

rumyantseva (approved):
LGTM

integration/helpers.go
```diff
@@ -166,6 +167,25 @@ func AssertEqualError(t testing.TB, expected mongo.CommandError, actual error) b
 	return assert.Equal(t, expected, a)
 }
 
+// AssertEqualErrorCode asserts error code, name and wrapped are the same.
+func AssertEqualErrorCode(t *testing.T, expected, actual error) bool {
```
AlekSi, line 171: I think we should remove `bool` result (we could always add it later) and rename function to something like `AssertSimilarError` or `AssertMatchesError` (please check existing helpers if there is some pattern) since we don't compare just error codes anymore

integration/query_compat_test.go
AlekSi: Let's create a new issue to think about that and what to do with existing `AssertEqualAltError`

AlekSi (commented):
(cursory review)

integration/query_compat_test.go
chilagrow: https://github.com/FerretDB/FerretDB/issues/1587

integration/helpers.go
chilagrow: Used `AssertMatchesCommandError`. Other existing helpers just were removing Raw fields, which is different for this case.

w84thesun (approved):

PR #1584, opened 2022-11-28
Simplify documents fetching for `delete`

### LINKED ISSUE #898: Offer a better design for fetching documents (iterator)

rumyantseva:
# Description

Closes #898.

## Readiness checklist


internal/handlers/pg/msg_delete.go
```diff
@@ -168,66 +175,85 @@ func (h *Handler) prepareDeleteParams(deleteDoc *types.Document) (*types.Documen
 	return filter, limit, nil
 }
 
-// execDelete fetches documents, filter them out and limiting with the given limit value.
+// execDelete fetches documents, filters them out, limits them (if needed) and deletes them.
+// If limit is true, only the first matched document is chosen for deletion, otherwise all matched documents are chosen.
 // It returns the number of deleted documents or an error.
-func (h *Handler) execDelete(ctx context.Context, sp *pgdb.SQLParam, filter *types.Document, limit int64) (int32, error) {
```
rumyantseva, line 173: I refactored this signature (instead of `limit int64` I set `limit bool`) because of two reasons:
- `limit` can be either 0 or 1, and it's confusing to have it defined as `int64`
- if someone calls with function with `limit` other then 0 or 1, it won't work correctly (and the doc of that function doesn't say anything about it)

chilagrow (approved):

w84thesun (approved):

PR #1585, opened 2022-11-29
Fix greater and less operators on array value comparison

### LINKED ISSUE #1569: Greater than comparison with array of integer returns unexpected result
### Versions

0.6.2

### What did you do?

Add `"array-numbers": bson.A{42, 43, 45}` to Composites shareddata. Comparison test to compare filter `bson.D{{"v", bson.D{{"$gt", int32(42)}}}}` returns false.

It's something related to array order and sorting and comparing against the largest value https://www.mongodb.com/docs/manual/reference/bson-type-comparison-order/#arrays

Please look for the issue number in the code as well.

Returns true.

### What did you see instead?

```
Compat (expected) IDs: [array-numbers array-two]
Target (actual)   IDs: [array-two]
```

chilagrow:
# Description

Closes #1569.

## Readiness checklist


internal/types/compare.go
```diff
@@ -70,7 +111,7 @@ func Compare(docValue, filterValue any) CompareResult {
 				continue
 			}
 
-			if res := compareScalars(docValue, filterValue); res != Incomparable {
+			if res := compareScalars(docValue, filterValue); cond(res) {
```
chilagrow, line 114: I didn't use `Max` or `Min`, this somehow got quite a bit complicated. For example comparing `$gte` with doc `bson.A{"c", "b", "a"}` with filter `bson.D{{"v", bson.D{{"$gte", 41.13}}}}`, I think it is greater because [string is greater than number](https://www.mongodb.com/docs/manual/reference/bson-type-comparison-order/#comparison-sort-order). But mongo DB doesn't think so according to the compat test. Anyway for now, let's skip max and min.

internal/types/compare.go
```diff
@@ -70,7 +123,7 @@ func Compare(docValue, filterValue any) CompareResult {
 				continue
 			}
 
-			if res := compareScalars(docValue, filterValue); res != Incomparable {
+			if res := compareScalars(docValue, filterValue); cond.satisfies(res) {
```
chilagrow, line 126: I didn't use Max or Min, this somehow got quite a bit complicated. For example comparing $gte with doc bson.A{"c", "b", "a"} with filter bson.D{{"v", bson.D{{"$gte", 41.13}}}}, I think it is greater because [string is greater than number](https://www.mongodb.com/docs/manual/reference/bson-type-comparison-order/#comparison-sort-order). But mongo DB doesn't think so according to the compat test. Anyway for now, let's skip max and min.

internal/types/compare.go
chilagrow: I was totally wrong :see_no_evil:

chilagrow: Okay, so it seems this way. If you have a docValue `bson.A{nil, "foo", 42}` and a filter `{"$gte", 43}`, compat returns less `<`. It seems to skip/ignores a string "foo" (which has [greater comparison order](https://www.mongodb.com/docs/manual/reference/bson-type-comparison-order/#comparison-sort-order)) but only compare against number 42. So getting max won't work because max is specific to type of value filter has.

internal/types/compare.go
```diff
@@ -40,12 +40,67 @@ const (
...
+	case Greater:
+		return c.greater
+	case Less:
+		return c.less
+	case Equal:
+		return c.equal
+	case Incomparable:
+		return false
+	default:
+		return false
```
rumyantseva, line 94: What could be other values apart from Greater, Less, Equal and Incomparable?

chilagrow: Removed this default added panic, just for the consistency with rest of the code

internal/types/compare.go
```diff
@@ -40,12 +40,67 @@ const (
...
+	return compare(docValue, filterValue, cond{greater: true, less: true, equal: true})
+}
+
+// CompareGreaterThan compares docValue for the greater than condition.
+func CompareGreaterThan(docValue, filterValue any) CompareResult {
+	return compare(docValue, filterValue, cond{greater: true})
+}
+
+// CompareGreaterThanOrEq compares docValue for the greater than or equal to condition.
+func CompareGreaterThanOrEq(docValue, filterValue any) CompareResult {
```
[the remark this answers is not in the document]
AlekSi: I think types package should export only `Compare` function and everything else should be an internal. I think we want to move towards a smaller exported APIs in general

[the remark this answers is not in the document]
chilagrow: At the moment we pass `Compare(bson.A{"c", "b", "a"}, 43)` for `Compare(docValue, filterValue)`. So we don't have information related to the operator and we need that for this case. Perhaps we can pass `Compare(bson.A{"c", "b", "a"}, bson.D{{"$gte", 41.13}})` which includes the operator info, dependent on document comparison though.

Any other way of how we could achieve this?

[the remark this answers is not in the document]
rumyantseva: @chilagrow hmm, I think I start understanding... So, when we call `Compare` we can't say "it's equal" or "it's greater" without knowing the operator? Like for different operators, the result of `Compare` could be different?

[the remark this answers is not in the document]
AlekSi: I think Compare(a, b) should compare values a and b; both a and b should not be expressions containing operators, but just values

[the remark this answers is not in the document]
AlekSi: And I think `Compare(docValue, filterValue any)` is a rudiment of an old comparison code that tried to handle subarrays. That really should be `Compare(a, b any)` if possible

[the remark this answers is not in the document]
chilagrow: I think what could possibly be done is to get `max` for  $gt, $gte and  `min` value for $lt,$lte before calling `Compare` and call `Compare(maxValue,filterValue)` for $gt for example. At the moment getting `max` or `min` is a bit tricky because when an array contains multiple types, it doesn't behave as expected.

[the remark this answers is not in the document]
chilagrow: I'll see how this can be simplified to  Compare(a, b any)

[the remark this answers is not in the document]
AlekSi: > At the moment getting max or min is a bit tricky because when an array contains multiple types, it doesn't behave as expected

So we just™ need to fix comparison, so Min and Max methods behave correctly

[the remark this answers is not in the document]
noisersup: I think I understand the logic of this PR and the solution looks reasonable to me. And yeah, I see a field to simplify our *existing* code, but I'm not sure if refactoring the code is in a scope of this issue.

integration/shareddata/composites.go
```diff
@@ -40,6 +40,9 @@ var Composites = &Values[string]{
 		"array-three-reverse": bson.A{nil, "foo", int32(42)},
 		"array-empty":         bson.A{},
 		"array-null":          bson.A{nil},
+		"array-numbers-asc":   bson.A{int32(42), int32(43), int32(45)},
+		"array-strings-desc":  bson.A{"c", "b", "a"},
+		"array-composite":     bson.A{nil, primitive.Timestamp{T: 42, I: 13}, false, "z", primitive.NilObjectID},
```
chilagrow, line 45: I wanted to add more concrete one but it failed on many tests so for now I created this issue https://github.com/FerretDB/FerretDB/issues/1607

internal/types/compare.go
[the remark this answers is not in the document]
chilagrow: I moved all the logic related to operator into filter. No change in `Compare` now.

integration/query_projection_test.go
```diff
@@ -63,61 +63,6 @@ func TestQueryProjection(t *testing.T) {
...
-	require.NoError(t, err)
-
-	for name, tc := range map[string]struct {
-		projection  any
-		expectedIDs []any
-	}{
-		"ElemMatch": {
-			projection: bson.D{{
-				"v",
-				bson.D{{"$elemMatch", bson.D{{"field", bson.D{{"$eq", 42}}}}}},
```
chilagrow, line 94: Already moved in https://github.com/FerretDB/FerretDB/blob/main/integration/query_projection_compat_test.go#L60 and now shareddata has array numbers in composit `"array-numbers-asc": bson.A{int32(42), int32(43), int32(45)}`

integration/query_projection_compat_test.go
```diff
@@ -64,21 +64,3 @@ func TestQueryProjectionCompatElemMatch(t *testing.T) {
 
 	testQueryCompat(t, testCases)
 }
-
-func TestQueryProjectionCompatSlice(t *testing.T) {
```
chilagrow, line 68: Somehow this was duplicated so removing it. See L50 to L66

integration/query_test.go
```diff
@@ -274,7 +274,7 @@ func TestQueryCount(t *testing.T) {
 	}{
 		"CountAllDocuments": {
 			command:  bson.D{{"count", collection.Name()}},
-			response: 45,
+			response: 48,
```
AlekSi, line 277: Those tests really should be moved to compat tests

internal/handlers/common/filter.go
```diff
@@ -348,9 +348,23 @@ func filterFieldExpr(doc *types.Document, filterKey string, expr *types.Document
...
 			// {field: {$gt: exprValue}}
-			if _, ok := exprValue.(types.Regex); ok {
+			switch exprValue.(type) {
+			case *types.Document, *types.Array:
+			case types.Regex:
 				msg := fmt.Sprintf(`Can't have RegEx as arg to predicate over field '%s'.`, filterKey)
 				return false, NewCommandErrorMsgWithArgument(ErrBadValue, msg, exprKey)
+			default:
+				if arrValue, ok := fieldValue.(*types.Array); ok {
+					// Filter array by only keeping the same type as exprValue.
```
rumyantseva, line 358: As discussed in the call, we need the "why" for this comment. 
For example, something like "only values of the same type can be compared, so we filter the array" (but maybe there is a better wording for it"

integration/shareddata/composites.go
```diff
@@ -40,6 +40,9 @@ var Composites = &Values[string]{
 		"array-three-reverse": bson.A{nil, "foo", int32(42)},
 		"array-empty":         bson.A{},
 		"array-null":          bson.A{nil},
+		"array-numbers-asc":   bson.A{int32(42), int32(43), int32(45)},
```
rumyantseva, line 43: @chilagrow as you mentioned in the call, those could be int32, int64, and float to check that we compare numbers correctly.

rumyantseva (approved):
The code looks good to me overall. Great job with all those tiny details! 

As discussed in the call, there are a couple of comments about documentation/test cases to improve.

rumyantseva (commented):
Sorry, I meant to leave "comment" (not "approve" yet, but almost! 👍)

internal/handlers/common/filter.go
w84thesun, line 358: Please rewrite this comment.

w84thesun (changes requested):
As discussed on the call comment about using `FIlterArrayByType` needs to be clarified a bit.

integration/update_test.go
```diff
@@ -96,13 +96,13 @@ func TestMultiFlag(t *testing.T) {
 				update:         bson.D{{"$set", bson.D{{"v", int32(43)}}}},
 				multi:          false,
 				stat:           bson.D{{"n", int32(1)}, {"nModified", int32(1)}, {"ok", float64(1)}},
-				expectedToFind: 5,
+				expectedToFind: 6,
```
chilagrow, line 99: Created and issue to move update tests to compat https://github.com/FerretDB/FerretDB/issues/1610

integration/query_test.go
chilagrow: fixed

integration/shareddata/composites.go
chilagrow: fixed.

internal/handlers/common/filter.go
chilagrow: I added example as well :pray:

chilagrow: added, please let me know it it's not clear :pray:

w84thesun (dismissed):

noisersup (dismissed):

integration/query_command_compat_test.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+// queryCountCompatTestCase describes command query compatibility test case.
+type queryCountCompatTestCase struct {
+	query   bson.D // optional
+	command bson.D // defaults to count collection
+}
+
+// testQueryCountCompat tests command query compatibility test cases.
+func testQueryCountCompat(t *testing.T, testCases map[string]queryCountCompatTestCase) {
```
rumyantseva, line 33: Hmm, if I'm not mistaken, `count_compat_test.go` has some similar things... Does it make sense to combine those two, so we don't have duplicate code?

rumyantseva (commented):
There is a possible duplicate of code for count compat tests (see `count_compat_test.go`), the rest looks good to me.

integration/query_command_compat_test.go
chilagrow: yes thanks for spotting this. I combined them

rumyantseva (dismissed):
Looks good to me overall, one tiny comment about test cases.

integration/count_compat_test.go
```diff
@@ -120,6 +131,19 @@ func TestCountCompat(t *testing.T) {
...
+		"IDBool": {
+			filter: bson.D{{"_id", "bool-true"}},
+		},
+		"FieldTrue": {
+			filter: bson.D{{"v", true}},
+		},
+		"FieldTypeArrays": {
+			filter: bson.D{{"v", bson.D{{"$type", "array"}}}},
+		},
+		"NonExistingCollection": {
```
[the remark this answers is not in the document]
chilagrow: Thanks, non-existing collections are already covered so removing test.

rumyantseva (approved):
LGTM

w84thesun (approved):

PR #1591, opened 2022-11-29
Fix `listIndexes` stub

noisersup:
# Description

There's an array inside the array by accident.

Also: should we send some message in response that this command is only a stub and the output is not relevant becuase we don't support indexes at this moment?

## Readiness checklist


AlekSi (dismissed):

AlekSi:
> should we send some message in response that this command is only a stub and the output is not relevant becuase we don't support indexes at this moment?

No, let's spend our time actually implementing indexes instead :)

AlekSi (commented):
Please do the same for Tigris

AlekSi (approved):

PR #1592, opened 2022-11-29
Improve `iterator.Interface` implementation in `pgdb.queryIterator`

### LINKED ISSUE #898: Offer a better design for fetching documents (iterator)

rumyantseva:
# Description

Closes #898.

## Readiness checklist


rumyantseva:
This PR is to check a particular design/implementation idea proposed by @AlekSi, so, @AlekSi's review is required and this is why "do not merge" label is set.

internal/handlers/pg/pgdb/query.go
```diff
@@ -110,15 +110,15 @@ func (pgPool *Pool) QueryDocuments(ctx context.Context, tx pgx.Tx, sp *SQLParam)
...
+// If the collection doesn't exist, it returns an empty iterator and no error.
 // If an error occurs, it returns nil and that error, possibly wrapped.
 func (pgPool *Pool) GetDocuments(ctx context.Context, tx pgx.Tx, sp *SQLParam) (
 	iterator.Interface[uint32, *types.Document], error,
 ) {
 	q, args, err := buildQuery(ctx, tx, sp)
 	if err != nil {
 		if errors.Is(err, ErrTableNotExist) {
-			return nil, nil
+			return newIterator(ctx, nil), nil
```
rumyantseva, line 121: @AlekSi here https://github.com/FerretDB/FerretDB/pull/1487/files#r1032112552 you proposed to return an empty iterator if the queried collection doesn't exist. Does my implementation in this PR cover your idea, or did you mean something else?

AlekSi:
I think that comment should be in the PR body before description

internal/handlers/pg/pgdb/query.go
AlekSi: I think so, but it will be more clear with Close method added

rumyantseva: @AlekSi you probably missed that, but `Close` was added and merged :)

internal/handlers/pg/pgdb/query_iterator.go
```diff
@@ -98,6 +98,9 @@ func (it *queryIterator) Close() {
 		return
 	}
 
-	it.rows.Close()
+	if it.rows != nil {
```
rumyantseva, line 101: @AlekSi changes in the `Close` method.

AlekSi, line 101: fwiw, it could be wrapped with sync.Once instead of mutex and flag

internal/handlers/pg/pgdb/query.go
AlekSi: Got confused by GH collapsing a diff at the worst possible place

internal/handlers/pg/pgdb/query_iterator.go
rumyantseva: Ah, that's the function I tried to remember but couldn't! I'll make changes in this PR then.

integration/update_field_test.go
```diff
@@ -86,7 +86,7 @@ func TestUpdateFieldCurrentDate(t *testing.T) {
 
 	t.Run("currentDate", func(t *testing.T) {
 		// maxDifference is a maximum amount of seconds can differ the value in placeholder from actual value
-		maxDifference := time.Duration(3 * time.Minute)
+		maxDifference := time.Duration(4 * time.Minute)
```
rumyantseva, line 89: This place started failing too often, and 3 min didn't seem to be enough anymore. 
We have some tasks that hopefully will improve performance, but for now it's out of the scope.

chilagrow (dismissed):

internal/handlers/pg/pgdb/query_iterator.go
rumyantseva: fixed

integration/update_field_test.go
AlekSi: Please add TODO with that task

internal/handlers/pg/pgdb/query_iterator.go
```diff
@@ -32,8 +32,8 @@ import (
 type queryIterator struct {
 	ctx         context.Context
 	rows        pgx.Rows
-	mxClosed    *sync.Mutex // protects closed
-	closed      bool        // indicates whether Close() was called
+	close       sync.Once   // to ensure that all Close() calls after the first will have no effect.
```
AlekSi, line 35:
```suggestion
	close       sync.Once   // to ensure that all Close() calls after the first will have no effect
```

We use dots only for complete and capitalized sentences.

no dots for sentence fragments

But the most idiomatic way would be to call that field `closeOnce` without any comment

internal/handlers/pg/pgdb/query_iterator.go
```diff
@@ -42,16 +42,12 @@ type queryIterator struct {
...
-		mxClosed: new(sync.Mutex),
+		ctx:  ctx,
+		rows: rows,
 	}
 
 	runtime.SetFinalizer(qi, func(qi *queryIterator) {
-		qi.mxClosed.Lock()
-		defer qi.mxClosed.Unlock()
-		if !qi.closed {
+		if !qi.closed.Load() {
```
AlekSi, line 50:
```suggestion
		it.close.Do(func() {
```

No need to use a separate flag. And please call it `it` everywhere for consistency

w84thesun (commented):
Nice, it will make usage of iterator easier! Please resolve conflicts.

AlekSi (dismissed):

internal/handlers/pg/pgdb/query_iterator.go
rumyantseva: Ah!

integration/update_field_test.go
rumyantseva: Added

AlekSi (approved):

w84thesun (approved):

PR #1596, opened 2022-11-29
Refactor `pg`'s `MsgListDatabases` and `pgdb`

### LINKED ISSUE #1056: Refactor `pg`'s `MsgListDatabases` and `pgdb`
`MsgListDatabases` calls `pg_total_relation_size` and `pg_database_size` directly. Code wrapping those functions should be incapsulated in `pgdb` package with proper transaction handling.

rumyantseva:
# Description

Closes #1056.

## Readiness checklist


internal/handlers/pg/pgdb/databases.go
```diff
@@ -117,3 +117,58 @@ func DropDatabase(ctx context.Context, tx pgx.Tx, db string) error {
...
+		fullName := pgx.Identifier{db, name}.Sanitize()
+		// If the table was deleted after we got the list of tables, pg_total_relation_size will return null.
+		// We use COALESCE to scan this null value as 0 in this case.
+		// Even though we run the query in a transaction, the current isolation level doesn't guarantee
+		// that the table is not deleted (see https://www.postgresql.org/docs/14/transaction-iso.html).
+		// PgPool (not a transaction) is used on purpose here. In this case, transaction doesn't lock
+		// relations, and it's possible that the table/schema is deleted between the moment we get the list of tables
+		// and the moment we get the size of the table. In this case, we might receive an error from the database,
+		// and transaction will be interrupted. Such errors are not critical, we can just ignore them, and
+		// we don't need to interrupt the whole transaction.
```
rumyantseva, line 154: An alternative approach would be to introduce something like

```go
// ErrTransactionConflict is returned when one of the queries in the transaction returned an error because
// of an unexpected conflict. The caller could retry such a transaction.
var ErrTransactionConflict = fmt.Errorf("transaction can't continue due to a conflict")

// InTransactionWithRetry wraps the given function f in a transaction.
// If f returns an error, the transaction is rolled back.
// Errors are wrapped with lazyerrors.Error,
// so the caller needs to use errors.Is to check the error,
// for example, errors.Is(err, ErrSchemaNotExist).
func (pgPool *Pool) InTransactionWithRetry(ctx context.Context, f func(pgx.Tx) error, maxRetries int, delay time.Duration) (err error) {
	for retry := 0; retry < maxRetries; retry++ {
		err := pgPool.InTransaction(ctx, f)

		switch {
		case err == nil:
			return nil
		case errors.Is(err, ErrTransactionConflict):
			time.Sleep(delay)
		default:
			return err
		}
	}
}
```

And then in the current function, instead of

```go
case pgerrcode.UndefinedTable, pgerrcode.InvalidSchemaName:
				// Table or schema was deleted after we got the list of tables, just ignore it
				continue
			}
```

we could return `ErrTransactionConflict` and retry the whole transaction.

I decided not to implement it because it looks like an unnecessary overhead, especially in this particular case.

rumyantseva: (cc @AlekSi, I added this comment because the initial task mentions "proper transaction handling")

chilagrow (approved):

w84thesun (approved):

internal/handlers/pg/pgdb/databases.go
AlekSi: We have a separate draft note in the project for `InTransactionRetry` helper

AlekSi (approved):

PR #1602, opened 2022-11-30
Fix validation for documents with duplicate keys

rumyantseva:
# Description

It was wrong to use `types.Document.Get` in validation because `types.Document.Get` can't work with duplicate keys correctly.

Luckily, now `types.Document.Get` panics in case of duplicate keys, so with the help of dance tests, we were able to catch this problem.

In principle, in such cases, we should att unit tests to make sure that the code is covered.

## Readiness checklist


chilagrow (dismissed):

noisersup (dismissed):

AlekSi (changes requested):
We should add tests for fixed bugs – in this repo. It is the first item in the readiness checklist

AlekSi (approved):

w84thesun (approved):

PR #1623, opened 2022-12-05
Use TLS for MongoDB in integration tests

### LINKED ISSUE #1594: Simplify Unix socket handling code
### What should be done?

Our current Unix socket code incorrectly uses `url.URL.Host` for the file path instead of `url.URL.Path`. That results in weird problems with case insensitivity, escaping, etc. We should fix and simplify integration tests.

### Where?

https://github.com/FerretDB/FerretDB/blob/main/integration/setup/common_unix.go and around

### LINKED ISSUE #1615: Use TLS for MongoDB in integration tests
**Scope:** when MongoDB is used we need to connect to MongoDB through TLS. For that MongoDB should be run with TLS.

Update Docker Compose / MongoDB configuration to prefer TLS connections and use CA/cert/key files in `build/certs` on the host.

Add `-compat-tls` integration testing flag to enable TLS with the same CA file.

`tlsInsecure` and other similar options should not be used; as far as our tests are concerned, MongoDB's TLS certificate should be valid (for the given CA). 

Update cert/key hosts if needed.

See #1612.

w84thesun:
# Description

Closes #1615.
Closes #1594.

## Readiness checklist


chilagrow (dismissed):
Looks great!

integration/setup/common.go
```diff
@@ -93,11 +98,65 @@ func SkipForPostgresWithReason(tb testing.TB, reason string) {
 	}
 }
 
+// buildMongoDBURIOpts represents buildMongoDBURI's options.
+type buildMongoDBURIOpts struct {
```
rumyantseva, line 102: I see in the `buildMongoDBURI` there is the logic that only one of `hostPort` or `unixSocketPath` needs to be set. Should we add any comments about it here? Like "it's up to X to decide how to use this configuration".

integration/setup/common.go
```diff
@@ -93,11 +98,65 @@ func SkipForPostgresWithReason(tb testing.TB, reason string) {
...
+		host = opts.hostPort
+		path = "/"
+	} else {
+		require.NotEmpty(tb, opts.unixSocketPath, "neither hostPort nor unixSocketPath are set")
+		path = opts.unixSocketPath
+	}
+
+	q := make(url.Values)
+
+	if opts.tls {
```
rumyantseva, line 123: How would tls work in case of unix socket configuration? Based on the current implementation, it looks like it's a possible case. Is it?

rumyantseva (commented):
The changes make sense to me. Asked a couple of questions for better understanding.

integration/setup/common.go
w84thesun: I've added the check for the Unix socket.

integration/setup/common.go
w84thesun: The logic that ensures that only one parameter should be set in `buildMongoDBURI` will fail test execution, so I think it's ok to leave it as is.

rumyantseva (approved):
LGTM

AlekSi (approved):

PR #1634, opened 2022-12-08
Fix comparison with unset fields

### LINKED ISSUE #1023: Comparison with unset fields works incorrectly
It seems that MongoDB differentiates between null and unset values, but we don't. See #1024.

## potential issues
- when we create filter documents we compare documents fetched from db with a filter given in the update query.
- when fields are not set in document we skip them, but we shouldn't

chilagrow:
# Description

Closes #1023.

This PR is about making unset work the same way as mongoDB. In the issue there is a link to a close #1024 which shows that unset comparison had issues. In this PR, we compare `unset` field with `null` field using compat dataset and tests. See `Unset` field being un-commented in the Scalar shareddata, so every compat test is run on `Unset` bson.

For a field `v` which is used in integration test, unset BSON looks like `{_id: "unset"}`, while null field BSON looks like `{_id: "null", v: null}`.

What was done before this PR:
We checked if filter was null filter, if it was it returned `true` indicating match, otherwise it returned `false`. So it returned early based on the filter value being `null` or not. And this would fail on operator such as `$gt`, `$lt`, `$nin` and `$in`. For example when comparing with filter `{v: {$gt: null}` on `{_id: "unset"}` we returned `true` before this PR. But compat test tells us it should be `false`.

What we change:
The field is set to `null` on unset value and let the comparison `Compare(a, b any)` do the work on returning result. 

## Readiness checklist


integration/query_test.go
```diff
@@ -30,238 +30,6 @@ import (
...
-
-	for name, tc := range map[string]struct {
-		sort        bson.D
-		expectedIDs []any
-		err         *mongo.CommandError
-	}{
-		"AscValueScalar": {
-			sort: bson.D{{"v", 1}, {"_id", 1}},
-			expectedIDs: []any{
-				"null",
```
chilagrow, line 169: It's quite tricky because `null` and `unset` are the equal value according to sort order, so I see they come in different order sometimes. Need test to not care about if `null` or `unset` comes in wrong order.

https://www.mongodb.com/docs/manual/reference/bson-type-comparison-order/#non-existent-fields

integration/query_test.go
```diff
@@ -166,6 +166,7 @@ func TestQuerySortValue(t *testing.T) {
 		"AscValueScalar": {
 			sort: bson.D{{"v", 1}, {"_id", 1}},
 			expectedIDs: []any{
+				"unset",
```
chilagrow, line 169: Created this issue https://github.com/FerretDB/FerretDB/issues/1637

internal/types/compare_test.go
```diff
@@ -37,18 +37,6 @@ func TestCompare(t *testing.T) {
 			b:        must.NotFail(NewArray(NullType{})),
 			expected: Less,
 		},
```
chilagrow, line 39: Removed because they are not the same for comparison, but for sorting they are considered that same. And this test is for comparison. See https://www.mongodb.com/docs/manual/reference/bson-type-comparison-order/#non-existent-fields

integration/update_field_compat_test.go
```diff
@@ -113,9 +113,13 @@ func TestUpdateFieldCompatMax(t *testing.T) {
...
 
 		"MultipleQueries": {
-			update:        bson.D{{"$max", bson.D{{"v", int32(39)}, {"a", int32(30)}}}},
+			update:        bson.D{{"$max", bson.D{{"a", int32(30)}, {"v", int32(39)}}}},
 			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/1061",
 		},
+		"MultipleQueriesSorted": {
+			update: bson.D{{"$max", bson.D{{"v", int32(39)}, {"a", int32(30)}}}},
+			skip:   "https://github.com/FerretDB/FerretDB/issues/1637",
+		},
```
chilagrow, line 122: Upon update, response was returned in reverse order for unset in `MultipleQueries`. For now I tag it as sorting issue and skip the test.
```
        	            	--- expected
        	            	+++ actual
        	            	@@ -3,8 +3,8 @@
        	            	     "_id",
        	            	-    "a",
        	            	-    "v"
        	            	+    "v",
        	            	+    "a"
        	            	   ],
        	            	   "_id": "unset",
        	            	-  "a": 30,
        	            	-  "v": 39
        	            	+  "v": 39,
        	            	+  "a": 30
        	            	 }
```

rumyantseva (commented):
I think I need more context to review the PR. What issues does it fix?
I see the one in `elemMatch` where we check `filterKey` in the document, what else?

noisersup (commented):
Yeah I'm also struggling a little bit with understanding the scope of this PR.
If I understand there's an issue that we handle unset values as nulls, and mongodb differentiate between them, right? Also there's a new `TestQueryCompatSort` test which is a result of finding some sorting bug in our compat test?

chilagrow:
@rumyantseva @noisersup thanks for the feedback, I added description of PR and comments in the code.

internal/handlers/common/update.go
```diff
@@ -36,9 +36,9 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
 	if update.Len() == 0 {
 		// replace to empty doc
 		for _, key := range doc.Keys() {
-			if key != "_id" {
-				changed = true
+			changed = true
```
chilagrow, line 39: Compat test tells if there is any key including `_id`, it should return true for changed. Added `Unset` to shareddata and this change passes compat test for `Unset`

chilagrow:
> Yeah I'm also struggling a little bit with understanding the scope of this PR. If I understand there's an issue that we handle unset values as nulls, and mongodb differentiate between them, right? Also there's a new `TestQueryCompatSort` test which is a result of finding some sorting bug in our compat test?

@noisersup if we see `null` in the filter and `unset` value in the docValue, our filter returned true before this PR. In reality, we should apply operator and returns the result based on the filter operator. This is what the PR is about. And yes, mongoDB differentiate the result based on the operator. The way I did it, most operators treat `null` and missing key equal [see] (https://www.mongodb.com/docs/manual/reference/bson-type-comparison-order/#non-existent-fields). I handle the operators that shouldn't treat `null` and `unset` equal upon getting error. Other operators should handle `null` and `unset` equally but each operator has to apply either logic, for that I set missing fieldValue as null to allow that.

chilagrow:
> I think I need more context to review the PR. What issues does it fix? I see the one in `elemMatch` where we check `filterKey` in the document, what else?

@rumyantseva fix in this PR is about errors that was produced by adding `Unset` to Scalar shareddata so they are used in compat test.
* upon updating `unset` document which only has `_id` key, it has to return that the collection was modified.
* before this PR is filter was `null` and document has missing value, it returned true. We no longer do that, instead we set null to document and let each operator do its logic. So we don't need special case such as `$all` anymore. `unset` and `null` are equivalent for comparison. But there are exception for `$not`, `$elemMatch` and `$exists` operators.
* `$type` operator if the filter is not `null` we let the operator handle errors, compat test caught that. If filter is `null` we return immediately because null and unset are not the same for `$type`
* For sorting `null` and missing field is equivalent. Compat caught that. 

And added a new test case `MultipleQueriesSorted` because `QueriesSorted` failed after adding unset and I think the reason is deeper, not sure yet.

rumyantseva (approved):
LGTM

w84thesun (approved):
Thanks for the extended explanation, it's made it easier to understand what's going on here 😃

PR #1646, opened 2022-12-14
Remove Incomparable

### LINKED ISSUE #1522: Remove `types.Incomparable`
Currently, `types.Compare` could return four results: less, greater, equal, and "incomparable". The latter is used in two cases: when comparing NaNs, and when comparing documents. Both those cases should no longer exist after #1151 and #1521. Let's remove `types.Incomparable`.

Split `types.Compare` into that function and `types.CompareOrder`.

Later: Come up with better names? Think about future changes.

chilagrow:
# Description

Closes #1522 .

To remove Incomparable, I made following changes.

*CompareOrder has 3 function exposed*

1. Used by update for `$max` operator, the sort order is the same implementation of `CompareOrder`. This uses array as part of sort type.
2. `CompareOrderForSort` is used by `.sort()` where comparison of array does not compare array in sort type, but uses min element of array for ascending sort and max element of array for descending sort. Empty array is less than Null for this.
3. `CompareOrderForOperator` is used by operators `$gt` `$gte` `$lt` `$lte`, where array is not compared in sort type but uses the same bson type for min element of array for `$lt` and `$lte` and and max element of array for `$gt` and `$gte`

*Numbers are sorted the same*

* `int32`, `int64`, `float64` are no longer differentiated in the sorting, they are the same in the comparison. This matches the same behaviour on compat.
## Readiness checklist


internal/types/compareresult_string.go
```diff
@@ -11,26 +11,16 @@ func _() {
...
 	_ = x[Less - -1]
 	_ = x[Greater-1]
-	_ = x[Incomparable-127]
 }
 
-const (
-	_CompareResult_name_0 = "<==>"
-	_CompareResult_name_1 = "≹"
-)
+const _CompareResult_name = "<==>"
```
chilagrow, line 16: Changed by running `stringer -linecomment -type CompareResult`

internal/types/compare_order.go
```diff
@@ -138,37 +110,128 @@ func CompareOrder(a, b any, order SortType) CompareResult {
...
-func compareNumberOrder(a, b any, order SortType) CompareResult {
+// CompareOrderForSort detects the data type for two values and compares them.
+// If a or b is array, the minimum element of each array
+// is used for Ascending sort, and maximum element for
+// Descending sort. Hence, an array is used
+// for type comparison and value comparison.
+// Empty Array is smaller than Null.
+//
+// This is used by sort operation.
+func CompareOrderForSort(a, b any, order SortType) CompareResult {
```
chilagrow, line 124: Naming can be improved.

internal/types/compare_order.go
```diff
@@ -138,37 +110,128 @@ func CompareOrder(a, b any, order SortType) CompareResult {
...
+
+	// invert the result for descending sort.
+	return compareInvert(result)
+}
+
+// CompareOrderForOperator detects the data type for two values and compares them.
+// If a is an array, the array is filtered by the same type as
+// b type.
+// It is used by $gt, $gte, $lt and $lte comparison.
+func CompareOrderForOperator(a, b any, order SortType) CompareResult {
```
chilagrow, line 189: Naming can also be improved.

internal/types/compare_order.go
```diff
@@ -81,34 +80,6 @@ func detectDataType(value any) compareTypeOrderResult {
 	}
 }
 
-// numberOrderResult represents the comparison order of numbers.
```
chilagrow, line 84: This is removed because compat test tells me the numbers are treated the same.

internal/types/array_test.go
```diff
@@ -116,7 +116,7 @@ func TestArrayMinMax(t *testing.T) {
 		"IntMin": {
 			arr:           intArray,
 			minOrMax:      min,
-			expectedValue: int32(2),
+			expectedValue: int64(2),
```
chilagrow, line 119: Due to removing number ordering because compat test indicated the number order by different types, expected value is the first smallest one on array if there are multiple same value,.

w84thesun (commented):
Changes are looking reasonable, some pg tests are failing though.

internal/types/compare.go
```diff
@@ -33,10 +33,9 @@ type CompareResult int8
 // Values match results of comparison functions such as bytes.Compare.
 // They do not match MongoDB SortType values where 1 means ascending order and -1 means descending.
 const (
-	Equal        CompareResult = 0   // ==
-	Less         CompareResult = -1  // <
-	Greater      CompareResult = 1   // >
-	Incomparable CompareResult = 127 // ≹
```
rumyantseva, line 39: 👍

w84thesun (approved):
LGTM!

noisersup (approved):
:100:

PR #1650, opened 2022-12-14
Improve the way of storing data about collections

### LINKED ISSUE #1347: Improve the way of storing data about collections
### What should be done?

Currently, we store information about all the collections in the row of the settings table. To simplify code, improve performance, and have fewer locks per operation, we should store the list of collections in a separate table, where one collection is represented by one row. With that, when we want to create settings for a new collection or delete settings for an existing one, we won't need to request the whole list and don't have to lock everything.

### Implementation details

We should reuse the same approach as we already use when work with PostgreSQL in other places (for example, the function `pgdb.InsertDocument` to insert documents into the settings table. It might be possible that such functions need to be refactored and/or extended. For example, if `insert ... on conflict do nothing` is needed, we need to extend `pgdb.InsertDocument` to support it.

### Necessary refactoring

How we deal with the settings table needs to be refactored.
There should be only three functions in `settings.go` (the signatures are provided as a starting point and might be changed if necessary):

#### addSettingsIfNotExists - to be used when collection needs to be created
```go
// addSettingsIfNotExists returns settings (table name) for the given database and collection.
// If such settings don't exist, it creates them.
func addSettingsIfNotExists(ctx context.Context, tx pgx.Tx, db, collection string) (string, error) 
```

#### getSettings - to be used when collection needs to be queried
```go
// getSettings returns settings (table name) for the given database and collection.
// If such settings don't exist, it returns an error.
func getSettings(ctx context.Context, tx pgx.Tx, db, collection string) (string, error)
```

#### removeSettings - to be used when collection needs to be deleted
```go
// removeSettings removes settings for the given database and collection.
// If such settings don't exist, it doesn't return an error.
func removeSettings(ctx context.Context, tx pgx.Tx, db, collection string) error
```

The functions should be as simple as possible and mustn't return unnecessary errors. For example, `createSettingsIfNotExist` mustn't return `AlreadyExists` or `removeSettings` mustn't return `NotExists`. 

### Unique constraint 
The field of `_jsonb` column representing the collection name in the settings table should have a unique constraint.
With that, we can use something like `INSERT INTO settings(_jsonb) (_jsonb = '{"collection": "hello", ...}') ON CONFLICT (_jsonb->>'collection') DO NOTHING` and keep the code of `addSettingsIfNotExists` simple. 

#### Other places in the code to check

We expect that the implementation of this task will improve performance as currently, settings-related functions (e.g. `getSettingsTable`) are called in every single operator. Check the places where the issue number is mentioned.

In particular in the test `TestUpdateFieldCurrentDate` we have `maxDifference := time.Duration(4 * time.Minute)`. After implementing the current task we expect that `maxDifference` could be moved back to `3 * time.Minute` (or even lower). If it's not enough, create a separate issue to investigate this particular problem with tests performance.

rumyantseva:
# Description

Closes #1347.

⚠️ The changes in this PR break backward compatibility. Please dump the existing databases before rolling out the changes and restore them afterwards.

## Readiness checklist


internal/handlers/pg/pgdb/transactions.go
```diff
@@ -90,20 +90,20 @@ func (pgPool *Pool) InTransaction(ctx context.Context, f func(pgx.Tx) error) (er
...
+		var tcErr *transactionConflictError
 
 		switch {
 		case err == nil:
 			return nil
 		case errors.As(err, &tcErr):
 			ctxutil.Sleep(ctx, delay)
 		default:
-			return err
+			return lazyerrors.Error(err)
```
rumyantseva, line 104: Added more `lazyErrors` to have more debug info, for example:

```
(InternalError) [transactions.go:108 pgdb.(*Pool).InTransactionRetry] [transactions.go:77 pgdb.(*Pool).InTransaction] [msg_create.go:97 pg.(*Handler).MsgCreate.func1] [collections.go:129 pgdb.CreateCollection] [settings.go:91 pgdb.upsertSettings] ERROR: deadlock detected (SQLSTATE 40P01)```

this is helpful as one can see the full chain of calls.

File: internal/handlers/pg/pgdb/transactions.go [hunk 2]
```diff
@@ -26,10 +26,10 @@ import (
 )
 
 // maxRetries is the maximum number of times to retry a transaction.
-const maxRetries = 3
+const maxRetries = 20
```
rumyantseva, line 29: The tricky part: this number of retries is enough for our tests (e.g. on my local machine there are ~66 parallel collections created), but if there are 100 or 1000 collections that are created in parallel, this number won't be enough.

internal/handlers/pg/pgdb/settings.go
```diff
@@ -42,193 +38,120 @@ const (
...
+//
+// It makes a document with _id and table fields and stores it in the settingsTableName table.
+// The given FerretDB collection name is stored in the _id field,
+// the corresponding PostgreSQL table name is stored in the table field.
+// For _id field it creates unique index.
+//
+// It returns a possibly wrapped error:
+//   - ErrInvalidDatabaseName - if the given database name doesn't conform to restrictions.
+//   - *transactionConflictError - if a PostgreSQL conflict occurs (the caller could retry the transaction).
+func ensureSettings(ctx context.Context, tx pgx.Tx, db, collection string) (tableName string, created bool, err error) {
```
rumyantseva, line 53: A potential alternative (proposed by Patryk) is to user the word "metadata" instead of "settings". I like that idea.

internal/handlers/pg/pgdb/transactions.go
```diff
@@ -26,18 +26,18 @@ import (
...
 
 // delay is the amount of time to wait before retrying a transaction.
-const delay = 10 * time.Millisecond
+const delay = 3 * time.Millisecond
```
chilagrow, line 32: Any particular reason we like 3ms?

internal/handlers/pg/pgdb/transactions.go
chilagrow: probably we will need a bit smarter retry strategy like exponential back off if we find 20 is too small.

internal/handlers/pg/pgdb/settings_test.go
```diff
@@ -0,0 +1,62 @@
...
+	pool := getPool(ctx, t)
+	databaseName := testutil.DatabaseName(t)
+	collectionName := testutil.CollectionName(t)
+	setupDatabase(ctx, t, pool, databaseName)
+
+	err := pool.InTransactionRetry(ctx, func(tx pgx.Tx) error {
+		nameCreated, _, err := ensureSettings(ctx, tx, databaseName, collectionName)
+		if err != nil {
+			return err
+		}
```
chilagrow, line 41: why can't we require no error here? :thinking: 
```suggestion
		require.NoError(t, err)
```

chilagrow (commented):
Some early comments :hugs: Looking good!

internal/handlers/pg/pgdb/collections.go
```diff
@@ -37,162 +38,136 @@ var validateCollectionNameRe = regexp.MustCompile("^[a-zA-Z_-][a-zA-Z0-9_-]{0,11
...
-//   - ErrTableNotExist - is the required FerretDB database does not exist.
+// CreateCollection creates a new FerretDB collection with the given name in the given database.
+// If the database does not exist, it will be created.
 //
-// Please use errors.Is to check the error.
+// It returns possibly wrapped error:
+//   - ErrInvalidDatabaseName - if the given database name doesn't conform to restrictions.
+//   - ErrInvalidCollectionName - if the given collection name doesn't conform to restrictions.
+//   - ErrAlreadyExist - if a FerretDB collection with the given name already exists.
+//   - *transactionConflictError - if a PostgreSQL conflict occurs (the caller could retry the transaction).
```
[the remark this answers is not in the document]
rumyantseva: There is only once case for it - `msg_create`.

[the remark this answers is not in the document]
chilagrow: Thanks I see, and that's when we return _collection already exist_ to the user :+1:

internal/handlers/pg/msg_update.go
```diff
@@ -67,18 +66,17 @@ func (h *Handler) MsgUpdate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 		return nil, err
 	}
 
-	created, err := pgdb.CreateCollectionIfNotExist(ctx, h.PgPool, sp.DB, sp.Collection)
+	err = h.PgPool.InTransactionRetry(ctx, func(tx pgx.Tx) error {
+		return pgdb.CreateCollectionIfNotExists(ctx, tx, sp.DB, sp.Collection)
+	})
 	if err != nil {
-		if errors.Is(pgdb.ErrInvalidTableName, err) ||
+		if errors.Is(pgdb.ErrInvalidCollectionName, err) ||
```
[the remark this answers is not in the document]
rumyantseva: It's not really possible to test it in FerretDB because we have custom rules for collection names (different than what MongoDB has), see how `ErrInvalidCollectionName` is used in `pgdb`.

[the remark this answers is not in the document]
rumyantseva: Or maybe we can add a case that could be invalid for both MongoDB and us. I think we can even add it to the list of collections when setup tests. I'll create an issue.

CHANGELOG.md
```diff
@@ -4,14 +4,22 @@
 
 ### What's Changed
```
AlekSi, line 5: FWIW, I don't think we should go into such details in user-visible changelog. Basically, we should say: "we changed storage internals; mongodump/mongoexport your data before upgrade, mongorestore/mongoimport after upgrade".

But that's not important right now, I will update changelog before release anyway.

rumyantseva: Thanks! I'm ok with either option. I was trying to follow a similar structure from one of the past release notes where we also gave quite a lot of context, but it it's not necessary, feel free to cut it.

internal/handlers/pg/pgdb/settings_test.go
rumyantseva: In the case of tests running in parallel, there could be (and there is) a situation when this code returns an error. It's, for example, when to test reach `ensureTest` and attempt to create indexes for the settings table in parallel. This is why we use `InTransactionRetry` here.

internal/handlers/pg/pgdb/transactions.go
rumyantseva: I was trying to come up with a small number (so we don't have big delays), and from the test logs, it looks to me that we don't need 10ms, we can choose a smaller number. Maybe even smaller than 3ms though...

chilagrow: For now I think 3ms is fine. Probably we will make change on this once we get more understanding by https://github.com/FerretDB/FerretDB/issues/1686 :+1:

internal/handlers/pg/pgdb/settings_test.go
chilagrow: Got it. Hence we are testing it within `InTransactionRetry` and checking that error instead of this error :pray:

internal/handlers/pg/pgdb/settings.go
```diff
@@ -42,193 +38,120 @@ const (
...
-	sql := fmt.Sprintf(`SELECT settings FROM %s`, pgx.Identifier{db, settingsTableName}.Sanitize())
-
-	if lock {
-		sql += " FOR UPDATE"
-	}
-
-	rows, err := tx.Query(ctx, sql)
+// getSettings returns PostgreSQL table name for the given FerretDB database and collection.
+//
+// If such settings don't exist, it returns a possibly wrapped ErrTableNotExist.
```
chilagrow, line 110:
```suggestion
// If such settings don't exist, it returns ErrTableNotExist.
```

internal/handlers/pg/msg_update.go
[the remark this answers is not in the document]
rumyantseva: https://github.com/FerretDB/FerretDB/issues/1688

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -0,0 +1,50 @@
...
+
+	sql := `CREATE` + unique + ` INDEX IF NOT EXISTS ` + pgx.Identifier{p.table + "_id"}.Sanitize() +
+		` ON ` + pgx.Identifier{p.schema, p.table}.Sanitize() +
+		` ((_jsonb->>'_id'))`
+
+	if _, err = tx.Exec(ctx, sql); err == nil {
+		return nil
+	}
+
+	return lazyerrors.Error(err)
```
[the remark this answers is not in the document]
rumyantseva: Well, my idea was that if we need to handle some `pgErrors` in the future, it would be easier. E.g., see `func insert`.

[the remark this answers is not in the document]
rumyantseva: fixed

internal/handlers/pg/pgdb/settings_test.go
rumyantseva: added a comment

internal/handlers/pg/pgdb/settings.go
rumyantseva: fixed

internal/handlers/pg/msg_update.go
noisersup, line 73: I know it was already there, but:
```
func errors.Is(err error, target error) bool

Is reports whether any error in err's chain matches target.
...
```

Shouldn't we use a different order of arguments here?

internal/handlers/pg/pgdb/database_metadata.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+	"github.com/FerretDB/FerretDB/internal/util/must"
+)
+
+const (
+	// Reserved prefix for database and collection names.
+	reservedPrefix = "_ferretdb_"
+
+	// Database metadata table name.
+	dbMetadataTableName = reservedPrefix + "database_metadata"
```
noisersup, line 35: cool!

noisersup (commented):
That's a lot of work! One question + there's still one little reference to "settings" inside the code.
PTAL: internal/handlers/pg/pgdb/pool.go:125

rumyantseva:
@noisersup pool.go:125 is not about settings table, it's about settings to connect to pg, so I believe it's fine to call it settings :)

internal/handlers/pg/msg_update.go
rumyantseva: hehehe, nice catch! (and another reason why we need to cover such things with tests)

rumyantseva: fixed

noisersup:
You right, thanks for checking it out!

w84thesun (approved):
I'm still not sure about the function names but couldn't think of something better

noisersup (approved):
LGTM!

rumyantseva:
@AlekSi your review is needed because of changelog. Feel free to suggest the text you'd like to see for the whole changelog, or we can also merge it as is and correct changelog afterwards.

AlekSi (approved):

PR #1670, opened 2022-12-21
Add `InTransactionRetry` helper

### LINKED ISSUE #1618: Add `InTransactionRetry` helper
### Final Design Deicsion

- [x] Retry parameters `maxRetries`, `delay` are constants
- [x] Context should only be checked in `inTransaction` (no need to check it in `InTransactioRetry`)
- [x] `ctxutil.Sleep(ctx, delay)` should be used for delays between attempts
- [x] `transactionConflictError` should wrap the original error
- [ ] ~`lazyerror` should be used for errors that are not expected / couldn't be processed~ `inTransaction` already wraps everything in `lazyError`.

---

### Initital Proposal (draft, might not compile)

Implement  `InTransactionRetry` function that would use `InTransaction` with the given retry in case the transaction failed.

**Details:** In principle, we need to retry the transaction only if there is a `pgerrcode.UniqueViolation`, `pgerrcode.DuplicateObject`, or a similar error. 

We can define a special error to be returned in such cases. For example, an error could be something like:

```go
// errTransactionConflict is returned when one of the queries in the transaction returned an error because
// of an unexpected conflict. The caller could retry such a transaction.
var errTransactionConflict = fmt.Errorf("transaction can't continue due to a conflict")
```

And then `InTransactionRetry` would be something like:

```go
func (pgPool *Pool) InTransactionRetry(ctx context.Context, f func(pgx.Tx) error, maxRetries int, delay time.Duration) (err error) {
	for retry := 0; retry < maxRetries; retry++ {

		// TODO: Maybe check if context is done  

		err := pgPool.InTransaction(ctx, f)

		switch {
		case err == nil:
			return nil
		case errors.Is(err, errTransactionConflict):
			time.Sleep(delay)
		default:
			return err
		}
	}
}
```

With that, the places where we check for pg errors could be like:

```go
	switch pgErr.Code {
	case pgerrcode.UniqueViolation, pgerrcode.DuplicateObject:
		// https://www.postgresql.org/message-id/CA+TgmoZAdYVtwBfp1FL2sMZbiHCWT4UPrzRLNnX1Nb30Ku3-gg@mail.gmail.com
		return errTransactionConflict
	default:
		return lazyerrors.Error(err)
	}
```

### Definition of done

- [x] Implement `InTransactionRetry`
- [x] Modify at least one handler (for example, `msg_insert`) to use `InTransactionRetry` in case of a pg conflict for `CreateCollectionIfNotExist`
- [x] Find places where we should use `InTransactionRetry` and create issue for them

rumyantseva:
# Description

Closes #1618.

## Readiness checklist

### DIFF
--- a/internal/handlers/pg/pgdb/pool.go
+++ b/internal/handlers/pg/pgdb/pool.go
@@ -24,7 +24,6 @@ import (
 	"github.com/jackc/pgx/v4/pgxpool"
 	"go.uber.org/zap"
 
-	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
 	"github.com/FerretDB/FerretDB/internal/util/state"
 )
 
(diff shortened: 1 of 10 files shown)


internal/handlers/pg/pgdb/pool.go
```diff
@@ -184,40 +183,3 @@ func (pgPool *Pool) SchemaStats(ctx context.Context, schema, collection string)
 		Name: schema,
 	}, nil
 }
-
-// InTransaction wraps the given function f in a transaction.
-// If f returns an error, the transaction is rolled back.
-// Errors are wrapped with lazyerrors.Error,
-// so the caller needs to use errors.Is to check the error,
-// for example, errors.Is(err, ErrSchemaNotExist).
-func (pgPool *Pool) InTransaction(ctx context.Context, f func(pgx.Tx) error) (err error) {
```
rumyantseva, line 193: Moved to `transactions.go`

noisersup (approved):
Finally resolved the parallel creates issue! :rocket:

w84thesun (approved):
LGTM!

Issue #200, opened 2021-12-21
Handle unordered `insert` with implicit transaction


AlekSi:
When inserting documents with `ordered` set to `true` (which should be set as a default value), insert documents one-by-one in seperate transactions, skip invalid documents (for example, with _id values that are already present) and return errors just for them.

With `ordered` set to false, attempt to insert all documents in one transaction.

See https://docs.mongodb.com/manual/reference/method/db.collection.insertMany/

ekalinin:
Please, correct me if I'm wrong, but it looks like MongoDB works oppositely at the moment.

From [insert's doc](https://docs.mongodb.com/manual/reference/method/db.collection.insert/#mongodb-method-db.collection.insert):

> If true, perform an ordered insert of the documents in the array, and if an error occurs with one of documents, MongoDB will return without processing the remaining documents in the array.
> If false, perform an unordered insert, and if an error occurs with one of documents, continue processing the remaining documents in the array.

From [insertMany's doc](https://docs.mongodb.com/manual/reference/method/db.collection.insertMany/#error-handling):

> Excluding Write Concern errors, ordered operations stop after an error, while unordered operations continue to process any remaining write operations in the queue.

Or did I miss something?

AlekSi:
The documentation is a bit confusing. I will write an integration test to see how MongoDB behaves.

AlekSi:
I added tests: https://github.com/FerretDB/dance/blob/main/tests/ferret/core_test.go In short, an unordered insert saves all valid documents, while an ordered insert stops on the first error while still saving previous documents. Both cases do not fit all-or-nothing PostgreSQL transactions.

In theory, we could optimize the performance of the happy path: insert all documents in a single implicit transaction; in case of error, retry inserts one by one. But that sounds weird.

@pboros WDYT?

AlekSi:
It looks like we could use [`ON CONFLICT`](https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT) `INSERT` clause to ignore some errors like duplicate IDs, but that would not work for some other errors (but what are they?)

noisersup:
As agreed on a call this one is moved to #940.

PR #1673, opened 2022-12-22
Support `ordered` argument for `insert` command

### LINKED ISSUE #940: Support `ordered` `insert`s

noisersup:
# Description

Closes #940.

## Readiness checklist


website/docs/reference/supported_commands.md
```diff
@@ -71,7 +71,7 @@ sidebar_position: 1
 |                   | `comment`                  | ⚠️  | Ignored in Tigris                                         |
 |                   | `let`                      | ⚠️  | Unimplemented                                             |
 |                   | `q`                        | ✅  |                                                           |
-|                   | `u`                        | ✅  | TODO check if u is an array of aggregation pipeline stages|
+|                   | `u`                        | ✅  | TODO check if u is an array of aggregation pipeline stages |
```
rumyantseva, line 74: fmt did it :)

integration/insert_compat_test.go
```diff
@@ -112,16 +117,54 @@ func TestInsertCompat(t *testing.T) {
 
 	testCases := map[string]insertCompatTestCase{
 		"InsertEmptyDocument": {
-			insert: bson.D{},
+			insert: []any{bson.D{}},
+			// TODO: this test fails with last byte being different as the value is always unique,
+			// see https://github.com/mongodb/mongo-go-driver/blob/e1bf8858dd9ba111e880f8e5a8b7e7b7da64cb54/bson/primitive/objectid.go#L51
+			// Target: &mongo.InsertManyResult{InsertedIDs:[]interface {}{primitive.ObjectID{0x63, 0xac, 0x15, 0x9e, 0xe6, 0x49, 0x0, 0x8, 0x7, 0x15, 0x28, 0x9a}}}
+			// Compat: &mongo.InsertManyResult{InsertedIDs:[]interface {}{primitive.ObjectID{0x63, 0xac, 0x15, 0x9e, 0xe6, 0x49, 0x0, 0x8, 0x7, 0x15, 0x28, 0x9b}}}
```
rumyantseva, line 124: @AlekSi this is the test case I mentioned (the one where `_id` for target and compat are different because the driver inserts a unique id every time). I suggest we just shouldn't compare if `_id`s are the same in this case.

integration/insert_compat_test.go
```diff
@@ -68,30 +74,43 @@ func testInsertCompat(t *testing.T, testCases map[string]insertCompatTestCase) {
...
-					targetCursor.Decode(&targetFindRes)
+					err = targetCursor.All(ctx, &targetFindRes)
+					require.NoError(t, err)
 
-					compatCursor, err := compatCollection.Find(ctx, bson.D{{}})
+					compatCursor, err := compatCollection.Find(ctx, bson.D{{}}, findOpts)
 					require.NoError(t, err)
 					defer compatCursor.Close(ctx)
-					compatCursor.Decode(&targetFindRes)
+					err = compatCursor.All(ctx, &compatFindRes)
```
rumyantseva, line 104: I added some more logic to find and compare data as there were not many actual checks in this test before.

integration/insert_run_command_compat_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.mongodb.org/mongo-driver/bson"
+
+	"github.com/FerretDB/FerretDB/integration/setup"
+)
+
+type insertRunCommandCompatTestCase struct {
```
rumyantseva, line 28: I'm not sure if `insertRunCommand` is a good name, but I didn't find a better one, so if you have a better name, please suggest it. The idea is this test uses `runCommand` and not `insertMany`, so more weird cases like empty document or invalid parameters  can be added.

internal/handlers/pg/msg_insert.go
```diff
@@ -62,34 +62,27 @@ func (h *Handler) MsgInsert(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 		return nil, err
 	}
 
-	var inserted int32
+	ordered := true
+	if ordered, err = common.GetOptionalParam(document, "ordered", ordered); err != nil {
```
rumyantseva, line 66: Test for this is added but skipped (I'll cover it in the same issue but as a separate PR to not make this PR too big).

Fashander (approved):

integration/insert_compat_test.go
```diff
@@ -111,17 +124,48 @@ func TestInsertCompat(t *testing.T) {
...
+			},
+			ordered:    false,
+			resultType: emptyResult,
+		},
+
+		"InsertOrderedOneError": {
+			insert: []any{
+				bson.D{{"_id", "1"}},
+				bson.D{{"_id", primitive.Regex{Pattern: "^regex$", Options: "i"}}},
+				bson.D{{"_id", "2"}},
```
noisersup, line 157: Okay, so here, if I understand correctly, both databases will:
1) insert _id: 1
2) return the error and don't rollback the first insert

and by the end our collection will look like: `[ {_id:1} ]` right?

AlekSi: See https://github.com/FerretDB/FerretDB/issues/200#issuecomment-1003748090 and the whole issue in general

integration/insert_compat_test.go
```diff
@@ -68,30 +72,39 @@ func testInsertCompat(t *testing.T, testCases map[string]insertCompatTestCase) {
...
-								setup.SkipForTigrisWithReason(t, targetErr.Error())
-							}
+						var e mongo.BulkWriteException
+						if errors.As(targetErr, &e) &&
+							e.HasErrorCodeWithMessage(121, "json schema validation failed for field") {
+							setup.SkipForTigrisWithReason(t, targetErr.Error())
 						}
 
-						assert.Equal(t, compatErr, targetErr)
+						assert.Equal(t, compatErr.Error(), targetErr.Error())
```
AlekSi, line 81: Why that change? Why do we compare strings instead of error values?

integration/insert_compat_test.go
```diff
@@ -23,12 +23,14 @@ import (
...
 	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
 
 	"github.com/FerretDB/FerretDB/integration/setup"
 )
 
 type insertCompatTestCase struct {
-	insert     bson.D                   // required
+	insert     []any                    // required, slice of bson.D to be insert
```
noisersup, line 32: If I understand, we use `[]any` here, and not `[]bson.D` because `InsertMany` function wants the `[]interface{}` as an argument. Then shouldn't we use `[]interface{}` here?  If I'm reasoning correctly, using `any` statement here mixes generics with interfaces

integration/insert_compat_test.go
rumyantseva: @noisersup yes, correct. This is how MongoDB works (I checked manually, and the compat tests prove it).

integration/insert_compat_test.go
```diff
@@ -55,10 +57,12 @@ func testInsertCompat(t *testing.T, testCases map[string]insertCompatTestCase) {
...
 
-					targetInsertRes, targetErr := targetCollection.InsertOne(ctx, insert)
-					compatInsertRes, compatErr := compatCollection.InsertOne(ctx, insert)
+					opts := options.InsertManyOptions{Ordered: &tc.ordered}
+					targetInsertRes, targetErr := targetCollection.InsertMany(ctx, insert, &opts)
+					compatInsertRes, compatErr := compatCollection.InsertMany(ctx, insert, &opts)
 
-					if targetInsertRes != nil || compatInsertRes != nil {
+					// If the result contains inserted ids, we consider the result non-empty.
+					if compatInsertRes != nil && len(compatInsertRes.InsertedIDs) > 0 {
```
AlekSi, line 65: We should check `targetInsertRes` too

AlekSi (dismissed):

integration/insert_compat_test.go
AlekSi: LGTM overall, but that change looks sus

integration/insert_compat_test.go
```diff
@@ -23,12 +23,14 @@ import (
...
 	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
 
 	"github.com/FerretDB/FerretDB/integration/setup"
 )
 
 type insertCompatTestCase struct {
-	insert     bson.D                   // required
+	insert     []any                    // required, slice of bson.D to be insert
+	ordered    bool                     // defaults to false
```
noisersup, line 33: As we use this struct for other insert test cases that are not `ordered` related, we could also make this a `*bool` and handle nil values as `true`. That's only suggestion and it's non-critical, though currently, because of defaulting to false, we run all other test cases as unordered and it could be misleading in the future, while adding other `insertMany` test cases

noisersup (commented):
Good job with making it work with tigris! :rocket:

Sorry for late review, but I spent more time on it than I thought. I realised that this PR removes completely the `'n'` field from the response and replaces it with `"nInserted"`. I was a little bit confused about that. Is nInserted field returned on the error or on every query? I've ran the mongosh against mongodb and this branch (with `task run-proxy`) and used this command based on one of test cases:
```
test> db.runCommand({"insert": "test","documents": [{"_id": "1"},{"_id": {$regex: "^regex$", $options: "i"}},{"_id": "2"}]})
{
  n: 1,
  writeErrors: [
    {
      index: 1,
      code: 52,
      errmsg: "_id fields may not contain '$'-prefixed fields: $regex is not valid for storage."
    }
  ],
  ok: 1
}
```

And that's the diff output:
```
--- res header
+++ proxy header
@@ -1 +1 @@
-length:   166, id:    8, response_to:   23, opcode: OP_MSG
+length:   185, id: 1415, response_to:   23, opcode: OP_MSG

Body diff:
--- res body
+++ proxy body
@@ -7,9 +7,7 @@
         "$k": [
-          "ok",
+          "n",
           "writeErrors",
-          "nInserted"
+          "ok"
         ],
-        "ok": {
-          "$f": 1
-        },
+        "n": 1,
         "writeErrors": [
@@ -22,7 +20,9 @@
             "index": 1,
-            "code": 2,
-            "errmsg": "invalid key: \"$regex\" (key must not contain '$' sign)"
+            "code": 52,
+            "errmsg": "_id fields may not contain '$'-prefixed fields: $regex is not valid for storage."
           }
         ],
-        "nInserted": 1
+        "ok": {
+          "$f": 1
+        }
       },
```

It seems to me that there are differences in output, and maybe tests don't check for that. Or I'm just missing something

Apart from that left a couple of comments

integration/insert_compat_test.go
AlekSi: `any` is an alias for `interface{}` since 1.18. It is not about generics. We should not use `interface{}` anywhere

noisersup: I understand the concept, though I remember we had similar usage of `any` on another issue and we said that we shouldn't do that, but maybe I misunderstood something. If that's correct usage then I'll remember that

integration/insert_compat_test.go
rumyantseva: @AlekSi that was a bad idea indeed! Fixing... (not only this thing but also the problems it hided)

integration/insert_compat_test.go
rumyantseva: Yeah, I was thinking about it, I like that suggestion :)

integration/insert_compat_test.go
rumyantseva: fixed

integration/insert_compat_test.go
rumyantseva: fixed

rumyantseva:
@noisersup nice catch with missing `n`! I'm not sure where I found `nInserted` 😅 I fixed it, so the PR is ready for another review round.

integration/insert_compat_test.go
rumyantseva: fixed

noisersup (commented):
:shipit:

rumyantseva (dismissed):
LGTM

integration/insert_compat_test.go
```diff
@@ -18,17 +18,20 @@ import (
...
 	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
 
 	"github.com/FerretDB/FerretDB/integration/setup"
 )
 
 type insertCompatTestCase struct {
-	insert     bson.D                   // required
+	insert     []any                    // required, slice of bson.D to be insert
+	ordered    *bool                    // defaults to true
```
AlekSi, line 34: FWIW, I think we should not use `*bool` that has three values if we use only two. The simplest solution could be `ordered bool // defaults to false` and `options.InsertMany().SetOrdered(tc.ordered)`

integration/insert_compat_test.go
AlekSi: We could simplify it with `FindAll` helper

integration/insert_compat_test.go
noisersup: Just in case this change was made after this [comment](https://github.com/FerretDB/FerretDB/pull/1673/files/aaf4b39f782bf7bb09a4a3bfe9ffb1eff31a900f#r1058492595). But yeah if we prefer simpler approach then let's do this!

rumyantseva: fixed

rumyantseva (approved):
:shipit:

AlekSi (approved):

integration/insert_compat_test.go
rumyantseva: (fixed)

PR #1683, opened 2022-12-22
Implement `iterator.Interface` for `types.Document` and `types.Array`

### LINKED ISSUE #1655: Implement `iterator.Interface` for `types.Array` and `types.Document`
### What should be done?

We already implement `iterator.Interface` where we query data. We now should use that interface where we iterate over queried data, replacing (in the future) many uses of `types.Document.Keys / Values`, etc.

`Next` method should return the next field name (for documents) or index (for arrays) and the value.
`Close` method should do nothing.

Update `MsgGetParameter` handlers (both `pg` and `tigris`) and called functions (`selectParam`, `selectUnit`) to use this interface. But do not refactor them beyond the scope of the task.

Unit tests are in the scope.
No changes in integration tests are expected.

### Where?

* https://github.com/FerretDB/FerretDB/blob/main/internal/types/array.go
* https://github.com/FerretDB/FerretDB/blob/main/internal/types/array_test.go
* https://github.com/FerretDB/FerretDB/blob/main/internal/types/document.go
* https://github.com/FerretDB/FerretDB/blob/main/internal/types/document_test.go
* https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/pg/msg_getparameter.go
* https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/tigris/msg_getparameter.go

w84thesun:
# Description

Closes #1655.

## Readiness checklist

* [x] I added unit tests for new functionality or bugfixes.
* [ ] I added integration tests for new functionality or bugfixes.
* [ ] I added compatibility tests for new functionality or bugfixes.
* [x] I made spot refactoring.
* [ ] I updated the documentation.


internal/types/document.go
```diff
@@ -428,6 +430,38 @@ func (d *Document) moveIDToTheFirstIndex() {
 	d.fields = slices.Delete(d.fields, idIdx+1, idIdx+2)
 }
 
+// Iterator returns an iterator over the document fields.
+func (d *Document) Iterator() iterator.Interface[string, any] {
```
rumyantseva, line 434: I'd move it to a separate file, smth like `document_iterator` to define a smaller context (this file is already quite long)

internal/types/document.go
```diff
@@ -428,6 +430,38 @@ func (d *Document) moveIDToTheFirstIndex() {
...
+	return newDocumentIterator(d)
+}
+
+// documentIterator represents an iterator over the document fields.
+type documentIterator struct {
+	doc         *Document
+	currentIter atomic.Uint32
```
rumyantseva, line 441: Do we need `currentInter` at all? If we work with field names, I don't think we need this number.

internal/handlers/tigris/msg_getparameter.go
```diff
@@ -68,17 +70,18 @@ func (h *Handler) MsgGetParameter(ctx context.Context, msg *wire.OpMsg) (*wire.O
...
+func selectParam(_, resDB *types.Document) (doc *types.Document, err error) {
 	doc = must.NotFail(types.NewDocument())
-	keys := document.Keys()
 
-	for _, k := range keys {
-		if k == "getParameter" || k == "comment" || k == "$db" {
-			continue
+	iter := resDB.Iterator()
+
+	for {
```
rumyantseva, line 78: I'm skeptical about `doc.Set(k, item)` being a part of this cycle. If there is a duplicate in keys, it'll panic.
Do we consider that duplicates are not possible in such a case, or should we use something else instead of `doc.Set`?

internal/types/document_test.go
```diff
@@ -161,4 +162,57 @@ func TestDocument(t *testing.T) {
...
+				document: must.NotFail(NewDocument("foo", "bar")),
+				expected: []fields{{k: "foo", v: "bar"}},
+			},
+			"two": {
+				document: must.NotFail(NewDocument("foo", "bar", "baz", "qux")),
+				expected: []fields{{k: "foo", v: "bar"}, {k: "baz", v: "qux"}},
+			},
+			"three": {
+				document: must.NotFail(NewDocument("foo", "bar", "baz", "qux", "quux", "quuz")),
+				expected: []fields{{k: "foo", v: "bar"}, {k: "baz", v: "qux"}, {k: "quux", v: "quuz"}},
```
rumyantseva, line 192: I'd suggest having at lease one duplicate key, as this is one of the reasons why we want to have iterator, something like:

```suggestion
				document: must.NotFail(NewDocument("foo", "bar", "baz", "qux", "foo", "quuz")),
				expected: []fields{{k: "foo", v: "bar"}, {k: "baz", v: "qux"}, {k: "foo", v: "quuz"}},
```

rumyantseva (commented):
Looks good to me overall, and I like that we test different possible situations!

Added one suggestion for the test and one question about the logic.

internal/types/document.go
w84thesun: Done.

internal/handlers/tigris/msg_getparameter.go
w84thesun: We iterate over the predefined document here and in `pg` handler, so I think it's safe to use `doc.Set` here

rumyantseva: Ok, sounds good to me!

rumyantseva (approved):
LGTM

noisersup (approved):
Cool! It was really straightforward to understand the code! I have 2 non-critical comments, but I think they could be done in seperate refactoring PR if needed

PR #1723, opened 2022-12-29
Add stub for getting client-specific connection

AlekSi:
# Description

To make the next PR smaller.

## Readiness checklist

### DIFF
--- a/internal/handlers/pg/msg_insert.go
+++ b/internal/handlers/pg/msg_insert.go
@@ -31,6 +31,11 @@ import (
 
 // MsgInsert implements HandlerInterface.
 func (h *Handler) MsgInsert(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
+	dbPool, err := h.DBPool(ctx)
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+
 	document, err := msg.Document()
 	if err != nil {
 		return nil, lazyerrors.Error(err)
@@ -67,7 +72,7 @@ func (h *Handler) MsgInsert(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, 
 		return nil, err
 	}
 
-	inserted, insErrors := h.insertMany(ctx, &sp, docs, ordered)
+	inserted, insErrors := h.insertMany(ctx, dbPool, &sp, docs, ordered)
 
 	replyDoc := must.NotFail(types.NewDocument(
 		"ok", float64(1),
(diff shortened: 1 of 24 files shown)


internal/handlers/pg/msg_insert.go
```diff
@@ -128,7 +132,7 @@ func (h *Handler) insertMany(ctx context.Context, sp *pgdb.SQLParam, docs *types
 }
 
 // insert prepares and executes actual INSERT request to Postgres.
-func (h *Handler) insert(ctx context.Context, sp *pgdb.SQLParam, doc any) error {
+func (h *Handler) insert(ctx context.Context, dbPool *pgdb.Pool, sp *pgdb.SQLParam, doc any) error {
```
rumyantseva, line 135: If we pass `dbPool` as param, should method `func (h *Handler) insert` be just `func insert` then?

internal/handlers/pg/pg.go
```diff
@@ -41,13 +48,45 @@ type NewOpts struct {
...
+
+	for k, pgPool := range h.pools {
+		pgPool.Close()
+		delete(h.pools, k)
+	}
+}
+
+// DBPool returns database connection pool for the given client connection.
+func (h *Handler) DBPool(ctx context.Context) (*pgdb.Pool, error) {
+	// TODO make real implementation; the current one is a stub
```
rumyantseva, line 69: I understand that this is not for the current PR, as the current PR is a stub.
As no issue is mentioned, I expect it to be fixed in the next PR (that we start passing auth data from the client).

internal/handlers/pg/pg.go
```diff
@@ -16,22 +16,29 @@
...
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
 	"github.com/FerretDB/FerretDB/internal/util/state"
 )
 
 // Handler implements handlers.Interface on top of PostgreSQL.
 type Handler struct {
 	*NewOpts
+
+	rw    sync.RWMutex
+	pools map[string]*pgdb.Pool
```
rumyantseva, line 36: The idea is that none of `Msg` methods should request this map directly, they should always call `h.DBPool(ctx)` instead. However, in principle such call is possible. Should we leave a comment that it is expected to use `h.DBPool`, but not this map + mutex directly? Or is it obvious?

rumyantseva (commented):
Looks good to me overall, left a couple of comments/questions.

internal/handlers/pg/pg.go
AlekSi: #1725

internal/handlers/pg/pg.go
AlekSi: I will add a comment in #1725, but someone should _really_ try to use that map…

internal/handlers/pg/msg_insert.go
AlekSi: Good point, will do in the next PR

rumyantseva (approved):
LGTM (with some expected additions in the next PR).

:shipit:

noisersup:
I was trying to review it _before merge_, but I don't understand the point of this. By client-specific we mean the connections from FerretDB to PostgreSQL, right? Do we have plans to support multiple postgresql connections in parallel here? Sorry if the answer is obvious but I feel lost here.

AlekSi:
@noisersup #1725 might make it clear

PR #1738, opened 2023-01-03
Report availability of newer versions

### LINKED ISSUE #1443: Report availability of newer versions
### What should be done?

Currently, if telemetry is enabled and a newer version of FerretDB is available, we log that information; but we do not expose it via `startupWarnings`. We should, so it could be visible to clients connecting with `mongosh` and other clients.

noisersup:
# Description

Closes #1443.

## Readiness checklist

### DIFF
--- a/internal/handlers/pg/msg_getlog.go
+++ b/internal/handlers/pg/msg_getlog.go
@@ -99,6 +99,10 @@ func (h *Handler) MsgGetLog(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, 
 				"The telemetry state is undecided; the first report will be sent soon. "+
 					"Read more about FerretDB telemetry and how to opt out at https://beacon.ferretdb.io.",
 			)
+		} else {
+			if state.LatestVersion != "" {
+
+			}
 		}
 
 		var log types.Array


internal/handlers/pg/msg_getlog.go
```diff
@@ -93,12 +93,19 @@ func (h *Handler) MsgGetLog(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 
 		// TODO https://github.com/FerretDB/FerretDB/issues/1443
```
rumyantseva, line 94: I see the review is requested, but the state of the PR is draft. I guess it's not ready yet :)

noisersup: Yeah, it was ready for review but I've started thinking about the definition of done (how should I modify tests and what with metrics), though I can see that you've spotted that I forgot removing TODOs, so thanks!

noisersup: As agreed on the meeting - let's skip tests subject for now, unless someone has some easy solution to make them

internal/util/state/metrics.go
```diff
@@ -66,6 +66,9 @@ func (mc *metricsCollector) Collect(ch chan<- prometheus.Metric) {
 		constLabels["telemetry"] = "undecided"
 	case *s.Telemetry:
 		constLabels["telemetry"] = "enabled"
+		if s.LatestVersion != v.Version {
+			constLabels["latest_version_available"] = s.LatestVersion
```
rumyantseva, line 70: Ideally, that would be nice (and seems to be possible) to cover it with a unit test.
However, with out current tests I don't see a simple way to do it, so let's create a separate issue to discuss if/how/what we can test here.

internal/handlers/pg/msg_getlog.go
```diff
@@ -91,14 +91,20 @@ func (h *Handler) MsgGetLog(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+		case state.Telemetry == nil:
 			startupWarnings = append(
 				startupWarnings,
 				"The telemetry state is undecided; the first report will be sent soon. "+
 					"Read more about FerretDB telemetry and how to opt out at https://beacon.ferretdb.io.",
 			)
+		case state.LatestVersion != mv.Version:
+			startupWarnings = append(
+				startupWarnings,
+				fmt.Sprintf("New version available! Latest version: %s; Current version: %s", state.LatestVersion, mv.Version),
```
rumyantseva, line 106: @Fashander, what do you think about this wording when we report that a new version is available?

rumyantseva (commented):
LGTM overall, I pinged @Fashander about the wording just in case. I'll approve when we have a separate issue for tests.

internal/handlers/pg/msg_getlog.go
noisersup: I don't think so, we don't document any update process anywhere, also it depends on the method of installation (docker users will have a different method of updating than package managers users (which would also have a different commands depending on the manager)) WDYT?

rumyantseva: I agree with Patryk, yeah

Fashander: That makes sense. Thanks.

internal/util/state/metrics.go
noisersup: [done](https://github.com/FerretDB/FerretDB/issues/1751)

PR #1739, opened 2023-01-03
Implement `distinct` command

### LINKED ISSUE #945: Implement `distinct` command
https://www.mongodb.com/docs/manual/reference/command/distinct/

In the scope:
* `distinct`, `key`, `query`, `comment` arguments;
* compatibility tests, etc (as usual).

Not in the scope:
* query statistics, query plan.

DoD:
- [x] usual things: both pg and tigris are in the scope, compat tests are in the scope, check various data types with compat tests, update docs about supported commands - https://docs.ferretdb.io/reference/supported_commands/
- [x] test an example where `key` uses dot notation

rumyantseva:
# Description

Closes #945.

## Readiness checklist

### DIFF
--- a/website/docs/reference/supported_commands.md
+++ b/website/docs/reference/supported_commands.md
@@ -281,10 +281,10 @@ Related epic - [Issue](https://github.com/FerretDB/FerretDB/issues/153)
 The epic - [Issue](https://github.com/FerretDB/FerretDB/issues/9).
 
 | Command     | Argument | Status | Comments                                                  |
-| ----------- | -------- | ------ | --------------------------------------------------------- |
+|-------------|----------|--------|-----------------------------------------------------------|
 | `aggregate` |          | ❌      | [Issue](https://github.com/FerretDB/FerretDB/issues/1410) |
 | `count`     |          | ✅      |                                                           |
-| `distinct`  |          | ❌      |  [Issue](https://github.com/FerretDB/FerretDB/issues/945) |
+| `distinct`  |          | ✅      |                                                           |
 
 <!-- markdownlint-disable MD001 MD033 -->
 <!-- That's the simplest way to remove those sections from the right menu. -->


internal/handlers/common/sort_array.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"sort"
+
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/must"
+)
+
+// SortArray sorts the values of given array.
+func SortArray(arr *types.Array, sortType types.SortType) {
```
rumyantseva, line 25: The idea is the same as for `common.SortDocuments`.

integration/distinct_compat_test.go
```diff
@@ -0,0 +1,157 @@
...
+			filter: bson.D{},
+		},
+		"DotNotation": {
+			field:  "v.foo",
+			filter: bson.D{},
+		},
+		"DotNotationArray": {
+			field:  "v.array.0",
+			filter: bson.D{},
+		},
```
chilagrow, line 153: I suggest one more edge test case. This one has challenge in other tests and was skipped https://github.com/FerretDB/FerretDB/blob/main/integration/update_field_compat_test.go#L208

```
field: "v.0.foo",
filter: bson.D{},
resultType: emptyResult,
```

chilagrow (commented):
Looks good! Just one additional edge case test suggestion :hugs:

integration/distinct_compat_test.go
rumyantseva: I added a test case, but here it's not `emptyResult`, because the Compat collection returns an empty array, not `nil`. I added one more check that if the len compat result is 0, the target result must be the same (so then we check that both of them give empty arrays and not something else).

chilagrow: Ah yes that makes sense, thanks for the explanation! :hugs:

chilagrow (approved):
Looks great

integration/distinct_compat_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+			var nonEmptyResults bool
+			for i := range targetCollections {
+				targetCollection := targetCollections[i]
+				compatCollection := compatCollections[i]
+				t.Run(targetCollection.Name(), func(t *testing.T) {
+					t.Helper()
+
+					targetRes, targetErr := targetCollection.Distinct(ctx, tc.field, tc.filter)
+					compatRes, compatErr := compatCollection.Distinct(ctx, tc.field, tc.filter)
```
noisersup, line 69: I know it's not mongosh, but still I assume it's mongo helper function. Are we sure that we want use it and not `runCommand`? 

From https://www.mongodb.com/docs/manual/reference/command/distinct/:
> In [mongosh](https://www.mongodb.com/docs/mongodb-shell/#std-program-mongosh), this command can also be run through the [db.collection.distinct()](https://www.mongodb.com/docs/manual/reference/method/db.collection.distinct/#mongodb-method-db.collection.distinct) helper method..
> 
> Helper methods are convenient for [mongosh](https://www.mongodb.com/docs/mongodb-shell/#std-program-mongosh) users, but they may not return the same level of information as database commands. In cases where the convenience is not needed or the additional return fields are required, use the database command.

integration/distinct_compat_test.go
```diff
@@ -0,0 +1,167 @@
...
+					// If compat result is an empty array, the target results must be empty array too
+					// (not nil or something else).
+					if len(compatRes) == 0 {
+						assert.Equal(t, compatRes, targetRes)
+					}
+
+					// We can't check the exact data types because they might be different.
+					// For example, if targetRes is [float64(0), int32(1)] and compatRes is [int64(0), int64(1)],
+					// we consider them equal. If different documents use different types to store the same value
+					// in the same field, it's hard to predict what type will be returned by distinct.
```
noisersup, line 94: Do we have any plans for future to investigate how MongoDB handles every case, or that's too much unnecessary effort?

internal/handlers/common/distinct.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// FilterDistinctValues returns distinct values from the given slice of documents with the given key.
+//
+// If the key is not found in the document, the document is ignored.
+//
+// If the key is found in the document, and the value is an array, each element of the array is added to the result.
+// Otherwise, the value itself is added to the result.
+func FilterDistinctValues(docs []*types.Document, key string) (*types.Array, error) {
+	distinct := types.MakeArray(len(docs))
```
noisersup, line 99: I understand that at the end we need to return the array, though why can't we use map here? As the `distinct` command runs against whole collection, iterating through it for every field could not scale well. Map would also remove requirement to sort everything afterwards. Is there any technical difficulty to do that now?

noisersup (commented):
I like the implementation, though I have one question related to it. If we don't want to change it, I'm ready to approve it!

integration/distinct_compat_test.go
rumyantseva: I think we had that question before, but let's discuss it again :) 

The idea is that we test compatibility with driver first. We don't care about monogsh, but we do care about mongo driver for Go. So, we use `Distinct` function because mongo driver offers it. For example, this is what we also did when we tested ordered inserts - we used `insertOne` and `insertMany` for main tests. It helped us reveal that we used to return a wrong response type. 

Imagine, you are a developer and you want to write an app that would use FerretDB and you need to call distinct. Most likely, as a developer, you will call `targetCollection.Distinct`, not `RunCommand`. This is why we want to test such cases first.

We will truly need to test with `runCommand`, but later. For example, we will need to cover cases when `key` is an array, or some other weird data is passed. We don't focus on it now because it'll take way too much time.

internal/handlers/common/distinct.go
rumyantseva: Unfortunately we can't use map to check for duplicates because a value of `Binary` type can't be a map key (it's not possible to get a hash from it).

But I like your question!

integration/distinct_compat_test.go
noisersup: You right! Yeah, let's focus on the drivers first. We could create an issue or todo for runCommand later!

internal/handlers/common/distinct.go
rumyantseva: Map will not remove the requirement to sort things because there is no guaranteed order of the keys in the map. And also the result must be of `types.Array`, so if we used map, we had to iterate through it to build an array and sort that array.

integration/distinct_compat_test.go
rumyantseva: That looks like an unnecessary effort to me, as it's such a tiny detail. I suggest keeping it as is and only implementing it if we have a user request. Also, maybe in some cases distinct can be moved to pushdown side, and this is another reason why I don't want to invest into it now. 

In principle, we can document it as a known difference and write a diff test for it. But again, that looks unnecessary to me.

Fashander (approved):

PR #1753, opened 2023-01-04
Support `$rename` field update operator

### LINKED ISSUE #626: Support `$rename` field update operator

noisersup:
# Description

Closes #626.

## Readiness checklist


internal/handlers/common/update.go
```diff
@@ -236,6 +243,71 @@ func processPopFieldExpression(doc *types.Document, update *types.Document) (boo
...
+		// Remove old document
+		doc.RemoveByPath(sourcePath)
+
+		// Set new path with old value
+		err = doc.SetByPath(targetPath, val)
+		if err != nil {
+			return changed, err
+		}
+
+		changed = true
```
chilagrow, line 305: Could we have a testcase with multiple rename fields with the first one succeeds and second one fails. Just want check we behave the same as mongoDB on setting `changed`

`bson.D{{"$rename", bson.D{{"v.foo", "boo"}}, bson.D{{"v.array", 1}}}}`
It will use dataset from here https://github.com/FerretDB/FerretDB/blob/main/integration/shareddata/composites.go#L33

internal/handlers/common/update.go
```diff
@@ -236,6 +243,71 @@ func processPopFieldExpression(doc *types.Document, update *types.Document) (boo
...
+			return changed, NewWriteErrorMsg(ErrEmptyName, "An empty update path is not valid.")
+		}
+
+		sourcePath := types.NewPathFromString(key)
+
+		renameValue, ok := renameRawValue.(string)
+		if !ok {
+			return changed, NewWriteErrorMsg(
+				ErrBadValue,
+				fmt.Sprintf("the 'to' field for $rename must be a string: %s: %v", key, renameRawValue),
```
chilagrow, line 272:
```suggestion
				fmt.Sprintf("The 'to' field for $rename must be a string: %s: %v", key, renameRawValue),
```
Let's be the same response as mongodb shell.

```
test> db.rename.insert({_id:"number1", v: 1})
DeprecationWarning: Collection.insert() is deprecated. Use insertOne, insertMany, or bulkWrite.
{ acknowledged: true, insertedIds: { '0': 'number1' } }
test> db.rename.updateMany({_id:"number1"},{$rename:{"v":1}})
MongoServerError: The 'to' field for $rename must be a string: v: 1
```

internal/handlers/common/update.go
```diff
@@ -236,6 +243,71 @@ func processPopFieldExpression(doc *types.Document, update *types.Document) (boo
...
+
+		// Get value to move
+		val, err := doc.GetByPath(sourcePath)
+		if err != nil {
+			// TODO refactor `getByPath` to make this check more elegant
+			if skipErrReg.MatchString(err.Error()) {
+				continue
+			}
+
+			return changed, NewWriteErrorMsg(ErrUnsuitableValueType, err.Error())
```
chilagrow, line 293: I beleive there was a test case which caused it to fail if we continue on any error so we handle `types.Document.Get: key not found` error specially? Could we create an error like ErrDocumentPathNotFound and handle it accordingly? Or would this cause too much change in codebase? :thinking:

integration/update_field_compat_test.go
```diff
@@ -481,6 +481,97 @@ func TestUpdateFieldCompatMin(t *testing.T) {
...
+			update:     bson.D{{"$rename", bson.D{{"v.array.0", ""}}}},
+			resultType: emptyResult,
+		},
+		"Multiple": {
+			update:        bson.D{{"$rename", bson.D{{"v.foo", "v.bar"}, {"v.42", "v.43"}}}},
+			skipForTigris: "schema violation",
+		},
+		"MultipleConflictDestSource": {
+			update: bson.D{{"$rename", bson.D{{"v", "foo"}, {"foo", "bar"}}}},
+			skip:   "Updating the path 'foo' would create a conflict at 'foo'",
```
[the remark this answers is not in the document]
rumyantseva: +1! :)

integration/update_field_compat_test.go
```diff
@@ -481,6 +481,97 @@ func TestUpdateFieldCompatMin(t *testing.T) {
...
+			resultType: emptyResult,
+		},
+		"FieldArray": {
+			update:     bson.D{{"$rename", bson.D{{"v", primitive.A{}}}}},
+			resultType: emptyResult,
+		},
+		"FieldNaN": {
+			update:     bson.D{{"$rename", bson.D{{"v", math.NaN()}}}},
+			resultType: emptyResult,
+			skip:       "CommandError returned",
```
rumyantseva, line 552: It we skip because it's a wrong response from FerretDB, we should create an issue.

internal/handlers/common/update.go
```diff
@@ -236,6 +243,71 @@ func processPopFieldExpression(doc *types.Document, update *types.Document) (boo
...
+				fmt.Sprintf("The source and target field for $rename must differ: %s: %#v", key, renameValue),
+			)
+		}
+
+		targetPath := types.NewPathFromString(renameValue)
+
+		// Get value to move
+		val, err := doc.GetByPath(sourcePath)
+		if err != nil {
+			// TODO refactor `getByPath` to make this check more elegant
```
rumyantseva, line 288: Ideally, this refactoring should be done in the scope of the current task (as refactoring is usually in the scope of the task).

rumyantseva (changes requested):
The main things are done, a couple of things need some improvements:
- The `GetByPath` refactoring that is mentioned should be a part of the current issue if it's needed ( 'm not sure about the need). As @chilagrow mentioned, we don't check the exact error type in the other places, do we really need to check it in the current one? What are different possible error types that could change the behavior?

-  Tests skipped for both PG and Tigris should be either unskipped or should have issues created for the reasons we skip them.

integration/update_field_compat_test.go
```diff
@@ -481,6 +481,97 @@ func TestUpdateFieldCompatMin(t *testing.T) {
...
+			resultType: emptyResult,
+		},
+		"FieldDoc": {
+			update:     bson.D{{"$rename", bson.D{{"v", primitive.D{}}}}},
+			resultType: emptyResult,
+		},
+		"FieldArray": {
+			update:     bson.D{{"$rename", bson.D{{"v", primitive.A{}}}}},
+			resultType: emptyResult,
+		},
```
rumyantseva, line 548: For those negative test cases I'm not sure if we really need test cases for all possible wrong types :)
Every subtest creates a separate collection (if not a separate database), and having too many of them would slow down tests.

integration/update_field_compat_test.go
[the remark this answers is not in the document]
noisersup: Should our tests be the same as mongodb, or error codes just should match? The compat tests fucntion does not compare strings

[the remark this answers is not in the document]
rumyantseva: @noisersup it depends! There are a few examples of various tests that compare errors if they are not exactly the same. For example, `TestInsertRunCommandCompat` has an alternative error message field to compare with. And then it's possible to use helpers like `AssertMatchesCommandError` and `AssertEqualAltError` to compare: https://github.com/FerretDB/FerretDB/blob/8ac94387d8819e6e8fb9f1f464e7f3fc677a0394/integration/insert_run_command_compat_test.go#L79.

And I think we had a few other test cases where we don't check the exact match.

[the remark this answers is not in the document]
noisersup: For this issue then, error handling seems complicated. MongoDB has different order of checking for errors and we don't return same errors as them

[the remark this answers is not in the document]
rumyantseva: If it's possible, we need to keep order the same. If it's not possible or if this problem remains in the other operators too, let's create a separate issue.

internal/handlers/common/update.go
noisersup: I was thinking to create some underlying error type like this one, the limitation I've seen though, is that creating new error variable in the same way as we do almost everywhere (`ErrDocumentPathNotFound = fmt.Errorf("...")`) limits our ability to add the invalid key inside the message, as we currently do and as MongoDB does. We could make some extended logic (like error codes), but that's more complex and is the reason why I left `TODO` here, so we can discuss the best approach :)

internal/handlers/common/update.go
noisersup: done

internal/handlers/common/update.go
noisersup: done

integration/update_field_compat_test.go
noisersup: done

internal/handlers/common/update.go
w84thesun: If I understand correctly what's going on, we need to skip the error if the key was not found in the document. To do so we can return `KeyNotFound` error type from `types.Document`.`Get` and check for it here.

w84thesun (commented):
I left a comment about error handling. Maybe it will help to process error.

integration/update_field_compat_test.go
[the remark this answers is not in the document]
noisersup: As agreed - moved different errors messages comparison to #1772

internal/handlers/common/update.go
noisersup: done

internal/handlers/common/update.go
noisersup: done

integration/update_compat_test.go
```diff
@@ -59,6 +59,8 @@ func testUpdateCompat(t *testing.T, testCases map[string]updateCompatTestCase) {
 				setup.SkipForTigrisWithReason(t, tc.skipForTigris)
 			}
 
+			skippedAllTigrisSubtests := true
```
chilagrow, line 62: I'm not sure the purpose of this :thinking:

internal/handlers/common/update.go
```diff
@@ -622,6 +698,57 @@ func extractValueFromUpdateOperator(op string, update *types.Document) (*types.D
 		return nil, NewWriteErrorMsg(ErrFailedToParse, "Modifiers operate on fields but we found another type instead")
 	}
 
+	if op == "$rename" {
+		iter := doc.Iterator()
+		keys := map[string]struct{}{}
+
+		for {
+			k, v, err := iter.Next()
```
chilagrow, line 706: Nice use of iterator :hugs:

chilagrow (commented):
Thanks for dot notation error handling improvement :muscle: 

Please see the comment about skipping tests :pray:

integration/update_compat_test.go
```diff
@@ -139,6 +143,10 @@ func testUpdateCompat(t *testing.T, testCases map[string]updateCompatTestCase) {
 				})
 			}
 
+			if skippedAllTigrisSubtests {
```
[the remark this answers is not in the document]
rumyantseva: If all the subtests are skipped, `tc.resultType` will not match because in case of Tigris the result will be empty, but expect result is "non-empty". As the result, the main test will fail even though all the subtests are skipped. Maybe there is a better way to skip All Tigris subtests in such cases, @chilagrow did you have anything similar while moving tests to compat tests?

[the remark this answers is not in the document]
chilagrow: Thanks for the explanation. I now see the problem you are solving. I see you want `non-empty` check to be skipped for tigris.
Tigris does not allow schema change so renaming document key is not something that would be supported by tigris if I understand correctly. So I would skip all $rename tests for tigris :hugs:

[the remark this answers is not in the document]
chilagrow: Let's do this https://github.com/FerretDB/FerretDB/pull/1753/files#r1066737442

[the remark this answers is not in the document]
noisersup: Technically we still support `$rename` for tigris, and it will work in some test cases that are here, so I don't know why would we remove tests for it

integration/update_field_compat_test.go
```diff
@@ -481,6 +481,76 @@ func TestUpdateFieldCompatMin(t *testing.T) {
 	testUpdateCompat(t, testCases)
 }
 
+func TestUpdateFieldCompatRename(t *testing.T) {
+	testCases := map[string]updateCompatTestCase{
```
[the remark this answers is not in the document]
noisersup: so we return to the previous solution? :smile:

What if some subtest will start failing? Skipping that for tigris will make it imperceptible

internal/types/path.go
```diff
@@ -26,6 +26,43 @@ import (
...
 )
 
+// DocumentPathErrorCode represents DocumentPathError error code.
+type DocumentPathErrorCode int
+
+const (
+	// ErrDocumentPathKeyNotFound indicates that key was not found in document.
+	ErrDocumentPathKeyNotFound = iota + 1
+	// ErrDocumentPathCantAccess indicates that path couldn't be accessed.
+	ErrDocumentPathCantAccess
```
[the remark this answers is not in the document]
noisersup: Looks good!

build/.trivyignore
```diff
@@ -8,3 +8,4 @@ CVE-2022-33987  exp:2023-03-01
 CVE-2022-37599  exp:2023-03-01
 CVE-2022-37603  exp:2023-03-01
 CVE-2022-46175  exp:2023-03-01
+CVE-2017-20165  exp:2023-03-01
```
noisersup, line 11:
```

┌─────────┬────────────────┬──────────┬───────────────────┬───────────────┬─────────────────────────────────────────────────┐
│ Library │ Vulnerability  │ Severity │ Installed Version │ Fixed Version │                      Title                      │
├─────────┼────────────────┼──────────┼───────────────────┼───────────────┼─────────────────────────────────────────────────┤
│ debug   │ CVE-2017-20165 │ LOW      │ 2.6.9             │ 3.1.0         │ debug Inefficient Regular Expression Complexity │
│         │                │          │                   │               │ vulnerability                                   │
│         │                │          │                   │               │ https://avd.aquasec.com/nvd/cve-2017-20165      │
└─────────┴────────────────┴──────────┴───────────────────┴───────────────┴─────────────────────────────────────────────────┘
```

internal/handlers/common/update.go
```diff
@@ -622,6 +698,57 @@ func extractValueFromUpdateOperator(op string, update *types.Document) (*types.D
 		return nil, NewWriteErrorMsg(ErrFailedToParse, "Modifiers operate on fields but we found another type instead")
 	}
 
+	if op == "$rename" {
```
w84thesun, line 701: We should probably update the function description or move `$rename` checks into a separate function.

w84thesun (changes requested):
Nicely done with the path error codes! Tigris tests still are failing though. Shouldn't we just skip the whole `$rename` test?

internal/handlers/common/update.go
noisersup: I've placed it in this function, because it also checks for duplicates afterwards and returns the error about path collisions, I'll add a comment about checking for collisions inside of it.

Btw, thanks to your catch I guess we found a bug in duplicates handling!

noisersup:
> Tigris tests still are failing though. Shouldn't we just skip the whole $rename test?

They were failing because of timeout. We agreed to skip tests that are related to tigris schema errors, I think that we don't need to skip tests that check for different errors

build/.trivyignore
noisersup: moved to seperate pr

integration/update_field_compat_test.go
```diff
@@ -481,6 +481,80 @@ func TestUpdateFieldCompatMin(t *testing.T) {
...
+			update:     bson.D{{"$rename", bson.D{{"v.array.0", ""}}}},
+			resultType: emptyResult,
+		},
+		"DotArrayNonExisting": {
+			update:     bson.D{{"$rename", bson.D{{"foo.0.baz", int32(1)}}}},
+			resultType: emptyResult,
+		},
+		"Multiple": {
+			update:        bson.D{{"$rename", bson.D{{"v.foo", "v.bar"}, {"v.42", "v.43"}}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/1776",
```
rumyantseva, line 527: 👍 Thanks for creating a separate issue! I think after my PR [about array checks](https://github.com/FerretDB/FerretDB/pull/1778), I have an idea of how to add some data to cover such test case. I like the idea of having a separate issue, so when this PR and my PR #1778 are merged, we can quickly add test cases in #1776 ;)

internal/types/path.go
```diff
@@ -26,6 +26,43 @@ import (
...
+	ErrDocumentPathCannotAccess
+	// ErrDocumentPathArrayInvalidIndex indicates that provided array index is invalid.
+	ErrDocumentPathArrayInvalidIndex
+	// ErrDocumentPathIndexOutOfBound indicates that provided array index is out of bound.
+	ErrDocumentPathIndexOutOfBound
+	// ErrDocumentPathCannotCreateField indicates that it's impossible to create a specific field.
+	ErrDocumentPathCannotCreateField
+)
+
+// DocumentPathError describes an error that could occur on document related operations.
```
rumyantseva, line 45:
```suggestion
// DocumentPathError describes an error that could occur on document path related operations.
```

internal/handlers/common/update.go
```diff
@@ -236,6 +244,67 @@ func processPopFieldExpression(doc *types.Document, update *types.Document) (boo
...
+func processRenameFieldExpression(doc *types.Document, update *types.Document) (bool, error) {
+	renameExpression := update.SortFieldsByKey()
+
+	var changed bool
+
+	for _, key := range renameExpression.Keys() {
+		renameRawValue, err := renameExpression.Get(key)
+		if err != nil {
+			// if $rename field does not exist, don't change anything
+			continue
```
rumyantseva, line 258: Interesting: I see codecov complains about this line, but there is a test case called `NonExistingField` that works with a field that doesn't exist. Do you understand why codecov complains? I wonder if we need a test case like `bson.D{{"$rename", bson.D{{"foo", "boo"}, {"v", "bar"}}}},` where one field doesn't exist and one field exists.

internal/handlers/common/update.go
```diff
@@ -236,6 +244,67 @@ func processPopFieldExpression(doc *types.Document, update *types.Document) (boo
...
+		}
+
+		if key == "" || renameRawValue == "" {
+			return changed, NewWriteErrorMsg(ErrEmptyName, "An empty update path is not valid.")
+		}
+
+		// this is covered in extractValueFromUpdateOperator
```
rumyantseva, line 265: Hmm, I don't understand this comment. What does it mean?
Also, I see codecov complains that the next few lines are not covered.

internal/handlers/common/update.go
```diff
@@ -622,6 +700,57 @@ func extractValueFromUpdateOperator(op string, update *types.Document) (*types.D
...
+		keys := map[string]struct{}{}
+
+		for {
+			k, v, err := iter.Next()
+			if err != nil {
+				if errors.Is(err, iterator.ErrIteratorDone) {
+					break
+				}
+
+				return nil, err
```
rumyantseva, line 714: Let's return a lazyerror here as, in principle, we don't expect to have an error here, and it would be nice to have it wrapped so we can follow the stack of functions being called.

internal/handlers/common/update.go
rumyantseva: I agree with Dmitry. We added a cyclomatic complexity with this `if`. I'm ok to call some function in this place, but I'd suggest moving the logic inside this `if` into a separate function, so the code is more readable.

internal/handlers/common/update.go
```diff
@@ -622,6 +700,57 @@ func extractValueFromUpdateOperator(op string, update *types.Document) (*types.D
...
+					ErrConflictingUpdateOperators,
+					fmt.Sprintf("Updating the '%s' would create a conflict at '%s'", k, k),
+				)
+			}
+
+			keys[k] = struct{}{}
+
+			if _, ok = keys[vStr]; ok {
+				return nil, NewWriteErrorMsg(
+					ErrConflictingUpdateOperators,
```
rumyantseva, line 745: Is it applicable for `$rename`?

internal/handlers/common/update.go
```diff
@@ -236,6 +244,67 @@ func processPopFieldExpression(doc *types.Document, update *types.Document) (boo
...
+
+		sourcePath := types.NewPathFromString(key)
+		targetPath := types.NewPathFromString(renameValue)
+
+		// Get value to move
+		val, err := doc.GetByPath(sourcePath)
+		if err != nil {
+			var dpe *types.DocumentPathError
+			if !errors.As(err, &dpe) {
+				panic("getByPath returned error with invalid type")
```
rumyantseva, line 283: Why did you decide to throw a panic here and not a lazy error?

rumyantseva (commented):
Great job with this issue! I especially like your attention to detail and all those corner cases covered with tests!

I asked a few questions about the places I didn't understand, but overall I like the PR.

chilagrow (dismissed):
There are a few lines of code we can remove since codecov thinks they are not being used.

Looks great :hugs:

internal/handlers/common/update.go
noisersup: I'm not sure what do you mean. This part of code is called only for `$rename` operator and it checks for duplicates between keys and values (same as mongodb does). It's not possible to use {$rename: {{"foo": "bar"}, { "bar": "baz"} }} and etc

internal/handlers/common/update.go
noisersup: Moved the code to seperate `validateRenameExpression` function

internal/handlers/common/update.go
noisersup: @chilagrow you right! Changed it to must.NotFail

internal/handlers/common/update.go
noisersup: We don't need to check if that statement is `ok` and if `renameRawValue` underlying type is the string, as we covered it in `extractValueFromUpdateOperator` (the same code is now in `validateRenameExpression`). Any error related to value type validation will be returned in that function and we don't need to handle this another time. I can rephrase it to something else like "The correctness of this type assertion is made in `validateRenameExpresssion`" or smth like that

internal/handlers/common/update.go
noisersup: Have I misunderstood something? :)
This is covered in [MultipleSecondInvalid](https://github.com/FerretDB/FerretDB/pull/1753/files#diff-02beb83969c6feb8b48d15aebcd299647e0aa2e8b1e5c531e74eb2c4f636ad85R537-R540)

chilagrow: Yes this is covered now. Thanks for adding it!

internal/handlers/common/update.go
noisersup: This part of code now is removed so I guess it's resolved

w84thesun (dismissed):
LGTM!

internal/handlers/common/update.go
```diff
@@ -635,6 +714,71 @@ func extractValueFromUpdateOperator(op string, update *types.Document) (*types.D
...
+	keys := map[string]struct{}{}
+
+	for {
+		k, v, err := iter.Next()
+		if err != nil {
+			if errors.Is(err, iterator.ErrIteratorDone) {
+				break
+			}
+
+			return err
```
rumyantseva, line 740: According to our previous agreements, this one should either be a lazy error or should be covered by a test.

internal/handlers/common/update.go
```diff
@@ -635,6 +714,71 @@ func extractValueFromUpdateOperator(op string, update *types.Document) (*types.D
...
+			return NewWriteErrorMsg(
+				ErrConflictingUpdateOperators,
+				fmt.Sprintf("Updating the '%s' would create a conflict at '%s'", k, k),
+			)
+		}
+
+		keys[k] = struct{}{}
+
+		if _, ok = keys[vStr]; ok {
+			return NewWriteErrorMsg(
```
rumyantseva, line 770: Can we come up with a test case that would cover this error?

internal/handlers/common/update.go
```diff
@@ -236,6 +244,63 @@ func processPopFieldExpression(doc *types.Document, update *types.Document) (boo
...
+		if key == "" || renameRawValue == "" {
+			return changed, NewWriteErrorMsg(ErrEmptyName, "An empty update path is not valid.")
+		}
+
+		// this is covered in validateRenameExpression
+		renameValue := renameRawValue.(string)
+
+		if key == renameRawValue {
+			return changed, NewWriteErrorMsg(
+				ErrBadValue,
```
rumyantseva, line 266: What is the test case for it? Is it something like `bson.D{{"v", "v"}}`? Can we add it to tests?

rumyantseva (commented):
Added a couple of comments about lines that codecov reported

internal/handlers/common/update.go
noisersup: Thanks! I was trying to find any occurences

rumyantseva: I think this is the only one in this PR :)

internal/handlers/common/update.go
noisersup: It's tested [here](https://github.com/FerretDB/FerretDB/pull/1753/files#diff-02beb83969c6feb8b48d15aebcd299647e0aa2e8b1e5c531e74eb2c4f636ad85R489-R492)

rumyantseva: Hmm... Hm... Do you have an idea why codecov didn't catch it? Personally, at first sight, I don't see it.

noisersup: My first guess it's because of the fact that we compare any value with a string (maybe even there's a bug here!!) I'll debug if this if statement works correctly and let you know

internal/handlers/common/update.go
noisersup: Good idea! Done https://github.com/FerretDB/FerretDB/pull/1753/files#diff-02beb83969c6feb8b48d15aebcd299647e0aa2e8b1e5c531e74eb2c4f636ad85R533-R536

rumyantseva (approved):
LGTM 👍

internal/handlers/common/update.go
```diff
@@ -611,6 +682,8 @@ func checkConflictingChanges(a, b *types.Document) error {
 // The result returned for "$setOnInsert" operator is
 //
 //	bson.D{{"v", nil}}.
+//
+// It also checks for path collisions and returns the error if there's any.
```
w84thesun, line 686: I believe we should remove that line now as the checks were moved to `validateRenameExpression`.

```suggestion
//
// It also checks for path collisions and returns the error if there's any.
```

w84thesun (changes requested):
Good, just one small fix left!

w84thesun (approved):

chilagrow (approved):

Taskfile.yml
```diff
@@ -185,6 +185,7 @@ tasks:
     cmds:
       - >
         go test -count=1 {{.RACEFLAG}} -tags={{.BUILDTAGS}} -shuffle=on -coverpkg=../...
+        -timeout=15m
```
chilagrow, line 188: Do we need this change since we didn't make change to `test-integration-tigris` after all?

noisersup: We've added another test cases and the time needed to run this tests increased. This value will be also overwritten with "20" in one of existing drafts

rumyantseva: I'd suggest to keep it to not re-request review again 🙈 , I'll resolve the conflict in my PR

AlekSi: That's fine (for now)

AlekSi (approved):

PR #1760, opened 2023-01-06
Support `$mul` field update operator

### LINKED ISSUE #625: Support `$mul` field update operator

chilagrow:
# Description

Closes #625. 

* Multiplication produces -0. This must be supported because not supporting such will cause too many exceptions. To do this, we now support -0 by always converting to 0.
* Dance PR handles testing `$mul` on max float value https://github.com/FerretDB/dance/pull/310
## Readiness checklist


internal/util/testutil/equal.go
```diff
@@ -241,9 +241,6 @@ func equalScalars(tb testing.TB, v1, v2 any) bool {
 		if math.IsNaN(s1) {
 			return math.IsNaN(s2)
 		}
-		if s1 == 0 && s2 == 0 {
-			return math.Signbit(s1) == math.Signbit(s2)
```
chilagrow, line 245: On compat `float64(-123456789)` * 0 return `-0`.
On target `float64(-123456789)` * 0 return `0`.

FerretDB does not support `-0`, and we let target and compat to equate `-0` == `0`

integration/shareddata/shareddata.go
```diff
@@ -83,6 +84,32 @@ func AllProviders() []Provider {
 	return maps.Values(res)
 }
 
+// Providers are array of providers.
+type Providers []Provider
+
+// Remove returns an array of providers without the specified provider.
+func (ps Providers) Remove(provider string) Providers {
```
rumyantseva, line 91: I like the idea of this function! With it, we can keep `AllProviders` in as many places as possible and exclude only those that are not supported because of a specific reason.

integration/shareddata/shareddata.go
```diff
@@ -83,6 +84,32 @@ func AllProviders() []Provider {
...
+	var found bool
+	res := make(map[string]Provider, len(ps)-1)
```
rumyantseva, line 93: An alternative way would be to iterate though the slice and call ([`slices.Remove`](https://pkg.go.dev/golang.org/x/exp/slices#Delete)) when the element is found. This way, we don't need to use an exrtra map.

integration/update_field_compat_test.go
```diff
@@ -793,3 +794,121 @@ func TestUpdateFieldCompatMixed(t *testing.T) {
 
 	testUpdateCompat(t, testCases)
 }
+
+func TestUpdateFieldCompatMul(t *testing.T) {
+	t.Parallel()
+
+	providers := shareddata.AllProviders().
+		Remove("BigDoubles").
```
rumyantseva, line 802: For each call of `Remove` let's add a comment with a reason why we remove it.

integration/shareddata/shareddata.go
rumyantseva: Maybe also the signature could be `Remove(provider ...string)`, so if a few providers are removed, we don't need to call `Remove` twice.

integration/shareddata/shareddata.go
chilagrow: Surprisingly this didn't work as straight forward because we alter would alter the caller's slice. I'll keep as it is for now.

rumyantseva: > because we alter would alter the caller's slice

I didn't understand this part. Does it modify the caller's slice? Or doesn't? I expect the caller's slice to be modified, yes. If we don't want it, we need to make a copy of it.

internal/handlers/tigris/msg_find.go
```diff
@@ -76,7 +76,7 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
 	}
 
 	var reply wire.OpMsg
-	must.NoError(reply.SetSections(wire.OpMsgSection{
+	if err := reply.SetSections(wire.OpMsgSection{
```
chilagrow, line 79: This used to panic on `-0`.

internal/types/document_validation.go
```diff
@@ -142,6 +142,11 @@ func (d *Document) ValidateData() error {
 					ErrValidation, fmt.Errorf("invalid value: { %q: %f } (infinity values are not allowed)", key, v),
 				)
 			}
+			if v == 0 && math.Signbit(v) {
```
chilagrow, line 145: Do not let `-0` upon update.

integration/shareddata/shareddata.go
chilagrow: Good idea :+1: Applied this

integration/shareddata/shareddata.go
chilagrow: We need to copy it anyway. So I created a new slice.

integration/update_field_compat_test.go
chilagrow: Applied :+1:

internal/handlers/common/update.go
```diff
@@ -526,6 +673,10 @@ func ValidateUpdateOperators(update *types.Document) error {
 		return err
 	}
 
+	if err = checkConflictingOperators(mul, currentDate, inc, min, max, pop, set, setOnInsert, unset); err != nil {
```
chilagrow, line 676: We could not do this. Because this is gonna get outdated very quickly.

While testing, I discovered each update operator conflict to each other and this is checking that,

integration/shareddata/shareddata.go
```diff
@@ -85,6 +86,31 @@ func AllProviders() []Provider {
...
+func (ps Providers) Remove(removeProviderNames ...string) Providers {
+	res := make([]Provider, 0, len(ps))
+
+	for _, p := range ps {
+		keep := true
+
+		for _, name := range removeProviderNames {
+			if p.Name() == name {
+				keep = false
+				break
```
rumyantseva, line 102:
```suggestion
continue
```

integration/shareddata/shareddata.go
```diff
@@ -85,6 +86,31 @@ func AllProviders() []Provider {
...
+			}
+		}
+
+		if keep {
```
rumyantseva, line 106: If we use `continue`, this check is not needed 
```suggestion
```

integration/shareddata/shareddata.go
chilagrow: Maybe I'm misunderstand, if I use `continue` here this will `continue` in the inner loop. I would like to stop iterating on inner loop. :thinking:

integration/shareddata/shareddata.go
chilagrow: I think this is still needed because we have nested loop :thinking:

integration/shareddata/shareddata.go
rumyantseva: ah! yes, you are right! sorry, I misread it

integration/shareddata/scalars.go
```diff
@@ -135,14 +134,32 @@ var Doubles = &Values[string]{
...
 		"double-max-overflow-verge": 9.223372036854776832e+18,
 		"double-min-overflow":       -9.223372036854776833e+18,
 		"double-min-overflow-verge": -9.223372036854776832e+18,
 	},
 }
 
+// BigDoubles contains double values which would overflow on
+// numeric update operation such as $mul. Upon such,
+// target returns error and compat returns +INF or -INF.
+// BigDoubles maybe excluded on such update tests and tested
```
w84thesun, line 147:
```suggestion
// BigDoubles may be excluded on such update tests and tested
```

w84thesun (changes requested):
Nice job! Added a tiny spelling fix.

internal/types/document_validation_test.go
```diff
@@ -102,12 +103,24 @@ func TestDocumentValidateData(t *testing.T) {
...
+		"InfFromUpdate": {
+			doc:      must.NotFail(NewDocument("v", math.Inf(1))),
+			isUpdate: true,
+			reason: errors.New(`update produces invalid value: { "v": +Inf } ` +
+				`(update operations that produce infinity values are not allowed)`),
+		},
+		"NegativeZeroFromUpdate": {
+			doc:      must.NotFail(NewDocument("v", math.Copysign(0, -1))),
+			isUpdate: true,
+			reason: errors.New(`update produces invalid value: { "v": -0 } ` +
```
rumyantseva, line 115: I guess, it's impossible to return just `0` in this case, right? Do we want to differentiate `-0` and `0` when we do operations? 
In diff.md it says "update operation producing -0 is converted to 0.", then why do we expect an error here?

internal/types/document_validation.go
```diff
@@ -138,10 +140,24 @@ func (d *Document) ValidateData() error {
...
+				}
+
 				return newValidationError(
 					ErrValidation, fmt.Errorf("invalid value: { %q: %f } (infinity values are not allowed)", key, v),
 				)
 			}
+
+			if v == 0 && math.Signbit(v) {
+				return newValidationError(
+					ErrValidation, fmt.Errorf("update produces invalid value: { %q: -0 } (-0 is not supported)", key),
```
rumyantseva, line 158: diff.md it says "update operation producing -0 is converted to 0." Is there some place that process that error and returns 0 if it happens?

rumyantseva (commented):
Great attention to detail and accuracy! 👍 
As usual, corner cases are the ones that take quite a lot of energy, but I like how we process and document them.

This is not a full review yet, but I left some questions I've had so far.

website/docs/diff.md
```diff
@@ -18,15 +18,18 @@ slug: /diff/
    * document keys must not contain `$` or `.` signs;
    * document fields of double type must not contain `Infinity` or `-Infinity` values.
 6. When insert command is called, insert documents must not have duplicate keys.
-7. Database and collection names restrictions:
+7. Update command restrictions:
+   * update operation producing `Infinity` or `-Infinity` is not supported.
+   * update operation producing -0 is converted to 0.
```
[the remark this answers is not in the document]
chilagrow: Good point. How does diff test work if ferretDB returns 0 and diff returns -0? Normally we check that ferretDB returns error but it's not gonna be the case :thinking:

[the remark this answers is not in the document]
rumyantseva: Well, it's up to us to decide what to check ;) So, for example, we can check that MongoDB returns `-0` and FerretDB returns `0`.

internal/util/testutil/equal_test.go
```diff
@@ -41,7 +41,6 @@ func TestEqual(t *testing.T) {
 	AssertNotEqual(t, math.Inf(-1), math.Inf(+1))
 
 	AssertEqual(t, 0.0, math.Copysign(0, +1))
-	AssertNotEqual(t, math.Copysign(0, +1), math.Copysign(0, -1))
```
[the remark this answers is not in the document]
chilagrow: I assume it was here because we used to support -0. I want to remove this because when compat returns -0 and target returns 0, we want to say they are equal. Should this cause issue? If this change is undesirable, we can test this solely in diff and undo this change.

internal/types/document_validation.go
chilagrow: Yeah diff.md is correct. However any future operator which may produce -0 by update operation may silently put the -0 in the DB. Having this check will avoid that.

chilagrow: Other way of safety check is also possible here, do you have other suggestion? :hugs:

website/docs/diff.md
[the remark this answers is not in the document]
chilagrow: Let me do that.

internal/types/document_validation.go
rumyantseva: I don't really understand how it works. I expect this error is checked somewhere, and `0` is returned instead of `-0`, but I don't see where.

internal/handlers/tigris/msg_find.go
rumyantseva: But `diff.md` says that in case of `-0` we return `0`. This code looks like we return an error.

internal/types/document_validation.go
```diff
@@ -62,7 +62,9 @@ func (e *ValidationError) Code() ValidationErrorCode {
 // ValidateData checks if the document represents a valid "data document".
 // It places `_id` field into the fields slice 0 index.
 // If the document is not valid it returns *ValidationError.
-func (d *Document) ValidateData() error {
+// The argument isUpdate is set true when validation is applied
+// upon update operation.
+func (d *Document) ValidateData(isUpdate bool) error {
```
rumyantseva, line 67: I think the initial idea of this function was to keep it as the one that can be used all the time, for any document, regardless of the stage or operation. This is the package `type` and it doesn't really know anything about inserts, updates, deletes, it works with documents themselves.

What if in the future we discover some special cases for `isDelete`, `isUpsert`, `isSomethingElse`? 
My suggestion is not to extend this function but to write a separate function to validate the result of update operation, and then this validate function should be in the place where update operation is happening.

internal/util/testutil/equal_test.go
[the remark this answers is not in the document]
rumyantseva: Well, even if we don't support -0 anymore, we probably want to make sure that if by mistake we return `-0` and compare it with `0`, the result of the comparison will tell us that they are not the same. With that, keeping this line of code looks reasonable to me.

internal/types/document_validation_test.go
rumyantseva: Or, as I wrote in my other comment, this validation doesn't have to be in `types`, and then we don't have this problem at all.

internal/types/document_validation.go
chilagrow: Thanks for the tip, let's do the validation in handler which is where update is happening :+1:

internal/types/document_validation.go
chilagrow: Before this change, update operation result `-0` was stored it to DB for the edge case that was not caught. Insert of `-0` is validated here so we can't insert it https://github.com/FerretDB/FerretDB/blob/main/internal/wire/validation.go#L74.

A document which may accidentally produce `-0` could be caught here and return error.

internal/handlers/tigris/msg_find.go
chilagrow: Since now it's handled by validation this doesn't need to change anymore. But I was surprised to see that updating DB with -0 was possible but it panic on trying to fetch value from DB. FYI, at the time I was implementing it, I didn't even know why negative zero was produced. :see_no_evil: I'll undo this change

internal/util/testutil/equal_test.go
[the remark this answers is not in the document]
chilagrow: Sure let's test this edge case in diff test and undo this change

internal/types/document_validation.go
chilagrow: Let's remove this validation and sets `-1` to `0`. I initially added this change because we sets -0 to 0 in $mul operator, but for other existing operators, edge case of inserting -0 might exist. And if it did update -0, it panic'd in https://github.com/FerretDB/FerretDB/pull/1760#discussion_r1068965770 which was undesirable.

internal/types/document_validation_test.go
chilagrow: We now change -0 to 0 in validation.

internal/types/document_validation_test.go
```diff
@@ -117,4 +117,40 @@ func TestDocumentValidateData(t *testing.T) {
 			}
 		})
 	}
+
+	for name, tc := range map[string]struct {
```
rumyantseva, line 121: Let's make the previous `for` and this `for` subtests, so we can run them in parallel.

rumyantseva (commented):
LGTM overall, great job! 👍 

Let's make unit tests subtests and parallel (so it's consistent with the other parts of code and can be run faster)

rumyantseva (approved):
LGTM

w84thesun (approved):
Nice job! 👍

chilagrow:
@Fashander could you review the documentation change please?

Fashander (approved):

PR #1769, opened 2023-01-09
Add `getMore` command

### LINKED ISSUE #1733: Support `getMore` command
### What should be done?

Test script: #2761

w84thesun:
# Description

Implementation of Tigris cursors is not in the scope of this PR. We should implement query iterators for Tigris first.

Closes #1733.

## Readiness checklist

* [x] I added unit tests for new functionality or bug fixes.


integration/getmore_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+
+	"github.com/FerretDB/FerretDB/integration/setup"
+	"github.com/FerretDB/FerretDB/integration/shareddata"
+	"github.com/FerretDB/FerretDB/internal/types"
+)
+
+func TestGetMoreErrors(t *testing.T) {
+	t.Skip("TODO: https://github.com/FerretDB/FerretDB/issues/1807")
```
w84thesun, line 30: Enabling this test would make MongoDB test flaky. Described what can be done with it in #1807.

website/docs/reference/supported_commands.md
```diff
@@ -56,7 +56,10 @@ sidebar_position: 1
 |                 | `hint`                     | ❌      | Ignored                                                    |
 |                 | `comment`                  | ⚠️      | Not implemented in Tigris                                  |
 |                 | `let`                      | ⚠️      | Unimplemented                                              |
-| `getMore`       |                            | ❌      | [Issue](https://github.com/FerretDB/FerretDB/issues/1733)  |
+| `getMore`       |                            | ✅      | Basic command if fully supported                           |
+|                 | `batchSize`                | ✅      |                                                            |
+|                 | `maxTimeMS`                | ⚠️     | https://github.com/FerretDB/FerretDB/issues/1808           |
```
w84thesun, line 61: I moved `maxTimeMS` into separate issue as this one is already big enough.

integration/shareddata/scalars.go
```diff
@@ -371,3 +373,27 @@ func tigrisSchema(typeString string) string {
 			}`
 	return strings.ReplaceAll(common, "%%type%%", typeString)
 }
+
+func generateBigMap(count int) map[string]any {
```
rumyantseva, line 377: Let's document this function (why and when it's needed)

integration/shareddata/scalars.go
```diff
@@ -371,3 +373,27 @@ func tigrisSchema(typeString string) string {
...
+	res := make(map[string]any, count)
+
+	for i := 0; i < count; i++ {
+		res[fmt.Sprintf("key-%d", i)] = rand.Int31()
+	}
+
+	return res
+}
+
+// Int32BigAmounts contains int32 values for tests.
```
rumyantseva, line 389: Let's add a comment to explain that this data provider is excluded by default and why (and when it needs to be used).

internal/handlers/tigris/msg_getmore.go
```diff
@@ -0,0 +1,131 @@
...
+				break
+			}
+
+			return nil, lazyerrors.Error(err)
+		}
+
+		resDocs.Append(doc)
+	}
+
+	id := int64(1)
```
rumyantseva, line 106: Do we know what `id` should be generated "in real life"? Do we need any follow-up tickets to support it? I guess https://github.com/FerretDB/FerretDB/issues/1807 is about it, right?

internal/handlers/tigris/msg_getmore.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/iterator"
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+	"github.com/FerretDB/FerretDB/internal/util/must"
+	"github.com/FerretDB/FerretDB/internal/wire"
+)
+
+// MsgGetMore implements handlers.Interface.
+func (h *Handler) MsgGetMore(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
+	document, err := msg.Document()
```
rumyantseva, line 33: Does it make sense to move the common logic (params reading) to the common handler? So it's not repeated for both `pg` and `tigris`. Examples: `common/find.go`, `common/distinct.go`

integration/query_compat_test.go
```diff
@@ -191,3 +196,24 @@ func TestQueryCompatSort(t *testing.T) {
...
+			batchSize: 1,
+		},
+		"BatchSize100": {
+			filter:    bson.D{},
+			batchSize: 100,
+		},
+		"BatchSize102": {
+			filter:    bson.D{},
+			batchSize: 102,
+		},
```
rumyantseva, line 215: Let's maybe add a comment about the choice of those numbers?
I see 100 is 1 less than default, and 102 is 1 more than default, I guess it's to test the logic around default value.

integration/query_test.go
```diff
@@ -503,3 +503,67 @@ func TestQueryNonExistingCollection(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, actual, 0)
 }
+
+func TestQueryBatchSize(t *testing.T) {
```
rumyantseva, line 507: Why it wasn't possible to write a compat test for this case?

rumyantseva (commented):
Not a full review yet, asked questions about the places I don't understand.

For now my biggest question - why for some cases it was possible to write compat tests, and for others it wasn't?

internal/handlers/tigris/msg_getmore.go
w84thesun: No, #1807 is not about it. I created an issue for that – #1811. Will add a TODO here.

internal/handlers/tigris/msg_getmore.go
w84thesun: I give it another thought and I think we can move it to the `common` package for now.

integration/query_compat_test.go
w84thesun: Added.

integration/shareddata/scalars.go
w84thesun: Added a comment.

integration/query_test.go
w84thesun: That test and `TestGetMoreErrors` are a bit different in the way how they work. We are expecting errors here and there, so we don't need to insert any data to run tests. The current compatibility tests setup assumes that there would be some data providers and makes it difficult to write proper tests. There are some similar tests (`TestQueryBadMaxTimeMSType`, `TestQueryBadSortType`) that were not moved to compat yet. We can work on that kind of tests in a separate issue.

internal/handlers/pg/msg_find.go
```diff
@@ -128,7 +148,7 @@ func (h *Handler) fetchAndFilterDocs(ctx context.Context, tx pgx.Tx, sqlParam *p
 		_, doc, err := iter.Next()
 		if err != nil {
 			if errors.Is(err, iterator.ErrIteratorDone) {
-				return resDocs, nil
+				break
```
w84thesun, line 151: Here I wanted to make it a bit easier to find where the result would be returned.

internal/clientconn/conninfo/conn_info.go
```diff
@@ -34,6 +36,16 @@ type ConnInfo struct {
 	rw       sync.RWMutex
 	username string
 	password string
+
+	curRW  sync.RWMutex
```
rumyantseva, line 40: Asking for educational purposes 🤓  I see `RWMutex` is set, could you explain the logic about reading/writing? I see in the code below, it uses write locks to set/remove the cursor, but I don't understand how `sync.RWMutex` benefits over `sync.Mutex`.

internal/clientconn/conninfo/conn_info.go
```diff
@@ -53,6 +65,27 @@ func (connInfo *ConnInfo) SetAuth(username, password string) {
...
+// SetCursor stores the cursor value.
+func (connInfo *ConnInfo) SetCursor(collection string, cursor iterator.Interface[int, any]) {
+	connInfo.curRW.Lock()
+	defer connInfo.curRW.Unlock()
+
+	connInfo.cursor[collection] = cursor
+}
+
+// RemoveCursor removes the cursor value stored.
+func (connInfo *ConnInfo) RemoveCursor(collection string) {
```
rumyantseva, line 82: Where do we call it from?

internal/clientconn/conninfo/conn_info.go
```diff
@@ -53,6 +65,27 @@ func (connInfo *ConnInfo) SetAuth(username, password string) {
 	connInfo.password = password
 }
 
+// Cursor returns the cursor value stored.
+func (connInfo *ConnInfo) Cursor(collection string) iterator.Interface[int, any] {
+	return connInfo.cursor[collection]
+}
+
+// SetCursor stores the cursor value.
```
rumyantseva, line 73: Maybe it worth to mention that "collection" string should contain both db and collection names not to cause a conflict between the same collections in different dbs.

integration/query_test.go
rumyantseva: Got it, thanks! Sounds reasonable! Let's add a comment about it? (To clarify that we don't plan to move it to compat tests)

integration/getmore_compat_test.go
```diff
@@ -0,0 +1,113 @@
...
+					require.NoError(t, compatResult.All(ctx, &compatRes))
+
+					assert.Equal(t, len(compatRes), len(targetRes), "result length mismatch")
+				})
+			}
+		})
+	}
+}
+
+func TestGetMoreCompat(t *testing.T) {
```
rumyantseva, line 97: Should we have a test that would cause storing two different iterators for two different collections in `ConnInfo`? As we use a map and a mutex there, that would be nice to have an example where we query two different collections in parallel from the same connection.

rumyantseva (commented):
Finished the review. Overall, it looks good to me as we implemented and tested a way to work with cursors. I also like that we have a follow-up task to clarify `id` field. 

I left a few comments with some questions/suggestions.

internal/clientconn/conninfo/conn_info.go
w84thesun: We call it from `MsgGetMore` when the iterator is done.

internal/clientconn/conninfo/conn_info.go
```diff
@@ -53,6 +65,27 @@ func (connInfo *ConnInfo) SetAuth(username, password string) {
 	connInfo.password = password
 }
 
+// Cursor returns the cursor value stored.
+func (connInfo *ConnInfo) Cursor(collection string) iterator.Interface[int, any] {
+	return connInfo.cursor[collection]
```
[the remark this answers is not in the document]
w84thesun: Yep, we should, I have missed it. Thanks!

integration/getmore_compat_test.go
w84thesun: Added a unit test for parallel read and write to cursor.

internal/clientconn/conninfo/conn_info.go
w84thesun: Renamed it to `key` and added a comment about usage.

internal/clientconn/conninfo/conn_info.go
w84thesun: As far as I can understand there could be `n` readers with the `RWMutex` apart from the regular `Mutex`.

rumyantseva: Why can't a regular mutex be used with those `n` readers?

integration/getmore_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+
+	"github.com/FerretDB/FerretDB/integration/setup"
+	"github.com/FerretDB/FerretDB/integration/shareddata"
+	"github.com/FerretDB/FerretDB/internal/types"
+)
+
+func TestGetMoreErrors(t *testing.T) {
```
chilagrow, line 29: Can we make this a compat test? Compat test is very useful for testing error

integration/getmore_compat_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/FerretDB/FerretDB/integration/shareddata"
+)
+
+type queryGetMoreCompatTestCase struct {
+	sort      bson.D
+	batchSize int
+	limit     int
+}
+
+func testGetMoreCompat(t *testing.T, testCases map[string]queryGetMoreCompatTestCase) {
```
chilagrow, line 35: I'm not seeing where it is calling getMore :thinking:

integration/query_test.go
chilagrow: Just noting that it might be easier to just write compat for testing if we are expecting errors here and there. We can setup compat the way we need it so empty collection should be easy too :hugs: 

Anyway I think now our test-integration-mongodb runs integration tests too so error itself is also tested against compat like system I think.

chilagrow (commented):
Great work!

Fashander (approved):

rumyantseva:
@w84thesun what does `task docs-fmt` say?

internal/clientconn/conninfo/conn_info.go
```diff
@@ -53,6 +65,33 @@ func (connInfo *ConnInfo) SetAuth(username, password string) {
...
+	connInfo.cursor[key] = cursor
+}
+
+// RemoveCursor removes the cursor value stored.
+// We use "db.collection" as the key to remove the cursor.
+func (connInfo *ConnInfo) RemoveCursor(key string) {
+	connInfo.curRW.Lock()
+	defer connInfo.curRW.Unlock()
+
+	delete(connInfo.cursor, key)
```
rumyantseva, line 92: What happens if we delete an iterator that wasn't closed? Is there a potential memory leak somewhere?

rumyantseva (commented):
Left some more comments/questions, would be nice to see codecov report just in case (when tests pass).

internal/clientconn/conninfo/conn_info.go
w84thesun: Added `cursor.Close`. I don't think we have to test it now as we iterate over `types.Array` and `Close` method doing nothing.

internal/handlers/tigris/msg_find.go
```diff
@@ -73,18 +74,37 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
-	firstBatch := types.MakeArray(len(resDocs))
-	for _, doc := range resDocs {
-		firstBatch.Append(doc)
+	batchSize := len(resDocs)
+	if len(resDocs) > int(params.BatchSize) {
+		batchSize = int(params.BatchSize)
+	}
+
+	firstBatch := types.MakeArray(batchSize)
+	resultDocumentsArray := types.MakeArray(0)
```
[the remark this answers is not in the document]
w84thesun: Fixed.

integration/getmore_compat_test.go
w84thesun: It's called by the driver when there is more document than in one batch. Added a comment about it where the `All` method is called.

integration/query_test.go
w84thesun: Moved to compat both tests.

internal/clientconn/conninfo/conn_info_test.go
```diff
@@ -72,3 +74,48 @@ func TestConnInfo(t *testing.T) {
...
+func TestConnInfoCursorParallelWork(t *testing.T) {
+	t.Parallel()
+
+	connInfo := NewConnInfo()
+
+	wg := sync.WaitGroup{}
+	start := make(chan struct{})
+
+	// Test parallel set of cursor.
+	for i := 0; i < 100; i++ {
```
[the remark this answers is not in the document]
w84thesun: Rewrote that test. Reading from deleted cursor seems to be not possible. We will get nil if it was already deleted.

internal/handlers/common/msg_getmore.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, NewCommandErrorMsg(
+			ErrTypeMismatch,
+			fmt.Sprintf(
+				`BSON field 'getMore.getMore' is the wrong type '%s', expected type 'long'`,
+				AliasFromType(cursorIDValue),
+			),
+		)
+	}
+
+	if cursorID != 1 {
```
chilagrow, line 69: I see, so get more cursorID can only be 1 at the moment :+1: . https://github.com/FerretDB/FerretDB/issues/1811

integration/getmore_compat_test.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	testGetMoreCompat(t, testCases)
+}
+
+type queryGetMoreErrorsCompatTestCase struct {
+	id         any
+	altMessage string
+	command    bson.D
+	err        bool
```
chilagrow, line 128:
```suggestion
	id         any // defaults to id fetched from running find command
	altMessage string // expects different error altMessage on target
	command    bson.D // defaults to empty
	err        bool // expects error if it is set to true
```

integration/getmore_compat_test.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+func testGetMoreCompatErrors(t *testing.T, testCases map[string]queryGetMoreErrorsCompatTestCase) {
```
chilagrow, line 131: This looks awesome! Thanks for this :hugs:

integration/getmore_compat_test.go
```diff
@@ -0,0 +1,290 @@
...
+
+	"github.com/FerretDB/FerretDB/integration/setup"
+	"github.com/FerretDB/FerretDB/integration/shareddata"
+	"github.com/FerretDB/FerretDB/internal/types"
+)
+
+type queryGetMoreCompatTestCase struct {
+	sort      bson.D
+	batchSize int
+	limit     int
```
chilagrow, line 37:
```suggestion
	sort      bson.D // defaults to bson.D{{"_id", 1}}
	batchSize int   // defaults to 0
	limit     int // defaults to 0
```

chilagrow (approved):
Super minor optional comments. Thanks for adding compat test :hugs:

internal/handlers/tigris/msg_find.go
```diff
@@ -73,18 +74,37 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+	firstBatch := types.MakeArray(batchSize)
+	moreResults := types.MakeArray(0)
+
+	for i := 0; i < len(resDocs); i++ {
+		if i < batchSize {
+			firstBatch.Append(resDocs[i])
+		} else {
+			moreResults.Append(resDocs[i])
+		}
+	}
```
rumyantseva, line 91: Maybe something like this (to have fewer cyclomatic complexity and fewer operations):

```suggestion
	for i := 0; i < batchSize; i++ {
		firstBatch.Append(resDocs[i])
	}
	
	for i := batchSize; i < len(resDocs); i++ {
		moreResults.Append(resDocs[i])
	}
```

Also I suggest moving the common logic from pg and tigris to common

internal/clientconn/conninfo/conn_info_test.go
```diff
@@ -72,3 +75,87 @@ func TestConnInfo(t *testing.T) {
...
+	for i := 0; i < runs; i++ {
+		wg.Add(1)
+
+		go func(i int) {
+			defer wg.Done()
+
+			ready <- struct{}{}
+
+			<-start
+			connInfo.SetCursor(fmt.Sprintf("cursor %d", i), nil)
```
rumyantseva, line 99:
```suggestion
			connInfo.SetCursor(fmt.Sprintf("cursor %d", i), nil)
			connInfo.Cursor(fmt.Sprintf("cursor %d", i))
```

Just because why not :) - to have more operations concurrently

internal/clientconn/conninfo/conn_info_test.go
```diff
@@ -72,3 +75,87 @@ func TestConnInfo(t *testing.T) {
...
+		}(i + 1000) // avoid setting the same cursor names.
+
+		go func(i int) {
+			defer wg.Done()
+
+			ready <- struct{}{}
+
+			<-start
+			connInfo.Cursor(fmt.Sprintf("cursor %d", i))
```
rumyantseva, line 154: Here we expect to receive some meaningful data, right (as `set` was called before)? Can we assert it?

internal/clientconn/conninfo/conn_info_test.go
```diff
@@ -72,3 +75,87 @@ func TestConnInfo(t *testing.T) {
...
+	for i := 0; i < runs/2; i++ {
+		wg.Add(2)
+
+		go func(i int) {
+			defer wg.Done()
+
+			ready <- struct{}{}
+
+			<-start
+			connInfo.SetCursor(fmt.Sprintf("cursor %d", i), nil)
```
rumyantseva, line 145: Let's set some data to check that we get it, as currently this test only tests concurrency bot not data correctness

internal/clientconn/conninfo/conn_info_test.go
```diff
@@ -72,3 +75,87 @@ func TestConnInfo(t *testing.T) {
 		})
 	}
 }
+
+func TestConnInfoCursorParallelWork(t *testing.T) {
```
rumyantseva, line 79: Let's add a comment to explain why exactly this test is needed and what we test here, this test doesn't have asserts, so it might make sense to explain why. Also, I think if it's possible, let's add some asserts to check data correctness.

rumyantseva (commented):
Looks good to me overall.

I added a couple of ideas and some questions about the cursor testing. Feel free to ping me for a call if there is something you'd like to discuss.

website/docs/reference/supported_commands.md
```diff
@@ -56,7 +56,10 @@ sidebar_position: 1
 |                 | `hint`                     | ❌      | Ignored                                                    |
 |                 | `comment`                  | ⚠️      | Not implemented in Tigris                                  |
 |                 | `let`                      | ⚠️      | Unimplemented                                              |
-| `getMore`       |                            | ❌      | [Issue](https://github.com/FerretDB/FerretDB/issues/1733)  |
+| `getMore`       |                            | ✅      | Basic command if fully supported                           |
```
Fashander, line 59:
```suggestion
| `getMore`       |                            | ✅      | Basic command is fully supported                           |
```

internal/clientconn/conninfo/conn_info_test.go
w84thesun: I added data to cursors and fill them with data that we assert when reading cursors.

internal/clientconn/conninfo/conn_info_test.go
w84thesun: Done.

internal/clientconn/conninfo/conn_info_test.go
w84thesun: Done.

integration/getmore_compat_test.go
```diff
@@ -0,0 +1,290 @@
...
+			compatNS, err := compatCursor.Get("ns")
+			require.NoError(t, err)
+
+			assert.Equal(t, targetNS, compatNS, "ns mismatch")
+		})
+	}
+}
+
+// getCursorID returns the cursor ID from a find command.
+func getCursorID(t *testing.T, ctx context.Context, targetCollection *mongo.Collection) any {
```
rumyantseva, line 220: Should it be in `helpers.go`?

rumyantseva (dismissed):
LGTM

integration/getmore_compat_test.go
w84thesun: Sure, moved it there.

rumyantseva (approved):
LGTM

chilagrow (approved):

Fashander (approved):

internal/clientconn/conninfo/conn_info.go
```diff
@@ -34,6 +36,16 @@ type ConnInfo struct {
 	rw       sync.RWMutex
 	username string
 	password string
+
+	curRW  sync.RWMutex
+	cursor map[string]iterator.Interface[int, any]
```
AlekSi, line 41: It is possible to have multiple cursor in one connection for the same collection:

```
test> for (var i = 0; i < 1000; i++) { db.test.insertOne({}) }
…

test> c1 = db.test.find({})
[
  { _id: ObjectId("63c91563be6127738499f256") },
  …
]

test> c2 = db.test.find({})
[
  { _id: ObjectId("63c91563be6127738499f256") },
  …
]

test> c1.next()
{ _id: ObjectId("63c91563be6127738499f26a") }

test> c1.next()
{ _id: ObjectId("63c91563be6127738499f26b") }

test> c2.next()
{ _id: ObjectId("63c91563be6127738499f26a") }

test> c2.next()
{ _id: ObjectId("63c91563be6127738499f26b") }
```

AlekSi (changes requested):

internal/clientconn/conninfo/conn_info.go
AlekSi: Let's use listener-scoped or handler-scoped atomic int64 as cursor ID and use it, not collection name

AlekSi: # 1733 says:
> and closes it when another command is issued

So let's do just that and keep one cursor per connection

w84thesun: Done.

Fashander (approved):

Fashander (approved):

internal/clientconn/conninfo/conn_info.go
```diff
@@ -53,6 +58,30 @@ func (connInfo *ConnInfo) SetAuth(username, password string) {
 	connInfo.password = password
 }
 
+// Cursor returns the cursor value stored.
+// We use "db.collection" as the key to get the cursor.
+func (connInfo *ConnInfo) Cursor(_ uint64) iterator.Interface[int, any] {
+	connInfo.curRW.RLock()
+	defer connInfo.curRW.RUnlock()
+
+	return connInfo.cursor
```
rumyantseva, line 67: It will panic if the cursor is not set, do we expect such behavior?

internal/clientconn/conn.go
```diff
@@ -127,15 +127,15 @@ func newConn(opts *newConnOpts) (*conn, error) {
 //
 // The caller is responsible for closing the underlying net.Conn.
 func (c *conn) run(ctx context.Context) (err error) {
-	var connInfo conninfo.ConnInfo
+	connInfo := new(conninfo.ConnInfo)
```
rumyantseva, line 130: this change doesn't seem to be needed anymore.

internal/clientconn/conninfo/conn_info_test.go
```diff
@@ -72,3 +77,39 @@ func TestConnInfo(t *testing.T) {
...
```
rumyantseva, line 81: Does the test name make sense? I don't see it checks parallel work

internal/handlers/common/msg_getmore.go
```diff
@@ -0,0 +1,159 @@
...
+		return nil, lazyerrors.Error(err)
+	}
+
+	if err = Unimplemented(document, "comment", "maxTimeMS"); err != nil {
+		return nil, err
+	}
+
+	db, err := GetRequiredParam[string](document, "$db")
+	if err != nil {
+		return nil, lazyerrors.Error(err)
```
rumyantseva, line 43: I guess it should be just smth like `return nil, err` like in other handlers

rumyantseva (commented):
Looks good to me overall, asked a couple of questions

internal/clientconn/conn.go
w84thesun: Fixed.

internal/clientconn/conninfo/conn_info_test.go
w84thesun: Fixed.

internal/clientconn/conninfo/conn_info.go
w84thesun: It will return `nil`, we test that behaviour in the unit test.

internal/clientconn/conninfo/conn_info.go
```diff
@@ -53,6 +58,30 @@ func (connInfo *ConnInfo) SetAuth(username, password string) {
 	connInfo.password = password
 }
 
+// Cursor returns the cursor value stored.
+// We use "db.collection" as the key to get the cursor.
```
chilagrow, line 62:
```suggestion
```

for now

internal/handlers/common/msg_getmore.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	batchSize, err := getBatchSize(document)
+	if err != nil {
+		return nil, err
+	}
+
+	connInfo := conninfo.Get(ctx)
+
+	cur := connInfo.Cursor(1)
```
chilagrow, line 80: Should this be the cursorID?
```suggestion
	cur := connInfo.Cursor(cursorID)
```

internal/handlers/common/msg_getmore.go
```diff
@@ -0,0 +1,159 @@
...
+			}
+
+			return nil, lazyerrors.Error(err)
+		}
+
+		resDocs.Append(doc)
+	}
+
+	// TODO: https://github.com/FerretDB/FerretDB/issues/1811
+	id := int64(1)
```
chilagrow, line 106: This is next cursor ID right? Suggest `nextCursorID`

chilagrow (commented):
I saw that multiple cursor in a collection is possible, but that can be implemented in the separate issue/PR if testing is complex.

internal/handlers/common/msg_getmore.go
w84thesun: For now it should be 1.

internal/handlers/common/msg_getmore.go
w84thesun: I suggest leaving it as is for now.

rumyantseva (approved):
In principle, I'm ready to merge it, especially as we have a follow-up task to clarify about cursor IDs.

chilagrow (approved):
LGTM :muscle:

Fashander (dismissed):

internal/handlers/handlers.go
```diff
@@ -155,4 +155,7 @@ type Interface interface {
 
 	// MsgWhatsMyURI returns peer information.
 	MsgWhatsMyURI(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error)
+
+	// MsgGetMore returns the next batch of documents from a cursor.
+	MsgGetMore(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error)
```
AlekSi, line 160: > // OP_MSG commands, sorted alphabetically

Please keep it this way

website/docs/reference/supported_commands.md
```diff
@@ -7,7 +7,7 @@ sidebar_position: 1
 ## Query commands
 
 | Command         | Argument                   | Status | Comments                                                   |
-| --------------- | -------------------------- | ------ | ---------------------------------------------------------- |
+| --------------- | -------------------------- | ------ |------------------------------------------------------------|
```
AlekSi, line 10: Note that other columns and all other tables have a padding space before and after `|`

internal/handlers/tigris/msg_find.go
```diff
@@ -73,18 +73,15 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
 
 	var reply wire.OpMsg
 	must.NoError(reply.SetSections(wire.OpMsgSection{
 		Documents: []*types.Document{must.NotFail(types.NewDocument(
 			"cursor", must.NotFail(types.NewDocument(
-				"firstBatch", firstBatch,
-				"id", int64(0), // TODO
+				"id", id,
 				"ns", fp.DB+"."+fp.Collection,
+				"firstBatch", firstBatch,
```
AlekSi, line 84: From what I can see, MongoDB returns fields in the old order: `firstBatch`, `id`, `ns`. By using the same order we make proxy's diff output much simpler to read

internal/handlers/pg/msg_find.go
AlekSi: You still need to read the whole function to check if there are other returns. But with `break`, you also need to read it backward from this point to check how many `for` there are.

`return` there is an extension of a general rule: return early. It is useful not only for error handling.

noisersup (dismissed):
That's a huge one! :1st_place_medal:

internal/clientconn/conninfo/conn_info.go
```diff
@@ -53,6 +58,29 @@ func (connInfo *ConnInfo) SetAuth(username, password string) {
...
+// SetCursor stores the cursor value.
+// We use "db.collection" as the key to store the cursor.
+func (connInfo *ConnInfo) SetCursor(iterator iterator.Interface[int, any]) {
+	connInfo.curRW.Lock()
+	defer connInfo.curRW.Unlock()
+
+	if connInfo.cursor != nil {
+		connInfo.cursor.Close()
+
+		connInfo.cursor = nil
```
AlekSi, line 78: That line does nothing, as `connInfo.cursor` is overwritten immediately after that

w84thesun:
> Why not connect the dots and make `fetchAndFilterDocs` return an iterator that we would use?

It would make this PR even bigger. Furthermore, it doesn't correspond with the idea of implementing this feature gradually, step by step.

website/docs/reference/supported_commands.md
w84thesun: Fixed.

internal/handlers/handlers.go
w84thesun: Fixed.

internal/handlers/tigris/msg_find.go
w84thesun: Restored order.

internal/clientconn/conninfo/conn_info.go
w84thesun: Removed that line.

internal/handlers/tigris/msg_find.go
w84thesun: I remembered why I changed that order. The compatibility test said that `firstBatch` should be in that order.

chilagrow (dismissed):
LGTM :muscle:

Fashander (approved):

rumyantseva:
> It would make this PR even bigger. Furthermore, it doesn't correspond with the idea of implementing this feature gradually, step by step.

@w84thesun is there a separate issue where these changes are mentioned to be implemented, or is it the same issue but a separate PR?

internal/handlers/common/msg_getmore.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	return &reply, nil
+}
+
+// getBatchSize returns the batch size from the document.
+func getBatchSize(doc *types.Document) (int64, error) {
+	batchSizeValue, err := doc.Get("batchSize")
+	if err != nil {
+		return 0, nil
```
rumyantseva, line 135: Why `0, nil` and not `0, err` (as we usually do)?

chilagrow: Because batchSize is optional parameter I think.

rumyantseva: Then I guess we should use `GetOptionalParam` instead (as we usually do).

rumyantseva: Though, I'm not sure if it's possible as is.

rumyantseva (commented):
LGTM overall, let's decide what's in the scope of the current issue and what's not.

For example, if Postgres iterator is for a separate issue, let's create such an issue and/or add a TODO to the code.

internal/handlers/common/msg_getmore.go
w84thesun: I introduced this function because we should return different errors when processing the `batchSize` parameter.

w84thesun:
> For example, if Postgres iterator is for a separate issue, let's create such an issue and/or add a TODO to the code.

We should talk about this on the DS call since there are several issues with Postgres cursor implementation.

internal/handlers/pg/msg_find.go
```diff
@@ -143,3 +193,66 @@ func (h *Handler) fetchAndFilterDocs(ctx context.Context, tx pgx.Tx, sqlParam *p
 		resDocs = append(resDocs, doc)
 	}
 }
+
+// getFirstBatchAndIterator fetches documents from the database and filters them using the provided sqlParam.Filter.
+func (h *Handler) getFirstBatchAndIterator(ctx context.Context, tx pgx.Tx, sqlParam *pgdb.SQLParam) (
+	[]*types.Document, iterator.Interface[int, *types.Document], error,
+) {
+	iter, err := pgdb.GetDocuments(ctx, tx, sqlParam)
```
chilagrow, line 201: Okay so this transaction can live long time potentially until all document from getMore is fetched. In terms of pg, this is not ideal.

And I suppose, it's here because our iterator assumes underling transaction exists.

Let me think on it. :thinking:

chilagrow (commented):
Just need time to think about transaction stuff with iterator. The rest looks good!

internal/handlers/pg/msg_find.go
AlekSi: That's fine: that transaction doesn't modifies data, and PostgreSQL's MVCC ensures that this transaction is relatively lightweight. It still consumes a connection, but that's true even without transaction, just with a cursor.

We still should be careful about closing both iterator and transaction, of course.

internal/clientconn/conninfo/conn_info.go
```diff
@@ -53,6 +77,116 @@ func (connInfo *ConnInfo) SetAuth(username, password string) {
...
+		}
+
+		if id < 0 {
+			id = int64(math.Abs(float64(id)))
+		}
+
+		_, ok := connInfo.cursors[id]
+		if !ok {
+			break
+		}
```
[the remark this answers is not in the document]
AlekSi: FWIW, I don't understand why we try so hard to generate a unique cursor ID without acquiring a lock. Just use a simple atomic counter, it costs nothing compared to network IO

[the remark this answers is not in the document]
AlekSi: (and I'm not sure why it is a method instead of a function)

internal/handlers/pg/msg_find.go
chilagrow: Good to know. Yes we need to be careful about closing them.

internal/clientconn/conninfo/conn_info.go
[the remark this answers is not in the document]
w84thesun: Changed to `atomic.AddInt64`.

chilagrow (dismissed):

rumyantseva (dismissed):
Wow wow!

I have a couple of questions (e.g. about `id` in `find`) but I don't want to block this PR, and I believe if something needs to be covered, we can cover it later. So, I'm ready to approve.

chilagrow (approved):

rumyantseva (approved):
LGTM

Fashander (approved):

noisersup (dismissed):
Checked again for things like data leaks etc and looks good to me! Everything looks heavily covered by tests (even parallel benchmarks)

AlekSi (commented):
Tests fail with:

expected: mongo.CommandError{Code:14, Message:"BSON field 'batchSize' is the wrong type 'array', expected type 'long'", Name:"TypeMismatch" [...]
actual  : mongo.CommandError{Code:43, Message:"cursor id 2 not found", Name:"CursorNotFound" [...]

internal/clientconn/conninfo/conn_info.go
```diff
@@ -34,6 +39,25 @@ type ConnInfo struct {
...
+func NewConnInfo() *ConnInfo {
+	return &ConnInfo{
+		cursors:       map[int64]Cursor{},
+		cursorCounter: 1,
+	}
+}
+
+// Cursor represents a cursor.
+type Cursor struct {
+	tx     pgx.Tx
```
AlekSi, line 58: With #1947, we could store only one iterator with a custom close function that would commit tx

integration/getmore_compat_test.go
```diff
@@ -0,0 +1,257 @@
...
+
+	targetCollection := s.TargetCollections[0]
+	compatCollection := s.CompatCollections[0]
+
+	ctx := s.Ctx
+
+	for name, tc := range testCases {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
```
b1ron, line 147: The unexpected errors don't seem to be related to sessions.

Maybe it's a goroutine processing data in parallel because removing the above call to `t.Parallel()` seems to avoid the issue, not sure.

PR #1778, opened 2023-01-11
Enable support for arrays in Tigris

### LINKED ISSUE #1704: Enable support for arrays in Tigris
This task is a continuation of #908.

After having arrays in `tjson`, we can integrate them and write tests.

A possible approach could be the following:
- Either creation of a collection or the first insertion of array should define the schema for a particular array field.
- If validation doesn't pass on Tigris side (an array contains various data types, attempt to insert an array of a type different than the one defined before etc), an error should be returned to the client.

### Testing

- Positive test cases can be covered with compat tests.
- Negative test cases can't be written unless we have validation for Mongo collections (where we can enable the same schema rules as the ones we have for Tigris).
- Diff tests are not possible because they don't work with Tigris.

Test case to check manually:
1. I insert an array to a new collection to a field called `tags`: `db.things.insertOne( { tags: ["a", "b", "c"] } );`
2. As the array contains data of the same type, insertion pass. The schema should describe string-based array.
3. I try to insert another array to the same field: `db.things.insertOne( { tags: [1, 2, 3] } );`
4. I should receive an error as the data type doesn't match (I try to insert `int`, not `string`).

### Definition of done
- [x] Add a valid data item to `sharedata.Composites`, something like `DocumentsDocuments`, but for arrays.
- [x] Add compat tests to work with arrays. Try to reuse existing array-related compat tests and shared data. Cases to cover (see TODOs in the code):
  - [x] create;
  - [x] insert, upsert, update;
  - [x] queries - array operations.
- [x] Add negative test cases (invalid or empty arrays) at least for unit tests.

rumyantseva:
# Description

Closes #1704.

ℹ️ I increased Tigris timeout to 20 min as with the enabled tests the default one wasn't enough.

There are four types of tests skipped for Tigris:
-  arrays with elements of mixed types (only arrays of the same type are supported in Tirgis)
- arrays and objects with `null` values (not supported in Tigris)
- fields with `array` name (not supported in Tigris as Tigris considers `array` a keyword)
- fields can't start from numbers (e.g. `42` is not a valid field name)

## Readiness checklist

* [x] I added integration tests for new functionality or bug fixes.


integration/query_array_compat_test.go
```diff
@@ -145,6 +139,7 @@ func TestQueryArrayCompatElemMatch(t *testing.T) {
 			},
 		},
 		"GtZeroWithTypeString": {
+			// A document like {"v":[42, "foo"]} matches this filter (there is an elem >0 and an elem of type string)
```
chilagrow, line 142: I need a bit more context on this. If I understand correctly `Composites` dataset is not used for tigris handler, not quite sure where `{"v":[42, "foo"]}` comes from :thinking:

integration/query_comparison_compat_test.go
```diff
@@ -121,43 +120,58 @@ func TestQueryComparisonCompatImplicit(t *testing.T) {
...
+			}}},
+			skipForTigris: "Tigris does not support mixed types in arrays",
 		},
 		"DocumentShuffledKeys": {
-			filter:     bson.D{{"v", bson.D{{"$eq", bson.D{{"42", "foo"}, {"array", bson.A{int32(42), "foo", nil}}, {"foo", int32(42)}}}}}},
+			filter: bson.D{{"v", bson.D{
+				{"$eq", bson.D{
+					{"42", "foo"}, {"array", bson.A{int32(42), "foo", nil}}, {"foo", int32(42)},
+				}},
+			}}},
```
chilagrow, line 139:
```suggestion
			}}},
			skipForTigris: "Tigris does not support mixed types in arrays",
```

chilagrow (commented):
Just tiny comments :hugs:

integration/query_array_compat_test.go
rumyantseva: The filter is set as
```
					{"$elemMatch", bson.D{
						{"$gt", int32(0)},
					}},
					{"$type", "string"},
```

it needs one element to match >0 and one element to be a string. It's not possible to define such array in Tigris. For PG, a case with smth like `{"v":[42, "foo",nill]}` would match.

chilagrow: Thanks, got it!

w84thesun (approved):
LGTM!

noisersup (approved):
LGTM! Nice Job with finding these tigris compatibility edge cases

AlekSi (approved):

PR #1790, opened 2023-01-12
Integrate `explain` into tests

### LINKED ISSUE #1279: Integrate `explain` into tests
We should integrate `explain` command invocation into our integration tests. That should show us that the existing simple query pushdown actually works.

### DoD

1. Extend `queryCompatTestCase` with `resultPushdown` field:

```go
type queryCompatTestCase struct {
	filter     bson.D                   // required
	sort       bson.D                   // defaults to `bson.D{{"_id", 1}}`
	projection bson.D                   // nil for leaving projection unset
	resultType compatTestCaseResultType // defaults to nonEmptyResult
	resultPushdown bool 
}
```

In the `testQueryCompat` test itself add calling `explain` for target system only. 
Compare the result of `explain` with `queryCompatTestCase.resultPushdown`.

For compat tests, we should not check `explain` results for the compat system (MongoDB).

2. Add `pushdown: true` to the explain result for Tigris/PG and compare with that field. 

```go
	err = reply.SetSections(wire.OpMsgSection{
		Documents: []*types.Document{must.NotFail(types.NewDocument(
			"queryPlanner", queryPlanner,
			"pushdown": /* */
			"explainVersion", "1",
			"command", cmd,
			"serverInfo", serverInfo,
			"ok", float64(1),
		))},
	})
```

The field "pushdown" should be set based on the presence of WHERE clause for pg and filter for Tigris.

3. Only queries are supported now, `insert` is not in the scope.
4. If this approach doesn't work, investigate a better approach and create another task that should be top prior.

noisersup:
# Description

Closes #1279.

I've realized that we don't need to do any refactoring on pgdb/tigris layer to check if pushdown occured, as all required data is provided by query planners.

There's also a small change in `NewPath` function, as it seemed more intuitive for its use case. If using "standard" slice notation there had any particular reason, please let me know! 

I've also increased tigris integration tests timeout to 20mins (same as in #1778), so @AlekSi approval might be necessary, depending on which one is going to be merged first.

## Readiness checklist


AlekSi:
> I've also increased tigris integration tests timeout to 20mins (same as in https://github.com/FerretDB/FerretDB/pull/1778)

Please update it to 25 so there is no merge conflict

internal/types/path.go
```diff
@@ -70,7 +70,7 @@ type Path struct {
 }
 
 // NewPath returns Path from a strings slice.
-func NewPath(path []string) Path {
+func NewPath(path ...string) Path {
```
rumyantseva, line 73: I like this change! To me this approach looks more readable 👍

internal/handlers/pg/msg_explain.go
```diff
@@ -79,6 +79,9 @@ func (h *Handler) MsgExplain(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 		return nil, err
 	}
 
+	// if filter was set, it means that pushdown was done
+	pushdown := queryPlanner.HasByPath(types.NewPath("Plan", "Filter"))
```
rumyantseva, line 83: I'm not sure if it's enough. Having a filter doesn't guarantee that pushdown was set. 
Pushdown was set if `pgdb.prepareWhereClause` returned a non-empty string. Otherwise, it wasn't.

Taskfile.yml
```diff
@@ -185,7 +185,7 @@ tasks:
     cmds:
       - >
         go test -count=1 {{.RACEFLAG}} -tags={{.BUILDTAGS}} -shuffle=on -coverpkg=../...
-        -timeout=15m
+        -timeout=20m
```
rumyantseva, line 188: I increased to 25 :)

internal/handlers/pg/msg_explain.go
noisersup: At the beginning I've started with this approach, but the fact that we need to return another value just for pushdowns (while it's unused in many places) bothered me a bit. I realised, that the `Plan.Filter` from PostrgreSQL EXPLAIN command shows exactly what filter (everything after WHERE clause) what passed, and most importantly - it is already provided to the msg_explain handler.

I don't understand in what cases we can't guarantee that the pushdown is not detected correctly with this approach

Taskfile.yml
noisersup: Ouch!

internal/handlers/pg/msg_explain.go
rumyantseva: Hmmm... Ok, got it!

Looks like it works! Then I like the simplicity of your choice! ;)


Will leave an example for more context. 

With `where`:
```
EXPLAIN SELECT * FROM test."ArrayStrings_7a18b54c" WHERE _jsonb='{}';
[
  {
    "QUERY PLAN" : "Seq Scan on \"ArrayStrings_7a18b54c\"  (cost=0.00..27.00 rows=7 width=32)"
  },
  {
    "QUERY PLAN" : "  Filter: (_jsonb = '{}'::jsonb)"
  }
]
```

Without `where`:

```
EXPLAIN SELECT * FROM test."ArrayStrings_7a18b54c";

Seq Scan on "ArrayStrings_7a18b54c"  (cost=0.00..23.60 rows=1360 width=32)
```

Taskfile.yml
noisersup: Fixed

internal/handlers/tigris/msg_explain.go
```diff
@@ -53,10 +53,15 @@ func (h *Handler) MsgExplain(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
 
+	queryFilter := string(h.db.BuildFilter(filter))
+
 	queryPlanner := must.NotFail(types.NewDocument(
-		"Filter", string(h.db.BuildFilter(filter)),
+		"Filter", queryFilter,
 	))
 
+	// if tigris query filter was set, it means, the pushdown was done
+	pushdown := queryFilter != "{}"
```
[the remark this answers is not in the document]
rumyantseva: Ok, and I'm not right here :) Because this is the exact case you check :)

integration/query_compat_test.go
```diff
@@ -32,7 +32,7 @@ type queryCompatTestCase struct {
 	sort           bson.D                   // defaults to `bson.D{{"_id", 1}}`
 	projection     bson.D                   // nil for leaving projection unset
 	resultType     compatTestCaseResultType // defaults to nonEmptyResult
-	resultPushdown bool                     // TODO https://github.com/FerretDB/FerretDB/issues/1279
+	resultPushdown bool                     // deafault to false
```
rumyantseva, line 35: Ah, now I got it! So, it's false by default, and in the test cases where pushdown is used, you added true!

rumyantseva (approved):
LGTM!

I like the solution! It took me a while to understand why it works, but I like the simplicity of it!

integration/query_compat_test.go
noisersup: Yeah I believe it's good for now, although I can see that if we start pushdown more generic filters, it will need to be set for too much test cases. But we can make this check more complex in any time :)

rumyantseva:
@noisersup I think you need `task fmt` ;)

rumyantseva (dismissed):
LGTM

integration/query_compat_test.go
chilagrow, line 35:
```suggestion
	resultPushdown bool                     // defaults to false
```

integration/query_compat_test.go
```diff
@@ -80,9 +80,17 @@ func testQueryCompat(t *testing.T, testCases map[string]queryCompatTestCase) {
...
+						{"find", targetCollection.Name()},
+						{"filter", filter},
+						{"sort", opts.Sort},
+						{"projection", opts.Projection},
+					}}}
+
+					var explainRes bson.D
+					require.NoError(t, targetCollection.Database().RunCommand(ctx, explainQuery).Decode(&explainRes))
+
+					assert.Equal(t, tc.resultPushdown, explainRes.Map()["pushdown"])
```
chilagrow, line 93: do we want to check the result found by push down query?

chilagrow (commented):
Just minor comments :)

integration/query_compat_test.go
noisersup: I believe it's not in the scope of this issue. We need a simple check to have any way to test the future pushdowns

rumyantseva (approved):
LGTM

integration/query_compat_test.go
chilagrow: :+1:

chilagrow (approved):

PR #1793, opened 2023-01-13
Simplify `types.Array` `Append` signature

w84thesun:
# Description

In this PR I remove `error` from the `Append` result parameters as it's always nil.

## Readiness checklist

### DIFF
--- a/internal/handlers/common/distinct.go
+++ b/internal/handlers/common/distinct.go
@@ -116,18 +116,12 @@ func FilterDistinctValues(docs []*types.Document, key string) (*types.Array, err
 
 				if !distinct.Contains(el) {
 					distinct.Append(el)
-					if err != nil {
-						return nil, lazyerrors.Error(err)
-					}
 				}
 			}
 
 		default:
 			if !distinct.Contains(v) {
 				distinct.Append(v)
-				if err != nil {
-					return nil, lazyerrors.Error(err)
-				}
 			}
 		}
 	}
--- a/internal/handlers/common/distinct.go
+++ b/internal/handlers/common/distinct.go
@@ -115,7 +115,7 @@ func FilterDistinctValues(docs []*types.Document, key string) (*types.Array, err
 				}
 
 				if !distinct.Contains(el) {
-					err := distinct.Append(el)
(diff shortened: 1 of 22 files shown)


internal/handlers/common/distinct.go
```diff
@@ -115,7 +115,7 @@ func FilterDistinctValues(docs []*types.Document, key string) (*types.Array, err
 				}
 
 				if !distinct.Contains(el) {
-					err := distinct.Append(el)
+					distinct.Append(el)
 					if err != nil {
```
AlekSi, line 119: I guess that `if` should go away too

AlekSi (commented):
TBH, I'm a bit skeptical about that PR. If we ever decide to return some errors from `Append` (for example, validation errors) it will be too easier to forget to add error checking to all invocations back

internal/handlers/common/distinct.go
rumyantseva: Yeah, this `if` is not needed.

(Someone should teach me how to do `if err := distinct.Append(el); err != nil {` next time 😅)

internal/handlers/common/distinct.go
```diff
@@ -124,7 +124,7 @@ func FilterDistinctValues(docs []*types.Document, key string) (*types.Array, err
 
 		default:
 			if !distinct.Contains(v) {
-				err := distinct.Append(v)
+				distinct.Append(v)
 				if err != nil {
```
rumyantseva, line 128: Another `err` that is not needed

rumyantseva (changes requested):
A couple of errors doesn't need to be checked anymore, but I like the idea of this improvement!

w84thesun:
> TBH, I'm a bit skeptical about that PR. If we ever decide to return some errors from `Append` (for example, validation errors) it will be too easier to forget to add error checking to all invocations back

I believe that linters could warn about that and it's better to keep that function simple as it is now.

chilagrow (approved):
Thanks for the simplification.

rumyantseva (approved):
LGTM

rumyantseva (approved):
LGTM

PR #1809, opened 2023-01-16
Support PostgreSQL pushdowns for numbers

### LINKED ISSUE #1755: Support pushdowns for equal operators on numbers
### What should be done?

Currently, we only pushdown queries where `_id` is equal to any `ObjectId` or `string`.
We should generalize pushdowns to all keys and most importantly - initialize pushdowns for a simple number comparison. 
- We should only pushdown queries where we check if numbers from all numeric types are equal (other comparison types are not in a scope)
- After we introduce pushdown tests in #1279, there shouldn't be any problem with testing.
- We shouldn't pushdown number comparison with any non numerical type, although we need to make sure that we don't omit any data by not including it in pushdown query response because of different types
- don't recurse yet, we shouldn't check for array items and subdocuments. I guess we could exclude every filter which tries to compare in array/subdocument, but please create test cases for that
 
## DoD
- Pushdown filters that check if a field is equal to any number from any numeric type
- Update and create compat tests to include enough edge cases related to type conversions (e.g. using fields of different types)
- Focus on tests related to comparing nonnumerical types with numerical to make sure that we don't leave any data (unit tests might be also a good idea)
- Implement the straightforward way, no type conversion or complicate cases. If we find that type conversion is needed or some cases work on Postgres/Tigris, create separate issue(-s)
- If without conversion noting is possible, we consider the current task a research task and create separate issues based on the research result

noisersup:
# Description

Closes #1755.

Have managed to pushdown all numeric types, strings and ObjectIDs for all the keys. 
Operators like "$eq", dot notation and tigris pushdowns are not covered in this PR.

## Readiness checklist

* [x] I added integration tests for new functionality or bug fixes.


integration/basic_test.go
```diff
@@ -223,6 +223,26 @@ func TestDeleteCommentQuery(t *testing.T) {
 	assert.Equal(t, expected, res)
 }
 
+func TestEmptyKey(t *testing.T) {
```
noisersup, line 226: Added integration test to make sure that we don't skip empty keys by accident because of current pushdowns or some future ones

internal/handlers/tigris/tigris.go
```diff
@@ -69,6 +69,7 @@ func New(opts *NewOpts) (handlers.Interface, error) {
 		NewOpts: opts,
 		db:      db,
 	}
+
```
noisersup, line 72:
```
internal/handlers/tigris/tigris.go:72:2: return statements should not be cuddled if block has more than two lines (wsl)
        return h, nil
```

internal/handlers/tigris/msg_getlog.go
```diff
@@ -83,7 +84,9 @@ func (h *Handler) MsgGetLog(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 		))
 
 	case "startupWarnings":
-		info, err := h.db.Driver.Info(ctx)
```
noisersup, line 86: internal/handlers/tigris/msg_getlog.go:86:9: shadow: declaration of "err" shadows declaration at line 36 (govet)
                info, err := h.db.Driver.Info(ctx)

integration/shareddata/scalars.go
```diff
@@ -174,7 +174,7 @@ var Strings = &Values[string]{
 					"foo": {"type": "integer", "format": "int32"}, 
 					"bar": {"type": "array", "items": {"type": "string"}},
 					"v": {"type": "string"},
-					"_id": {"type": "string"}
+					"_id": {"type": "string"},
```
rumyantseva, line 177: That doesn't look like a valid JSON, I will be surprised if it doesn't fail somewhere :)

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,20 +246,37 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
+					// If it's not, but the value under the key k is an array - select if it contains the value equal to v.
+					` OR (jsonb_typeof(_jsonb->%[1]s) = 'array' AND`+
+					` EXISTS (SELECT 1 FROM jsonb_array_elements(_jsonb->%[1]s) as elem WHERE (elem::jsonb = %[2]s)))`,
+				p.Next(), // $1 = k
+				p.Next(), // $2 = v
+			)
+
+			filters = append(filters, sql)
+			args = append(args, k)
+			args = append(args, string(must.NotFail(pjson.MarshalSingleValue(v))))
```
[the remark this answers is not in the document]
noisersup: Yes! Thanks! :smile:

integration/shareddata/scalars.go
noisersup: Nice catch! It failed

chilagrow:
Renamed PR because we don't support equal operators (`$eq`) yet

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,20 +246,38 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
+		// don't pushdown query operators and dot notation yet
+		if len(k) != 0 && (k[0] == '$' || types.NewPathFromString(k).Len() > 1) {
+			continue
+		}
 
-			case types.ObjectID:
+		// don't iterate through array for _id keys to simplify the query
+		if k == "_id" {
+			switch v := v.(type) {
+			case types.ObjectID, string:
```
chilagrow, line 257: It looks to me that we only support string and ObjectID query pushdown for `_id` while we support more for other field? :thinking: Maybe I'm misunderstanding

integration/query_comparison_compat_test.go
```diff
@@ -68,23 +68,55 @@ func TestQueryComparisonCompatImplicit(t *testing.T) {
 			filter:     bson.D{{"v.some.0", bson.A{42}}},
 			resultType: emptyResult,
 		},
+		"Int32": {
```
chilagrow, line 71: Do we have a test for if _id is pushed down? I think we should have for each data type of ID.

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,20 +246,38 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
 	for k, v := range sqlFilters.Map() {
-		switch k {
-		case "_id":
-			switch v := v.(type) {
-			case string:
-				filters = append(filters, fmt.Sprintf(`((_jsonb->'_id')::jsonb = %s)`, p.Next()))
-				args = append(args, string(must.NotFail(pjson.MarshalSingleValue(v))))
+		// don't pushdown query operators and dot notation yet
+		if len(k) != 0 && (k[0] == '$' || types.NewPathFromString(k).Len() > 1) {
+			continue
```
chilagrow, line 251: Shall we create an issue and add todo here :hugs:

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,20 +246,38 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
 				filters = append(filters, fmt.Sprintf(`((_jsonb->'_id')::jsonb = %s)`, p.Next()))
 				args = append(args, string(must.NotFail(pjson.MarshalSingleValue(v))))
 			}
-		default:
+
 			continue
 		}
+
+		switch v := v.(type) {
+		case string, int32, int64, types.ObjectID, float32, float64:
```
chilagrow, line 266: Let's order them to align with our convention and we don't have float32.
https://github.com/FerretDB/FerretDB/blob/main/CONTRIBUTING.md#code-style-and-conventions.

I suggest listing all data types and `continue` on the types we don't support yet. :hugs: That way, it's easier to see what's not supported

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,20 +246,38 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
+			sql := fmt.Sprintf(
+				`((_jsonb->%[1]s)::jsonb = %[2]s)`+ // Select if value under the key k is equal to value v.
+					// If it's not, but the value under the key k is an array - select if it contains the value equal to v.
+					` OR (jsonb_typeof(_jsonb->%[1]s) = 'array' AND`+
+					` EXISTS (SELECT 1 FROM jsonb_array_elements(_jsonb->%[1]s) as elem WHERE (elem::jsonb = %[2]s)))`,
+				p.Next(), // $1 = k
+				p.Next(), // $2 = v
```
chilagrow, line 273: Can we escape or sanitize before using p.Next()?

chilagrow (commented):
Some early comments.

internal/handlers/pg/pgdb/query.go
noisersup: Exactly. As for `_id`s we support only strings and ObjectIDs, there's no need to send the SQL query that would check if `_id` field is an array.

This technically doesn't change much in performance, as PostgreSQL won't evaluate the whole query if it doesn't need to, although from logging perspecitve, and further debugging, it seems reasonable to keep queries as simple as possible

wdyt?

internal/handlers/pg/pgdb/query.go
noisersup: Yeah, I was planning to add further issues but wanted to discuss it as I'm not sure what's currently in scope of our roadmap. Thanks for noticing that we could add TODO here!

internal/handlers/pg/pgdb/query.go
noisersup: Whoops! I leave too many thing while I'm debugging :smile: nice catch!

internal/handlers/pg/pgdb/query.go
noisersup: Where? We don't push data directly to the query, but use the arguments. We've also used the same approach before this PR. I guess I don't understand

integration/query_comparison_compat_test.go
noisersup: https://github.com/FerretDB/FerretDB/pull/1809/files#diff-eee22386d31e852a2d632d26269d2856e7c3fa7b58d03407c9ff6d344aba1b66R176 We do

internal/handlers/pg/pgdb/query.go
chilagrow: `p.Next()` must be escaped or sanitized, so it does not contain harmful query.

>  We've also used the same approach before this PR. I guess I don't understand
Please fix them as well please :pray:

chilagrow: I see, I misunderstood initially. it's value is `$2`, not `v` which I thought was a value of something :see_no_evil:.

rumyantseva (dismissed):
The solution looks reasonable to me!

I also like the comments that explain the query. 
The tests proof that it works.

w84thesun (approved):
Nice job! 💪

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,20 +247,42 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
+			// TODO $eq and $ne https://github.com/FerretDB/FerretDB/issues/1840
+			// TODO dot notation https://github.com/FerretDB/FerretDB/issues/1841
+			continue
+		}
 
-			case types.ObjectID:
+		// don't iterate through array for _id keys to simplify the query
+		if k == "_id" {
+			switch v := v.(type) {
+			case string, types.ObjectID:
```
AlekSi, line 259: Shouldn't we also include numbers there?

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,20 +247,42 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
-		default:
+
 			continue
 		}
+
+		switch v := v.(type) {
+		case *types.Document, *types.Array, types.Binary, bool, time.Time, types.NullType, types.Regex, types.Timestamp:
+			continue
+
+		case float64, string, types.ObjectID, int32, int64:
```
AlekSi, line 271: If out intention is to enumerate all types there, let's add a `default:` that would panic

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,20 +247,42 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
+			sql := fmt.Sprintf(
+				`((_jsonb->%[1]s)::jsonb = %[2]s)`+ // Select if value under the key k is equal to value v.
```
AlekSi, line 273: Should that part include a condition that $1 is _not_ an array?

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,20 +247,42 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
+					// If it's not, but the value under the key k is an array - select if it contains the value equal to v.
+					` OR (jsonb_typeof(_jsonb->%[1]s) = 'array' AND`+
+					` EXISTS (SELECT 1 FROM jsonb_array_elements(_jsonb->%[1]s) as elem WHERE (elem::jsonb = %[2]s)))`,
```
AlekSi, line 276: It think that could be simplified using `@>` operator. See https://www.postgresql.org/docs/current/datatype-json.html#JSON-CONTAINMENT
> SELECT '["foo", "bar"]'::jsonb @> '"bar"'::jsonb;

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,20 +247,42 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
+				p.Next(), // $1 = k
+				p.Next(), // $2 = v
+			)
+
+			filters = append(filters, sql)
+			args = append(args, k,
+				string(must.NotFail(pjson.MarshalSingleValue(v))),
+			)
```
AlekSi, line 284: See #593

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,20 +247,42 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
+
+		switch v := v.(type) {
+		case *types.Document, *types.Array, types.Binary, bool, time.Time, types.NullType, types.Regex, types.Timestamp:
+			continue
+
+		case float64, string, types.ObjectID, int32, int64:
+			sql := fmt.Sprintf(
+				`((_jsonb->%[1]s)::jsonb = %[2]s)`+ // Select if value under the key k is equal to value v.
+					// If it's not, but the value under the key k is an array - select if it contains the value equal to v.
+					` OR (jsonb_typeof(_jsonb->%[1]s) = 'array' AND`+
```
chilagrow, line 275: Do we have tests for array? We should have one that matches and one that doesn't match this condition. Also empty array should be tested :hugs:

integration/query_comparison_compat_test.go
```diff
@@ -69,23 +69,55 @@ func TestQueryComparisonCompatImplicit(t *testing.T) {
...
+			filter:             bson.D{{"v", int32(42)}},
+			resultPushdown:     true,
+			skipTigrisPushdown: true,
+		},
+		"Int64": {
+			filter:             bson.D{{"v", int64(42)}},
+			resultPushdown:     true,
+			skipTigrisPushdown: true,
+		},
```
chilagrow, line 81:
```suggestion
		},
		"Double": {
			filter:             bson.D{{"v", 42}},
			resultPushdown:     true,
			skipTigrisPushdown: true,
		},
```
This looks like interesting case, because I think it should match.

internal/handlers/pg/pgdb/query.go
noisersup: We already test for that. For example when we run int64 or int32 tests they'll search for top level `42` values but also for arrays with `42`:

```
=== RUN   TestQueryComparisonCompatImplicit/Int64/TestQueryComparisonCompatImplicit_ArrayInt32s
=== CONT  TestQueryComparisonCompatImplicit/Int64/TestQueryComparisonCompatImplicit_ArrayInt32s
    query_comparison_compat_test.go:181: Compat (expected) IDs: [array-int32-one array-int32-three array-int32-two]
    query_comparison_compat_test.go:181: Target (actual)   IDs: [array-int32-one array-int32-three array-int32-two]
        --- PASS: TestQueryComparisonCompatImplicit/Int64/TestQueryComparisonCompatImplicit_ArrayInt32s (0.01s)
```

integration/query_comparison_compat_test.go
chilagrow: The interesting thing here is that if int64(42) in [int64](https://github.com/FerretDB/FerretDB/blob/main/integration/shareddata/scalars.go#L97) and int32(42) in [int32](https://github.com/FerretDB/FerretDB/blob/main/integration/shareddata/scalars.go#L86) matches `float64(42)` or not. As I understand float64(42) is same as 42. We have test for double 42.13 but not for double 42.

chilagrow: Actually the test I'm looking for is already there `ValueNumber` :see_no_evil: My bad

internal/handlers/pg/pgdb/query.go
chilagrow: Thank :+1:

internal/handlers/pg/pgdb/query.go
noisersup: Looks much simpler!

internal/handlers/pg/pgdb/query.go
noisersup: fixed

integration/query_comparison_compat_test.go
chilagrow: I think the filter you meant to add was `_id` instead of `v`. Something like this?
`filter: bson.D{{"_id", primitive.ObjectID{0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x10, 0x11}},`

`filter: bson.D{{"_id", primitive.ObjectID{0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01}},`

Also lets add one for int64, int32 and float64 :pray:

internal/handlers/pg/pgdb/query.go
noisersup: fixed

internal/handlers/pg/pgdb/query.go
noisersup: Yeah we should, fixed and covered with tests

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,19 +247,46 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
+			// TODO $eq and $ne https://github.com/FerretDB/FerretDB/issues/1840
+			// TODO dot notation https://github.com/FerretDB/FerretDB/issues/1841
+			continue
+		}
 
-			case types.ObjectID:
+		// don't iterate through array for _id keys to simplify the query
+		if k == "_id" {
+			switch v := v.(type) {
+			case float64, string, types.ObjectID, int32, int64:
```
chilagrow, line 259: Let's do the same as below by `panic` on `default`, for the consistency :hugs:

noisersup: good point! fixed

chilagrow (approved):

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,19 +247,50 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
+			// TODO dot notation https://github.com/FerretDB/FerretDB/issues/1841
+			continue
+		}
 
-			case types.ObjectID:
+		// don't iterate through array for _id keys to simplify the query
+		if k == "_id" {
+			switch v := v.(type) {
+			case *types.Document, *types.Array, types.Binary, bool, time.Time, types.NullType, types.Regex, types.Timestamp:
+				// Don't do pushdown for invalid _id types
```
rumyantseva, line 260: Why do we consider those `_id` types invalid? If I'm not mistaken, `Array` is invalid, others are possible (maybe not all of them make sense, but I wouldn't use the word "invalid").

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,19 +247,50 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
+		switch v := v.(type) {
+		case *types.Document, *types.Array, types.Binary, bool, time.Time, types.NullType, types.Regex, types.Timestamp:
 			continue
+
+		case float64, string, types.ObjectID, int32, int64:
+			sql := fmt.Sprintf(
+				// Select if value under the key k is equal to value v.
+				`((_jsonb->%[1]s)::jsonb = %[2]s)`+
+					// If it's not, but the value under the key k is an array - select if it contains the value equal to v.
+					` OR (jsonb_typeof(_jsonb->%[1]s) = 'array' AND`+
```
rumyantseva, line 280: What happens if we use `(_jsonb->%[1]s)::jsonb @> %[2]s)` condition without  `jsonb_typeof(_jsonb->%[1]s) = 'array'`?

rumyantseva (commented):
LGTM overall, asked a couple of extra questions for my better understanding.

internal/handlers/pg/pgdb/query.go
noisersup: Oh, you're right. It's possible to use some of them as _ids

internal/handlers/pg/pgdb/query.go
noisersup: It looks correct! Fixed that

integration/query_comparison_compat_test.go
noisersup: Yeah, it seems so, although it's still worth to make a check for ObjectID for other keys than `_id`.

The checks for all the types for `_id` field are already implemented:

https://github.com/FerretDB/FerretDB/pull/1809/files#diff-b3f97e3a3514a7a7ff2a9104a0fd6e913c90b0eab5d079ea9e2ebcdeeb4e4b31R138-R164

rumyantseva (approved):
LGTM

integration/query_compat_test.go
```diff
@@ -28,13 +28,14 @@ import (
...
-	skipForTigris  string                   // skip test for Tigris
-	skip           string                   // skip test for all backends, myst have issue number mentioned
+	filter             bson.D                   // required
+	sort               bson.D                   // defaults to `bson.D{{"_id", 1}}`
+	projection         bson.D                   // nil for leaving projection unset
+	resultType         compatTestCaseResultType // defaults to nonEmptyResult
+	resultPushdown     bool                     // defaults to false
+	skipTigrisPushdown bool                     // defaults to false
+	skipForTigris      string                   // skip test for Tigris
+	skip               string                   // skip test for all backends, myst have issue number mentioned
```
chilagrow, line 38:
```suggestion
	skip               string                   // skip test for all backends, must have issue number mentioned
```

chilagrow (commented):
Some comments :hugs:

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,19 +247,50 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
-				filters = append(filters, fmt.Sprintf(`((_jsonb->'_id')::jsonb = %s)`, p.Next()))
-				args = append(args, string(must.NotFail(pjson.MarshalSingleValue(v))))
+		if len(k) != 0 && (k[0] == '$' || types.NewPathFromString(k).Len() > 1) {
+			// TODO $eq and $ne https://github.com/FerretDB/FerretDB/issues/1840
+			// TODO dot notation https://github.com/FerretDB/FerretDB/issues/1841
+			continue
+		}
 
-			case types.ObjectID:
+		// don't iterate through array for _id keys to simplify the query
```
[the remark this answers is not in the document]
noisersup: we generate different queries for `_id`s that doesn't check for arrays, as _id cannot be the array. It's purely because of simplification reasons

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,19 +247,50 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
 			continue
+
+		case float64, string, types.ObjectID, int32, int64:
+			sql := fmt.Sprintf(
+				// Select if value under the key k is equal to value v.
+				`((_jsonb->%[1]s)::jsonb = %[2]s)`+
+					// If it's not, but the value under the key k is an array - select if it contains the value equal to v.
+					` OR (_jsonb->%[1]s)::jsonb @> %[2]s`,
+				p.Next(), // $1 = k
+				p.Next(), // $2 = v
```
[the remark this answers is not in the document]
noisersup: good idea!

rumyantseva (approved):
LGTM

internal/handlers/pg/pgdb/query.go
```diff
@@ -246,19 +247,50 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
+		case *types.Document, *types.Array, types.Binary, bool, time.Time, types.NullType, types.Regex, types.Timestamp:
+			// type not supported for pushdown
 			continue
+
+		case float64, string, types.ObjectID, int32, int64:
+			sql := fmt.Sprintf(
+				// Select if value under the key k is equal to value v.
+				`((_jsonb->%[1]s)::jsonb = %[2]s)`+
+					// If it's not, but the value under the key k is an array - select if it contains the value equal to v.
+					` OR (_jsonb->%[1]s)::jsonb @> %[2]s`,
```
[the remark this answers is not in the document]
noisersup: containment operator "@>" while it's working on jsonb values will check if the value on the left contain at the top level the value on the right. If the left value (let's say field 'v') is an array it will check if it contains the value from the right, if it's the jsonb document like `{v:{"foo":42,"boo":32}}` then it will check if the value on the right is present inside it. 

If the value on the left is the jsonb value like '"foo"' or '42', then the @> comparator will compare it with the value on the right, it doesn't result in any error, so *technically* it's possible to simplify the where clause to just `WHERE (_jsonb->%[1]s)::jsonb @> %[2]s`, although I wouldn't do that, as it's not the proper usage of the operator, we're not sure how this comparison behaves in edge cases, and it's harder to understand when you read the code.

integration/query_compat_test.go
```diff
@@ -90,7 +91,13 @@ func testQueryCompat(t *testing.T, testCases map[string]queryCompatTestCase) {
...
 					require.NoError(t, targetCollection.Database().RunCommand(ctx, explainQuery).Decode(&explainRes))
 
-					assert.Equal(t, tc.resultPushdown, explainRes.Map()["pushdown"])
+					// temporarily make possible to skip tigris pushdown checks
+					if tc.skipTigrisPushdown && setup.IsTigris(t) {
+						require.True(t, tc.resultPushdown, "Cannot use skipTigrisPushdown when resultPushdown is false'")
+						assert.Equal(t, false, explainRes.Map()["pushdown"], "Tigris pushdown check was skipped, but it was actually true")
+					} else {
+						assert.Equal(t, tc.resultPushdown, explainRes.Map()["pushdown"])
+					}
```
w84thesun, line 100: That part is not really clear to me. The comment above says we want to skip Tigris pushdown checks but we still have some asserts.

Could it be simplified to something like this?

```suggestion
					if !(tc.skipTigrisPushdown && setup.IsTigris(t)) {
						assert.Equal(t, tc.resultPushdown, explainRes.Map()["pushdown"])
					}
```

integration/setup/common.go
```diff
@@ -87,6 +87,12 @@ func SkipForTigrisWithReason(tb testing.TB, reason string) {
 	}
 }
 
+// IsTigris returns if tests are running against the tigris handler.
```
w84thesun, line 90:
```suggestion
// IsTigris returns if tests are running against the Tigris handler.
```

integration/query_compat_test.go
```diff
@@ -90,7 +91,13 @@ func testQueryCompat(t *testing.T, testCases map[string]queryCompatTestCase) {
 					var explainRes bson.D
 					require.NoError(t, targetCollection.Database().RunCommand(ctx, explainQuery).Decode(&explainRes))
 
-					assert.Equal(t, tc.resultPushdown, explainRes.Map()["pushdown"])
+					// temporarily make possible to skip tigris pushdown checks
```
w84thesun, line 94:
```suggestion
					// temporarily make it possible to skip Tigris pushdown checks
```

w84thesun (commented):
Overall good, asked about the pushdown check in the compatibility test.

integration/query_compat_test.go
```diff
@@ -90,7 +91,13 @@ func testQueryCompat(t *testing.T, testCases map[string]queryCompatTestCase) {
...
+					if tc.skipTigrisPushdown && setup.IsTigris(t) {
+						require.True(t, tc.resultPushdown, "Cannot use skipTigrisPushdown when resultPushdown is false'")
```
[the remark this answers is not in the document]
noisersup: It tests scenarios when we expect that we don't support pushdowns for tigris yet, but we actually do. Similar thing is done with `emptyResults`. If some change in the code made the result of the test non-empty, we return the error that our expectations are different than actual response

integration/query_compat_test.go
noisersup: This simplification removes some assumptions:
1. We shouldn't skip pushdown checks for tigris if we don't expect any pushdowns (if `tc.resultPushdown` is not set)
2. We shouldn't skip pushdown checks for tigris if we actually do a pushdown for tigris in a testcase

The `tc.skipTigrisPushdown` was introduced to make possible introducing PostgreSQL pushdowns without Tigris ones, as they seem more complicated.

noisersup: Although I can see some inconsistency with the naming, right. When we use `skipForTigris`, we just want to skip the whole test regardless of whether it pass or fails, and here we validate if our expectations are correct. Although, I couldn't find any better name... `dontDoTigrisPushdown`, `tigrisNotPushdowned`... Any ideas? :smile: 

We could split `resultPushdowns` to `pushdownPg` and `pushdownTrigris`, but imo in the future it will create too many lines, as we aim to do pushdowns for both handlers, so personally I would stay with current approach where we treat lack of pushdown for Tigris as exception.

w84thesun: Ok, after your explanation I understood what was going on. I think the comment should be changed in a way that we can see what those checks are for then.

w84thesun: Something like: `If tc.skipTigrisPushdown is set check that pushdown actually wasn't made.`

w84thesun (approved):
👍

rumyantseva (approved):
LGTM

internal/handlers/pg/pgdb/query.go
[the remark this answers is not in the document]
AlekSi: > If the left value is an array
> If the left value is a jsonb document
> However, if the left value is a jsonb number

Please also consider strings. For example, how that would work given a document `{v: "foo"}` and filter `{v: "fo"}`, and vice versa?

chilagrow (approved):
Looks good to me, thanks for the detailed comment!

internal/handlers/pg/pgdb/query.go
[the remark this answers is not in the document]
noisersup: Checked that after the call with @chilagrow 

```SQL
select _jsonb from test."Composites_83e0b565" where (_jsonb->'v'->'42')::jsonb @> '"foo"';
```

While this query has some results, as we have 2 documents with `v.42 = "foo"` in collection

```SQL
select _jsonb from test."Composites_83e0b565" where (_jsonb->'v'->'42')::jsonb @> '"f"';
```

this one doesn't return anything

PR #1814, opened 2023-01-17
Fix `$set` operator to apply correct comparison

### LINKED ISSUE #1662: `$set` operator does not set modified count for changing to same value
### Versions

0.7.1

### What did you do?

Moved `TestUpdateFieldCompatSet` test `ArrayNil` and `SetSameValueInt` test case to compat.

The same as compat.

### What did you see instead?

The compat expect 1 modification, but target has 0 modification.
```
        --- FAIL: TestUpdateFieldCompatSet/ArrayNil/TestUpdateFieldCompatSet-ArrayNil_Scalars (0.98s)
            --- FAIL: TestUpdateFieldCompatSet/ArrayNil/TestUpdateFieldCompatSet-ArrayNil_Scalars/null (0.02s)
                update_field_compat_test.go:387: 
                        Error Trace:    /home/chi/go/src/github.com/chilagrow/FerretDB/integration/update_compat_test.go:123
                        Error:          Not equal: 
                                        expected: &mongo.UpdateResult{MatchedCount:1, ModifiedCount:1, UpsertedCount:0, UpsertedID:interface {}(nil)}
                                        actual  : &mongo.UpdateResult{MatchedCount:1, ModifiedCount:0, UpsertedCount:0, UpsertedID:interface {}(nil)}
                                        
                                        Diff:
                                        --- Expected
                                        +++ Actual
                                        @@ -2,3 +2,3 @@
                                          MatchedCount: (int64) 1,
                                        - ModifiedCount: (int64) 1,
                                        + ModifiedCount: (int64) 0,
                                          UpsertedCount: (int64) 0,
                        Test:           TestUpdateFieldCompatSet/ArrayNil/TestUpdateFieldCompatSet-ArrayNil_Scalars/null
```

chilagrow:
# Description

Closes #1662.

## Readiness checklist


internal/types/compare_order.go
```diff
@@ -110,7 +137,19 @@ func CompareOrder(a, b any, order SortType) CompareResult {
 		return result
 	}
 
-	return Compare(a, b)
+	c := &comparer{
+		compareTypeOrder: compareType,
```
chilagrow, line 141: Within array and within document, we must also check for different number type, since we pass this `compareType` around.

internal/types/compare_order.go
```diff
@@ -91,9 +92,35 @@ const (
...
+
+const (
+	_ numberOrderResult = iota
+	doubleDT
+	int32DT
+	int64DT
+)
+
+// detectNumberType returns a sequence for float64, int32 and int64 types.
+func detectNumberType(value any) numberOrderResult {
```
rumyantseva, line 106: I see where this function is used, but I don't understand how it works. The result of this function is compared with `numbersDataType` which is another constant, and there are no other use cases for it.

Should this function be called something like `isNumber` then and include a comparison with `numbersDataType`?

internal/types/compare_order.go
```diff
@@ -91,9 +92,35 @@ const (
...
+type numberOrderResult uint8
+
+const (
+	_ numberOrderResult = iota
+	doubleDT
+	int32DT
+	int64DT
+)
+
+// detectNumberType returns a sequence for float64, int32 and int64 types.
```
rumyantseva, line 105: What does it mean "a sequence", I see `numberOrderResult` is a `uint8`, is there a way we interpret it as a sequence?

integration/update_field_compat_test.go
```diff
@@ -718,12 +733,36 @@ func TestUpdateFieldCompatSetArray(t *testing.T) {
...
 		"EmptyArray": {
 			update: bson.D{{"$set", bson.D{{"v", bson.A{}}}}},
 		},
+		"ArrayNumbersAsc": {
+			update:        bson.D{{"$set", bson.D{{"v", bson.A{int32(42), int64(43), 45.5}}}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/1668",
+		},
+		"ArrayStringsDesc": {
+			update:        bson.D{{"$set", bson.D{{"v", bson.A{"c", "b", "a"}}}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/1816",
```
rumyantseva, line 748: This one should work for tigris, see https://github.com/FerretDB/FerretDB/blob/71ba4d0c578ba0777b9464e17f979610e2a42f88/integration/shareddata/composites.go#L133

integration/update_field_compat_test.go
```diff
@@ -666,9 +666,24 @@ func TestUpdateFieldCompatSet(t *testing.T) {
 			update:        bson.D{{"$set", bson.D{{"foo", int32(12)}, {"v", nil}}}},
 			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/1676",
 		},
-		"SetSameValueInt": {
-			update: bson.D{{"$set", bson.D{{"v", int32(42)}}}},
-			skip:   "https://github.com/FerretDB/FerretDB/issues/1662",
+		"SameValueInt32Type": {
+			update:        bson.D{{"$set", bson.D{{"v", int32(42)}}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/1668",
```
rumyantseva, line 671: This one should work for Tigris, just an int32 number, should be fine.

integration/update_field_compat_test.go
```diff
@@ -666,9 +666,24 @@ func TestUpdateFieldCompatSet(t *testing.T) {
...
+		},
+		"SameValueInt64Type": {
+			update:        bson.D{{"$set", bson.D{{"v", int64(42)}}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/1668",
```
rumyantseva, line 675: Same, should work for Tigris, it's just a number

integration/update_field_compat_test.go
```diff
@@ -666,9 +666,24 @@ func TestUpdateFieldCompatSet(t *testing.T) {
...
+		},
+		"SameValueDoubleType": {
+			update:        bson.D{{"$set", bson.D{{"v", 42.0}}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/1668",
```
rumyantseva, line 679: Same, as far as I understand it should work for Tigris

integration/update_field_compat_test.go
```diff
@@ -666,9 +666,24 @@ func TestUpdateFieldCompatSet(t *testing.T) {
...
+		},
+		"DocSameNumberType": {
+			update: bson.D{{"$set", bson.D{{"v", bson.D{{"foo", int32(42)}}}}}},
+		},
+		"DocDifferentNumberType": {
+			update:        bson.D{{"$set", bson.D{{"v", bson.D{{"foo", int64(42)}}}}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/1668",
```
rumyantseva, line 686: This one doesn't need an issue number but an explanation that it can't be set for Tigris. If we attempt to set a different type, Tigris will return an error.

integration/update_field_compat_test.go
```diff
@@ -718,12 +733,36 @@ func TestUpdateFieldCompatSetArray(t *testing.T) {
 			resultType: emptyResult,
 		},
 		"ArrayNil": {
-			update: bson.D{{"$set", bson.D{{"v", bson.A{nil}}}}},
-			skip:   "https://github.com/FerretDB/FerretDB/issues/1662",
+			update:        bson.D{{"$set", bson.D{{"v", bson.A{nil}}}}},
+			skipForTigris: "tigris produce empty result because composites dataset is not applicable",
```
rumyantseva, line 737: I think this case is possible in `beta.27`, so it's a TODO

integration/update_field_compat_test.go
```diff
@@ -718,12 +733,36 @@ func TestUpdateFieldCompatSetArray(t *testing.T) {
...
 		},
 		"EmptyArray": {
 			update: bson.D{{"$set", bson.D{{"v", bson.A{}}}}},
 		},
+		"ArrayNumbersAsc": {
+			update:        bson.D{{"$set", bson.D{{"v", bson.A{int32(42), int64(43), 45.5}}}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/1668",
```
rumyantseva, line 744: This one is not possible for Tigris, see `skipForTigris: "Tigris does not support mixed types in arrays",` in the code

integration/update_field_compat_test.go
```diff
@@ -718,12 +733,36 @@ func TestUpdateFieldCompatSetArray(t *testing.T) {
...
+		},
+		"ArrayStringsDesc": {
+			update:        bson.D{{"$set", bson.D{{"v", bson.A{"c", "b", "a"}}}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/1816",
+		},
+		"ArrayDifferentNumberType": {
+			update:        bson.D{{"$set", bson.D{{"v", bson.A{int64(42), int64(43), 45.5}}}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/1816",
```
rumyantseva, line 752: Tigris does not support mixed types in arrays

integration/update_field_compat_test.go
```diff
@@ -718,12 +733,36 @@ func TestUpdateFieldCompatSetArray(t *testing.T) {
...
+		},
+		"ArraySameNumberType": {
+			update:        bson.D{{"$set", bson.D{{"v", bson.A{int32(42), int64(43), 45.5}}}}},
+			skipForTigris: "https://github.com/FerretDB/FerretDB/issues/1816",
```
rumyantseva, line 756: Tigris does not support mixed types in arrays

integration/update_field_compat_test.go
```diff
@@ -718,12 +733,36 @@ func TestUpdateFieldCompatSetArray(t *testing.T) {
...
+		},
+		"DocSameNumberType": {
+			update:        bson.D{{"$set", bson.D{{"v", bson.D{{"foo", int32(42)}, {"42", "foo"}, {"array", bson.A{int32(42), "foo", nil}}}}}}},
+			skipForTigris: "tigris produce empty result because composites dataset is not applicable",
```
rumyantseva, line 760: See `Tigris does not support field names started from numbers (`42`)` messages. Composites are possible in general.

integration/update_field_compat_test.go
```diff
@@ -718,12 +733,36 @@ func TestUpdateFieldCompatSetArray(t *testing.T) {
...
+		},
+		"DocDifferentNumberType": {
+			update:        bson.D{{"$set", bson.D{{"v", bson.D{{"foo", int32(42)}, {"42", "foo"}, {"array", bson.A{int64(42), "foo", nil}}}}}}},
+			skipForTigris: "tigris produce empty result because composites dataset is not applicable",
```
rumyantseva, line 764: Same here. Composites are possible, fields started with numbers are not possible.

rumyantseva (commented):
Left some comments about tests

internal/types/compare.go
```diff
@@ -52,17 +53,31 @@ func Compare(docValue, filterValue any) CompareResult {
...
-	switch docValue := docValue.(type) {
+	c := &comparer{
+		compareTypeOrder: compareTypeOrder,
+	}
+
+	return c.compare(docValue, filterValue)
+}
+
+// comparer uses compareTypeOrder to perform type comparison.
+type comparer struct {
```
rumyantseva, line 64: Why a struct is needed? Is it used as a dependency somewhere?

internal/types/compare_order.go
```diff
@@ -253,6 +292,31 @@ func compareTypeOrder(a, b any) CompareResult {
...
+		aType := detectNumberType(a)
+		bType := detectNumberType(b)
+
+		switch {
+		case aType < bType:
+			return Less
+		case aType > bType:
+			return Greater
+		default:
+			return Equal
```
rumyantseva, line 313: Are there any use cases for it? If I understand the task correctly, it's enough to know whether the types are equal or not.

internal/types/compare_order.go
```diff
@@ -253,6 +292,31 @@ func compareTypeOrder(a, b any) CompareResult {
 	}
 }
 
+// compareType compares compareTypeOrder then compares
+// number types. The number types int32, int64 and float64
+// are considered different.
```
rumyantseva, line 297: Ther are not only considered different, there is also quite a specific logic in how we sort them.

internal/types/compare.go
chilagrow: This is used for checking if number type is equal or not within nested documents, we need to pass how to compare type order when it is called within recursion. Definitely need to improve documentation

For standard comparison by `Compare(a, b)`, 
a doc `{"v": {"foo": {"bar": [int32(1)]}}}` should equal
a doc `{"v": {"foo": {"bar": [float64(1)]}}}`
For `CompareOrder(a,b,Ascending)`, above should not equal.

integration/update_field_compat_test.go
chilagrow: For the same as above, this can only be tested on double data, I updated the test.

integration/update_field_compat_test.go
chilagrow: :+1:

integration/update_field_compat_test.go
chilagrow: Yes it works on `ArrayStrings` :+1: , I just need to skip `ArrayDoubles`, `ArrayInt32s` and `ArrayRegexes`

integration/update_field_compat_test.go
chilagrow: Applied :+1:

integration/update_field_compat_test.go
chilagrow: Applied :+1:

integration/update_field_compat_test.go
chilagrow: Thanks. I didn't try to enable tigris handler for `Composites` but I see that it would fail on "42" :fearful:

integration/update_field_compat_test.go
chilagrow: :pray: updating skip reason

internal/types/compare_order.go
chilagrow: No longer applicable now. We use separate func Identical for this work

internal/types/identical.go
```diff
@@ -0,0 +1,175 @@
...
+		aIter, bIter := a.Iterator(), b.Iterator()
+		defer aIter.Close()
+		defer bIter.Close()
+
+		for {
+			_, aField, err := aIter.Next()
+			if errors.Is(err, iterator.ErrIteratorDone) {
+				return true
+			} else if err != nil {
+				return false
```
rumyantseva, line 48: As this function doesn't have an error return value, I suggest we should panic in all those cases if `iter.Next` returns an error (for objects and arrays). Currently, we just ignore it, and this could cause unexpected behavior if such an error becomes possible in the future.

integration/update_field_compat_test.go
```diff
@@ -709,6 +794,9 @@ func TestUpdateFieldCompatSet(t *testing.T) {
 func TestUpdateFieldCompatSetArray(t *testing.T) {
 	t.Parallel()
 
+	stringsProvider := shareddata.AllProviders().
+		Remove("ArrayDoubles", "ArrayInt32s", "ArrayRegexes")
```
rumyantseva, line 798: It looks to me that if we want to specify string providers, they should be included, not excluded. Otherwise, every time we add a new (non-string) provider we need to come to this part of the code and exclude it here.

Also, do we need to specify them? Can't we ignore Tigris-related tests/errors?

rumyantseva (commented):
I like the solution!

I added a couple of thoughts/questions.

integration/update_field_compat_test.go
chilagrow: Yes that's an option, let me ignore errors :+1:

internal/types/identical.go
chilagrow: Sounds sensible :hugs: let me do that

rumyantseva (approved):
LGTM

w84thesun (approved):
👍

PR #1842, opened 2023-01-20
Support Tigris pushdowns for numbers

### LINKED ISSUE #1755: Support pushdowns for equal operators on numbers

noisersup:
# Description

Closes #1755.

## Readiness checklist


noisersup:
This PR is halted, the current state is that it works for most cases, but there are a few edge cases as the fact that filtering the `""` doesn't return the empty strings (maybe the tigris bug?), or float comparisons on big numbers (though we don't expect the float comparison will work correctly)

internal/handlers/tigris/tigrisdb/query.go
```diff
@@ -95,25 +110,45 @@ func (tdb *TigrisDB) BuildFilter(filter *types.Document) driver.Filter {
 	res := map[string]any{}
 
 	for k, v := range filter.Map() {
-		// filter only by _id for now
-		if k != "_id" {
+		// don't pushdown empty strings yet
```
noisersup, line 113: For those cases, I believe that the empty string filter is the last one in terms of incompatibility. They work currently on API, but not on driver. I assume it'll be shortly fixed so I'll cover that on dotnotation task.

rumyantseva: @noisersup let's create a separate issue and add a TODO here?

internal/handlers/tigris/tigrisdb/query.go
```diff
@@ -95,25 +110,45 @@ func (tdb *TigrisDB) BuildFilter(filter *types.Document) driver.Filter {
...
-		switch v.(type) {
-		case string:
-			// filtering by string values is complicated if the storage supports encodings, collations, etc,
-			// but Tigris does not support any of these
-		case types.ObjectID:
-			// filtering by ObjectID is always safe
-		default:
-			// skip other types for now
+		if k != "" {
+			// don't pushdown $comment and dot notation yet
```
rumyantseva, line 119: Let's add TODO and an issue number

internal/handlers/tigris/tigrisdb/query.go
```diff
@@ -95,25 +110,45 @@ func (tdb *TigrisDB) BuildFilter(filter *types.Document) driver.Filter {
...
+				continue
+			}
+		}
+
+		// _id field supports only specific types
+		if k == "_id" {
+			switch v.(type) {
+			case *types.Document, *types.Array, types.Binary, bool, time.Time, types.NullType, types.Regex, types.Timestamp:
+			// type not supported for pushdown
+
```
rumyantseva, line 130: Formatting in this switch-case looks a bit insocnistent to me: 
- Should each `case` be ended with an empty line or not?
- Should a comment be shifter in this case (like in the next one)?

In the switch below, the same things are formated differently.

internal/handlers/tigris/tigrisdb/query.go
```diff
@@ -95,25 +110,45 @@ func (tdb *TigrisDB) BuildFilter(filter *types.Document) driver.Filter {
...
+				panic(fmt.Sprintf("Unexpected type of field %s: %T", k, v))
+			}
+
 			continue
 		}
 
-		// filter by the exact _id value
-		id := must.NotFail(tjson.Marshal(v))
-		res["_id"] = json.RawMessage(id)
+		switch v.(type) {
```
rumyantseva, line 142: What is the difference between `k == "_id"` and other fields? Both switches look quite identical to me.

rumyantseva (commented):
Great job overall! I asked a few questions.

internal/handlers/tigris/tigrisdb/query.go
rumyantseva: Well, at the moment, we don't validate queries. If someone tries to set `_id` as an array, what will be the result? Nothing will be found, and it's an expected result. Why should we validate queries then?

noisersup: valid point, going to fix that

internal/handlers/tigris/tigrisdb/query.go
chilagrow: > Should each case be ended with an empty line or not?

I think if we agree we can put that logic into our linter :hugs:

chilagrow (approved):

internal/handlers/tigris/tigrisdb/query.go
```diff
@@ -85,7 +88,19 @@ func (tdb *TigrisDB) QueryDocuments(ctx context.Context, param *FetchParam) ([]*
...
 
-	return res, iter.Err()
+	err = iter.Err()
+
+	switch {
+	case err == nil:
+		fallthrough
+	case IsInvalidArgument(err):
+		// Skip errors from filtering invalid types
+		break
```
AlekSi, line 98: @noisersup Why do we do that? IIUC, we just hide what could be a real error

noisersup: InvalidArgument error occures while filtering the schema with wrong type. For example if `v` is a string, but we use filter `{"v":42}` we get this error. MongoDB on this scenario result with no error, as it just means that filter don't match the value in database

AlekSi: ic, thanks

PR #1852, opened 2023-01-23
Set default `pool_max_conns` to 20 for PostgreSQL

### LINKED ISSUE #1844: Set default `pool_max_conns=20` for PostgreSQL
### What should be done?

If `pool_max_conns` is not present in the PostgreSQL URL, assume it is 20, not pgx's default.

https://github.com/FerretDB/FerretDB/blob/d479051013c0cdd7b92469dc06eb1845c15295f9/internal/handlers/pg/pgdb/pool.go#L60

jkoenig134:
# Description

Closes #1844

## Readiness checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/chore.yml
+++ b/.github/ISSUE_TEMPLATE/chore.yml
@@ -29,9 +29,9 @@ body:
       label: Definition of Done
       description: What should be done to consider this issue done? List everything that applies.
       value: |
-        - unit tests added;
-        - integration tests added;
-        - compatibility tests added;
+        - all handlers updated;
+        - unit tests added/updated;
+        - integration/compatibility tests added/updated;
         - spot refactorings done;
         - user documentation updated;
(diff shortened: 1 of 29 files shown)


jkoenig134:
@AlekSi I assumed that like setting user + password `pg.go` should be responsible for the URI manipulation and in this method I do have a `url.URL` object instead of only a string like in `pool.go`.

AlekSi:
Here we only use per-connection username and password. All other parameters, like host, port, and all query parameters are the same.

jkoenig134:
What about setting it in the NewOpts constructor?

AlekSi:
Well, that's possible, but I really don't see a problem in using `u, err := url.Parse(uri)` and `uri = u.String()` in `NewPool` :) Seems to be the most natural place

AlekSi (approved):

AlekSi:
Thank you!

PR #1857, opened 2023-01-24
Rework on integration test setup

### LINKED ISSUE #1568: Rework integration test setup and auth
### What should be done?

At the moment, we fetch flag values but each test can't assign value for the flag. Additionally `-handler` flag is used incorrectly for validator.

* Validators should be used only for target. When we run compatibility tests with target=tigris and compat=MongoDB, we get validators for `tigris` handler for both databases and then ignore MongoDB's error. In general, `-handler`'s flag value is used incorrectly. https://github.com/FerretDB/FerretDB/blob/f0ca7811ed2aae21ce5063bd3cab533d56cf6cfb/integration/setup/setup_compat.go#L131
* Integration setup uses flag's value but unable to specify flags used per test cases. Enable test cases to test authentication. https://github.com/FerretDB/FerretDB/blob/f0ca7811ed2aae21ce5063bd3cab533d56cf6cfb/integration/setup/common.go#L231
* Test different flags values of env data `TestEnvData`.
* Test different flag values of `postgresql-url` in `TestCommandsAuthenticationSASLStart` 
* Provider is compatible with both compat and target, but validator is only compatible with target. Fix this definition https://github.com/FerretDB/FerretDB/blob/62039bf08d8cb4cd19bbc39c0c75008f13e3a48c/integration/shareddata/shareddata.go#L33-L39
* pg uri used for testing is set from creating a new listener or using existing connection on port. Rework on how uri is built and used. Currently attempting all possible combinations. Probably this wouldn't help for speed. https://github.com/FerretDB/FerretDB/blob/cc7b47db1efce9c164f5922eb564ea6b7a8592ef/integration/setup/common.go#L170-L181
* Remove code like https://github.com/FerretDB/FerretDB/blob/56a2dd17b77bcb6afe1fc64b592a7985c4df15b7/integration/create_test.go#L70-L81

Keep note that maybe we may want test data per provider.
### Where?

DoD:
- [x] in compat test, validators are correctly applied for target but not for compat.
- [x] tests added for `TestCommandsAuthenticationSASLStart`
  - [x] authentication with PLAIN
  - [ ] authentication with invalid credential
  - [x] authentication with valid credential
- [x] uri for specific target/compat is being built.
- [x] validator is not used for compat
- [ ] address issue numbers mentioned in the code

---

- [x] rework validations in shareddata
- [x] fix `task env-data` (#2200)
- [x] fix test script

chilagrow:
# Description

Closes #1568.

This is one of few PR.
* It uses `-tigris-urls` from Taskfile.yml.
* Removed `-handler`, it uses `-postgresql-url` or `-tigris-urls` to interpret the handler. If neither `-postgresql-url` nor `-tigris-urls` is set, it uses no handler (example `test-integration-mongodb`)
* validator is applied only for target
* authentication URL is built specifically for the target and compat. `PLAIN` for target with `TLS`, `SCRAM` (by default) for compat with `TLS`.

## Readiness checklist


integration/setup/common.go
```diff
@@ -98,123 +112,59 @@ func SkipForPostgresWithReason(tb testing.TB, reason string) {
...
-	}{{
-		user:          nil,
-		authMechanism: "",
-	}, {
-		user:          url.UserPassword("username", "password"),
-		authMechanism: "PLAIN",
-	}, {
-		user:          url.UserPassword("username", "password"),
-		authMechanism: "", // defaults to SCRAM when username is set
-	}} {
```
chilagrow, line 182: TCP uses
```
		user:          nil,
		authMechanism: "",
```

FerretDB TSL uses 
```
		user:          url.UserPassword("username", "password"),
		authMechanism: "PLAIN",
```
Compat with TSL uses  
```
		user:          url.UserPassword("username", "password"),
		authMechanism: "", // defaults to SCRAM when username is set
```
Let me know if I'm wrong.

integration/commands_authentication_test.go
```diff
@@ -15,9 +15,113 @@
...
+				"unable to authenticate using mechanism \"PLAIN\": (InternalError) " +
+				"[msg_saslstart.go:66 pg.(*Handler).MsgSASLStart] [pg.go:114 pg.(*Handler).DBPool] " +
+				"pgdb.NewPool: failed to connect to `host=127.0.0.1 user=invalid database=ferretdb`: " +
+				"server error (FATAL: role \"invalid\" does not exist (SQLSTATE 28000))",
+		},
+		"TLSWrongPassword": {
+			listenerTLS:     true,
+			clientTLS:       true,
+			invalidPassword: true,
+			skip:            "TODO: Expects error but connection is established successfully",
```
chilagrow, line 64: Let's validate if really this is the case, then create an issue

integration/setup/common.go
```diff
@@ -104,123 +117,57 @@ func SkipForPostgresWithReason(tb testing.TB, reason string) {
 	}
 }
 
-// checkMongoDBURI returns true if given MongoDB URI is working.
-func checkMongoDBURI(tb testing.TB, ctx context.Context, uri string) bool {
```
chilagrow, line 108: This is essentially the same func as `setupClient(...` 
https://github.com/FerretDB/FerretDB/pull/1857/files#diff-e44bf8c258ee6bba05163a94822b0605ba181d26efa013179df8087fa3366e5dR261

We don't need to create client in `checkMongoDBURI` then create another client in `setupClient` in every  compat test.

integration/commands_authentication_test.go
```diff
@@ -15,9 +15,110 @@
...
+			})
+
+			ctx := s.Ctx
+
+			// s.MongoDBURI looks like `mongodb://username:password@127.0.0.1:35697/?authMechanism=PLAIN`.
+			// For testing invalid parameter, we replace each part of it.
+			clientURI := s.MongoDBURI
+
+			if tc.invalidUsername {
+				clientURI = strings.Replace(clientURI, "username", "invalid", 1)
```
noisersup, line 95: I believe the safer way to do that is to use net/url, although it's just a test, so I believe it's okay

integration/setup/common.go
```diff
@@ -104,123 +104,57 @@ func SkipForPostgresWithReason(tb testing.TB, reason string) {
...
 type buildMongoDBURIOpts struct {
-	hostPort       string
+	host           string
 	unixSocketPath string
 	tls            bool
+	authMechanism  string
+	user           *url.Userinfo
 }
 
 // buildMongoDBURI builds MongoDB URI with given URI options and validates that it works.
```
w84thesun, line 116: We removed validation, so we need to update the description.

w84thesun (commented):
Good job! Added some comments.

noisersup (dismissed):
LGTM!

integration/commands_authentication_test.go
chilagrow: Nope, this works on mongosh-secure

integration/commands_authentication_test.go
chilagrow: Thanks good idea, I ended up using options package from the driver to set it :hugs:

integration/setup/common.go
chilagrow: Updated :+1:

integration/commands_authentication_test.go
```diff
@@ -15,9 +15,110 @@
...
+	for name, tc := range testcases {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			if tc.skip != "" {
+				setup.SkipForPostgresWithReason(t, tc.skip)
+			}
+
+			postgreSQLURL := "postgres://username:password@127.0.0.1:5432/ferretdb?pool_min_conns=1"
+			port := 0
+			unix := false
```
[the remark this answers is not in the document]
chilagrow: Thanks for spotting this :+1:

integration/setup/common.go
```diff
@@ -104,123 +104,61 @@ func SkipForPostgresWithReason(tb testing.TB, reason string) {
...
-func buildMongoDBURI(tb testing.TB, ctx context.Context, opts *buildMongoDBURIOpts) string {
-	tb.Helper()
+// buildMongoDBURI builds MongoDB URI with given URI options.
+func buildMongoDBURI(tb testing.TB, opts *buildMongoDBURIOpts) string {
+	q := make(url.Values)
 
-	var host string
+	if opts.tls {
+		require.Empty(tb, opts.unixSocketPath, "unixSocketPath cannot be used with TLS")
+		q.Set("tls", "true")
```
chilagrow, line 122: Now setting tls true, because it's being set in https://github.com/FerretDB/FerretDB/blob/main/Taskfile.yml#L389

integration/commands_authentication_test.go
noisersup: :100:

noisersup (commented):
It looks like we need to increase the timeout I guess!

chilagrow:
> It looks like we need to increase the timeout I guess!

This is very strange, I need to check newly added tests are not making it slow. :thinking:

integration/embedded_test.go
```diff
@@ -1,113 +0,0 @@
...
-	"go.mongodb.org/mongo-driver/bson/primitive"
-	"go.mongodb.org/mongo-driver/mongo"
-	"go.mongodb.org/mongo-driver/mongo/options"
-
-	"github.com/FerretDB/FerretDB/ferretdb"
-	"github.com/FerretDB/FerretDB/integration/setup"
-	"github.com/FerretDB/FerretDB/internal/util/testutil"
-)
-
-func TestEmbedded(t *testing.T) {
```
chilagrow, line 34: This is covered in `TestCommandsAuthenticationSASLStart`

integration/setup/flags.go
```diff
@@ -0,0 +1,172 @@
...
+		case "compat-port":
+			compatPort, ok := v.(int)
+			require.True(tb, ok, "%s is not int: %T", k, v)
+
+			f.compatPort = &compatPort
+		case "compat-tls":
+			compatTLS, ok := v.(bool)
+			require.True(tb, ok, "%s is not bool: %T", v)
+
+			f.targetTLS = &compatTLS
```
rumyantseva, line 74: Is this enough as `bool`? Don't we need something like "target-auth-mechanism" and "compat-auth-mechanism" to know how exactly we should connect to target and compat in case of TLS/auth being set? Or is it just for the beginning?

integration/setup/flags.go
```diff
@@ -0,0 +1,172 @@
...
+			targetPort, ok := v.(int)
+			require.True(tb, ok, "%s is not int: %T", k, v)
+
+			f.targetPort = &targetPort
+		case "target-tls":
+			targetTLS, ok := v.(bool)
+			require.True(tb, ok, "%s is not bool: %T", v)
+
+			f.targetTLS = &targetTLS
+		case "handler":
```
rumyantseva, line 50: I remember there were some handler-related questions in this issue. How is it used now? Do we use `handler` for `target` only? How does it work in case of compat collections setup? Should we call it "target-handler"?.. Do we need this flag at all having in mind that there are flags for postgresql-url and tigris-url, so we already know if we work with pg or Tigris?..

rumyantseva (commented):
Great job with refactoring and structuring!

I asked some questions about the places I didn't understand, I guess at least some of them are for a separate PR though.

integration/commands_authentication_test.go
```diff
@@ -15,9 +15,139 @@
...
+		"NotPlainTLS": {
+			listenerTLS:          true,
+			clientTLS:            true,
+			invalidAuthMechanism: true,
+			dbErr:                "auth error",
+		},
+		"TLSWithNonTLSClient": {
+			listenerTLS: true,
+			clientTLS:   false,
+			skip:        "driver attempts to reconnect until time out",
```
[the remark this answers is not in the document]
chilagrow: You are right, this skip is permanent. So let's remove this test. They are not the ones from embedded tests, the embedded tests cases are `TLS` and `TCL`.

Due to driver reconnecting again and again this test keeps running for long time until timeout, which is not what we want in our testing.

integration/commands_authentication_test.go
```diff
@@ -15,9 +15,139 @@
...
+		},
+		"TCPWithTLSClient": {
+			listenerTLS: false,
+			clientTLS:   true,
+			skip:        "driver attempts to reconnect until time out",
```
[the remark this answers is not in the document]
chilagrow: Removing this too.

Due to driver reconnecting again and again this test keeps running for long time until timeout, which is not what we want in our testing.

integration/commands_authentication_test.go
```diff
@@ -15,9 +15,139 @@
...
+			// upon Connect clientURI.String() value of auth is overridden with SetAuth.
+			client, err := mongo.Connect(ctx, clientOpts)
+			require.NoError(t, err)
+
+			t.Cleanup(func() {
+				err = client.Disconnect(ctx)
+				require.NoError(t, err)
+			})
+
+			dbs, err := client.ListDatabaseNames(ctx, bson.D{})
```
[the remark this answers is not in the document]
chilagrow: I think listing database is okay for now. A collection being setup by `SetupWithOpts` appears to be not visible to the client accessing from different connection. Looking at usage, maybe creating collection and databases were prepared specifically for admin tests?

Anyway, using this 
```
			s := setup.SetupWithOpts(t, &setup.SetupOpts{
				Flags: map[string]any{
					"target-tls":         tc.listenerTLS,
					"target-port":        0,
					"target-unix-socket": false,
					"postgresql-url":     "postgres://username:password@127.0.0.1:5432/ferretdb?pool_min_conns=1",
				},
				CollectionName: ""test-collection",
				DatabaseName:   "test-db",
			})

```

and fetching collection names returns empty.
```
collectionNames, err := client.Database("test-db").ListCollectionNames(ctx, bson.D{})
```

integration/setup/flags.go
chilagrow: > How is it used now?

The handler still defaults to pg. This will change in next PR. validators are now only applied to target.

> Do we use handler for target only?

Yes for the validator, the handler is used only for target. The compat does not use validator since we know compat is always compatible with all provider.

> Should we call it "target-handler"?

I  like this suggestion, let me handle this in another PR. For example, we have test-integration-mongodb and which target is mongodb equivalent but handler is not applicable here because it's not `pg` or `tigris`. From my understanding handler is what we implement, so if we can't create listener for it, we shouldn't call it an handler. What do you think?

> Do we need this flag at all having in mind that there are flags for postgresql-url and tigris-url, so we already know if we work with pg or Tigris?

Great point, let me see if I can remove it fully. But let's handle the handler logic in next PR.

integration/setup/flags.go
chilagrow: First you found bug, it should be assigned to `f.compatTLS = &compatTLS`. I added tests for the flags. The compatTLS and targetTLS should be separate.

For authentication mechanism, it's based on if ferretdb is connecting or mongodb is connecting. see this for my understanding of TLS.
https://github.com/FerretDB/FerretDB/pull/1857#discussion_r1086350806

As far as I know, we share the certificate between target and compat, using flag for it is sufficient. @w84thesun is that right?

w84thesun: Yes, we have one set for both MongoDB and FerretDB.

w84thesun (approved):
Great job! 👍 I like all the comments, they will make it easier to understand our test setup.

rumyantseva (dismissed):
I'm not sure if we fully decided on everything, but I'm ok to merge this PR as it as and continue making improvements in separate PRs if needed.

Maybe @AlekSi's look could help.

w84thesun (approved):
👍

integration/setup/flags.go
```diff
@@ -0,0 +1,152 @@
...
+package setup
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// flags are used to override flags set from cli with test setup option.
+type flags struct {
```
AlekSi, line 24: Let's remove that change from that PR

chilagrow:
So actually it makes change in `startup`, I'll handle necessary merge conflict :pray:

w84thesun (dismissed):

w84thesun (dismissed):

w84thesun (approved):

rumyantseva (dismissed):
LGTM

integration/create_compat_test.go
```diff
@@ -61,21 +61,14 @@ func TestCreateCompat(t *testing.T) {
...
-		if errorTextContains(targetErr,
-			`support for field "validator" is not implemented yet`,
-		) {
-			targetErr = targetDB.CreateCollection(s.Ctx, collName)
-		}
 
-		require.NoError(t, targetErr)
+	// Set $tigrisSchemaString for tigris only.
+	var opts options.CreateCollectionOptions
+	if setup.IsTigris(t) {
```
chilagrow, line 67: I think this is still better than how it was done before this change.

AlekSi (approved):
I think this is great

w84thesun (approved):
👍

PR #1863, opened 2023-01-24
Add simple `otel` tracing to collect data from tests

rumyantseva:
# Description

Closes FerretDB/engineering#33.

Something to start with, we can add more traces/details when we continue investigating. 

## Readiness checklist


integration/setup/common.go
```diff
@@ -351,5 +370,31 @@ func startup() {
 		} else {
 			zap.S().Infof("Compat system: port %d.", p)
 		}
+
+		// Set open telemetry tracer if jaeger endpoint is provided.
+		if *jaegerEndpointF != "" {
```
chilagrow, line 375: Could we have this in separate function like `startupTracer(tb testing.TB)` and call it from here? I'm refactoring this file too so it would make conflict handling easier :hugs:

integration/setup/common.go
```diff
@@ -108,11 +116,16 @@ func SkipForPostgresWithReason(tb testing.TB, reason string) {
 func checkMongoDBURI(tb testing.TB, ctx context.Context, uri string) bool {
 	tb.Helper()
 
+	_, span := otel.Tracer("").Start(ctx, "checkMongoDBURI")
```
chilagrow, line 119: What does `Tracer("")` do? A bit strange to have empty tracer arg. Example has name space in it but I'm not sure the impact for reporting https://github.com/open-telemetry/opentelemetry-go/blob/main/example/namedtracer/foo/foo.go#L34

chilagrow (commented):
Just few comments :hugs:

integration/setup/common.go
w84thesun: It will use the default name for tracer https://github.com/open-telemetry/opentelemetry-go/blob/c0fb8dec4cf2183f852ebd7e663053614e044bf4/trace.go#L23

noisersup (dismissed):
Tested everything locally, looks good to me!

integration/setup/common.go
rumyantseva: @chilagrow Dmitry mentioned the exact reason. I suggest not using named tracers unless we really need them. If it solves the question, could you resolve the thread, please?

integration/setup/common.go
rumyantseva: Yeah, I like this idea! 👍

chilagrow (approved):

w84thesun (dismissed):
LGTM 👍

integration/setup/common.go
rumyantseva: done

chilagrow (dismissed):

w84thesun (approved):
👍 `otel` shorthand is a bit confusing at first 😄

noisersup (dismissed):
Checked the changes - LGTM!

integration/setup/common.go
```diff
@@ -29,6 +29,12 @@ import (
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
+	"go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/exporters/jaeger"
+	"go.opentelemetry.io/otel/sdk/resource"
+	tracesdk "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
```
AlekSi, line 37: Uh. UH!

Aside from that horrible package path… why 1.12 and not 1.17?

CONTRIBUTING.md
```diff
@@ -181,6 +181,19 @@ let's not fight over it.)
...
+#### Observability in tests
+
+Optionally, for those who prefer [OpenTelemetry tracing approach](https://opentelemetry.io/docs/concepts/signals/traces/),
+it is possible to use [Jaeger](https://www.jaegertracing.io/) in tests.
+
+Run `task jaeger` to start all-in-one Jaeger Docker container (the traces will be stored in memory and will be available
+while the container is running).
+
+Then run `task test-integration-pg JAEGER_ENDPOINT=http://localhost:14268/api/traces`
+to run `pg` handler tests with tracing enabled.
```
AlekSi, line 193: Why is that configurable and not always enabled? It seems like that is always useful without even more knobs

integration/setup/common.go
```diff
@@ -108,11 +116,16 @@ func SkipForPostgresWithReason(tb testing.TB, reason string) {
...
 
+	_, span := otel.Tracer("").Start(ctx, "checkMongoDBURI")
+	defer span.End()
+
 	defer trace.StartRegion(ctx, "checkMongoDBURI").End()
 	trace.Log(ctx, "checkMongoDBURI", uri)
 
 	clientOpts := options.Client().ApplyURI(uri)
 
+	clientOpts.Monitor = otelmongo.NewMonitor()
```
AlekSi, line 127: Please use `clientOpts.SetMonitor` just like you did below for consistency

integration/setup/common.go
```diff
@@ -334,7 +353,7 @@ func setupClient(tb testing.TB, ctx context.Context, uri string) *mongo.Client {
 }
 
 // startup initializes things that should be initialized only once.
-func startup() {
+func startup(tb testing.TB) {
```
AlekSi, line 356: `startup` can't accept `testing.TB` because that TB would be from one random test, and that completely messes logging and other things

integration/setup/common.go
```diff
@@ -351,5 +370,36 @@ func startup() {
...
+		tracesdk.WithSampler(tracesdk.AlwaysSample()),
+		tracesdk.WithResource(resource.NewSchemaless(
+			semconv.ServiceNameKey.String("FerretDB"),
+		)),
+	)
+
+	// Register TracerProvider globally to use it by default
+	otel.SetTracerProvider(tp)
+
+	tb.Cleanup(func() {
```
AlekSi, line 399: That runs when the test that first called `startup` exits, not when all tests exit

AlekSi (changes requested):

internal/util/testutil/testutil.go
```diff
@@ -29,9 +30,16 @@ import (
...
 
+	// root span
+	ctx, span := otel.Tracer("").Start(context.Background(), tb.Name())
+
 	// make one task per top-level test for nicer histograms
-	ctx, task := trace.NewTask(context.Background(), strings.Split(tb.Name(), "/")[0])
+	ctx, task := trace.NewTask(ctx, strings.Split(tb.Name(), "/")[0])
+
 	tb.Cleanup(task.End)
+	tb.Cleanup(func() {
```
AlekSi, line 40: `span.End()` can be called just like `task.End()` above

integration/setup/common.go
rumyantseva: Hmmm... Could you recommend me what I can use instead of this? I was looking for something like `teardown` similar to `startup`, but I don't see it.

w84thesun (approved):

Issue #1828, opened 2023-01-18
`distinct` for array -> object -> array doesn't work correctly


rumyantseva:
### Versions

Current main

### What did you do?

```
db.pampam.insert({
    _id: 'array-documents-nested',
    v: [
      { foo: [ { bar: 'hello' }, { bar: 'world' } ] }
    ]
  })

db.pampam.distinct("v.0.foo")
db.pampam.distinct("v.foo")
```

Check other places where dot notation was used.

Also, see #1724.

### What did you expect to see?

Both `db.pampam.distinct("v.0.foo")` and `db.pampam.distinct("v.foo")` should give the same response:

```
[ { bar: 'hello' }, { bar: 'world' } ]
```

### What did you see instead?

`db.pampam.distinct("v.foo")` gives an empty response

rumyantseva:
Similar use case with dot notation: https://github.com/FerretDB/FerretDB/issues/1828

rumyantseva:
Estimation is removed because it should be covered in #1724 (which is already M).

AlekSi:
Let's wait for Tigris to support `foo.0.bar` access

AlekSi:
It works now

Issue #1744, opened 2023-01-03
Update should return `WriteError` when given a bad update path


b1ron:
After we fix panic in #1954 / #1873, let's return correct error in this issue.

### Versions

v0.8.0

### What did you do?

This only seems to happen with `$set`, see repro in #1743.

### What did you expect to see?

```
WriteResult({
        "nMatched" : 0,
        "nUpserted" : 0,
        "nModified" : 0,
        "writeError" : {
                "code" : 56,
                "errmsg" : "The update path '0..' contains an empty field name, which is not allowed."
        }
})
```

### What did you see instead?

```
dance_ferretdb  | 2023-01-03T23:04:10.275+0545  DPANIC  // 172.29.0.1:45710 -> 172.29.0.3:27017         clientconn/conn.go:158  err must be non-nil
dance_ferretdb  | (err = <nil>)
dance_ferretdb  | github.com/FerretDB/FerretDB/internal/clientconn.(*conn).run.func2
dance_ferretdb  |       github.com/FerretDB/FerretDB/internal/clientconn/conn.go:158
dance_ferretdb  | runtime.gopanic
dance_ferretdb  |       runtime/panic.go:890
dance_ferretdb  | github.com/FerretDB/FerretDB/internal/clientconn.(*conn).run.func5
dance_ferretdb  |       github.com/FerretDB/FerretDB/internal/clientconn/conn.go:230
dance_ferretdb  | runtime.gopanic
dance_ferretdb  |       runtime/panic.go:890
dance_ferretdb  | github.com/FerretDB/FerretDB/internal/types.NewPath
dance_ferretdb  |       github.com/FerretDB/FerretDB/internal/types/path.go:42
dance_ferretdb  | github.com/FerretDB/FerretDB/internal/types.NewPathFromString
dance_ferretdb  |       github.com/FerretDB/FerretDB/internal/types/path.go:54
dance_ferretdb  | github.com/FerretDB/FerretDB/internal/handlers/common.processSetFieldExpression
dance_ferretdb  |       github.com/FerretDB/FerretDB/internal/handlers/common/update.go:157
dance_ferretdb  | github.com/FerretDB/FerretDB/internal/handlers/common.UpdateDocument
dance_ferretdb  |       github.com/FerretDB/FerretDB/internal/handlers/common/update.go:60
dance_ferretdb  | github.com/FerretDB/FerretDB/internal/handlers/pg.(*Handler).MsgUpdate.func2
[...]

PR #1864, opened 2023-01-24
Pushdown SQL queries with dot notation

### LINKED ISSUE #1841: Pushdown dot notation queries
### What should be done?

We should pushdown filters with keys specified in dot notation format for both pg and tigris handlers

noisersup:
# Description

Closes #1841.

## Readiness checklist


internal/handlers/pg/pgdb/query.go
```diff
@@ -274,21 +299,16 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
-				`((_jsonb->%[1]s)::jsonb = %[2]s)`+
-					// If it's not, but the value under the key k is an array - select if it contains the value equal to v.
-					` OR (_jsonb->%[1]s)::jsonb @> %[2]s`,
-				p.Next(), // placeholder $1 used for field key k for preventing SQL injections
+			filters = append(filters, fmt.Sprintf(
+				// If the value under the key is not equal to v,
+				// but the value under the key k is an array - select if it contains the value equal to v.
+				sql+` OR (_jsonb%[1]s%[2]s)::jsonb @> %[3]s`,
+				operator, // -> or #>
+				p.Next(), // placeholder $1 used for field key or it's path for preventing SQL injections
```
chilagrow, line 307:
```suggestion
		// Select if value under the key is equal to provided value
		// If the value under the key is not equal to v,
		// but the value under the key k is an array - select if it contains the value equal to v.
		sql := `((_jsonb%[1]s%[2]s)::jsonb = %[3]s) OR (_jsonb%[1]s%[2]s)::jsonb @> %[3]s`
		
		// operator is -> for non-array, #> for array
		// placeholder p.Next() is used to prevent SQL injections.
		// The first p.Next() gets $1 and used for key,
		// the second p.Next() gets $2 and used for value.
		filters = append(filters, fmt.Sprintf(sql, operator,  p.Next(), p.Next())
```
Because you ask for suggestion :see_no_evil:. I know you've separated SQL to keep common element, it's much easier to find args %1 %2 %3 if it's right next to each other. 

I didn't test this. So it might not even compile :sweat_smile:

internal/handlers/pg/pgdb/query.go
```diff
@@ -247,20 +247,45 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
+
 			case float64, string, types.ObjectID, int32, int64:
-				filters = append(filters, fmt.Sprintf(`((_jsonb->'_id')::jsonb = %s)`, p.Next()))
-				args = append(args, string(must.NotFail(pjson.MarshalSingleValue(v))))
+				filters = append(filters, fmt.Sprintf(
+					sql,      // simple query that checks for equal value under the key
+					operator, // -> or #>
+					p.Next(), // placeholder $1 used for field key or it's path for preventing SQL injections
+					p.Next(), // placeholder $2 used for field value v for preventing SQL injections
+				))
```
chilagrow, line 285:
```suggestion
		// Select if value under the key is equal to provided value
		sql := `((_jsonb%[1]s%[2]s)::jsonb = %[3]s)`
		
		// operator is -> for non-array, #> for array
		// placeholder p.Next() is used to prevent SQL injections.
		// The first p.Next() gets $1 and used for key,
		// the second p.Next() gets $2 and used for value.
		filters = append(filters, fmt.Sprintf(sql, operator,  p.Next(), p.Next())
```
Let's bring `sql` closer to this. It's just hard to see what operator is applied to which part of sql.

Maybe you won't like this :see_no_evil:

internal/handlers/pg/pgdb/query.go
```diff
@@ -247,20 +247,45 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
+				key = path.Slice()     // '{v,foo}'
+				prefix = path.Prefix() // 'v'
+			}
+		}
+
+		// Select if value under the key is equal to provided value
+		sql := `((_jsonb%[1]s%[2]s)::jsonb = %[3]s)` // %1 - operator; %2 - key; %3 - value
+
+		// Handle _id with a simpler query, as it can't be an array
+		if k == "_id" || prefix == "_id" {
```
chilagrow, line 274: Wow what a weird edge case `filter: bson.D{{"_id.foo", "bar"}},` :see_no_evil: 

Nice catch!

integration/query_test.go
```diff
@@ -473,6 +474,10 @@ func TestDotNotation(t *testing.T) {
 			filter:      bson.D{{"wsx.0.edc.0.rfv", int32(1)}},
 			expectedIDs: []any{"document-deeply-nested"},
 		},
+		"FieldID": {
+			filter:      bson.D{{"_id.foo", "bar"}},
+			expectedIDs: []any{bson.D{{"foo", "bar"}}},
```
chilagrow, line 479: Is it possible to add this to compat test? We prefer to compare result against compat unless we have reason why not to.

integration/query_array_compat_test.go
```diff
@@ -84,12 +84,14 @@ func testQueryArrayCompatDotNotation() map[string]queryCompatTestCase {
 			resultType: emptyResult,
 		},
 		"Field": {
-			filter:        bson.D{{"v.array", int32(42)}},
-			skipForTigris: "Tigris does not support language keyword 'array' as field name",
+			filter:         bson.D{{"v.array", int32(42)}},
+			skipForTigris:  "Tigris does not support language keyword 'array' as field name",
+			resultPushdown: true,
```
chilagrow, line 89: We have `"array-documents": bson.A{bson.D{{"field", int32(42)}}, bson.D{{"field", int32(44)}}},` in `Composites`. This might be interesting because we might find out if array -> object -> array doesn't work https://github.com/FerretDB/FerretDB/issues/1828 or array -> object doesn't work.

 
Could you try to push down to match one of the field like ` int32(42)`? I think query would be like this, but please double check. :hugs: 
```
			filter:         bson.D{{"v.array.0.field", int32(42)}},
```

If it doesn't work we can update the issue 1828 :hugs:

chilagrow (commented):
Awesome work catching edge cases too. Thanks for detail comments. I have suggestions.

internal/handlers/pg/pgdb/query.go
```diff
@@ -247,20 +247,45 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any) {
...
 
-		// don't iterate through array for _id keys to simplify the query
-		if k == "_id" {
+		operator := "->" // operator is the operator that is used to access the field. (->/#>)
+
+		var key any = k   // key can be either a string '"v"' or path '{v,foo}'
+		var prefix string // prefix is the first key in path, if the filter key is not a path - the prefix is empty
+
+		// If the key is in dot notation use path operator (#>)
+		if len(k) != 0 {
```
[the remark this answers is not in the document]
noisersup: >  Is this because we can have empty key like {"","foo"}?
Exactly, it's possible to have an empty key.

It's also worth to mention that if we try to put an empty key `""` inside the `NewPathFromString` function, it will panic, so we need to nest it inside the if statement

internal/handlers/pg/pgdb/query.go
noisersup: So we want to stay with previous approach where we have predefined queries for both cases, right?

internal/handlers/pg/pgdb/query.go
noisersup: Extracting comments on top of it seems much more readable, thanks!

integration/query_test.go
noisersup: We might need to add the _id field into the tests, but from my understanding for all of them we can only set "v" key, right?

internal/handlers/pg/pgdb/query.go
chilagrow: Yes please, I prefer having both part of sql query together. It's just I don't have to scroll up and down to read :hugs:

internal/handlers/pg/pgdb/query.go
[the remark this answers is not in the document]
chilagrow: >It's also worth to mention that if we try to put an empty key "" inside the NewPathFromString function, it will panic, so we need to nest it inside the if statement

Please create an issue for it or return the same error as compat :pray: Panic on something user can accidentally do is not ideal.

integration/query_test.go
chilagrow: You are right looks like the setup only allows _id to be comparable https://github.com/FerretDB/FerretDB/blob/main/integration/shareddata/shareddata.go#L149. So cannot be bson for now.

internal/handlers/pg/pgdb/query.go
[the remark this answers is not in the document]
noisersup: I meant that this panic is expected as we don't want to use `NewPathFromString` function on empty strings. The goal of this if statement is to don't call the function to don't panic on empty keys. Sorry if I didn't clarify correctly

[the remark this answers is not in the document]
chilagrow: I understand what you are trying to do here :hugs: 

But I'm concerned that `NewPathFromString` is panicing when it shouldn't. It is used in 20+ places and each time if we try to use it, we have to check for empty keys or it panics. It seems to me that `NewPathFromString` should return `BadValue` error. So I think you discovered a bug here. We should create an issue to handle it. Because other 20+ places don't have `len(k) != 0` logic to avoid panic. :see_no_evil:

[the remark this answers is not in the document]
noisersup: I see your point, and it seems that making empty check inside the path might be a good approach. 

I'm not sure about the "bug" part though. At least 9 of those calls are inside the tests, and it seems that we always try to check how operators and commands work with empty keys in test cases. 

>  don't have len(k) != 0 logic to avoid panic.

I think in most cases those check are made inside other functions (for example distinct command does it while fetching the parameter from document.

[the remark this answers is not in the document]
chilagrow: I think this is related https://github.com/FerretDB/FerretDB/issues/1744 :+1:

w84thesun (commented):
Can you please add a test for `prepareWhereClause`? There is a lot of stuff happening inside, so it may be easier to understand it with a unit test on hand.

chilagrow (approved):
Thanks for the change, looks great to me!

noisersup:
@w84thesun That's actually a great idea thanks! I'm bit worried about the time of implementation, but I can start from basic test cases and add more in next issues.

integration/query_array_compat_test.go
rumyantseva: Well, `v.array.0.field` will work in the current PR, but `v.array.field` is for #1828, yeah. 

I see there is `v.array.0.foo` case listed in this file, it covers this logic. I don't think we need any extra cases.

integration/query_array_compat_test.go
```diff
@@ -99,9 +101,45 @@ func testQueryArrayCompatDotNotation() map[string]queryCompatTestCase {
...
+			skipForTigris:  "Tigris does hot support characters as field name",
+			resultPushdown: true,
+		},
+		"FieldAsterix": {
+			filter:         bson.D{{"v.*", int32(42)}},
+			skipForTigris:  "Tigris does hot support characters as field name",
+			resultPushdown: true,
+		},
+		"FieldAt": {
+			filter:         bson.D{{"v.@", int32(42)}},
```
rumyantseva, line 129: Nice example!

integration/query_array_compat_test.go
```diff
@@ -99,9 +101,45 @@ func testQueryArrayCompatDotNotation() map[string]queryCompatTestCase {
...
+			skipForTigris:  "No suitable Tigris-compatible provider to test this data",
+			resultPushdown: true,
+		},
+		"DocumentDotNotationArrayDocumentNoIndex": {
+			filter: bson.D{{"v.array.foo", int32(42)}},
+			skip:   "https://github.com/FerretDB/FerretDB/issues/1828",
+		},
+		"FieldArrayIndex": {
+			filter:         bson.D{{"v.foo[0]", int32(42)}},
+			skipForTigris:  "Tigris does hot support characters as field name",
```
rumyantseva, line 115:
```suggestion
			skipForTigris:  "Tigris does not support characters as field name",
```

Here and below ;)

integration/shareddata/composites.go
```diff
@@ -31,9 +31,17 @@ var Composites = &Values[string]{
 	data: map[string]any{
 		"document":                   bson.D{{"foo", int32(42)}},
 		"document-composite":         bson.D{{"foo", int32(42)}, {"42", "foo"}, {"array", bson.A{int32(42), "foo", nil}}},
+		"document-array-document":    bson.D{{"array", bson.A{bson.D{{"foo", int32(42)}}, bson.D{{"foo", int32(41)}}}}},
```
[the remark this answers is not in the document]
rumyantseva: Hmm, do we really need it? Why `ArrayDocuments` is not enough?

rumyantseva (commented):
The changes look reasonable to me, but I doubt `document-array-document` is needed as we already have `array-documents-nested`, which seems to be pretty similar. 

Also, why did you decide to extend `Composites` in general? It's not really recommended as it won't work for Tigris (if we need more data, it's better to add new vars to this file).

chilagrow (approved):

w84thesun (approved):
:shipit:

PR #1871, opened 2023-01-25
Improve insert in Tigris by using single query for multiple documents

rumyantseva:
# Description

Closes FerretDB/engineering/issues/37.

Transaction don't help to make it faster (see a comment in FerretDB/engineering/issues/37).

I tried to take a look at the other places, but I don't see quick wins to insert faster.

## Readiness checklist


internal/handlers/tigris/msg_insert.go
```diff
@@ -107,6 +107,12 @@ func insertMany(ctx context.Context, dbPool *tigrisdb.TigrisDB, fp *tigrisdb.Fet
 	var inserted int32
 	var insErrors common.WriteErrors
 
+	// Attempt to insert all the documents in the same request to make insert faster.
+	if err := dbPool.InsertManyDocuments(ctx, fp.DB, fp.Collection, docs); err == nil {
+		return int32(docs.Len()), &insErrors
```
chilagrow, line 112:
```suggestion
		return int32(docs.Len()), nil
```

The `err` is nil here.

internal/handlers/tigris/msg_insert.go
```diff
@@ -107,6 +107,12 @@ func insertMany(ctx context.Context, dbPool *tigrisdb.TigrisDB, fp *tigrisdb.Fet
...
+	}
+
+	// If the transaction failed, attempt to insert each document separately.
```
chilagrow, line 115: Does `InsertManyDocuments` fail for any integration test setup? I assume not, but since we have fallback, just wanted to double check :hugs:

chilagrow (commented):
Just have a question :hugs:

internal/handlers/tigris/msg_insert.go
rumyantseva: No, as far as I can see, it doesn't. I expect it to fail for ordered insert tests.

internal/handlers/tigris/msg_insert.go
rumyantseva: It will not work here because the caller expects `insErrors` to be not `nil`. I decided to keep it as is and not refactor it.

chilagrow: That was unexpected but I see the logic now :hugs:

internal/handlers/tigris/msg_insert.go
chilagrow: Cool, I guess if `InsertManyDocuments` starts to fail for whatever reason, we will be able to see it on tigris docker logs.

chilagrow (approved):

noisersup (approved):
:shipit:

PR #1878, opened 2023-01-27
Use multiple Tigris instances to run tests

### LINKED ISSUE #1568: Rework integration test setup and auth

chilagrow:
# Description

Closes #1568.

Use 4 additional instances for running tests for tigris exposed on ports 8082, 8083, 8085 and 8086. Skipping 8084 because it's used by something else.

On CI, this change appears to runs tigris tests in 6 to 7 minutes. I checked recent merges to main which tigris tests appear to run in 10 to 11min. I don't know how much more $$ it might use.
 
## Readiness checklist


integration/setup/startup.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// startupInitializer creates an instance of startupInitializer.
+func newStartupInitializer() *startupInitializer {
+	nPortCalls := uint64(0)
+	return &startupInitializer{nPortCalls: &nPortCalls}
+}
+
+// getNextTigrisPort gets the next port number of Tigris to be used
+// for testing in Round Robin fashion.
```
noisersup, line 38: Nice usage!

noisersup (approved):
Tigris tests faster than PostgreSQL? :laughing: I guess it's because of number of skips

LGTM!

integration/setup/startup.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Package setup provides integration tests setup helpers.
+package setup
+
+import (
+	"sync/atomic"
+)
+
+// ports are available port of Tigris.
+var ports = []uint16{8081, 8082, 8083, 8085, 8086}
```
rumyantseva, line 23: I like this idea, but I'm a bit skeptical about these ports being defined here as the exact number of exact ports and not being configurable.
Is there a plan to make those things configurable (like `tigris-url` itself) - maybe in separate PRs?

rumyantseva (approved):
I like the idea! Asked a question about configuration.

integration/setup/startup.go
chilagrow: I've looked at it briefly and decided I would not do it for this, currently there were several places where ports are defined, env tool, docker-compose.yml, and within integrations tests. So it's nice to have configurable but also it touches more than the integration test parameter.  For this PR, I'm gonna do quick win. :see_no_evil:

rumyantseva: Oh, `envtool` is a nice catch as we wait for `8081` only, and not for the other ports...  
I'm ok to merge it as is to have a quick win, and maybe we can reconsider this and make it configurable after #1857 (so, it'll be easier to deal with configuration).

w84thesun (approved):
👍

cmd/envtool/envtool.go
```diff
@@ -151,33 +151,36 @@ func setupPostgresSecured(ctx context.Context, logger *zap.SugaredLogger) error
 func setupTigris(ctx context.Context, logger *zap.SugaredLogger) error {
 	logger = logger.Named("tigris")
 
-	err := waitForPort(ctx, logger, 8081)
-	if err != nil {
-		return err
-	}
+	// Skip 8084 because since it is used by userland proxy.
```
AlekSi, line 154: By what "userland proxy"?

integration/setup/startup.go
AlekSi: I think we should start paying a bit more attention to our code comments and not leave explanations on GitHub only, where they are unlikely to be found.

In this case, cross-linking those ports to the Docker Compose configuration would significantly improve [understanding](https://www.osnews.com/story/19266/wtfsm/).

chilagrow: Yeah agree. I've added comments. I also foresee the number of instances are just a guess and need adjusting hence a new issue https://github.com/FerretDB/FerretDB/issues/1887

w84thesun (dismissed):
👍

noisersup (approved):
:shipit:

rumyantseva (approved):
LGTM

integration/setup/startup.go
AlekSi: See bd05d83c9d5df8e36f41de8aff7a95bb60ed215b

AlekSi (approved):

chilagrow:
Thank you for the clean up @AlekSi :hugs:

PR #1880, opened 2023-01-27
Pushdown `$eq` query operator

### LINKED ISSUE #1840: Pushdown `$eq` and `$ne` query operators
We should pushdown `$eq` and `$ne` for numbers, strings and ObjectIDs for `pg` (both) and `tigris` (`$eq` only) handlers.

The dot notation is in the scope.

Check compression of large float and integer numbers.
* $eq operator does not use regex match
* implicit operator does use regex match

noisersup:
# Description

Closes #1840.

- `$eq` for pg supported
- `$eq` for tigris supported, `$ne` moved to another issue
- smaller ~~and larger~~ refactorings

~~Due to the complex SQL queries, this PR introduces larger refactor inside `pgdb` package.~~

~~Making "interface" to the comparison logic by separating query related logic into the builder struct will make introducing more complicated operators easier, make code less repetitive, even dot notation should be much easier to re-implement.~~

~~As the complexity of building queries on Tigris will grow, we should also introduce similar approach for building queries there.~~

## Readiness checklist


internal/handlers/pg/pgdb/query.go
```diff
@@ -268,61 +267,133 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any, error) {
...
+
+				case "$ne":
+					switch docVal := docVal.(type) {
+					case *types.Document, *types.Array, types.Binary, bool,
+						time.Time, types.NullType, types.Regex, types.Timestamp:
+						// type not supported for pushdown
+					case float64, string, types.ObjectID, int32, int64:
+						builder.addFilter(k, ne, docVal)
+					default:
+						panic(fmt.Sprintf("Unexpected type of value: %v", v))
```
chilagrow, line 305: This part has a lot of repeat of `$eq`, since supported types are the same, could we not repeat it all?

internal/handlers/pg/pgdb/query.go
```diff
@@ -268,61 +267,133 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any, error) {
...
-			// If the value under the key is not equal to v,
-			// but the value under the key k is an array - select if it contains the value equal to v.
-			sql := `(_jsonb->%[1]s)::jsonb = %[2]s OR (_jsonb->%[1]s)::jsonb @> %[2]s`
-
-			// placeholder p.Next() returns SQL argument references such as $1, $2 to prevent SQL injections.
-			// placeholder $1 is used for field key,
-			// placeholder $2 is used for field value v.
-			filters = append(filters, fmt.Sprintf(sql, p.Next(), p.Next()))
-			args = append(args, k, string(must.NotFail(pjson.MarshalSingleValue(v))))
+			builder.addFilter(k, eq, v)
```
chilagrow, line 319: I think this is logic is specifically for implicit filter operator. It's quite hard to understand as it is though.

internal/handlers/pg/pgdb/query.go
```diff
@@ -268,61 +267,133 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any, error) {
...
+		fb.args = append(fb.args, key, string(must.NotFail(pjson.MarshalSingleValue(val))))
+
+	case ne:
+		sql := `NOT ( ` +
+			// does document contain the key,
+			// it is necessary, as NOT won't work correctly if the key does not exist.
+			`_jsonb ? %[1]s AND ` +
+
+			// does the value under the key is equal to the filter
+			`(_jsonb->%[1]s)::jsonb ` + eqOperator + ` %[2]s AND ` +
```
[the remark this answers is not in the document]
chilagrow: I'm not sure how it works. Because `fb.addFilter(key,eq,val)` just add a filter, here we would want to produce a string.

integration/query_comparison_compat_test.go
```diff
@@ -86,6 +86,32 @@ func testQueryComparisonCompatImplicit() map[string]queryCompatTestCase {
...
+		"Int64Min": {
+			filter:         bson.D{{"v", int64(math.MinInt64)}},
+			resultPushdown: true,
+			skip:           "https://github.com/FerretDB/FerretDB/issues/2057",
+		},
+		"Int64DoubleBig": {
+			filter:         bson.D{{"v", int64(2 << 60)}},
+			resultPushdown: true,
+			skip:           "https://github.com/FerretDB/FerretDB/issues/2057",
+		},
```
chilagrow, line 113: Do you have a test for string being regular expression? The Implicit and $eq behave differently on that.

internal/handlers/pg/pgdb/query.go
```diff
@@ -268,61 +267,133 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any, error) {
...
+
+			// does the value type is equal to the filter's one
+			`(_jsonb->'$s'->'p'->%[1]s->'t')::jsonb = '"` + pjson.GetTypeOfValue(val) +
```
chilagrow, line 376: So I'm not sure if this is working. For example, if val is `int(1)`, db has `double(1)` and `long(1)` they are considered equal. For example, `{v: {$ne: 1}` on values `{v: int(1)}`, `{v: double(1)}` and `{v: long(1)}` should be false. 
But tests in integration are passing so I don't think I'm understanding how it works.

internal/handlers/pg/pgdb/query.go
```diff
@@ -268,61 +267,133 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any, error) {
...
-				// placeholder $1 is used for field key,
-				// placeholder $2 is used for field value v.
-				filters = append(filters, fmt.Sprintf(sql, p.Next(), p.Next()))
-				args = append(args, k, string(must.NotFail(pjson.MarshalSingleValue(v))))
-
-			default:
-				panic(fmt.Sprintf("Unexpected type of value: %v", v))
+		switch v := v.(type) {
+		case *types.Document:
+			iter := v.Iterator()
```
chilagrow, line 272: This part becomes difficult to read. I think multiple level of switch makes it hard to follow.it was difficult to tell that documents were iterated for $eq and $ne

chilagrow (changes requested):
To me the current implementation is hard to follow, but I raise two point which are about
* implicit and $eq handling string as regular expression needs to be addressed
* $ne on different number type and how it is working

AlekSi (changes requested):
See Slack

internal/handlers/pg/pgdb/query.go
```diff
@@ -268,61 +267,133 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any, error) {
...
+	query, args := builder.generateWhereClause()
+
+	return query, args, nil
+}
+
+// builderOperator represents available filtersBuilder operators.
+type builderOperator uint8
+
+const (
+	eq builderOperator = iota
```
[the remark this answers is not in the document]
noisersup: > We have key like $eq and $ne.

In this example builderOperators `eq` and `ne` are not the same thing as `$eq` and `$ne`.
`$eq` is a FerretDB operator that handles different types in a specific way, it can contain much more FerretDB specific logic inside. 
On the other hand `eq` is representation of a simple PostgreSQL clause, which might be useful for `$eq` and other operators to work.

> Why do we need to introduce this?

It was an attempt to wrap some basic query building logic into the separate function. 
Currently for every operator we need to:
- check if the key is `_id` because we use different SQL operator for those
- check if the key is dot notation, because we use different SQL operator for those either
- write whole query for each operator
- append to the filters and args variables to generate actual SQL query afterwards

We need to do a lot of stuff just for the actual SQL query generation, but that's not all, 
as we also need to handle some FerretDB operators logic, such as handling different types in different way,
everything in a single function.

Imo it's good to move database level comparison into different place and make it easy to just call it with single lines of code. 
- It will prevent us from code repetition, as all the `_id`/dot notation logic can be handled inside of it.
- We wouldn't need to write more and more SQL code for every operator, but rather just "build" those queries with existing blocks
(equal operation is useful for many other cases than just `$eq`)
- `prepareWhereClause` function would just check for types and operator specific logic, and call appropriate builder functions
- It will also prevent us from introducing more and more of nested `switch`es, loops and `if`s.

And yes, as we agreed, we don't restructure this before GA, but afterwards I would strongly recommend to take a look at this, as
with re-introducing dot notation and adding a couple of additional operators, `prepareWhereClause` will be just painful to read...

internal/handlers/pg/pgdb/query.go
noisersup: Thanks for spotting that! I'm investigating that

internal/handlers/pg/pgdb/query.go
noisersup: That's why we need to split this function. Apart from separating builder logic, I would suggest also to keep the switch from `*types.Document` case inside another function, that would handle all operators. Wdyt?

internal/handlers/pg/pgdb/query.go
noisersup: I see your point, we could find some way to make those type switches less repetitive, although I believe it's just a matter of couple new operators/types, to introduce changes that handle different types in a different way.

internal/handlers/pg/pgdb/query.go
noisersup: For those types, implicit eq, and `$eq` share the same logic. I just simplified the SQL query from
```SQL
(_jsonb->$1)::jsonb = $2 OR (_jsonb->$1)::jsonb @> $2
```

to just

```SQL
(_jsonb->$1)::jsonb @> $2
```

as now we are sure that it's [noted in PostgreSQL docs](https://www.postgresql.org/docs/15/datatype-json.html#JSON-CONTAINMENT) that `@>` containment operator for a scalar values works the same as `=` equal operator.

In the future we could check explain cost of both of them though , but currently, I would suggest to keep it as simple as possible.

integration/query_comparison_compat_test.go
noisersup: Resolved on pair programming!

internal/handlers/pg/pgdb/query.go
noisersup: I've inserted those documents:
```
db.values.insertMany([{v:"1"},{v:1},{v:Int32("1")},{v:1.0},{v:Long("1")},{v:Double("1")}])
```
and it seems that they are converted to the decimal.
```
test> db.values.find()
[
  { _id: ObjectId("6400ae417fa60685f6b68b42"), v: '1' },
  { _id: ObjectId("6400ae417fa60685f6b68b43"), v: 1 },
  { _id: ObjectId("6400ae417fa60685f6b68b44"), v: 1 },
  { _id: ObjectId("6400ae417fa60685f6b68b45"), v: 1 },
  { _id: ObjectId("6400ae417fa60685f6b68b46"), v: Long("1") },
  { _id: ObjectId("6400ae417fa60685f6b68b47"), v: 1 },
  { _id: ObjectId("6400ae587fa60685f6b68b48"), v: 1 },
  { _id: ObjectId("6400ae717fa60685f6b68b49"), v: 1 },
  { _id: ObjectId("6400aeeb7fa60685f6b68b4a"), v: 1 },
  { _id: ObjectId("6400af347fa60685f6b68b4b"), v: 1 }
]
```

I even added more examples (such as v:1.1), and afterwards used `{$inc{v:-0.1}}`, and it seems that every time when the value can be decimal, it converts to decimal. I did not occur any diffs between FerretDB and MongoDB, while running `$ne` on those cases.

internal/handlers/pg/pgdb/query.go
w84thesun: Will we enable pushdown for different operators with different types? For `$eq` enable support for `time.Time` and for `$ne` or another operator `types.ObjectID`?
If that's not the case we can put that type checks into a small helper function like `isPushdownEnabledForValue(v any) bool`.

noisersup: I believe, that for different filter types, we might send different queries. If we are talking about `$eq`: after enabling `*types.Array`, we will send different queries than for strings or numbers.

w84thesun: That's unfortunate. 😞

internal/handlers/pg/pgdb/query.go
```diff
@@ -251,12 +251,13 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any, error) {
 			return "", nil, lazyerrors.Error(err)
 		}
 
-		if k != "" {
+		switch {
```
rumyantseva, line 254: I like this switch-case approach, it looks more readable than a chain of `if`s!

rumyantseva (approved):
The logic looks quite complicated to me, but I don't see a way to simplify it. Also, I like that there are unit tests, they help with the understanding of examples.

w84thesun (dismissed):
:shipit:

internal/handlers/pg/pgdb/query.go
chilagrow: Thanks a lot for the detailed testing!

internal/handlers/pg/pgdb/query.go
chilagrow: `isPushdownEnabledForOperatorValue(op ,v any) bool` would work though, but let's not do it in this issue. Next operator maybe

chilagrow (approved):

internal/handlers/pg/pgdb/query.go
chilagrow: for now let's keep it like this

internal/handlers/pg/pgdb/query.go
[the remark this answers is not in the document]
chilagrow: Thanks for the explanation. I think filter building of repetitiveness is actually fine. There are somethings that are easy to read when it is repetitive, I think sql is one of them. Because you are only focusing on one set of sql at a time. The nested switch is another story, yeah I agree it would be nice to prevent.

internal/handlers/pg/pgdb/query.go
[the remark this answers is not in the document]
chilagrow: > at this point we still build whole ne query from scratch 

I think this is very nice. I know we have tests to check that the produced sql is what we expect, but to make it readable it's easier to have it the way it is. SQL to me is easier to read if entire query is in one place :hugs:

integration/query_comparison_compat_test.go
chilagrow: For anyone interested, implicit and $eq is different when the filter value is regular expression. Since we do not support regular expression pushdown yet, nothing need to be address on this. I've learnt this too

internal/handlers/pg/pgdb/query.go
```diff
@@ -251,12 +251,13 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any, error) {
...
+		case k == "":
```
AlekSi, line 255: How is that possible?

internal/handlers/pg/pgdb/query.go
AlekSi: It is, but do we really need that chain? The first condition looks sus, and if it's not actually needed, that's not a chain anymore

internal/handlers/pg/pgdb/query.go
```diff
@@ -268,46 +269,81 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any, error) {
...
-			default:
-				panic(fmt.Sprintf("Unexpected type of value: %v", v))
-			}
+		// check if values are equal or if the left value contains the right one
+		eqOperator := "@>"
 
-			continue
+		if k == "_id" {
+			// check if values are equal
+			eqOperator = "="
```
AlekSi, line 277: Why we need that?

internal/handlers/pg/pgdb/query.go
```diff
@@ -268,46 +269,81 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any, error) {
...
 		}
 
 		switch v := v.(type) {
-		case *types.Document, *types.Array, types.Binary, bool, time.Time, types.NullType, types.Regex, types.Timestamp:
+		case *types.Document:
+			iter := v.Iterator()
+			defer iter.Close()
+
+			for {
+				docKey, docVal, err := iter.Next()
```
AlekSi, line 286: In general, longer names should indicate larger scope. `k` and `v` are shorter, but more prominent. I think is is worth renaming them and those variables

internal/handlers/pg/pgdb/query.go
```diff
@@ -268,46 +269,81 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any, error) {
...
+
+				switch docKey {
+				case "$eq":
+					switch v := docVal.(type) {
+					case *types.Document, *types.Array, types.Binary, bool,
+						time.Time, types.NullType, types.Regex, types.Timestamp:
+						// type not supported for pushdown
+					case float64, string, types.ObjectID, int32, int64:
+						// Select if value under the key is equal to provided value.
+						sql := `(_jsonb->%[1]s)::jsonb ` + eqOperator + ` %[2]s`
```
AlekSi, line 303:
```suggestion
						sql := `(_jsonb->` + p.Next() + `)::jsonb ` + eqOperator + ` ` + p.Next()
```

Let's not mix string concatenation and sprintf.

And btw, why we need `::jsonb`?

internal/handlers/pg/pgdb/query.go
```diff
@@ -268,46 +269,81 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any, error) {
...
+						// type not supported for pushdown
+					case float64, string, types.ObjectID, int32, int64:
+						sql := `NOT ( ` +
+							// does document contain the key,
+							// it is necessary, as NOT won't work correctly if the key does not exist.
+							`_jsonb ? %[1]s AND ` +
+							// does the value under the key is equal to the filter
+							`(_jsonb->%[1]s)::jsonb ` + eqOperator + ` %[2]s AND ` +
+							// does the value type is equal to the filter's one
+							`(_jsonb->'$s'->'p'->%[1]s->'t')::jsonb = '"` + pjson.GetTypeOfValue(v) +
```
AlekSi, line 324: Why do we check type only for `$ne`, but not for `$eq`?

AlekSi (commented):
There are merge conflicts

internal/handlers/pg/pgdb/query.go
noisersup: What exactly? Empty keys?

Here's small test from mongodb:
```
test> db.values.insertOne({"":1})
{
  acknowledged: true,
  insertedId: ObjectId("6401bf27047d38a39d55655c")
}
test> db.values.find({"":1})
[ { _id: ObjectId("6401bf27047d38a39d55655c"), '': 1 } ]
test>
```

internal/handlers/pg/pgdb/query.go
noisersup: As `_id` cannot be an array, containment operator is not needed for equal operations on it.

internal/handlers/pg/pgdb/query.go
noisersup: Yeah good point, will change that.

internal/handlers/pg/pgdb/query.go
noisersup: If you run `{v:{$eq:42}}` against the collection that contains int(42) and Timestamp(42), it's okay to prefetch both of them, as we filter those on our side.

For `$ne` we can't apply the same logic. If we don't check for type for `{v:{$ne:42}`, PostgreSQL will match both int and Timestamp documents, and invert results, so both of them won't be passed to FerretDB.

internal/handlers/pg/pgdb/query.go
noisersup: > And btw, why we need ::jsonb?

We don't... We use them from beginning, but now I see that there's no point of conversion into the jsonb type, when you select from jsonb column...

internal/handlers/pg/pgdb/query.go
noisersup:
```
	--- FAIL: TestQueryCompatRunner/NeDoubleZero/TestQueryCompatRunner_DateTimes (0.02s)
		query_compat_test.go:193: Compat (expected) IDs: [datetime datetime-epoch datetime-null datetime-year-max datetime-year-min]
		query_compat_test.go:193: Target (actual)   IDs: [datetime datetime-null datetime-year-max datetime-year-min]
	--- FAIL: TestQueryCompatRunner/NeDoubleZero/TestQueryCompatRunner_Scalars (0.04s)
		query_compat_test.go:193: Compat (expected) IDs: [binary binary-empty bool-false bool-true datetime datetime-epoch datetime-year-max datetime-year-min double double-1 double-2 double-3 double-4 double-5 double-6 double-7 double-big double-max double-max-overflow double-max-overflow-verge double-min-overflow double-min-overflow-verge double-smallest double-whole int32 int32-1 int32-2 int32-3 int32-max int32-min int64 int64-1 int64-2 int64-3 int64-big int64-max int64-min null objectid objectid-empty regex regex-empty string string-double string-empty string-whole timestamp timestamp-i unset]
		query_compat_test.go:193: Target (actual)   IDs: [binary binary-empty bool-false bool-true datetime datetime-year-max datetime-year-min double double-1 double-2 double-3 double-4 double-5 double-6 double-7 double-big double-max double-max-overflow double-max-overflow-verge double-min-overflow double-min-overflow-verge double-smallest double-whole int32 int32-1 int32-2 int32-3 int32-max int32-min int64 int64-1 int64-2 int64-3 int64-big int64-max int64-min null objectid objectid-empty regex regex-empty string string-double string-empty string-whole timestamp timestamp-i unset]
	--- FAIL: TestQueryCompatRunner/NeInt64Zero/TestQueryCompatRunner_DateTimes (0.11s)
		query_compat_test.go:193: Compat (expected) IDs: [datetime datetime-epoch datetime-null datetime-year-max datetime-year-min]
		query_compat_test.go:193: Target (actual)   IDs: [datetime datetime-null datetime-year-max datetime-year-min]
	--- FAIL: TestQueryCompatRunner/NeInt64Zero/TestQueryCompatRunner_Scalars (0.04s)
		query_compat_test.go:193: Compat (expected) IDs: [binary binary-empty bool-false bool-true datetime datetime-epoch datetime-year-max datetime-year-min double double-1 double-2 double-3 double-4 double-5 double-6 double-7 double-big double-max double-max-overflow double-max-overflow-verge double-min-overflow double-min-overflow-verge double-smallest double-whole int32 int32-1 int32-2 int32-3 int32-max int32-min int64 int64-1 int64-2 int64-3 int64-big int64-max int64-min null objectid objectid-empty regex regex-empty string string-double string-empty string-whole timestamp timestamp-i unset]
		query_compat_test.go:193: Target (actual)   IDs: [binary binary-empty bool-false bool-true datetime datetime-year-max datetime-year-min double double-1 double-2 double-3 double-4 double-5 double-6 double-7 double-big double-max double-max-overflow double-max-overflow-verge double-min-overflow double-min-overflow-verge double-smallest double-whole int32 int32-1 int32-2 int32-3 int32-max int32-min int64 int64-1 int64-2 int64-3 int64-big int64-max int64-min null objectid objectid-empty regex regex-empty string string-double string-empty string-whole timestamp timestamp-i unset]
	--- FAIL: TestQueryCompatRunner/NeString/TestQueryCompatRunner_Regexes (0.02s)
		query_compat_test.go:193: Compat (expected) IDs: [regex regex-empty regex-null]
		query_compat_test.go:193: Target (actual)   IDs: [regex-empty regex-null]
	--- FAIL: TestQueryCompatRunner/NeString/TestQueryCompatRunner_ArrayRegexes (0.02s)
		query_compat_test.go:193: Compat (expected) IDs: [array-regex]
		query_compat_test.go:193: Target (actual)   IDs: []
```

internal/handlers/pg/pgdb/query.go
noisersup: > Let's not mix string concatenation and sprintf.

How do we want to do that in `$ne`?

```go
p1, p2 := p.Next(), p.Next()

sql := `NOT ( ` +
	// does document contain the key,
	// it is necessary, as NOT won't work correctly if the key does not exist.
	`_jsonb ? ` + p1 + ` AND ` +
	// does the value under the key is equal to the filter
	`_jsonb->` + p1 + ` ` + eqOperator + ` ` + p2 + ` AND ` +
	// does the value type is equal to the filter's one
	`_jsonb->'$s'->'p'->` + p1 + `->'t' = '"` + pjson.GetTypeOfValue(v) +
	`"')`
```

If we don't want to mix concatenation with sprintf, I would rather go the opposite way and include eqOperator to `Sprintf` values. What do you think?

AlekSi: > Let's not mix string concatenation and sprintf.

If we can't use just concatenation, let use just sprintf

internal/handlers/pg/pgdb/query.go
AlekSi: But is it worth making our code more complicated, given that index could be built for `@>` operators, so it would make sense to use only it?

internal/handlers/pg/pgdb/query.go
AlekSi: So we have a test for that?

noisersup: https://github.com/FerretDB/FerretDB/blob/1f076a5d8b3fd006a6f8570c593fdbc4518b1cf0/integration/basic_test.go#L231

AlekSi: Should that test check that pushdown was performed?

internal/handlers/pg/pgdb/query.go
noisersup: Previously we've used longer query for all other field, and simpler one for `_id`, so it made sense then. But yes, right now the difference is not big, and we can remove `_id` check for now. 

Maybe there are some differences in how effective Postgres evaluates both operators, but let's not focus on that right now. I will remove this check.

internal/handlers/pg/pgdb/query.go
noisersup: We can do it.

integration/basic_test.go
```diff
@@ -235,18 +235,29 @@ func TestEmptyKey(t *testing.T) {
...
 	require.NoError(t, err)
 
-	res, err := collection.Find(ctx, bson.D{{"", "foo"}})
+	var explainRes bson.D
+	require.NoError(t, collection.Database().RunCommand(ctx, explainQuery).Decode(&explainRes))
+
+	expectedPushdown := !setup.IsPushdownDisabled()
+	assert.Equal(t, expectedPushdown, explainRes.Map()["pushdown"], "Unexpected pushdown result")
+
+	res, err := collection.Find(ctx, filter)
```
noisersup, line 254: With a larger amount of key specific tests we could consider creating shareddata collection with key corner cases

internal/handlers/pg/pgdb/query.go
chilagrow: `NewPathFromString`  won't panic, it returns `ErrDocumentPathEmptyKey` error. So that check now can be removed :hugs: 

`k[0] == '$'` can be replaced with `strings.HasPrefix(k,'$')` to avoid panic too.

chilagrow (approved):
Looks good to me!

rumyantseva (approved):
LGTM

internal/handlers/pg/pjson/schema.go
```diff
@@ -62,6 +62,40 @@ const (
 	elemTypeLong      elemType = "long"
 )
 
+// GetTypeOfValue returns pjson type of supported value.
+func GetTypeOfValue(v any) string {
```
AlekSi, line 66: That's one bit I'm concerned about, but it is not used in this PR, so we could move to that other PR

AlekSi (approved):
I will let the team do the full review, but LGTM overall

internal/handlers/pg/pjson/schema.go
noisersup: Forgot about this one.

internal/handlers/pg/pgdb/query.go
rumyantseva: It looks fixed to me!

rumyantseva (approved):
LGTM

internal/handlers/pg/pgdb/query.go
```diff
@@ -251,78 +252,91 @@ func prepareWhereClause(sqlFilters *types.Document) (string, []any, error) {
...
-				// placeholder $1 is used for field key,
-				// placeholder $2 is used for field value v.
-				filters = append(filters, fmt.Sprintf(sql, p.Next(), p.Next()))
-				args = append(args, k, string(must.NotFail(pjson.MarshalSingleValue(v))))
-
-			default:
-				panic(fmt.Sprintf("Unexpected type of value: %v", v))
+		case errors.As(err, &pe):
+			if pe.Code() == types.ErrDocumentPathEmptyKey {
+				break
```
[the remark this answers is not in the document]
noisersup: Simplified it a bit, and added a comment. Previously we had more "flat" logic, as we checked if the key is empty on this function layer. Although, I guess it's `NewPathFromString` responsibility to do this check.

rumyantseva (approved):

rumyantseva (approved):

rumyantseva (approved):

chilagrow (approved):
Great work!

internal/handlers/pg/pgdb/query.go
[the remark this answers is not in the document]
chilagrow: The refactored logic looks very clear to me :)

noisersup:
Resolved merge conflicts.

rumyantseva (approved):

PR #1886, opened 2023-01-30
Fix querying for an embedded field in array

### LINKED ISSUE #1724: Querying for an embedded field in array does not behave correctly
### Versions

v0.7.1-29-gaf17fa7-dirty

### What did you do?

See repro script in #1722.

1 document returned.

### What did you see instead?

0 documents and an assertion.

### Environment

```markdown
- OS: Ubuntu
- Architecture: x86_64
- Version: 22.10
- Go: go1.19.3 linux/amd64
- Git: 2.37.2
- Deployment: Docker
```

### DoD

That should be supported by the common Go filtering code. Pushdown for such filters is not in the scope.

- [ ] Add a test data item to `Composites` to support `{a: [{}, {b: 1}]}` example (object -> array -> object)
- [ ] Use `ArrayDocuments`, it has this particular case
- [ ] Run existent test and see if smth needs to be modified because of the new test data item
- [ ] Add compat tests for find, count, findAndModify, update, delete for filter `{'a.b': 1}`
- [ ] If needed, create new issue for problems irrelevant to the current issue
- [ ] Update user docs (dot notation)
- [ ] It may also fix #1828, check it
- [ ] Don't push down queries that have dot notation

### LINKED ISSUE #1828: `distinct` for array -> object -> array doesn't work correctly

rumyantseva:
# Description

Closes #1724.
Closes #1828. 

See also my comment in #1724, it has some examples of this logic.

This is the first PR for this task, without pushdown.  Pushdown will be covered in a separate PR.

## Readiness checklist


integration/findandmodify_compat_test.go
```diff
@@ -316,7 +329,7 @@ func testFindAndModifyCompat(t *testing.T, testCases map[string]findAndModifyCom
 					targetErr = targetCollection.Database().RunCommand(ctx, targetCommand).Decode(&targetMod)
 					compatErr = compatCollection.Database().RunCommand(ctx, compatCommand).Decode(&compatMod)
 					require.Equal(t, targetErr, compatErr)
-					AssertEqualDocuments(t, targetMod, compatMod)
+					AssertEqualDocuments(t, compatMod, targetMod)
```
rumyantseva, line 332: 🙈

internal/types/path.go
```diff
@@ -156,40 +156,119 @@ func RemoveByPath[T CompositeTypeInterface](comp T, path Path) {
...
 
 	return next, nil
 }
 
+// getAllByPath returns all matching values by path - a sequence of indexes and keys.
+func getAllByPath[T CompositeTypeInterface](comp T, path Path) ([]any, error) {
+	next := []any{comp}
+
+	for _, p := range path.Slice() {
+		var newNext []any
```
chilagrow, line 198: Just a suggestion, doesn't need to be applied but I feel this can be read easier if we have `current` and `next` :innocent:

internal/types/document.go
```diff
@@ -287,15 +287,29 @@ func (d *Document) Remove(key string) any {
...
 }
 
-// GetByPath returns a value by path - a sequence of indexes and keys.
+// GetExactByPath returns a value by path - a sequence of indexes and keys.
 // If the Path has only one element, it returns the value for the given key.
-func (d *Document) GetByPath(path Path) (any, error) {
-	return getByPath(d, path)
+//
+// Use this method when you need the exact path to match, for example when you need
+// to get a value by path and modify it.
```
chilagrow, line 299:
```suggestion
// to get a value by path and modify it.
// For example, if document is "foo": [{a: 1}, {a: 2}, {b: 3}], then d.GetAllByPath(Path({"foo", "a"})) will return ErrDocumentPathKeyNotFound.
```
Correct?

internal/types/document.go
```diff
@@ -287,15 +287,29 @@ func (d *Document) Remove(key string) any {
...
+}
+
+// GetAllByPath returns all values by path - a sequence of indexes and keys.
+//
+// It should be used for querying and filtering.
+//
+// For arrays of objects if index is not specified, all matching objects are returned.
+// For example, if document is "foo": [{a: 1}, {a: 2}, {b: 3}], then d.GetAllByPath(Path({"foo", "a"})) will return [1, 2].
+func (d *Document) GetAllByPath(path Path) ([]any, error) {
+	return getAllByPath(d, path)
```
chilagrow, line 312: Should it be called `FilterByPath(...)`? The amazing doc string indicates it should only be called for querying and filtering but not for getting from the path :hugs:

internal/types/path.go
```diff
@@ -156,40 +156,119 @@ func RemoveByPath[T CompositeTypeInterface](comp T, path Path) {
...
+				if err != nil {
+					return nil, newDocumentPathError(ErrDocumentPathKeyNotFound, fmt.Errorf("types.getAllByPath: %w", err))
+				}
+
+				newNext = append(newNext, v)
+
+			case *Array:
+				index, err := strconv.Atoi(p)
+
+				switch err {
```
chilagrow, line 213: I think switch is not necessary, a `if err != nil { break}` should be sufficient here. :hugs:

internal/types/array.go
```diff
@@ -76,9 +76,14 @@ func (a *Array) Get(index int) (any, error) {
...
 
-// GetByPath returns a value by path - a sequence of indexes and keys.
-func (a *Array) GetByPath(path Path) (any, error) {
-	return getByPath(a, path)
+// GetExactByPath returns a value by the exact path - a sequence of indexes and keys.
+func (a *Array) GetExactByPath(path Path) (any, error) {
+	return getExactByPath(a, path)
+}
+
+// GetAllByPath returns all values that match the path - a sequence of indexes and keys.
```
chilagrow, line 84: Doc string needs updating.

internal/handlers/common/filter.go
```diff
@@ -58,31 +58,52 @@ func filterDocumentPair(doc *types.Document, filterKey string, filterValue any)
...
+		for _, docValue := range docValues {
+			switch docValue := docValue.(type) {
+			case *types.Document:
+				doc = docValue
+				filterKey = path.Suffix()
+			case *types.Array:
+				p := path.Suffix()
+				index, err := strconv.Atoi(p)
+
+				switch err {
```
chilagrow, line 75: Hmm, switch with `case nil:` and `default:` seems like it can be `if err != nil` :innocent:

internal/handlers/common/filter.go
```diff
@@ -58,31 +58,52 @@ func filterDocumentPair(doc *types.Document, filterKey string, filterValue any)
...
-				return false, err
-			}
+					doc = must.NotFail(types.NewDocument(filterKey, value))
 
-			doc = must.NotFail(types.NewDocument(filterKey, value))
+				default:
+					// If p is not an array index, it could be a key of a document inside the array.
+					// We need to iterate over all documents in the array and check if any of them
+					// matches the filter.
+					for i := 0; i < docValue.Len(); i++ {
```
chilagrow, line 92: If docValue length is zero what should happen?

chilagrow (commented):
I will take another look again, great work!

integration/count_compat_test.go
```diff
@@ -129,6 +129,9 @@ func TestCountCompat(t *testing.T) {
 		"FieldTypeArrays": {
 			filter: bson.D{{"v", bson.D{{"$type", "array"}}}},
 		},
+		"DotNotation": {
+			filter: bson.D{{"v.food.bar", "world"}},
```
chilagrow, line 133: Does this supposed to be `v.food.bar` or `v.foo.bar`? :hugs:

integration/query_comparison_compat_test.go
```diff
@@ -355,6 +355,9 @@ func testQueryComparisonCompatEq() map[string]queryCompatTestCase {
 		"NoSuchFieldNull": {
 			filter: bson.D{{"no-such-field", bson.D{{"$eq", nil}}}},
 		},
+		"DotNotation": {
+			filter: bson.D{{"v.food.bar", bson.D{{"$eq", "world"}}}},
```
chilagrow, line 359: I don't see anything in shareddata which matches `v.food.bar` but it's not using `emptyResult` so what is supposed to match this one?

chilagrow (commented):
Thanks for the awesome work!

noisersup (changes requested):
That's a lot of refactoring! Thanks for your research and findings about update cases.

Few comments from me and 1 question!

w84thesun (commented):
Discussed implementation details on the call.
Looks like there could be 2 possible options:
* combine `getAllByPath` and `getExactByPath` adding a parameter for the kind of result we are looking for and returning a slice of values;
* refactor those functions and turn `getExactByPath` into something like `filterByPath` which would be used for queries and filtering. 

Also as I understand there is an undocumented feature of dot notation which we probably want to write down somewhere.
It is so that when you are filtering with dot notation, the path `v.foo.bar` is acting more like regex `v.*foo.*bar` resulting in matches with the path `v.0.foo.0.bar`.
Although when you are trying to update a document with the dot notation path `v.foo.bar` it should be matched exactly in order to perform the correct update operation.

integration/query_comparison_compat_test.go
chilagrow: Still it should return `emptyResult` if it's incorrect array field. Or I am being blind on what's being matched. After all, both target and compat is matching _something_. So this test proves that implementation matches compat. But I don't what data it matches.

internal/handlers/common/filter.go
chilagrow: @noisersup which test covers this scenario :hugs:? If not, could you suggest please?

chilagrow:
> Discussed implementation details on the call. Looks like there could be 2 possible options:
> 
> * combine `getAllByPath` and `getExactByPath` adding a parameter for the kind of result we are looking for and returning a slice of values;
> * refactor those functions and turn `getExactByPath` into something like `filterByPath` which would be used for queries and filtering.
> 
> Also as I understand there is an undocumented feature of dot notation which we probably want to write down somewhere. It is so that when you are filtering with dot notation, the path `v.foo.bar` is acting more like regex `v.*foo.*bar` resulting in matches with the path `v.0.foo.0.bar`. Although when you are trying to update a document with the dot notation path `v.foo.bar` it should be matched exactly in order to perform the correct update operation.

@rumyantseva @w84thesunmMind blown with the dot notation :see_no_evil:. Particularly regrex part. I think having specific function for filtering is way to go here. Because getting path and getting filters are sufficiently different. But let's ask others what they think.

Yes, documentation would be great help here!

Finally, I have a question. If I have a query `v.food.bar` it seems it matches newly added document at `v.0.foo.0.bar` since the test returns nonEmptyResult. I suppose, regex is applied there.
What about if I query `v.fo.bar`, does it match `v.0.foo.0.bar`?

rumyantseva:
@noisersup @w84thesun @chilagrow thank you, everyone, for your comments! I think I now have some ideas about what I should refactor (both tests and code need some changes). I think I should also leave some comments in tests to explain how they work, or maybe at least I can include the exact `_id` to the filter, so it's easier to understand which documents are matched in which tests.

rumyantseva:
I close this one for now as it doesn't make much sense, but I'd like to reuse some intermediate commits. I'll open a new PR with more actual approach.

PR #1907, opened 2023-02-01
Fix `$pop` operator error handling of non-existent path

### LINKED ISSUE #1663: `$pop` with non-existing dot notation path does not fail
### Versions

0.7.1

### What did you do?

Moved `TestUpdateFieldCompatPop` test `PopDotNotation` to compat. 

Same error as compat

```
    update_field_compat_test.go:495: 
        	            	write exception: write errors: [Cannot use the part (array) of (v.array) to traverse the element ({v: /foo/i})]
        	Test:       	TestUpdateFieldCompatPop/PopDotNotation/TestUpdateFieldCompatPop-PopDotNotation_Regexes/regex
        	Messages:   	compat error; target returned no error
            --- FAIL: TestUpdateFieldCompatPop/PopDotNotation/TestUpdateFieldCompatPop-PopDotNotation_Regexes/regex 
```

### What did you see instead?

No error on scalar shareddata.

chilagrow:
# Description

Closes #1663.

## Readiness checklist


internal/handlers/common/update_array_operators.go
```diff
@@ -91,6 +97,40 @@ func processPopArrayUpdateExpression(doc *types.Document, update *types.Document
...
+		val := must.NotFail(doc.Get(prefix))
+		if prefixDoc, ok := val.(*types.Document); ok {
+			// recursively check if document contains prefix.
+			return ensureNonExistentPrefix(prefixDoc, prefix, path.TrimPrefix())
+		}
+
+		return commonerrors.NewWriteErrorMsg(
+			commonerrors.ErrUnsuitableValueType,
+			fmt.Sprintf(
+				"Cannot use the part (%s) of (%s) to traverse the element ({%s: %v})",
```
chilagrow, line 121: Message is not the same, please suggest if you know an easy fix.

```
compat -   Message: (string) (len=99) "Cannot use the part (array) of (v.array.foo.array) to traverse the element ({v: [ \"c\", \"b\", \"a\" ]})",
target  +   Message: (string) (len=92) "Cannot use the part (array) of (v.array.foo.array) to traverse the element ({v: &{[c b a]}})",
```

```
compat -   Message: (string) (len=84) "Cannot use the part (array) of (v.array.foo.array) to traverse the element ({v: []})",
target +   Message: (string) (len=87) "Cannot use the part (array) of (v.array.foo.array) to traverse the element ({v: &{[]}})",
```

internal/handlers/common/update_array_operators.go
```diff
@@ -91,6 +97,40 @@ func processPopArrayUpdateExpression(doc *types.Document, update *types.Document
 	return changed, nil
 }
 
+// ensureNonExistentPrefix checks if a given path and its prefix
+// does not exist in doc. It returns error if any prefix exists in the path.
```
rumyantseva, line 101: Let's specify what type of error it returns as it looks important (that's the error we want to return in the response).

internal/handlers/common/update_array_operators.go
```diff
@@ -91,6 +97,40 @@ func processPopArrayUpdateExpression(doc *types.Document, update *types.Document
 	return changed, nil
 }
 
+// ensureNonExistentPrefix checks if a given path and its prefix
```
rumyantseva, line 100: Based on the description I don't understand what's the difference between this func and `HasByPath`.

rumyantseva (commented):
I guess it's something `$pop`-specific, then probably it's fine... But I don't really understand the difference between this case and `hasByPath`. If `hasByPath` returns false, we check if the prefix exists? Can't we reuse `hasByPath` for it once again but only for prefix?

Also, the title of this PR doesn't look suitable for release notes, I believe it should mention `$pop` as an operator (like we do in other PRs), not just the word "Pop".

chilagrow:
> I guess it's something $pop-specific, then probably it's fine... But I don't really understand the difference between this case and hasByPath. If hasByPath returns false, we check if the prefix exists? Can't we reuse hasByPath for it once again but only for prefix?
Also, the title of this PR doesn't look suitable for release notes, I believe it should mention $pop as an operator (like we do in other PRs), not just the word "Pop". The title doesn't describe if we fix the bug or what actually is happening. As a user who doesn't have any idea about the issue you solve, I'd be confused seeing such a sentence in release notes.

can't use doc.HasByPath(path) because pop does not error on non-existent path, pop returns error if other value type exists on part of path.

internal/handlers/common/update_array_operators.go
chilagrow: Thanks I updated the comment

internal/handlers/common/update_array_operators.go
chilagrow: Added comment in the code :+1:

w84thesun (commented):
Overall good, proposed value format suggestion.

internal/handlers/common/update_array_operators.go
chilagrow: Thanks amazing!

w84thesun (dismissed):
👍

internal/handlers/common/update_array_operators.go
```diff
@@ -91,6 +96,44 @@ func processPopArrayUpdateExpression(doc *types.Document, update *types.Document
 	return changed, nil
 }
 
+// checkUnsuitableValueError returns ErrUnsuitableValueType when part of
+// the path is used by non document value type.
+// If path is "v.foo" and
+// - doc is {v: 42} it returns ErrUnsuitableValueType, v is used by unsuitable value type.
+// - doc is {c: 10} it returns no error since the path does not exist.
```
rumyantseva, line 103:
```suggestion
// checkUnsuitableValueError returns ErrUnsuitableValueType if 
// if finds an element that contains a non-document value. 
// If it fiends an element that does not exist, it returns nil.
//
// For example, if the path is "v.foo" and
// - doc is {v: 42}, it returns ErrUnsuitableValueType, v is used by unsuitable value type;
// - doc is {c: 10}, it returns no error since the path does not exist.
```

I'm not sure if my doc suggestion is better, so feel free to skip it

rumyantseva (commented):
I like the comments and explanations, but I'm not fully convinced it can't be a part of `GetByPath`. Maybe I don't understand some other corner cases. I don't press "request the changes", maybe others would agree with this approach, and then I'm ok to merge it as is. Feel free to schedule a pair programming call with me if it's easier to discuss things synchronously.

rumyantseva (approved):
After some thinking and experimentation, I actually decided to approve the PR, if we have better ideas around `GetByPath`, we can implement them.

internal/handlers/common/update_array_operators.go
```diff
@@ -53,13 +54,17 @@ func processPopArrayUpdateExpression(doc *types.Document, update *types.Document
...
-		if !doc.HasByPath(path) {
-			continue
-		}
-
 		val, err := doc.GetByPath(path)
 		if err != nil {
-			return false, err
+			// If any part of the path exists in doc, pop returns ErrUnsuitableValueType.
+			// Cannot use doc.HasByPath(path) because pop does not error on non-existent
+			// path, pop returns error if other value type exists on part of path.
```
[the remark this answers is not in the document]
chilagrow: Thanks for the suggestion, applied :+1:

internal/handlers/common/update_array_operators.go
chilagrow: I took part of it. Let me know what you think :pray:

internal/handlers/common/update_array_operators.go
```diff
@@ -91,6 +96,44 @@ func processPopArrayUpdateExpression(doc *types.Document, update *types.Document
 	return changed, nil
 }
 
+// checkUnsuitableValueError returns ErrUnsuitableValueType when part of
+// the path is used by non document value type.
+// If path is "v.foo" and
+// - doc is {v: 42} it returns ErrUnsuitableValueType, v is used by unsuitable value type.
+// - doc is {c: 10} it returns no error since the path does not exist.
+func checkUnsuitableValueError(doc *types.Document, key string, path types.Path) error {
```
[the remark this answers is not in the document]
chilagrow: Thanks for experimenting. I also attempted, but came to the conclusion that `getPath` is sufficiently different. Because `getPath` expects to return error if it does not exist, on the other hand the user case for this is no error if  the path doesn't exist

chilagrow:
Thanks a lot for input everyone :pray:

w84thesun (approved):

rumyantseva (approved):
LGTM

PR #1946, opened 2023-02-09
Fix item ordering on arrays replacing operations

### LINKED ISSUE #1736: Replace should not sort keys
MongoDB doesn't sort data for both upsert and non-upsert updates without any operator, but FerretDB [does](https://github.com/FerretDB/FerretDB/pull/1735), we need to fix that.

Fixing this issue should not be complicated, we just need to remove the [sort function call ](https://github.com/FerretDB/FerretDB/blob/3e28163606eedc26d5a4dd9e1213671577d672b1/internal/handlers/common/update.go#L121) and check if there was any reason behind it. We also need to create test cases for this and integrate them into compatibility tests if possible.

For update's `$set` and probably other update operator both MongoDB and FerretDB sorts the data before insert.

There is an issue with using the update command from the mongo-go-driver or mongosh, as it requires providing an atomic operator such as $set. However, it is possible to reproduce this issue by using the replaceOne command. The test case is provided in the comments.

Testing things via `mongo-test` might be blocked by #1850 but unnecessary. I've provided temporary fix that make running `mongo-test` possible in comments.

### DoD
- [ ] Make FerretDB to not sort documents on replace operations
- [ ] Create test cases for that

### Versions

v0.8.0

### What did you do?

See repro in https://github.com/FerretDB/FerretDB/pull/1735

The order of keys unchanged. 

### What did you see instead?

That the keys were sorted.

### Environment

```markdown
- OS: Ubuntu
- Architecture: x86_64
- Version: 22.10
- Go: go1.19.3 linux/amd64
- Git: 2.37.2
- Deployment: Docker
```

noisersup:
# Description

Closes #1736.

## Readiness checklist

### DIFF
--- a/integration/update_field_test.go
+++ b/integration/update_field_test.go
@@ -17,6 +17,7 @@ package integration
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
@@ -82,4 +83,26 @@ func TestUpdateFieldSet(t *testing.T) {
 			AssertEqualDocuments(t, tc.expected, actual)
 		})
 	}
+}
+
+func TestReplaceKeepOrder(t *testing.T) {
+	setup.SkipForTigris(t)
+
+	t.Parallel()
+	ctx, collection := setup.Setup(t, shareddata.Scalars)
+
+	expected := bson.D{{"_id", int32(1)}, {"c", int32(1)}, {"b", int32(2)}, {"a", int32(3)}}
+
+	_, err := collection.InsertOne(ctx, bson.D{{"_id", 1}})
+	require.NoError(t, err)
+
+	_, err = collection.ReplaceOne(ctx, bson.D{{"_id", 1}}, expected)
+	require.NoError(t, err)
+
+	res := collection.FindOne(ctx, bson.D{{"_id", 1}})
+
+	var actual bson.D
+	require.NoError(t, res.Decode(&actual))
+
+	assert.Equal(t, expected, actual)
 }
--- a/internal/handlers/common/update.go
+++ b/internal/handlers/common/update.go
@@ -141,10 +141,7 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
 				}
 			}
 
-			setDocKeys := setDoc.Keys()
-			sort.Strings(setDocKeys)
-
-			for _, setKey := range setDocKeys {
+			for _, setKey := range setDoc.Keys() {
 				setValue := must.NotFail(setDoc.Get(setKey))
 				doc.Set(setKey, setValue)
 			}


integration/update_field_test.go
```diff
@@ -83,3 +84,25 @@ func TestUpdateFieldSet(t *testing.T) {
 		})
 	}
 }
+
+func TestReplaceKeepOrder(t *testing.T) {
```
noisersup, line 88: It was tricky to implement compat test for that, so I've stayed with integration one. Please let me know if you see better option.

w84thesun:
```go
func TestReplaceKeepOrderCompat(t *testing.T) {
	setup.SkipForTigrisWithReason(t, "Schema violation")

	t.Parallel()

	s := setup.SetupCompatWithOpts(t, &setup.SetupCompatOpts{Providers: shareddata.Providers{shareddata.Scalars}})

	ctx := s.Ctx
	targetCollection := s.TargetCollections[0]
	compatCollection := s.CompatCollections[0]

	replace := bson.D{{"_id", int32(1)}, {"c", int32(1)}, {"b", int32(2)}, {"a", int32(3)}}

	_, err := targetCollection.InsertOne(ctx, bson.D{{"_id", 1}})
	require.NoError(t, err)
	_, err = compatCollection.InsertOne(ctx, bson.D{{"_id", 1}})
	require.NoError(t, err)

	_, err = targetCollection.ReplaceOne(ctx, bson.D{{"_id", 1}}, replace)
	require.NoError(t, err)
	_, err = compatCollection.ReplaceOne(ctx, bson.D{{"_id", 1}}, replace)
	require.NoError(t, err)

	targetResult := targetCollection.FindOne(ctx, bson.D{{"_id", 1}})
	require.NoError(t, targetResult.Err())
	compatResult := compatCollection.FindOne(ctx, bson.D{{"_id", 1}})
	require.NoError(t, compatResult.Err())

	var targetDoc, compatDoc bson.D
	require.NoError(t, targetResult.Decode(&targetDoc))
	require.NoError(t, compatResult.Decode(&compatDoc))

	assert.Equal(t, compatDoc, targetDoc)
}
```

noisersup: Does it actually make sens to initialize all providers just to insert the document with _id:1 and send replace query that will only match the _id:1?

w84thesun: If you don't need providers for this test you can change the setup line to `s := setup.SetupCompatWithOpts(t, &setup.SetupCompatOpts{AddNonExistentCollection: true})`

w84thesun: And it will initialize only with `Scalars`.

noisersup: Thanks!

rumyantseva (dismissed):
From my understanding of the code, this fix should fix it!
(I checked the places where we call `sort.Sort`, and only this one seems to be the one we don't need)

w84thesun (approved):
👍

integration/update_compat_test.go
```diff
@@ -485,3 +485,37 @@ func TestUpdateCompatMultiFlagCommand(t *testing.T) {
 
 	testUpdateCommandCompat(t, testCases)
 }
+
+func TestReplaceKeepOrderCompat(t *testing.T) {
+	setup.SkipForTigrisWithReason(t, "Schema violation")
```
chilagrow, line 490: Minor comment. We don't need to skip tigris here. The violation here comes from using not string `_id` and using random provider.

I think we can make sure tigris schema is not violated here by selecting provider. For example, `shareddata.Int32s` provider would work for your int data if we update tigris schema.
```
r := setup.SetupCompatWithOpts(t, &setup.SetupCompatOpts{Providers: []shareddata.Provider{shareddata.Int32s}})
```
Adding schema for`a`, `b`, `c` would be enough. `foo` is already there though.
https://github.com/FerretDB/FerretDB/blob/458bc77112ad35fb1653b989e3cfba1b96451d28/integration/shareddata/scalars.go#L377
```
"a": {%%type%%},
"b": {%%type%%},
"c": {%%type%%},
```

And `_id`s have to be string to not violate tigris schema we set in collections :sweat_smile:

rumyantseva (approved):
LGTM. 

It seems that we need to unify field names we use for Tigris collections, so we don't need to extend this schema even more in the future.

noisersup:
> It seems that we need to unify field names we use for Tigris collections, so we don't need to extend this schema even more in the future.

It would be cool to make some helper function that would get schema as an argument, to make some test specific schemas if needed.

w84thesun (approved):
👍

internal/handlers/common/update.go
```diff
@@ -141,10 +141,7 @@ func UpdateDocument(doc, update *types.Document) (bool, error) {
 				}
 			}
 
-			setDocKeys := setDoc.Keys()
-			sort.Strings(setDocKeys)
```
AlekSi, line 145: There are two other places where we call `sort.Strings`:

https://github.com/FerretDB/FerretDB/blob/09eb7ee5ca2f1313179a525610044e7d8ca6bb3d/internal/handlers/common/update.go#LL162C2-L162C13

https://github.com/FerretDB/FerretDB/blob/09eb7ee5ca2f1313179a525610044e7d8ca6bb3d/internal/handlers/common/update.go#L596

Should we update them too?

PR #1947, opened 2023-02-09
Add `iterator.WithClose` helper

AlekSi:
# Description

It could be used to track both rows and transaction in a single iterator.

## Readiness checklist

### DIFF
--- a/internal/handlers/tigris/tigrisdb/query_iterator.go
+++ b/internal/handlers/tigris/tigrisdb/query_iterator.go
@@ -34,14 +34,13 @@ var queryIteratorProfiles = pprof.NewProfile("github.com/FerretDB/FerretDB/inter
 
 // queryIterator implements iterator.Interface to fetch documents from the database.
 type queryIterator struct {
-	ctx context.Context
-
-	iter   driver.Iterator
+	ctx    context.Context
 	schema *tjson.Schema
 
-	stack []byte     // formatted stack trace that is used to format finalizer error message
-	m     sync.Mutex // protects call to iter.Next(), iter.Close and n
-	n     int        // number of current iteration starting from 0
+	m     sync.Mutex
+	iter  driver.Iterator
+	stack []byte // not really under mutex, but placed there to make struct smaller (due to alignment)
+	n     int
 }
 
 // newIterator returns a new queryIterator for the given driver.Iterator.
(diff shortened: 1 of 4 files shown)


internal/handlers/tigris/tigrisdb/query_iterator.go
```diff
@@ -102,9 +101,12 @@ func (iter *queryIterator) Next() (int, *types.Document, error) {
...
 		switch {
 		case err == nil:
-			fallthrough
+			// nothing
 		case IsInvalidArgument(err):
-			// Skip errors from filtering invalid types
+			// Skip errors from filtering different types.
+			// For example, given document {v: 42} and filter {v: "42"},
+			// MongoDB would skip that document because the type is different.
+			// Tigris returns a schema error in such cases that we ignore.
```
noisersup, line 109: Thank you!

internal/util/iterator/close.go
```diff
@@ -0,0 +1,42 @@
...
+// withClose wraps an iterator with a custom close function.
+type withClose[K, V any] struct {
+	iter  Interface[K, V]
+	close func()
+}
+
+// WithClose wraps an iterator with a custom close function.
+//
+// That function should call Close() method of the wrapped iterator.
+func WithClose[K, V any](iter Interface[K, V], close func()) Interface[K, V] {
```
noisersup, line 26: I don't understand the usage, can you provide some basic example? I've tried to came up with something locally but still doesn't understand. The close function provided by user should close the transaction, but what's the point if we do that already with InTransaction function?

w84thesun (approved):
👍

internal/util/iterator/close.go
```diff
@@ -0,0 +1,42 @@
...
+
+// withClose wraps an iterator with a custom close function.
+type withClose[K, V any] struct {
+	iter  Interface[K, V]
+	close func()
+}
+
+// WithClose wraps an iterator with a custom close function.
+//
+// That function should call Close() method of the wrapped iterator.
```
noisersup, line 25: Can't we implement that logic inside `withClose.Close` function, so it will call both user provided `close` func, and iterator `Close` one?

internal/util/iterator/close.go
AlekSi: See https://github.com/FerretDB/FerretDB/pull/1769/files#r1101318420

internal/util/iterator/close.go
AlekSi: We don't know the right order of those call there

noisersup (approved):

PR #1963, opened 2023-02-13
Add a flag for disabling query pushdown

### LINKED ISSUE #1912: Add a flag for disabling query pushdown
### What should be done?

Out compatibility tests ensure that pushdown works correctly, but we can't test _everything_. Eventually, our users or we ourselves will find a problem in our Go filtering code or pushed down queries, and it would not be easy to differentiate between the two. So we should introduce a way to diagnose those problems easier.

Let's add `--test-disable-pushdown` flag for that. When it is `true`, we should ignore `pgdb.SQLParam.Filter` and `tigrisdb.FetchParam.Filter` fields, fetch all documents, and filter them only on the Go side. (While we are on it, let's refactor `pgdb` and `tigrisdb` to make them more similar to each other; for example, it seems like `SQLParam` and `FetchParam` should not be called that differently.)

We should also add this flag to integration testing. When true, all pushdowns should be disabled, and tests that check them should be updated.

Let's not add a test configuration (Taskfile target, CI job) for that yet, but it should be possible – all tests should pass with and without that flag.

### Where?

All handlers, all integration tests, `pgdb` / `tigrisdb` unit tests.

### Definition of Done

- all handlers updated;
- unit tests added/updated;
- spot refactorings done;
- user documentation for flags updated.

noisersup:
# Description

Closes #1912.

- Implemented `-test-disable-pushdown` flag and tested it on both handler
- Made it possible to disable pushdowns on integration tests
- renamed both `SQLParams` and `FetchParams` to `QueryParams`
- made BuildFilter "standalone" function without assigning to TigrisDB struct
- Renamed pgdb `GetDocuments` to `QueryDocuments`, now in both handlers we have **Query**Params that are used inside **Query**Documents functions. In pgdb it also looks consistent with `queryById`!
- Other smaller refactors

## Readiness checklist


internal/handlers/tigris/msg_explain.go
```diff
@@ -28,7 +29,8 @@ import (
 
 // MsgExplain implements HandlerInterface.
 func (h *Handler) MsgExplain(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
-	dbPool, err := h.DBPool(ctx)
+	// DBPool is unused here, and the connection is established just for health check
+	_, err := h.DBPool(ctx)
```
rumyantseva, line 33: @noisersup awesome! I'll close my PR then!

w84thesun (approved):
A lot of changes here! Good job, I like how it turned out!

rumyantseva (approved):
👍

PR #1994, opened 2023-02-14
Refactor `types.NewPathFromString` to return `(Path, error)`

### LINKED ISSUE #1954: Refactor `types.NewPathFromString` to return `(Path, error)`
### What should be done?

`types.NewPathFromString` should return `(Path, error)` and not panic.

What errors are returned is not important in this issue. That's for #1744.

See #1873 for the original issue.

### Where?

https://github.com/FerretDB/FerretDB/blob/main/internal/types/path.go

### Definition of Done

- all handlers updated;
- unit tests added/updated;
- spot refactorings done;
- code comments updated (to at least explain what paths are invalid).

w84thesun:
# Description

Closes #1954.

## Readiness checklist

### DIFF
--- a/integration/update_field_compat_test.go
+++ b/integration/update_field_compat_test.go
@@ -52,18 +52,18 @@ func TestUpdateFieldCompatCurrentDate(t *testing.T) {
 		},
 		"BoolTrue": {
 			update: bson.D{{"$currentDate", bson.D{{"v", true}}}},
-			paths:  []types.Path{types.NewPathFromString("v")},
+			paths:  []types.Path{must.NotFail(types.NewPathFromString("v"))},
 		},
 		"BoolTwoTrue": {
 			update: bson.D{{"$currentDate", bson.D{{"v", true}, {"nonexistent", true}}}},
 			paths: []types.Path{
-				types.NewPathFromString("v"),
-				types.NewPathFromString("nonexistent"),
+				must.NotFail(types.NewPathFromString("v")),
+				must.NotFail(types.NewPathFromString("nonexistent")),
 			},
 		},
 		"BoolFalse": {
 			update: bson.D{{"$currentDate", bson.D{{"v", false}}}},
-			paths:  []types.Path{types.NewPathFromString("v")},
+			paths:  []types.Path{must.NotFail(types.NewPathFromString("v"))},
 		},
 		"Int32": {
 			update:     bson.D{{"$currentDate", bson.D{{"v", int32(1)}}}},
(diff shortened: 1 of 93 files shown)


integration/update_field_compat_test.go
```diff
@@ -52,34 +52,34 @@ func TestUpdateFieldCompatCurrentDate(t *testing.T) {
 		},
 		"BoolTrue": {
 			update: bson.D{{"$currentDate", bson.D{{"v", true}}}},
-			paths:  []types.Path{types.NewPathFromString("v")},
+			paths:  []types.Path{must.NotFail(types.NewPathFromString("v"))},
```
rumyantseva, line 55: I guess `NewStaticPath` should also work for all those changes with `must.NotFail`, but I don't think it's critical.

internal/types/path.go
```diff
@@ -74,12 +74,13 @@ type Path struct {
...
 func NewStaticPath(path ...string) Path {
-	return must.NotFail(newPathFromString(strings.Join(path, ".")))
+	return must.NotFail(NewPathFromString(strings.Join(path, ".")))
 }
 
-// TODO merge into NewPathFromString
-// https://github.com/FerretDB/FerretDB/issues/1954
-func newPathFromString(s string) (Path, error) {
+// NewPathFromString returns Path from path string and error.
+// Error return for empty path and empty path elements.
```
rumyantseva, line 81:
```suggestion
// It returns an error if the path is empty or contains empty elements.
```

internal/handlers/common/filter.go
```diff
@@ -56,7 +56,11 @@ func FilterDocument(doc, filter *types.Document) (bool, error) {
 func filterDocumentPair(doc *types.Document, filterKey string, filterValue any) (bool, error) {
 	if strings.ContainsRune(filterKey, '.') {
 		// {field1./.../.fieldN: filterValue}
-		path := types.NewPathFromString(filterKey)
+		path, err := types.NewPathFromString(filterKey)
+		if err != nil {
+			return false, err
```
rumyantseva, line 61: I belive with the current implementation for `NewPathFromString` we should return lazyerrors everywhere (like in `distinct`) to trace more details. 

Or otherwise we should add tests with wrong paths and see what "compat" returns and return the same.

rumyantseva (commented):
The changes look to me in accordance with the issue.

What is unclear to me (and it's unclear in the issue - do we want to check for compatibility and return similar errors, or the current approach with lazyerrors is enough).

integration/update_field_compat_test.go
w84thesun: Makes sense, changed it.

internal/handlers/common/filter.go
w84thesun: Added call to `lazyerrors.Error` when processing `NewPathFromString` errors.

rumyantseva (approved):
LGTM!

rumyantseva (approved):
I see the changes in `prepareWhereClause`, I partly understand why we added them (a case when a map is not enough seems possible to me).

chilagrow (approved):
Great work!

PR #1998, opened 2023-02-15
Store cursors in `ConnInfo`

### LINKED ISSUE #1733: Support `getMore` command

AlekSi:
# Description

Extracted from #1769.

Closes #1733.

## Readiness checklist

### DIFF
--- a/internal/clientconn/conninfo/conn_info.go
+++ b/internal/clientconn/conninfo/conn_info.go
@@ -12,28 +12,142 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package conninfo provides a ConnInfo struct that is used to handle connection-specificinfo
-// and can be shared through context.
+// Package conninfo provides access to connection-specific information.
 package conninfo
 
 import (
 	"context"
+	"math/rand"
+	"runtime"
+	"runtime/pprof"
 	"sync"
+	"sync/atomic"
+
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/debugbuild"
+	"github.com/FerretDB/FerretDB/internal/util/iterator"
 )
 
 // contextKey is a special type to represent context.WithValue keys a bit more safely.
 type contextKey struct{}
 
-// connInfoKey stores the key for withConnInfo context value.
-var connInfoKey = contextKey{}
+var (
+	// Context key for WithConnInfo/Get.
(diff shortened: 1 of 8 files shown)


internal/clientconn/conninfo/conn_info.go
```diff
@@ -12,28 +12,142 @@
...
+		lastCursorID.Store(rand.Uint32())
+	}
+}
 
 // ConnInfo represents connection info.
 type ConnInfo struct {
 	PeerAddr string
 
 	rw       sync.RWMutex
+	cursors  map[int64]Cursor
```
rumyantseva, line 57: Don't you wan't to use `uint32` as the map key?

internal/clientconn/conninfo/conn_info.go
```diff
@@ -12,28 +12,142 @@
...
+func (connInfo *ConnInfo) StoreCursor(iter iterator.Interface[int, *types.Document], filter *types.Document) int64 {
+	connInfo.rw.Lock()
+	defer connInfo.rw.Unlock()
+
+	var id int64
+
+	// use global, sequential, positive, short cursor IDs to make debugging easier
+	for {
+		id = int64(lastCursorID.Add(1))
+		if _, ok := connInfo.cursors[id]; id != 0 && !ok {
```
rumyantseva, line 128: This approach looks helpful indeed!

rumyantseva (approved):
Looks helpful!

w84thesun (approved):

internal/clientconn/conninfo/conn_info.go
AlekSi: It is int64 in the wire protocol. While we could save a few bytes per cursor, that would make more code depend on the ID generation. For example, consider this code:

```
// Cursor returns cursor by ID, or nil.
func (connInfo *ConnInfo) Cursor(id int64) *Cursor {
	connInfo.rw.RLock()
	defer connInfo.rw.RUnlock()

	c, ok := connInfo.cursors[uint32(id)]
	if !ok {
		return nil
	}

	return &c
}
```

It is subtly wrong: there are two int64 values that correspond to the same uint32 value. So we have to check that id > 0 and <= math.MaxUint32. That would split that logic into two functions. Not sure if it is worth it

rumyantseva: ok, makes sense!

PR #2003, opened 2023-02-15
Disable query pushdown if the corresponding flag is set

### LINKED ISSUE #1912: Add a flag for disabling query pushdown

rumyantseva:
# Description

Closes #1912.

# In this PR

Implementation:
- disable query pushdown for all the cases where queries are in use;
- add user documentaiton.

Spot refactoring in addition to the implementation:
- `pgdb.QueryParam` -> `pgdb.QueryParams`;
- `tigrisdb.QueryParam` -> `tigrisdb.QueryParams`.
- simplified `msg_delete` (for both `pg` and `tigris` handlers)

## Readiness checklist


cmd/ferretdb/main.go
```diff
@@ -201,7 +201,11 @@ func setupLogger(stateProvider *state.Provider) *zap.Logger {
 
 	// Similarly to Prometheus, unless requested, don't add UUID to all messages, but log it once at startup.
 	if !cli.Log.UUID {
-		startupFields = append(startupFields, zap.String("uuid", logUUID))
+		startupFields = append(
+			startupFields,
+			zap.String("uuid", logUUID),
+			zap.Bool("disable-pushdown", cli.Test.DisablePushdown),
```
AlekSi, line 207: It should not be under that `if`

internal/handlers/tigris/tigrisdb/query.go
```diff
@@ -30,16 +30,15 @@ import (
...
+// QueryParams represents options/parameters used by the fetch/query.
+type QueryParams struct {
 	Filter     *types.Document
 	DB         string
 	Collection string
 }
 
 // QueryDocuments fetches documents from the given collection.
-func (tdb *TigrisDB) QueryDocuments(ctx context.Context, param *QueryParam) (iterator.Interface[int, *types.Document], error) {
+func (tdb *TigrisDB) QueryDocuments(ctx context.Context, param *QueryParams) (iterator.Interface[int, *types.Document], error) {
```
AlekSi, line 41:
```suggestion
func (tdb *TigrisDB) QueryDocuments(ctx context.Context, qp *QueryParams) (iterator.Interface[int, *types.Document], error) {
```

cmd/ferretdb/main.go
rumyantseva: Ah! 😅

internal/handlers/common/filter.go
```diff
@@ -55,6 +55,11 @@ func FilterDocument(doc, filter *types.Document) (bool, error) {
 
 // filterDocumentPair handles a single filter element key/value pair {filterKey: filterValue}.
 func filterDocumentPair(doc *types.Document, filterKey string, filterValue any) (bool, error) {
+	if strings.HasPrefix(filterKey, "$") {
```
rumyantseva, line 58: This check is moved to the top of this function to handle cases like `path.$something`. Without this change, `-disable-pushdown=false` and `-disable-pushdown=true` would give different results (cc @chilagrow, it's relevant to your task a bit).

internal/handlers/tigris/tigrisdb/query.go
rumyantseva: fixed

cmd/ferretdb/main.go
rumyantseva: fixed

website/docs/flags.md
```diff
@@ -41,6 +41,8 @@ Some default values are overridden in [our Docker image](quickstart_guide/docker
 | `proxy-addr`           | See [Operation modes/Proxy](operation_modes.md#proxy)                     | `FERRETDB_PROXY_ADDR`           |                                         |
 | **Telemetry**          |                                                                           |                                 |                                         |
 | `telemetry`            | See [Configure telemetry](telemetry.md#configure-telemetry)               | `FERRETDB_TELEMETRY`            | `undecided`                             |
+| **Testing**            |                                                                           |                                 |                                         |
+| `test-disable-pushdown` | Disable pushing down queries to the backend (to only filter on FerretDB)  | `FERRETDB_TEST_DISABLE_PUSHDOWN` | `false`                                 |
```
rumyantseva, line 45: @AlekSi please suggest a better description (if any)

AlekSi: That's fine for now

cmd/ferretdb/main.go
```diff
@@ -196,6 +196,7 @@ func setupLogger(stateProvider *state.Provider) *zap.Logger {
 		zap.String("package", info.Package),
 		zap.Bool("debugBuild", info.DebugBuild),
 		zap.Any("buildEnvironment", info.BuildEnvironment.Map()),
+		zap.Bool("disable-pushdown", cli.Test.DisablePushdown),
```
AlekSi, line 199:
```suggestion
		zap.Bool("disablePushdown", cli.Test.DisablePushdown),
```

Note that it uses camelCase

internal/handlers/pg/msg_findandmodify.go
```diff
@@ -69,7 +69,7 @@ func (h *Handler) MsgFindAndModify(ctx context.Context, msg *wire.OpMsg) (*wire.
 		ctx = ctxWithTimeout
 	}
 
-	queryParam := pgdb.QueryParam{
+	queryParam := pgdb.QueryParams{
```
AlekSi, line 72:
```suggestion
	queryParams := pgdb.QueryParams{
```

internal/handlers/pg/msg_findandmodify.go
```diff
@@ -214,7 +214,7 @@ func (h *Handler) MsgFindAndModify(ctx context.Context, msg *wire.OpMsg) (*wire.
 type upsertParams struct {
 	hasUpdateOperators bool
 	query, update      *types.Document
-	queryParam         *pgdb.QueryParam
+	queryParam         *pgdb.QueryParams
```
AlekSi, line 217:
```suggestion
	queryParams         *pgdb.QueryParams
```

internal/handlers/tigris/msg_find.go
```diff
@@ -95,8 +95,16 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
 	return &reply, nil
 }
 
-// fetchAndFilterDocs fetches documents from the database and filters them using the provided QueryParam.Filter.
-func fetchAndFilterDocs(ctx context.Context, dbPool *tigrisdb.TigrisDB, qp *tigrisdb.QueryParam) ([]*types.Document, error) {
+// fetchAndFilterDocs fetches documents from the database and filters them using the provided QueryParams.Filter.
+func (h *Handler) fetchAndFilterDocs(ctx context.Context, dbPool *tigrisdb.TigrisDB, qp *tigrisdb.QueryParams) ([]*types.Document, error) { //nolint:lll // argument list is too long
```
AlekSi, line 99: The problem with converting those functions back to methods (as they were before) is that receiver has a lot of extra information, including more `*tigrisdb.TigrisDB` connection pools. I think we should move to functions, not into the opposite direction

rumyantseva: @AlekSi so then we pass whatever we need from `Handler` as params? Do we need to introduce a new struct to pass params, or should we just extend the list of params?

AlekSi: Up to you

w84thesun (commented):
Overall looks good, not much to add to @AlekSi's comments.

internal/handlers/pg/msg_findandmodify.go
rumyantseva: fixed

internal/handlers/pg/msg_findandmodify.go
rumyantseva: fixed

internal/handlers/tigris/msg_find.go
rumyantseva: fixed

internal/handlers/pg/msg_delete.go
```diff
@@ -180,8 +180,17 @@ func (h *Handler) prepareDeleteParams(deleteDoc *types.Document) (*types.Documen
 // execDelete fetches documents, filters them out, limits them (if needed) and deletes them.
 // If limit is true, only the first matched document is chosen for deletion, otherwise all matched documents are chosen.
 // It returns the number of deleted documents or an error.
-func execDelete(ctx context.Context, dbPool *pgdb.Pool, qp *pgdb.QueryParam, limit bool) (int32, error) {
+func (h *Handler) execDelete(ctx context.Context, dbPool *pgdb.Pool, qp *pgdb.QueryParams, limit bool) (int32, error) {
```
w84thesun, line 183: Should we remove the receiver here too?

rumyantseva: Yeah, sounds reasonable!

internal/handlers/common/filter.go
chilagrow: Thanks!

internal/handlers/pg/msg_find.go
```diff
@@ -110,9 +110,24 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
-func (h *Handler) fetchAndFilterDocs(ctx context.Context, tx pgx.Tx, sqlParam *pgdb.QueryParam) ([]*types.Document, error) {
-	iter, err := pgdb.QueryDocuments(ctx, tx, sqlParam)
+func fetchAndFilterDocs(ctx context.Context, fp *fetchParams) ([]*types.Document, error) {
+	// filter is used to filter documents on the FerretDB side,
+	// qp.Filter is used to filter documents on the PostgreSQL side (query pushdown).
+	filter := fp.qp.Filter
+
+	if fp.disablePushdown {
+		fp.qp.Filter = nil
+	}
```
chilagrow, line 128: Do you need a unit test for this and tigris? Integration test isn't in scope as I read the issue.

chilagrow (dismissed):

internal/handlers/pg/msg_find.go
rumyantseva: I believe not for now. When we set `disable-pushdown` flag with running tests, there will be checks (implemented in a separate PR).

Fashander (dismissed):

internal/handlers/pg/msg_delete.go
rumyantseva: fixed

w84thesun (dismissed):
👍

chilagrow (dismissed):
Great work!

internal/handlers/common/filter.go
AlekSi: That is actually quite concerning: `path.$something` should return the same result with pushdown enabled or disabled – an error about invalid path / field name. And I think we should a missing check for in the right place and add tests, not just move `if`

AlekSi (commented):
https://github.com/FerretDB/FerretDB/pull/2003/files#r1113844095

Fashander (dismissed):

internal/handlers/common/filter.go
rumyantseva: Interesting... I moved this back, and I can't find this case anymore. So, maybe it was something else. 

Then I don't introduce this change, but I added a couple of more tests with `$`.

Fashander (dismissed):

w84thesun (approved):

internal/handlers/pg/msg_delete.go
```diff
@@ -167,23 +161,40 @@ func (h *Handler) prepareDeleteParams(deleteDoc *types.Document) (*types.Documen
...
 			"limit",
 		)
 	}
 
-	return filter, limit, nil
+	return filter, limit == 1, nil
+}
+
+// deleteParams contains parameters for execDelete function.
+type deleteParams struct {
```
noisersup, line 175: I really like idea of creating parameter structs for those functions, although the name here seems misleading to me. Now we have `deleteParams` struct which is used to provide arguments to `execDelete` function, but in the same file we also provide `prepareDeleteParams` function which extracts parameter for mongo query. Can we rename it to let's say `execDeleteParams`?

noisersup (changes requested):
Looks great! Setting filter to nil on handler level seems much better! One small comment about struct naming.

w84thesun (approved):
👍

internal/handlers/pg/msg_delete.go
rumyantseva: Fixed, please take another look

noisersup (approved):
:shipit:

Fashander (approved):

PR #2004, opened 2023-02-15
Add support for `$addToSet` array update operator

### LINKED ISSUE #825: Support `$addToSet` array update operator
https://www.mongodb.com/docs/manual/reference/operator/update/addToSet/

DoD - to cover:

- [x] existing field - new value non-array, new value array
- [x] existing field - existing value
- [x] non-existent field - new value non-array, new value array
- [ ] invalid field name
- [x] dot notation

Not in scope:
- `$each` - https://github.com/FerretDB/FerretDB/issues/828

w84thesun:
# Description

Closes #825.

I didn't add any test cases with array values, as they would result in embedded arrays. Also can't come up with any invalid name that would return error from MongoDB.

## Readiness checklist


internal/handlers/common/update_array_operators.go
```diff
@@ -189,3 +190,114 @@ func processPushArrayUpdateExpression(doc *types.Document, update *types.Documen
...
+
+				value, err = array.Get(i)
+				if err != nil {
+					return false, lazyerrors.Error(err)
+				}
+
+				compareResult := types.Compare(value, addToSetValueRaw)
+
+				if compareResult == types.Equal {
+					shouldAdd = false
```
rumyantseva, line 269:
```suggestion
					shouldAdd = false
					break
```

I guess we don't need to iterate through the rest of the array then

internal/handlers/common/update_array_operators.go
```diff
@@ -189,3 +190,114 @@ func processPushArrayUpdateExpression(doc *types.Document, update *types.Documen
...
+
+		switch addToSetValueRaw := addToSetValueRaw.(type) {
+		case *types.Document, float64, string, types.Binary, types.ObjectID, bool,
+			time.Time, types.NullType, types.Regex, int32, types.Timestamp, int64:
+			shouldAdd := true
+
+			for i := 0; i < array.Len(); i++ {
+				var value any
+
+				value, err = array.Get(i)
```
rumyantseva, line 261: I think we can use `must.NotFail` here (not super critical).

internal/handlers/common/update_array_operators.go
```diff
@@ -189,3 +190,114 @@ func processPushArrayUpdateExpression(doc *types.Document, update *types.Documen
...
+		default:
+			panic(fmt.Sprintf("unhandled type %T", addToSetValueRaw))
+		}
+
+		// No values to append to the array.
+		if len(appendValues) == 0 {
+			continue
+		}
+
+		for _, appendValue := range appendValues {
```
rumyantseva, line 291: Can't we just do `array.Append(appendValues)` instead of the whole loop?

integration/update_array_compat_test.go
```diff
@@ -120,3 +120,47 @@ func TestUpdateArrayCompatPush(t *testing.T) {
 
 	testUpdateCompat(t, testCases)
 }
+
+func TestUpdateArrayCompatAddToSet(t *testing.T) {
+	t.Parallel()
+
+	testCases := map[string]updateCompatTestCase{
+		"DuplicateKeys": {
+			update:     bson.D{{"$pop", bson.D{{"v", 1}, {"v", 1}}}},
```
rumyantseva, line 129: Why `$pop`?

integration/update_array_compat_test.go
```diff
@@ -120,3 +120,47 @@ func TestUpdateArrayCompatPush(t *testing.T) {
 
 	testUpdateCompat(t, testCases)
 }
+
+func TestUpdateArrayCompatAddToSet(t *testing.T) {
+	t.Parallel()
+
+	testCases := map[string]updateCompatTestCase{
```
rumyantseva, line 127: I know we usually don't use filters in updates, but can we use one case filtered by `_id` where we try to `addToSet` the value that is already in the set? Or if it's covered by one of the current test cases, let's add a comment that clarifies which document (with which `_id`) covers it.

rumyantseva (commented):
In general, looks consistenly with the other parts of the code.

Added a few questions/suggestions.

integration/update_array_compat_test.go
w84thesun: Fixed.

internal/handlers/common/update_array_operators.go
w84thesun: Rewrote this part as we actually will add only 1 value to the set.

internal/handlers/common/update_array_operators.go
w84thesun: Fixed.

integration/update_array_compat_test.go
w84thesun: Added a comment.

Fashander (dismissed):

rumyantseva (dismissed):
LGTM

rumyantseva:
@w84thesun I see Tigris fails for `addToSet` tests, it is worth checking why (feel free to ping me if something is unclear about Tigris behavior).

rumyantseva (approved):
Looks consistent with the other array update tests.

@AlekSi the tests might be interesting having in mind what we discussed. We need to decide what we do with tests for Tigris in this case. My suggestion - continue doing what we do now (skip tests for tigris), but in principle we can come up with data set with some extended schemas that would work for Tigris).

chilagrow (approved):
looks good :hugs:

rumyantseva (approved):

chilagrow (approved):

Fashander (approved):

PR #2030, opened 2023-02-21
Fix querying an embedded field in array

### LINKED ISSUE #1724: Querying for an embedded field in array does not behave correctly

chilagrow:
# Description

Closes #1724.

## Readiness checklist


internal/handlers/common/filter.go
```diff
@@ -55,85 +55,191 @@ func FilterDocument(doc, filter *types.Document) (bool, error) {
...
 		}
 
-		// we pass the path without the last key because we want {fieldN: *someValue*}, not just *someValue*
-		docValue, err := doc.GetByPath(path.TrimSuffix())
-		if err != nil {
-			return false, nil // no error - the field is just not present
+		if filterSuffix, docs = getDocumentsAtSuffix(doc, path); len(docs) == 0 {
+			// When no document is found at suffix, use an empty one.
+			// So operators such as $nin is applied to the empty document.
+			docs = append(docs, types.MakeDocument(0))
```
chilagrow, line 70: Tested by `DocumentDotNotationArrayDocumentNoIndexNin` test case

rumyantseva (approved):
I don't have a better way to implement it, so I'm ready to approve.

As discussed in the call, maybe we can add some unit tests to showcase the implemented logic (but those unit tests should be excluded from code coverage because we can't trust they have the same logic as MongoDB).

internal/handlers/common/filter.go
```diff
@@ -55,85 +55,191 @@ func FilterDocument(doc, filter *types.Document) (bool, error) {
...
-	}
+	// If we got here, it means that none of the documents matched the filter.
+	return false, nil
+}
 
-	switch filterValue := filterValue.(type) {
-	case *types.Document:
-		// {field: {expr}} or {field: {document}}
-		return filterFieldExpr(doc, filterKey, filterValue)
+// getDocumentsAtSuffix finds documents at the suffix of the given path.
```
w84thesun, line 144: Let's add a comment that we go iteratively through the document to find all those matches.

w84thesun (commented):
Thanks for all the explanations! I'm pretty happy with the changes, but as discussed small comment for the `getDocumentsAtSuffix` function would help.

noisersup (dismissed):
Thank you for explanation. Great job with implementing this logic.

I see that pg integration tests are failing due to the timeout (same as in other PRs), so it might be necessary to increase it.

internal/handlers/common/filter.go
chilagrow: I added comment and sample example cases in doc string.

chilagrow:
> As discussed in the call, maybe we can add some unit tests to showcase the implemented logic (but those unit tests should be excluded from code coverage because we can't trust they have the same logic as MongoDB).

@rumyantseva  For now, I added show case in the doc string, downside is it might get outdated. I think we maybe able to use `Example....` instead of `Test...` https://go.dev/blog/examples, but not gonna figure out the code coverage related thing in this issue :see_no_evil:

rumyantseva (approved):
This is one of the most complicated path-related logical parts in FerretDB, I hope the docs will help our future contributors!

w84thesun (approved):
Thanks for the detailed comment!

noisersup (approved):
Examples help a lot!

Issue #1509, opened 2022-11-21
Implement `createIndexes` command


b1ron:
https://www.mongodb.com/docs/manual/reference/command/createIndexes/

DoD:

Types to consider:
- [x] Single field indexes
- [x] Multiple fields (compound) indexes

Various cases to test:
- [x] Sort on indexes (including multiple fields indexes)
   - [x] Test cases for various order settings in index specificaiton
   - [x] Test cases for invalid index spec
- [ ] Show indexes in query analyzer - in a separate issue #2297
- [x] Test cases for index intersection
- [x] Test case when we try to create two indexes with different names for the same keys (field + order pairs)
- [x] Test case when we try to create an index with the name that already exists
- [x] Test wether `{ a: 1, b: 1 }` and `{ b: 1, a: 1 }` are considered different indexes and could exist together
- [x] Test cases should use "create indexes -> list indexes" approach to check that list command returns the correct response
- [x] Check index name generation: check how if works index is created through the driver and though runCommand with a custom name
- [x] Check todos in the code
- [x] Update supported commands doc

**Not** in the scope:

- Tigris
- Unique indexes, see #2045
- Sparse indexes
- Special types (geo, text etc)
- Hidden indexes
- Partial indexes
- TTL indexes
- Background index building
- Collation
- Wildcard projection

PR #2048, opened 2023-02-23
Create unqiue index for `_id` automatically

### LINKED ISSUE #1384: Create unique index for `_id` automatically
### Versions

November 9, 2022

for example, https://github.com/FerretDB/FerretDB/commit/9e34c247112d8d2bc03f260a49a04a6b500284cf

### What did you do?

Call `db.pcs.insertOne({"_id": "foo", "value": "bar"})` twice

The second attempt to insert entry with the same value in the `_id` field should return an error.

```
test> db.pcs.insertOne({"_id": "foo", "value": "bar"})
{ acknowledged: true, insertedId: 'foo' }
test> db.pcs.insertOne({"_id": "foo", "value": "bar"})
 E11000 duplicate key error collection: test.pcs index: _id_ dup key: { _id: "foo" }
```

### What did you see instead?

Both entries were inserted:

```
test> db.pcs.insertOne({"_id": "foo", "value": "bar"})
{ acknowledged: true, insertedId: 'foo' }
test> db.pcs.insertOne({"_id": "foo", "value": "bar"})
{ acknowledged: true, insertedId: 'foo' }
test> db.pcs.find();
[ { _id: 'foo', value: 'bar' }, { _id: 'foo', value: 'bar' } ]
```

P.S. This is how it works for Postgres. For Tigris it returns an error, however, the error is not correct (we should return a proper error, but an internal error is returned instead).

DoD:
- See also: TODOs in the code
- Create index
- Not enforce id uniqueness in Go code (indexes should be enough for it) - lazyerror is good enough for now
- List indexes should return - just returning _id index is enough (no need to query metadata yet)

```[tasklist]
### Tasks
- [ ] #2044
```

### LINKED ISSUE #2044: Format index names in PG, so the length limit is not reached
With table names like `TestInsertCompat-InsertUnorderedOneError-InsertMany_St_4c6d95fa` the "default" index name could be something like `TestInsertCompat-InsertUnorderedOneError-InsertMany_St_4c6d95fa_field`, and then we reach the limit of index name. In such case, PG automatically truncates it, and it produces the same name for everything.

We should find a way to create indexes for long table names. 
One possible option is not format index names in a similar way to how we format collection->table names, so we ensure that the index name is not too long.

Metadata to consider storing:

- v
- key
- name
- unique

rumyantseva:
# Description

Closes #2044.
Closes #1384.

Refactoring: `type metadata struct ` is introduced in `pgdb` to group metadata-related methods.

## Readiness checklist


internal/handlers/pg/pgdb/database_metadata_indexes.go
```diff
@@ -0,0 +1,121 @@
...
+	return
+}
+
+// formatIndexName returns index name in form <shortened_name>_<name_hash>_idx.
+// Changing this logic will break compatibility with existing databases.
+func formatIndexName(collection, index string) string {
+	name := collection + "_" + index
+
+	hash32 := fnv.New32a()
+	_ = must.NotFail(hash32.Write([]byte(name)))
```
w84thesun, line 111: This probably could be simplified:
```suggestion
	must.NotFail(hash32.Write([]byte(name)))
```

w84thesun (commented):
Overall changes look reasonable, although it seems that there are some problems with locks.

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -19,28 +19,46 @@ import (
...
-// indexParams describes the parameters for creating an index.
+// indexParams contains parameters for creating an index.
 type indexParams struct {
-	schema   string // pg schema name
-	table    string // pg table name
-	isUnique bool   // whether the index is unique
+	db         string          // FerretDB database name
+	collection string          // FerretDB collection name
+	index      string          // FerretDB index name
+	key        *types.Document // Index specification (pairs of field names and sort orders)
```
chilagrow, line 31: `key` is currently not used for index creation in postgres, which is okay I guess.

chilagrow (commented):
Aside from what's already mentioned in TODO https://github.com/FerretDB/FerretDB/issues/1509

Changes look good to me!

AlekSi (commented):
Please take a look on a list of functions/methods in pgdb and make it simpler, smaller, and more consistent. In particular, I don't understand why we keep adding more helpers instead of removing them. For example, `setById` is different from `SetDocumentByID`?

internal/handlers/pg/pgdb/database_metadata_indexes.go
```diff
@@ -0,0 +1,121 @@
...
+
+	nameSymbolsLeft := maxIndexNameLength - hash32.Size()*2 - 5 // 5 is for "_" delimiter and "_idx" suffix
+	truncateTo := len(name)
+
+	if truncateTo > nameSymbolsLeft {
+		truncateTo = nameSymbolsLeft
+	}
```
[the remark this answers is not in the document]
rumyantseva: Yeah, let me add a test to showcase this example.

rumyantseva:
@AlekSi because `SetDocumentByID` works with metadata table, and if we need to store data for the metadata table we can't use it. Ther is not metadata table for metadata table. I'll schedule a pair programming session with you, as it's not straightforward to me what exactly you consider useless.

rumyantseva:
@w84thesun could you specify the problems with the locks you see? I see PG times out, but it's quite expected: we started adding `_id` index for all the collections we create, so tests became slower. I think we should just increase the timeout for PG tests.

AlekSi:
I did not say "useless", though. But there is a lot opportunity for refactoring (a.k.a. technical debt), and we should reduce it, not increase it. I don't believe that there is no opportunity for us to not have two very similar function doing two very similar things in the same package. We can do better, and we should.

w84thesun:
> @w84thesun could you specify the problems with the locks you see? I see PG times out, but it's quite expected: we started adding `_id` index for all the collections we create, so tests became slower. I think we should just increase the timeout for PG tests.

Yes, that's about timeout, not locks, I just started to read the output of the test from the end 😄

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -19,28 +19,46 @@ import (
...
+	if isUnique {
 		unique = " UNIQUE"
 	}
 
-	sql := `CREATE` + unique + ` INDEX IF NOT EXISTS ` + pgx.Identifier{p.table + "_id"}.Sanitize() +
-		` ON ` + pgx.Identifier{p.schema, p.table}.Sanitize() +
-		` ((_jsonb->>'_id'))`
+	sql := `CREATE` + unique + ` INDEX IF NOT EXISTS ` + pgx.Identifier{index}.Sanitize() +
+		` ON ` + pgx.Identifier{schema, table}.Sanitize() +
+		` ((_jsonb->>'_id'))` // TODO Provide ability to set fields https://github.com/FerretDB/FerretDB/issues/1509
```
noisersup, line 61: We access the value by text `->>` and not by jsonb  `->` here. I believe that's the reason why `{_id: 1}` is now considered the same as `{_id:"1"}`.

**FerretDB**:
```
test> db.values.insertOne({_id:1})
test> db.values.insertOne({_id:"1"})
MongoServerError: E11000 duplicate key error collection: test.values index: _id_ dup key: { _id: "1" }
```

**MongoDB**:
```
test> db.values.insertOne({_id:1})
test> db.values.insertOne({_id:"1"})
{ acknowledged: true, insertedId: '1' }
test> db.values.insertOne({_id:"1"})
```

noisersup (changes requested):
It seems like a great idea to use an index for that!

I left a comment about the query. Also it seems that comma in the description "Closes" tag did not include all the issues to the Development section.

internal/handlers/pg/pgdb/update.go
```diff
@@ -32,21 +32,26 @@ func SetDocumentByID(ctx context.Context, tx pgx.Tx, qp *QueryParams, id any, do
 		return 0, err
 	}
 
-	table, err := getMetadata(ctx, tx, qp.DB, qp.Collection)
+	table, err := newMetadata(tx, qp.DB, qp.Collection).getTableName(ctx)
```
rumyantseva, line 35: Codecov complains about a missing test, an issue for it is https://github.com/FerretDB/FerretDB/issues/2149

internal/handlers/pg/pgdb/update.go
```diff
@@ -32,21 +32,26 @@ func SetDocumentByID(ctx context.Context, tx pgx.Tx, qp *QueryParams, id any, do
...
 
-	if qp.Comment != "" {
-		qp.Comment = strings.ReplaceAll(qp.Comment, "/*", "/ *")
-		qp.Comment = strings.ReplaceAll(qp.Comment, "*/", "* /")
+	if comment != "" {
+		comment = strings.ReplaceAll(comment, "/*", "/ *")
+		comment = strings.ReplaceAll(comment, "*/", "* /")
 
-		sql += `/* ` + qp.Comment + ` */ `
+		sql += `/* ` + comment + ` */ `
```
rumyantseva, line 51: Another place where codecov complains - https://github.com/FerretDB/FerretDB/issues/2150

internal/handlers/pg/pgdb/database_metadata_indexes.go
rumyantseva: fixed

internal/handlers/pg/pgdb/query.go
```diff
@@ -60,7 +60,7 @@ func Explain(ctx context.Context, tx pgx.Tx, qp *QueryParams) (*types.Document,
 		return nil, lazyerrors.Error(ErrTableNotExist)
 	}
 
-	table, err := getMetadata(ctx, tx, qp.DB, qp.Collection)
+	table, err := newMetadata(tx, qp.DB, qp.Collection).getTableName(ctx)
```
chilagrow, line 63: At the moment for pushdown check, we check if [`Filter` exists](https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/pg/msg_explain.go#L88), but now we use index condition so it looks like those cases it uses `Index Cond` for explain plan. See the doc https://www.postgresql.org/docs/current/using-explain.html

debug output to show what is available in the plan:
image

So to check pushdown is done, it is when any of `Filter`, `Index Cond` or any other index is present in explain plan.

chilagrow (commented):
I like the way it is refactored with metadata!

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -22,25 +22,61 @@ import (
...
+	if err := createIndexIfNotExists(ctx, tx, ip.db, pgTable, pgIndex, true); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 // createIndexIfNotExists creates a new index for the given params if it does not exist.
-func createIndexIfNotExists(ctx context.Context, tx pgx.Tx, p *indexParams) error {
+func createIndexIfNotExists(ctx context.Context, tx pgx.Tx, schema, table, index string, isUnique bool) error {
```
noisersup, line 69: It seems that the list of arguments here is pretty long. I see that we modified the `indexParams` to describe the FerretDB layer logic, and not the PostgreSQL for the `createIndex` function. If I understand, we don't plan to use this function in lot of places and it will be rather something internal for `pgdb`, so I guess it's not a big problem.

However the naming here looks a bit misleading to me. There's `createIndexIfNotExists` function, which creates an index based on PostgreSQL related arguments.

Afterwards we have `createIndex`, which is a wrapper on top of `createIndexIfNotExists`, it will be the exported function and (If I understand) preferably used in our code.

Both functions create index if it doesn't exist, but they differ from each other only by "If not exists" part.

I would suggest to differentiate those function based on their layers, or - just keep the name the same, but make one exported (`createIndex` for pgdb layer and `CreateIndex`for FerretDB layer).

What do you think?

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -22,25 +22,61 @@ import (
...
+	if isUnique {
 		unique = " UNIQUE"
 	}
 
-	sql := `CREATE` + unique + ` INDEX IF NOT EXISTS ` + pgx.Identifier{p.table + "_id"}.Sanitize() +
-		` ON ` + pgx.Identifier{p.schema, p.table}.Sanitize() +
-		` ((_jsonb->>'_id'))`
+	sql := `CREATE` + unique + ` INDEX IF NOT EXISTS ` + pgx.Identifier{index}.Sanitize() +
+		` ON ` + pgx.Identifier{schema, table}.Sanitize() +
+		` ((_jsonb->'_id'))` // TODO Provide ability to set fields https://github.com/FerretDB/FerretDB/issues/1509
```
noisersup, line 79: Sorry for reviewing the same line second time but we could try to simplify it to one pair of brackets, or even no brackets

```suggestion
		` _jsonb->'_id'` // TODO Provide ability to set fields https://github.com/FerretDB/FerretDB/issues/1509
```

internal/handlers/pg/pgdb/query.go
```diff
@@ -224,6 +193,10 @@ func buildIterator(ctx context.Context, tx pgx.Tx, p *iteratorParams) (iterator.
 
 	query += where
 
+	if p.forUpdate {
+		query += ` FOR UPDATE`
+	}
```
noisersup, line 198: That will be useful!

noisersup (commented):
I really like the idea behind separating metadata to the struct and calling metadata functions in that way. It's much easier to understand the code.

I left some comments, but apart from them LGTM

internal/handlers/pg/pgdb/indexes.go
rumyantseva: @noisersup did you try to run it locally? I tried your suggestion but it failed as syntax error. So I think double brackets are needed.

internal/handlers/pg/pgdb/query.go
rumyantseva: Thanks a lot!

integration/create_test.go
```diff
@@ -405,3 +405,16 @@ func TestCreateTigris(t *testing.T) {
 		})
 	}
 }
+
+// TestCreateShortName tests how a collection with a short name works.
+// For example, for such collections hashed table names (Postgres table names) are the same as collection names.
+func TestCreateShortName(t *testing.T) {
```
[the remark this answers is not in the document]
rumyantseva: Nice catch! And nice reason to refactor `TestCollectionName` too! ;)

internal/handlers/pg/pgdb/indexes.go
rumyantseva: I renamed the unexported one to `createPGIndexIfNotExists` to clarify that it works with PG. The future-exported `CreateIndex` is needed to work with index as FerretDB abstraction (metadata + PG index).

I also renamed `createTableIfNotExists` to `createPGTableIfNotExists` to keep the naming consistent.

Taskfile.yml
```diff
@@ -5,7 +5,7 @@ env:
   GORACE: halt_on_error=1,history_size=2
 
 vars:
-  INTEGRATIONTIME: 12m
+  INTEGRATIONTIME: 15m
```
rumyantseva, line 8: Timeout increase is expected because index creation takes time

integration/basic_test.go
```diff
@@ -315,78 +315,76 @@ func TestCollectionName(t *testing.T) {
...
-			"WithADashAtBeginning": {
-				collection: "-collection_name",
+			alt: fmt.Sprintf("Invalid collection name: 'testcollectionname.%s'", collectionName300),
+		},
+		"LongEnough": {
+			collection: collectionName100,
+		},
+		"Short": {
+			collection: "a",
+		},
```
chilagrow, line 345: Amazing, thanks for the refactoring of this test!

chilagrow (approved):
Great work!

internal/handlers/pg/pgdb/indexes.go
AlekSi: What about single brackets?

rumyantseva: syntax error.

rumyantseva: ` ERROR: syntax error at or near "->" (SQLSTATE 42601)`

internal/handlers/pg/pgdb/indexes.go
noisersup: Looks good!

AlekSi (approved):

PR #2099, opened 2023-03-02
Add aggregation pipeline compat tests and fix errors

### LINKED ISSUE #1410: Implement `aggregate` command
### What should be done?

After #1891 and #1439, it should be possible to add integration / compat tests for aggregation pipelines. Let's do that.

Add compatibility test helper similar to `testQueryCompat` that accepts aggregation pipeline instead of filter, sort, and projection. It should add `$sort` stage if it is not present in the test case for stable results.

Explain/pushdown checks are not in the scope because they are not possible yet.

Changing existing aggregation pipelines to return correct errors is in the scope. Stage-specific bug fixes are in the scope unless those bugs affect `find` and other commands too.

### Definition of Done

- stages updated to return correct errors;
- integration/compatibility tests added;
- spot refactorings done.

chilagrow:
# Description

Closes #1410.

## Readiness checklist


integration/aggregate_compat_test.go
```diff
@@ -0,0 +1,324 @@
...
+		"ID": {
+			pipeline: bson.A{bson.D{{"$match", bson.D{{"_id", "string"}}}}},
+		},
+		"Int": {
+			pipeline: bson.A{
+				bson.D{{"$match", bson.D{{"v", 42}}}},
+				// sort by _id because mongoDB does not have deterministic natural order.
+				bson.D{{"$sort", bson.D{{"_id", 1}}}},
+			},
+		},
```
chilagrow, line 254: See 

```
Error:      	Not equal: 
        	            	expected: [
        	            	  {
        	            	    "$k": [
        	            	      "_id",
        	            	      "v"
        	            	    ],
        	            	    "_id": "array-numbers-asc",
        	            	    "v": [
        	            	      42,
        	            	      {
        	            	        "$l": "43"
        	            	      },
        	            	      {
        	            	        "$f": 45.5
        	            	      }
        	            	    ]
        	            	  },
        	            	  {
        	            	    "$k": [
        	            	      "_id",
        	            	      "v"
        	            	    ],
        	            	    "_id": "array-composite",
        	            	    "v": [
        	            	      {
        	            	        "$f": 42.13
        	            	      },
        	            	      "foo",
        	            	      {
        	            	        "$b": "KgAN",
        	            	        "s": 128
        	            	      },
        	            	      {
        	            	        "$o": "000102030405060708091011"
        	            	      },
        	            	      true,
        	            	      {
        	            	        "$d": 1635761922123
        	            	      },
        	            	      null,
        	            	      {
        	            	        "$r": "foo",
        	            	        "o": "i"
        	            	      },
        	            	      42,
        	            	      {
        	            	        "$t": "180388626445"
        	            	      },
        	            	      {
        	            	        "$l": "41"
        	            	      }
        	            	    ]
        	            	  },
        	            	  {
        	            	    "$k": [
        	            	      "_id",
        	            	      "v"
        	            	    ],
        	            	    "_id": "array-three",
        	            	    "v": [
        	            	      42,
        	            	      "foo",
        	            	      null
        	            	    ]
        	            	  },
        	            	  {
        	            	    "$k": [
        	            	      "_id",
        	            	      "v"
        	            	    ],
        	            	    "_id": "array",
        	            	    "v": [
        	            	      42
        	            	    ]
        	            	  },
        	            	  {
        	            	    "$k": [
        	            	      "_id",
        	            	      "v"
        	            	    ],
        	            	    "_id": "array-three-reverse",
        	            	    "v": [
        	            	      null,
        	            	      "foo",
        	            	      42
        	            	    ]
        	            	  }
        	            	]
        	            	actual  : [
        	            	  {
        	            	    "$k": [
        	            	      "_id",
        	            	      "v"
        	            	    ],
        	            	    "_id": "array",
        	            	    "v": [
        	            	      42
        	            	    ]
        	            	  },
        	            	  {
        	            	    "$k": [
        	            	      "_id",
        	            	      "v"
        	            	    ],
        	            	    "_id": "array-three-reverse",
        	            	    "v": [
        	            	      null,
        	            	      "foo",
        	            	      42
        	            	    ]
        	            	  },
        	            	  {
        	            	    "$k": [
        	            	      "_id",
        	            	      "v"
        	            	    ],
        	            	    "_id": "array-numbers-asc",
        	            	    "v": [
        	            	      42,
        	            	      {
        	            	        "$l": "43"
        	            	      },
        	            	      {
        	            	        "$f": 45.5
        	            	      }
        	            	    ]
        	            	  },
        	            	  {
        	            	    "$k": [
        	            	      "_id",
        	            	      "v"
        	            	    ],
        	            	    "_id": "array-composite",
        	            	    "v": [
        	            	      {
        	            	        "$f": 42.13
        	            	      },
        	            	      "foo",
        	            	      {
        	            	        "$b": "KgAN",
        	            	        "s": 128
        	            	      },
        	            	      {
        	            	        "$o": "000102030405060708091011"
        	            	      },
        	            	      true,
        	            	      {
        	            	        "$d": 1635761922123
        	            	      },
        	            	      null,
        	            	      {
        	            	        "$r": "foo",
        	            	        "o": "i"
        	            	      },
        	            	      42,
        	            	      {
        	            	        "$t": "180388626445"
        	            	      },
        	            	      {
        	            	        "$l": "41"
        	            	      }
        	            	    ]
        	            	  },
        	            	  {
        	            	    "$k": [
        	            	      "_id",
        	            	      "v"
        	            	    ],
        	            	    "_id": "array-three",
        	            	    "v": [
        	            	      42,
        	            	      "foo",
        	            	      null
        	            	    ]
        	            	  }
        	            	]
```

integration/aggregate_compat_test.go
```diff
@@ -0,0 +1,350 @@
...
+func TestAggregateCompatMatch(t *testing.T) {
+	testCases := map[string]aggregateStageCompatTestCase{
+		"ID": {
+			pipeline: bson.A{bson.D{{"$match", bson.D{{"_id", "string"}}}}},
+		},
+		"Int": {
+			pipeline: bson.A{
+				bson.D{{"$match", bson.D{{"v", 42}}}},
+				// sort by _id because compat does not have deterministic order.
+				bson.D{{"$sort", bson.D{{"_id", 1}}}},
```
AlekSi, line 253: We should do it in `testAggregateStageCompat` itself; see

https://github.com/FerretDB/FerretDB/blob/852ebf4002669578fd332c90d5aff8f9af08a66f/integration/query_compat_test.go#L67-L70

There we should not use a separate field, but we could check if `pipeline` includes `$sort`

chilagrow: Added check to find $sort in the pipeline, and adding $sort if that doesn't exist :+1:

integration/aggregate_compat_test.go
```diff
@@ -0,0 +1,369 @@
...
+
+// aggregateStageCompatTestCase describes aggregate compatibility test case.
+type aggregateStageCompatTestCase struct {
+	skip       string                   // skip test for all handlers, must have issue number mentioned
+	pipeline   bson.A                   // required, unspecified $sort appends bson.D{{"$sort", bson.D{{"_id", 1}}}}
+	resultType compatTestCaseResultType // defaults to nonEmptyResult
+}
+
+// testAggregateStageCompat tests aggregate stage compatibility test cases.
+func testAggregateStageCompat(t *testing.T, testCases map[string]aggregateStageCompatTestCase) {
```
noisersup, line 37: If we use this helper for in tests with single stages but also multiple stages, then I would suggest to make it plural
```suggestion
func testAggregateStagesCompat(t *testing.T, testCases map[string]aggregateStageCompatTestCase) {
```

integration/aggregate_compat_test.go
```diff
@@ -0,0 +1,369 @@
...
+			t.Helper()
+
+			if tc.skip != "" {
+				t.Skip(tc.skip)
+			}
+
+			pipeline := tc.pipeline
+			require.NotNil(t, pipeline, "pipeline should be set")
+
+			t.Parallel()
```
noisersup, line 54:
```suggestion
                        t.Parallel()
                        
			pipeline := tc.pipeline
			require.NotNil(t, pipeline, "pipeline should be set")			
```

integration/aggregate_compat_test.go
```diff
@@ -0,0 +1,369 @@
...
+		"AscendingID": {
+			pipeline: bson.A{bson.D{{"$sort", bson.D{{"_id", 1}}}}},
+		},
+		"DescendingID": {
+			pipeline: bson.A{bson.D{{"$sort", bson.D{{"_id", -1}}}}},
+		},
+		"AscendingValue": {
+			pipeline: bson.A{bson.D{{"$sort", bson.D{
+				{"v", 1},
+				{"_id", 1}, // sort by _id when v is the same.
```
noisersup, line 332: We could remove this, as it's the default test value, but I assume it's better to show it inside the test case as we test for `$sort` here particularly. :+1:

noisersup (commented):
Some small changes suggested from me. Apart from that it looks good!

integration/aggregate_compat_test.go
rumyantseva, line 332: What if we don't set it and only have one case with `v`? I think as a user, if I want to sort by `v`, I don't really care about setting an `_id` as well. Should this comment be a TODO for sorting implementation?

rumyantseva (approved):
The tests looks good to me (left one minor comment), so I'm ready to approve. It looks to me like we need a separate issue to implement proper sorting.

integration/aggregate_compat_test.go
chilagrow: So if `_id` is not set, we let compat (and target) sort by any order when `v` are the same. For example, documents containing the same value `v` might be different order each execution. See the doc 
https://www.mongodb.com/docs/manual/reference/operator/aggregation/sort/#sort-consistency

We want to avoid inconsistency particularly between compat and target hence we sort by _id for those with same `v`. Does this make sense?

As a user, if you want to have consistently same order, it's recommended to use `_id` sort at the end too :hugs:

integration/aggregate_compat_test.go
```diff
@@ -0,0 +1,369 @@
...
+
+			pipeline := tc.pipeline
+			require.NotNil(t, pipeline, "pipeline should be set")
+
+			t.Parallel()
+
+			var hasSortStage bool
+			for _, stage := range pipeline {
+				stage, ok := stage.(bson.D)
+				require.True(t, ok, "stage needs to be bson.D type")
```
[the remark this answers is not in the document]
chilagrow: Actually I did tried that as well and driver handles it nicely with an error. So I'm removing this check!

integration/aggregate_compat_test.go
chilagrow: So this one isn't default value, but `AscendingID` test it the default one. Let's keep it there because we would want to have a test case for it explicitly!

integration/aggregate_compat_test.go
chilagrow: Yea sure

integration/aggregate_compat_test.go
chilagrow: Thanks, let's do that.

noisersup (approved):

integration/aggregate_compat_test.go
rumyantseva: Yeah, makes total sense! This looks like something we should mention in our documentation (cc @Fashander).

rumyantseva (approved):
:shipit:

PR #2116, opened 2023-03-03
Add blog post - How MongoDB compares array

Fashander:
# Description

Closes #{issue_number}.

## Readiness checklist


website/blog/2023-02-21-ferretdb-compare-arrays.md
```diff
@@ -0,0 +1,380 @@
...
+  { _id: 'scalar-string', v: 'a' },
+  { _id: 'array-mixed', v: [ 1, 'z', 'b' ] }
+]
+```
+
+The documents selected are `scalar-string` and `array-mixed`.
+
+The `scalar-string` document has a smaller value compared to the filter argument.
+
+The array record `array-mixed` was selected because the smallest element `b` is
```
chilagrow, line 145: I think we need to bring the comparison between array and non array if want to use this example, we haven't explained how `array-mixed` got selected.

https://github.com/FerretDB/FerretDB/blob/bc761c53bb0ba74f2ce9e10870c6a223c6382c58/website/blog/2023-03-01.ferretdb-compare-arrays.md?plain=1#L123

Another option is remove everything array and nested array from the example.

Fashander: Thanks for pointing this out. I shouldn't have removed it. I'll bring it back now

website/blog/2023-02-21-ferretdb-compare-arrays.md
```diff
@@ -0,0 +1,380 @@
...
+  { _id: 'nested-empty', v: [ [] ] },
+  { _id: 'nested-mixed', v: [ [ 'c' ], [ 1, 'z', 'b' ] ] }
+]
+```
+
+No scalar values were selected, because array argument only selects arrays.
+The filter argument selected arrays that are less
+and nested arrays which contain an array that is less.
+
+So _contains_ condition is applied on documents containing less than the filter argument, but not the other way around.
```
chilagrow, line 177: This condition needs to be explained in the earlier section with example I think.

website/blog/2023-02-21-ferretdb-compare-arrays.md
```diff
@@ -0,0 +1,380 @@
...
+In the query, `array-bool` and `array-string` were selected since they contain values that is
+greater than the filter argument of `c` at its first index.
+
+The arrays are compared index to index using BSON comparison order then by value comparison.
+
+When dealing with nested arrays, using the `$gt` operator on an array filter argument does not select the largest element from a nested array, unlike how `$lt` operates.
+
+Instead, the `$gt` operator compares the array filter argument and the nested array in the collection index to index, using the BSON comparison order.
+
+If the BSON type of the array in the collection is greater than the BSON type of the filter argument (which is a string), then all nested arrays are selected.
```
chilagrow, line 316:
```suggestion
Because BSON type of the array in the collection is greater than the BSON type of the filter argument (which is a string), all nested arrays are selected.
```

This was my reasoning due to my finding

website/blog/2023-02-21-ferretdb-compare-arrays.md
```diff
@@ -0,0 +1,380 @@
...
+  { _id: 'array-bool', v: [ false ] },
+  { _id: 'nested-mixed', v: [ [ 'c' ], [ 1, 'z', 'b' ] ] },
+  { _id: 'nested-bool', v: [ [ true ] ] }
+]
+```
+
+No scalar values from the collection were selected since the scalar values are not array type.
+
+The selected array `array-bool` contains a boolean value, which has a greater BSON comparison order than that of an array type.
+Therefore, when comparing `['c']` with the items in the array, the largest item from the array (`false`) was used for the comparison.
```
chilagrow, line 347:
```suggestion
```

Actually this was due to index to index comparison.

website/blog/2023-02-21-ferretdb-compare-arrays.md
```diff
@@ -0,0 +1,380 @@
...
+This is due to the complexities caused by
+differences in the comparison of array and other BSON types.
+
+In particular, differentiating between the cases where the BSON comparison order of an array is used versus an element from the comparison array was not entirely straightforward.
+
+Since then, we’ve taken time to understand the inner workings of comparison for various operations.
+
+After investigating the behaviour of the `$lt` and `$gt` operators with array and nested array filter arguments, we discovered that the comparison process is different between the two cases.
+
+However, given an array argument filter and nested array collection, `$lt` uses the smallest element from the nested array and `$gt` on the other hand compares index to index of nested array collection.
```
chilagrow, line 374:
```suggestion
However, given an array argument filter and array collection, `$lt` uses the smallest element from the array and `$gt` on the other hand compares index to index of array collection.
```
It also works like that for nested array as well. But important to mention that array along is difficult

website/blog/2023-02-21-ferretdb-compare-arrays.md
Fashander: I can't seem to find where this was mentioned

chilagrow: It's this one https://github.com/FerretDB/FerretDB/blob/bc761c53bb0ba74f2ce9e10870c6a223c6382c58/website/blog/2023-03-01.ferretdb-compare-arrays.md?plain=1#L123

chilagrow: Maybe we need $eq to explain it. https://github.com/FerretDB/FerretDB/blob/bc761c53bb0ba74f2ce9e10870c6a223c6382c58/website/blog/2023-03-01.ferretdb-compare-arrays.md?plain=1#L240

Fashander: I'm hoping we don't expand it too much as-is. So, can we simply explain it in a sentence or two what that contain condition does without necessarily providing an example?

Fashander:
@AlekSi PTAL

AlekSi:
I will come up with a structure for a series of blog posts about sorting and filtering this week. On pause for now.

rumyantseva (commented):
Should we decide what to do with this topic? It feels like it might be the right time to cover this topic now!

w84thesun (commented):
Please re-request review when the post would be ready

noisersup (commented):
Leaving comment for now as it's in a draft stage

PR #2156, opened 2023-03-07
Support dot notation in sorting

### LINKED ISSUE #2101: Support dot notation in sorting
_See the comments for implementation details_

### Versions

0.9.2

### What did you do?

https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/common/sort_documents.go#L57

Sorting currently does not handle dot notation.

Example: this won't work
```
[{"$sort": {"v.foo": 1}]
```

DOD:
* add test case `v.0` for each command
* add test case `v.foo` for each command

Dot notation is used for sorting

### What did you see instead?

See the link above

### Environment

```markdown
- OS:
- Architecture:
- Version:
- Go:
- Git:
- Deployment:
- Deployment details:
```

rumyantseva:
# Description

Closes #2101.

## Readiness checklist


integration/aggregate_compat_test.go
```diff
@@ -345,7 +345,12 @@ func TestAggregateCompatSort(t *testing.T) {
 				{"v.foo", 1},
 				{"_id", 1}, // sort by _id when v is the same.
 			}}}},
-			skip: "https://github.com/FerretDB/FerretDB/issues/2101",
```
rumyantseva, line 348: I added a few more tests in addition to this one. So, now it covers a bit more cases related to dot notation. If you see some other examples that need to be covered, don't hesitate to propose them.

Also, I see Tigris started failing with timeout. I didn't add too many cases, so I'm not sure why. Maybe someone has an idea how to handle it (or should I add one more Tigris instance maybe?)

internal/handlers/tigris/msg_aggregate.go
```diff
@@ -103,7 +104,7 @@ func (h *Handler) MsgAggregate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
 
 		var s aggregations.Stage
 		if s, err = aggregations.NewStage(d); err != nil {
-			return nil, err
+			return nil, lazyerrors.Error(err)
```
rumyantseva, line 107: I added more `lazyerror` errors to simplify debugging (to see when things fail when they fail). If you think some of these errors shouldn't be "lazy", please correct me.

integration/aggregate_compat_test.go
```diff
@@ -365,6 +370,12 @@ func TestAggregateCompatSort(t *testing.T) {
 			pipeline:   bson.A{bson.D{{"$sort", bson.D{}}}},
 			resultType: emptyResult,
 		},
+		"Location15998": { // aka DotNotationMissingField
```
rumyantseva, line 373: In this test, test cases are called after error codes, so I keep the same naming. However, in the other places, such test was called `DotNotationMissingField`, so for this one I call it `Location15998`, but in the other places I call it `DotNotationMissingField` - as before. And I added a comment to be able to find this test case when search by "DotNotationMissingField".

w84thesun: Is there a way to replace those test case names with meaningful names? 
Location15976 – no keys provided
Location15975 – invalid sort order value provided
Location15973 – invalid sort order value type

w84thesun (commented):
Looks good overall, I added the comment about test case names.

integration/aggregate_compat_test.go
rumyantseva: @chilagrow what do you think? To me, the naming `LocationXYZ` makes sense as it's easy to understand which cases are covered by it, but on the other hand, this is not how such tests are named in the other parts of code. As you are the owner of this naming strategy, I need your permission to rename it to what Dmitry suggests :)

chilagrow: So I started naming them `Location..` because I wanted to have short and easy to map test names to the error it was expecting to receive. But it's not so easy to tell apart. Of the top of my head, meaningful is something like this? Anyway agree with @w84thesun on having something more meaningful :hugs: 

Location15976 – no keys provided produces `ErrSortMissingKey` -> `SortMissingKey`
Location15975 – invalid sort order value provided produces `ErrSortBadOrder` -> `SortBadOrder`
Location15973 – invalid sort order value type produces `ErrSortBadExpression` -> `SortBadExpression`

Also if the test case name is too long, tests fail because test name is used for the collection name :see_no_evil:.

internal/handlers/common/aggregations/sort.go
```diff
@@ -54,6 +54,8 @@ func newSort(stage *types.Document) (Stage, error) {
 }
 
 // Process implements Stage interface.
+//
+// If sort path is invalid, it returns a possibly wrapped types.DocumentPathError.
```
chilagrow, line 58: I think this is where we want to return `ErrPathContainsEmptyElement` error instead of `types.DocumentPathError`. This error is specific to `sort` stage as far as I can tell, and if we return `ErrPathContainsEmptyElement` here, we don't need to add the logic in both tigris and pg. What do you think? :hugs:

internal/handlers/tigris/msg_aggregate.go
chilagrow: I'm not particularly accustomed to using much lazyerror but it's probably okay :hugs:

chilagrow (commented):
Great work just have a couple of comments!

internal/handlers/common/aggregations/sort.go
rumyantseva: I like the idea of returning `ErrPathContainsEmptyElement`, but I don't understand how it will help us not add the logic in Tigris and PG handlers as we need to return `CommandError` there. But it seems that I might miss something.

integration/aggregate_compat_test.go
rumyantseva: @chilagrow hmm, do you have a link where it shows which errors it returns on CI? It passes locally, so if it fails on CI, it's something interesting. I see that CI returns a timeout from time to time for Postgres, and I wonder why, as I didn't add too many tests.

chilagrow: See this one https://github.com/FerretDB/FerretDB/actions/runs/4372348012/jobs/7649165466
It failed on 3 tests and one of which is this test.

```
2023-03-09T08:21:13.1577454Z     --- FAIL: TestQueryCompatRunner/ImplicitEqBoolTrue (0.37s)
2023-03-09T08:21:13.1577916Z         --- FAIL: TestQueryCompatRunner/ImplicitEqBoolTrue/TestQueryCompatRunner_Bools (0.02s)
2023-03-09T08:21:13.1578383Z             query_compat_test.go:193: Compat (expected) IDs: [bool-true]
2023-03-09T08:21:13.1578709Z             query_compat_test.go:193: Target (actual)   IDs: []
2023-03-09T08:21:13.1578942Z             query_compat_test.go:193: 
2023-03-09T08:21:13.1579512Z                    Error Trace:    /home/runner/work/FerretDB/FerretDB/internal/util/testutil/equal.go:66
2023-03-09T08:21:13.1580310Z                                                            /home/runner/work/FerretDB/FerretDB/integration/helpers.go:152
2023-03-09T08:21:13.1581205Z                                                            /home/runner/work/FerretDB/FerretDB/integration/query_compat_test.go:126
2023-03-09T08:21:13.1581440Z                    Error:          Not equal: 
2023-03-09T08:21:13.1581720Z                                    expected: [
2023-03-09T08:21:13.1581942Z                                      {
2023-03-09T08:21:13.1582224Z                                        "$k": [
2023-03-09T08:21:13.1582519Z                                          "_id",
2023-03-09T08:21:13.1582797Z                                          "v"
2023-03-09T08:21:13.1583040Z                                        ],
2023-03-09T08:21:13.1583468Z                                        "_id": "bool-true",
2023-03-09T08:21:13.1583769Z                                        "v": true
2023-03-09T08:21:13.1583988Z                                      }
2023-03-09T08:21:13.1584190Z                                    ]
2023-03-09T08:21:13.1584477Z                                    actual  : []
2023-03-09T08:21:13.1584797Z                                    --- expected
2023-03-09T08:21:13.1585065Z                                    +++ actual
2023-03-09T08:21:13.1585408Z                                    @@ -1,10 +1 @@
2023-03-09T08:21:13.1585641Z                                    -[
2023-03-09T08:21:13.1585901Z                                    -  {
2023-03-09T08:21:13.1586221Z                                    -    "$k": [
2023-03-09T08:21:13.1586557Z                                    -      "_id",
2023-03-09T08:21:13.1586864Z                                    -      "v"
2023-03-09T08:21:13.1587132Z                                    -    ],
2023-03-09T08:21:13.1587558Z                                    -    "_id": "bool-true",
2023-03-09T08:21:13.1587901Z                                    -    "v": true
2023-03-09T08:21:13.1588163Z                                    -  }
2023-03-09T08:21:13.1588389Z                                    -]
2023-03-09T08:21:13.1588605Z                                    +[]

```

integration/query_comparison_compat_test.go
```diff
@@ -452,10 +452,8 @@ func testQueryComparisonCompatGt() map[string]queryCompatTestCase {
 						{"array", bson.A{int32(42), "foo", nil}}, {"42", "foo"}, {"foo", int32(42)},
 					}},
 				}},
-				{"_id", bson.D{{"$ne", "array-documents-nested"}}}, // satisfies the $gt condition
 			},
-			resultType:     emptyResult,
-			resultPushdown: true,
+			resultPushdown: false,
```
rumyantseva, line 456: @noisersup I need your opinion regarding this one. I added one more value to shareddata composites - `array-documents-nested-duplicate`. I needed one more value to test that sort works correctly for dot notation (as I need more than one document). After that, this test started failing. I'm not sure why exactly `array-documents-nested` was excluded (I guess, to only work with "simpler" documents, not this one). I included it back, and after that, the test started working without pushdown (so I set `resultPushdown` as `false`). Alternatively, I can exclude both `array-documents-nested` and `array-documents-nested-duplicate` from this test. Still, such exclusion makes the test a bit more complicated, so I wonder if the current approach is good enough from the pushdown point of view. Is it correct that it returns `false`?

integration/shareddata/composites.go
```diff
@@ -301,5 +301,14 @@ var ArrayDocuments = &Values[string]{
 				},
 			}},
 		},
+		"array-documents-nested-duplicate": bson.A{ // duplicate value is needed to test sorting
```
rumyantseva, line 304: It looks to me, that this is the main reason why tests became slower - in every test we insert one more document in the test's collection.

internal/handlers/common/aggregations/sort.go
rumyantseva: Fixed

integration/query_comparison_compat_test.go
noisersup: It's because this PR removes line 455 with the `$ne` operator, so we don't perform any pushdown here, as we don't support `$gt` yet

noisersup: It's okay to remove it! Everything that covers `$eq` should be inside `Eq` tests, and in this test case we do pushdown "by accident"

integration/aggregate_compat_test.go
rumyantseva: Fixed in all `integration`, so the naming is consistent

w84thesun (dismissed):
👍

noisersup (approved):
Looks good to me! Test cases look reasonable!

chilagrow (approved):

integration/findandmodify_compat_test.go
```diff
@@ -174,6 +174,44 @@ func TestFindAndModifyCompatUpdate(t *testing.T) {
 	testFindAndModifyCompat(t, testCases)
 }
 
+// TestFindAndModifyCompatSort tests how various sort orders are handled.
+//
+// TODO Add more tests for sort: https://github.com/FerretDB/FerretDB/issues/2168
```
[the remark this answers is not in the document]
chilagrow: I'm assuming there was a reason why those were not relevant test for here.

internal/handlers/common/aggregations/sort.go
```diff
@@ -54,9 +56,20 @@ func newSort(stage *types.Document) (Stage, error) {
...
 func (m *sort) Process(ctx context.Context, in []*types.Document) ([]*types.Document, error) {
 	if err := common.SortDocuments(in, m.fields); err != nil {
-		return nil, err
+		var pathErr *types.DocumentPathError
+		if errors.As(err, &pathErr) && pathErr.Code() == types.ErrDocumentPathEmptyKey {
+			return nil, commonerrors.NewCommandErrorMsgWithArgument(
+				commonerrors.ErrPathContainsEmptyElement,
+				"FieldPath field names may not be empty strings.",
+				"$sort (stage)",
```
rumyantseva, line 68: Like in https://github.com/FerretDB/FerretDB/pull/2198/files

chilagrow (approved):

w84thesun (approved):
:shipit:

AlekSi (approved):

PR #2163, opened 2023-03-08
Fix pushdown for large double values

### LINKED ISSUE #2057: Fix pushdowns for large double values
### What should be done?

### Where?

## Definition Of Done
- [ ] Fix the bug

noisersup:
# Description

Closes #2057.

## Readiness checklist

### DIFF
--- a/integration/query_comparison_compat_test.go
+++ b/integration/query_comparison_compat_test.go
@@ -89,12 +89,10 @@ func testQueryComparisonCompatImplicit() map[string]queryCompatTestCase {
 		"DoubleBigInt64": {
 			filter:         bson.D{{"v", float64(2 << 61)}},
 			resultPushdown: true,
-			skip:           "https://github.com/FerretDB/FerretDB/issues/2057",
 		},
 		"DoubleBigInt64PlusOne": {
 			filter:         bson.D{{"v", float64(2<<61 + 1)}},
 			resultPushdown: true,
-			skip:           "https://github.com/FerretDB/FerretDB/issues/2057",
 		},
 		"Int64Max": {
 			filter:         bson.D{{"v", int64(math.MaxInt64)}},
@@ -103,12 +101,10 @@ func testQueryComparisonCompatImplicit() map[string]queryCompatTestCase {
 		"Int64Min": {
 			filter:         bson.D{{"v", int64(math.MinInt64)}},
 			resultPushdown: true,
-			skip:           "https://github.com/FerretDB/FerretDB/issues/2057",
 		},
 		"Int64DoubleBig": {
 			filter:         bson.D{{"v", int64(2 << 60)}},
 			resultPushdown: true,
-			skip:           "https://github.com/FerretDB/FerretDB/issues/2057",
 		},
 
 		"DoubleBig": {
@@ -291,12 +287,10 @@ func testQueryComparisonCompatEq() map[string]queryCompatTestCase {
 		"DoubleBigInt64": {
 			filter:         bson.D{{"v", bson.D{{"$eq", float64(2 << 61)}}}},
 			resultPushdown: true,
-			skip:           "https://github.com/FerretDB/FerretDB/issues/2057",
 		},
 		"DoubleBigInt64PlusOne": {
 			filter:         bson.D{{"v", bson.D{{"$eq", float64(2<<61 + 1)}}}},
 			resultPushdown: true,
-			skip:           "https://github.com/FerretDB/FerretDB/issues/2057",
 		},
 		"String": {
 			filter:         bson.D{{"v", bson.D{{"$eq", "foo"}}}},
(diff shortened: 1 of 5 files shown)


internal/handlers/pg/pgdb/query_test.go
```diff
@@ -399,14 +406,21 @@ func TestPrepareWhereClause(t *testing.T) {
 			)),
 			args:     []any{`v`, fmt.Sprint(float64(2 << 61))},
```
AlekSi, line 407:
```suggestion
			args:     []any{`v`, float64(2 << 61)},
```

Float values are no longer stored as strings

noisersup: This test checks for arguments passed to the PostgreSQL, which are passed as strings. The `42` value is passed as `"42"` and the `"foo"` value is passed as `"\"foo\""`

AlekSi: … but they should not be passed as strings

AlekSi: We should not call `MarshalSingleValue` in `prepareWhereClause`. Instead, we should provide a value as it is stored in PostgreSQL: underlying numerical type, string, etc.

noisersup: It's done by `pjson.MarshalSingleValue` function. From last investigations I highly believe that the function might be the reason of the bug of this PR so I will cover this

rumyantseva (approved):
I'm glad to see that we can pushdown more and more thing!
The query looks scarier and scarier, but I think it's good for now (we need this query in any case).

integration/shareddata/scalars.go
```diff
@@ -94,14 +95,15 @@ var Scalars = &Values[string]{
...
-		"int64-big":  int64Big,
-		"int64-1":    int64(1099511628000),
-		"int64-2":    int64(281474976700000),
-		"int64-3":    int64(72057594040000000),
+		"int64":            int64(42),
+		"int64-zero":       int64(0),
+		"int64-max":        int64(math.MaxInt64),
+		"int64-min":        int64(math.MinInt64),
+		"int64-big":        longBig,
+		"int64-double-big": int64(1 << 61), // duplicate?
```
chilagrow, line 103: It looks like a duplicate. It's fine for now.

integration/shareddata/scalars.go
```diff
@@ -339,11 +361,30 @@ var Int64s = &Values[string]{
...
+		"int64-prec-max":       int64(doubleMaxPrec),
+		"int64-prec-max-plus":  int64(doubleMaxPrec) + 1,
+		"int64-prec-max-minus": int64(doubleMaxPrec) - 1,
+
+		// negative long big values ~ -(1<<61)
+		"int64-neg-big":       -longBig,
+		"int64-neg-big-plus":  -(longBig + 1),
+		"int64-neg-big-minus": -(longBig - 1),
+
+		// long representation of double max precision ~ (1<<53)
```
chilagrow, line 384:
```suggestion
		// long representation of double max precision ~ -(1<<53-1)
```
I'm pretty sure it's just a typo

internal/handlers/tigris/tigrisdb/query.go
```diff
@@ -193,3 +201,41 @@ func BuildFilter(filter *types.Document) (string, error) {
...
+		types.NullType, types.Regex, types.Timestamp:
+		// type not supported for pushdown
+
+	case float64:
+		switch {
+		case docVal > types.MaxSafeDouble:
+			filter = map[string]any{"$gt": types.MaxSafeDouble}
+		case docVal < -types.MaxSafeDouble:
+			// skip pushdown as`$lt` operator works incorrectly
+			// for negative numbers
```
chilagrow, line 219: Informative :+1:

internal/handlers/tigris/tigrisdb/query.go
```diff
@@ -193,3 +201,41 @@ func BuildFilter(filter *types.Document) (string, error) {
 
 	return string(result), nil
 }
+
+// filterEqual returns the tigris filter with arguments that filters documents
+// where the value is equal to v.
+func filterEqual(v any) (filter any, err error) {
+	switch docVal := v.(type) {
+	case *types.Document, *types.Array, types.Binary,
+		types.NullType, types.Regex, types.Timestamp:
```
chilagrow, line 210: I think this line is just impossible to reach, because we only call `filterEqual` on types `float64, string, types.ObjectID, bool, time.Time, int32, int64`

integration/query_comparison_compat_test.go
```diff
@@ -558,7 +674,22 @@ func TestQueryComparisonCompatGt(t *testing.T) {
...
+			filter: bson.D{{"v", bson.D{{"$gt", int64(1<<61) - 1}}}},
+		},
+		"Int64NegBig": {
+			filter: bson.D{{"v", bson.D{{"$gt", -int64(1 << 61)}}}},
+		},
+		"Int64NegBigPlusOne": {
+			filter: bson.D{{"v", bson.D{{"$gt", -int64(1<<61) + 1}}}},
+		},
+		"Int64NegBigMinusOne": {
+			filter: bson.D{{"v", bson.D{{"$gt", -int64(1<<61) - 1}}}},
```
chilagrow, line 692: Nice addition of tests, even we don't yet pushdown $gt. :hugs:

chilagrow (approved):
Just really minor stuff. Amazing work on pushdown!

rumyantseva (approved):

chilagrow (approved):

PR #2187, opened 2023-03-10
Add support for `$each` modifier to `$addToSet` array update operator

### LINKED ISSUE #828: Support `$each` array update operator's modifier
Depends on #503 and #825

https://www.mongodb.com/docs/manual/reference/operator/update/each/

DoD:
- [ ] `$addToSet`
- [ ] `$push`
- [ ] test case for non existing field

w84thesun:
# Description

Closes #828.

## Readiness checklist


integration/update_array_compat_test.go
```diff
@@ -238,3 +238,60 @@ func TestUpdateArrayCompatPullAll(t *testing.T) {
...
+
+func TestUpdateArrayCompatAddToSetEach(t *testing.T) {
+	t.Parallel()
+
+	testCases := map[string]updateCompatTestCase{
+		"Document": {
+			update: bson.D{{"$addToSet", bson.D{{"v", bson.D{
+				{"$each", bson.A{bson.D{{"field", int32(42)}}}},
+			}}}}},
+			skipForTigris: "Mixed types are not supported for Tigris.",
```
rumyantseva, line 250: So we skip all the tests for Tigris? Don't data sets like `ArrayInt32s` fit with a bit different query -- smth like
`{"$each", bson.A{int32(42), jnt32(43)}}`? I like the idea of the test with `field`, but I think we can have at least one case that would fit Tigris too.

rumyantseva (approved):
I asked a few questions about test cases and implementation.

rumyantseva (changes requested):
Ok, I press "Request changes" because if I press "Comment", GitHub approves the PR. Looks like a bug on GitHub side.

integration/update_array_compat_test.go
w84thesun: If we want to test only `Int32` we should filter by `array-int32s` then, because there are arrays with doubles which would produce different result from MongoDB (Tigris would convert int32 to float). If we ok with that, then I can apply some filters.

integration/update_array_compat_test.go
```diff
@@ -238,3 +238,60 @@ func TestUpdateArrayCompatPullAll(t *testing.T) {
...
+		},
+		"String": {
+			update:        bson.D{{"$addToSet", bson.D{{"v", bson.D{{"$each", bson.A{"foo"}}}}}}},
+			skipForTigris: "Mixed types are not supported for Tigris.",
+		},
+		"Int32": {
+			update:        bson.D{{"$addToSet", bson.D{{"v", bson.D{{"$each", bson.A{int32(1), int32(2)}}}}}}},
```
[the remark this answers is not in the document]
w84thesun: Sure, added.

internal/handlers/common/update_array_operators.go
```diff
@@ -251,69 +272,82 @@ func processAddToSetArrayUpdateExpression(doc, update *types.Document) (bool, er
...
-		switch addToSetValueRaw := addToSetValueRaw.(type) {
-		case *types.Document, float64, string, types.Binary, types.ObjectID, bool,
-			time.Time, types.NullType, types.Regex, int32, types.Timestamp, int64:
-			shouldAdd := true
+	return changed, nil
+}
 
-			for i := 0; i < array.Len(); i++ {
-				var value any
+// valueInSet checks if the value is already in the array.
```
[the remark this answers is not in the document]
w84thesun: No difference actually, I missed that function. Thanks!

noisersup (dismissed):
Looks good to me!

integration/update_array_compat_test.go
```diff
@@ -238,3 +238,60 @@ func TestUpdateArrayCompatPullAll(t *testing.T) {
...
+	testCases := map[string]updateCompatTestCase{
+		"Document": {
+			update: bson.D{{"$addToSet", bson.D{{"v", bson.D{
+				{"$each", bson.A{bson.D{{"field", int32(42)}}}},
+			}}}}},
+			skipForTigris: "Mixed types are not supported for Tigris.",
+		},
+		"String": {
+			update:        bson.D{{"$addToSet", bson.D{{"v", bson.D{{"$each", bson.A{"foo"}}}}}}},
+			skipForTigris: "Mixed types are not supported for Tigris.",
```
chilagrow, line 254: This should work for tigris, so no need to skip right? We can have at least one test that covers tigris.

chilagrow (dismissed):
Great work, just think we can not skip String test for tigris

integration/update_array_compat_test.go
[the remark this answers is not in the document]
rumyantseva: Hmm, I don't see it's added. I mean something like `{"$each", bson.A{int32(1), int32(42), int32(2)}}`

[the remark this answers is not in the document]
w84thesun: I remember adding it but for some reason, it wasn't added. I fixed it now.

integration/update_array_compat_test.go
w84thesun: Yes, the strings provider is working. Unskipped this test case.

chilagrow (dismissed):

integration/update_array_compat_test.go
```diff
@@ -238,3 +238,61 @@ func TestUpdateArrayCompatPullAll(t *testing.T) {
...
+			resultType: emptyResult,
+		},
+		"EmptyArray": {
+			filter:     bson.D{{"_id", "array-documents-nested"}},
+			update:     bson.D{{"$addToSet", bson.D{{"v", bson.D{{"$each", bson.A{}}}}}}},
+			resultType: emptyResult,
+		},
+		"ArrayValuesExists": {
+			update:        bson.D{{"$addToSet", bson.D{{"v", bson.D{{"$each", bson.A{int32(42), int32(43)}}}}}}},
+			skipForTigris: "Tigris schema validation would fail.",
```
rumyantseva, line 272: `skipForTigris` reason is not correct, the test case is quite similar to `Int32` test case, what value does it bring?

w84thesun: Fixed.

integration/update_array_compat_test.go
```diff
@@ -238,3 +238,61 @@ func TestUpdateArrayCompatPullAll(t *testing.T) {
...
+
+func TestUpdateArrayCompatAddToSetEach(t *testing.T) {
+	t.Parallel()
+
+	testCases := map[string]updateCompatTestCase{
+		"Document": {
+			update: bson.D{{"$addToSet", bson.D{{"v", bson.D{
+				{"$each", bson.A{bson.D{{"field", int32(42)}}}},
+			}}}}},
+			skipForTigris: "Mixed types are not supported for Tigris.",
```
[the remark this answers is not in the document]
w84thesun: Changed the reason.

rumyantseva (approved):
LGTM

chilagrow (approved):

PR #2199, opened 2023-03-14
Support query pushdown for `$match` aggregation stage

### LINKED ISSUE #1894: Support query pushdown for `$match` aggregation stage
### What should be done?

If `$match` is the first stage, it should be pushed down.

How to do tests it without explain? Do we have explain?

Investigation:
- [x] [MsgExplain](https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/pg/msg_explain.go) needs to get `pipeline` optional parameter for aggregate command. If the `$match` is the first stage, set `QueryParams.Filter` to the $match query for checking pushdown. 
- [x] In aggregation pipeline, set filter when `$match` is the first stage, this pushes down the query.
- [x] Compat test need to be updated to call `explain aggregate` before calling aggregate.

DoD
- [x] refactoring explain is in scope
- [x] test explain unhappy path

chilagrow:
# Description

Closes #1894.

## Readiness checklist

### DIFF
--- a/integration/aggregate_compat_test.go
+++ b/integration/aggregate_compat_test.go
@@ -28,9 +28,10 @@ import (
 
 // aggregateStagesCompatTestCase describes aggregation stages compatibility test case.
 type aggregateStagesCompatTestCase struct {
-	skip       string                   // skip test for all handlers, must have issue number mentioned
-	pipeline   bson.A                   // required, unspecified $sort appends bson.D{{"$sort", bson.D{{"_id", 1}}}}
-	resultType compatTestCaseResultType // defaults to nonEmptyResult
+	skip           string                   // skip test for all handlers, must have issue number mentioned
+	pipeline       bson.A                   // required, unspecified $sort appends bson.D{{"$sort", bson.D{{"_id", 1}}}}
+	resultType     compatTestCaseResultType // defaults to nonEmptyResult
+	resultPushdown bool                     // defaults to false
 }
 
 // testAggregateStagesCompat tests aggregation stages compatibility test cases with all providers.
@@ -88,6 +89,21 @@ func testAggregateStagesCompatWithProviders(t *testing.T, providers shareddata.P
 				compatCollection := compatCollections[i]
 				t.Run(targetCollection.Name(), func(t *testing.T) {
 					t.Helper()
+
+					explainCommand := bson.D{{"explain", bson.D{
+						{"aggregate", targetCollection.Name()},
+						{"pipeline", pipeline},
+					}}}
+					var explainRes bson.D
+					require.NoError(t, targetCollection.Database().RunCommand(ctx, explainCommand).Decode(&explainRes))
+
+					var msg string
+					if setup.IsPushdownDisabled() {
+						tc.resultPushdown = false
+						msg = "Query pushdown is disabled, but target resulted with pushdown"
+					}
+
+					assert.Equal(t, tc.resultPushdown, explainRes.Map()["pushdown"], msg)
 
 					targetCursor, targetErr := targetCollection.Aggregate(ctx, pipeline)
 					compatCursor, compatErr := compatCollection.Aggregate(ctx, pipeline)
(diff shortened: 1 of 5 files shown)


w84thesun (dismissed):
Nice job! I like how `Explain` in `pgdb` looks now!

integration/aggregate_compat_test.go
```diff
@@ -241,16 +258,43 @@ func TestAggregateCompatStages(t *testing.T) {
...
+	providers := []shareddata.Provider{
+		// for testing empty pipeline use a collection with single document,
+		// because sorting will not matter.
+		shareddata.Unsets,
+	}
+
+	testCases := map[string]aggregateStagesCompatTestCase{
+		"Empty": {
+			pipeline: bson.A{},
+			noSort:   true,
```
rumyantseva, line 291: I'm confused by this condition. What do we test by setting `noSort` specifically? What would change if we didn't have `noSort` condition at all?

rumyantseva (commented):
The implementation looks reasonable to me.

I'm quite confused by having `noSort` param in test cases. I don't really understand why this param is needed. I see there is only one test case that uses it, and by reading the code I don't understand what would change if `noSort` didn't exist. Overall, it seems to me that `noSort` complicates the understanding of testing. E.g., as a contributor, in which new test cases I should set `noSort` as true?

integration/aggregate_compat_test.go
chilagrow: It's because of https://github.com/FerretDB/FerretDB/blob/main/integration/aggregate_compat_test.go#L32, $sort is always added to pipeline. But we do not want it to test this condition that pipeline is empty, 
https://github.com/FerretDB/FerretDB/pull/2199/files#diff-d5be5b46004f6aaf1d90e271efea546e6a16800f6ebef0f9b67049bc8f240bfdR26

integration/aggregate_compat_test.go
```diff
@@ -28,9 +28,14 @@ import (
...
-	pipeline   bson.A                   // required, unspecified $sort appends bson.D{{"$sort", bson.D{{"_id", 1}}}}
-	resultType compatTestCaseResultType // defaults to nonEmptyResult
+	skip           string                   // skip test for all handlers, must have issue number mentioned
+	pipeline       bson.A                   // required
+	resultType     compatTestCaseResultType // defaults to nonEmptyResult
+	resultPushdown bool                     // defaults to false
+	// when noSort is false, it appends bson.D{{"$sort", bson.D{{"_id", 1}}}}
+	// to pipeline, only if $sort stage does not exist.
+	// This ensures compat and target has the same order of results.
+	noSort bool
```
rumyantseva, line 38: Let's better say when `noSort` should be true, it's already false by default, but as a developer when do I need to set it true?

integration/aggregate_compat_test.go
```diff
@@ -77,8 +82,9 @@ func testAggregateStagesCompatWithProviders(t *testing.T, providers shareddata.P
 				}
 			}
 
-			if !hasSortStage {
-				// add sort stage to sort by _id because compat does not have deterministic order.
+			if !hasSortStage && !tc.noSort {
+				// add sort stage to sort by _id because compat and target
+				// would be ordered differently otherwise.
 				pipeline = append(pipeline, bson.D{{"$sort", bson.D{{"_id", 1}}}})
```
rumyantseva, line 88: To be honest, I'm not a fan of `noSort` variable in tests. Maybe we should just explicitly set sort in each test where it's needed and remove this line completely? Then we don't need `hasSortStage` nor `noSort`?

rumyantseva (dismissed):
I approve as I don't want to postpone the task, and I like the implementation. However, the idea of `noSort` wasn't really sold to me.

integration/aggregate_compat_test.go
chilagrow: Removed noSort in preference for just checking empty pipeline

rumyantseva (approved):
LGTM

w84thesun (approved):
LGTM

PR #2213, opened 2023-03-15
Large double values filters are handled incorrectly

noisersup:
# Description

I was working on some demo of the current problems of #2163 but while working on Javascript test I've 

## Readiness checklist

### DIFF
--- a/build/legacy-mongo-shell/test.js
+++ b/build/legacy-mongo-shell/test.js
@@ -3,5 +3,66 @@
 (function() {
   'use strict';
 
+  const coll = db.test;
+
+  coll.drop();
+
+  // As double's mantissa is 53bits long, we expect that every number that's <= 2<<53-1 can be represented precisely.
+  //
+  // Every Javscript Number type is stored as IEEE754 double floating point.
+  let init = [
+    {_id: 'double-max-prec', v: Number.MAX_SAFE_INTEGER}, // Double 2**53-1 is a last precise number = 9007199254740991.0
+    {_id: 'double-max-prec-long', v: NumberLong(Number.MAX_SAFE_INTEGER)}, // Long 2**53-1 is precise but not last = 9007199254740991
+  ];
+
+  coll.insertMany(init);
+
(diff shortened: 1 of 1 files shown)


build/legacy-mongo-shell/test.js
```diff
@@ -3,5 +3,66 @@
...
+  // 2305843009213694000.0 == 2305843009213694000.0
+  //
+  // Technically speaking, here following statement should be true:
+  // 2305843009213694000.0 == 2305843009213693953
+  //
+  // Although at this point test on mongodb fails. I don't know how is it possible that 
+  // 2305843009213693952 is rounded to 2305843009213694000.0 correctly, but
+  // 2305843009213693953 maybe not?
+  actual = coll.find({v: 2 * Math.pow(2,60)}).toArray();
+  assert.eq((new Array).concat(bigNumbers, bigNumbersDontMatch), actual);
```
noisersup, line 53: I'm not sure why it fails against MongoDB. It might be because of javascript double/long conversion issues, although why it works against Long("2305843009213693952") but not Long("2305843009213693953")? Dunno.

I've tested that also directly from `mongosh` by using `2305843009213693952.0` and `2305843009213693952.012312`, but both result in the same output as here, only `double-big` and `double-big-long` are matched.

What's even more strange is that:
```
test> db.test.find({v:2305843009213693953.0})
[
  { _id: 'double-big', v: 2305843009213694000 },
  { _id: 'double-big-long', v: Long("2305843009213693952") }
]
```

AlekSi: Could we use constants instead of math to make test simpler? That should eliminate one potential source of problems

b1ron: Chiming in with possibly unrelated information. I have less context on this issue.

I think it will convert numbers that exceed `Number.MAX_SAFE_INTEGER` to `NumberLong` in the shell before inserting the value. But apparently numbers that do exceed `Number.MAX_SAFE_INTEGER` will lead to unexpected behavior. So, `NumberLong` does not accept 1<<63 - 1 is not an int64.

AlekSi:
The test passed now. Great!

PR #2214, opened 2023-03-15
Make `listIndexes` use data from the database

### LINKED ISSUE #1509: Implement `createIndexes` command

rumyantseva:
# Description

Closes #1509.

This is a part of the main issue: we start getting data for `listIndexes` command from the metadata table. 
Please suggest alternative naming or structuring if you have any ideas.

## Readiness checklist


internal/handlers/pg/pgdb/indexes.go
```diff
@@ -16,49 +16,118 @@ package pgdb
...
+	metadata, err := newMetadata(tx, db, collection).get(ctx, false)
+	if err != nil {
+		return nil, err
+	}
+
+	if !metadata.Has("indexes") {
+		return []IndexParams{}, nil
+	}
+
+	indexes := must.NotFail(metadata.Get("indexes")).(*types.Array)
```
chilagrow, line 67: So here if indexes is not array, it panics. I think this is good since it can only happen if it was modified directly on the database.

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -16,49 +16,118 @@ package pgdb
...
+				var value any
+				field, value, err = keyIter.Next()
+
+				switch {
+				case err == nil:
+					res[i].Key = append(res[i].Key, IndexKeyPair{
+						Field: field,
+						Order: IndexOrder(value.(int32)),
+					})
+				case errors.Is(err, iterator.ErrIteratorDone):
```
chilagrow, line 101:
```suggestion
				case errors.Is(err, keyIter.ErrIteratorDone):
```

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -16,49 +16,118 @@ package pgdb
...
+			idx := idx.(*types.Document)
+			key := must.NotFail(idx.Get("key")).(*types.Document)
+
+			res[i] = IndexParams{
+				Index:  must.NotFail(idx.Get("name")).(string),
+				Unique: must.NotFail(idx.Get("unique")).(bool),
+				Key:    make([]IndexKeyPair, 0, key.Len()),
+			}
+
+			keyIter := key.Iterator()
```
chilagrow, line 88:
```suggestion
			keyIter := key.Iterator()
			defer keyIter.Close()
```
Just for the consistency.

chilagrow (commented):
Some comments!

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -16,49 +16,119 @@ package pgdb
...
+			defer keyIter.Close()
```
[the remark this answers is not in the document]
AlekSi: It is safe to call `keyIter.Close()` only there. Two calls below are not needed

internal/handlers/pg/pgdb/indexes.go
chilagrow: Indeed, I think I misread something totally.. Ignore my comment.

internal/handlers/pg/pgdb/indexes.go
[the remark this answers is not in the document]
rumyantseva: @AlekSi where "there"? :) Here?

internal/handlers/pg/pgdb/indexes.go
chilagrow: By the way, why is this line not covered by the test? codecov thinks so.

internal/handlers/pg/pgdb/indexes.go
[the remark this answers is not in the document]
AlekSi: … on line 89. I will avoid using "there" and "here" from now on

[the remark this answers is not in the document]
rumyantseva: Fixed & added a comment

internal/handlers/pg/pgdb/indexes.go
rumyantseva: Because we exit the loop before we come here:

```
for i := 0; i < key.Len(); i++
```

And I need `i` to address `res[i]`. But I don't like this duplication of iteration... Do you have an idea how to improve it?

rumyantseva: Maybe I can just ignore this error actually :D

rumyantseva: fixed

chilagrow (approved):

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -16,49 +16,114 @@ package pgdb
...
-// indexOrder constants.
+// IndexOrder constants.
 const (
-	indexOrderAsc  indexOrder = 1
-	indexOrderDesc indexOrder = -1
+	IndexOrderAsc  IndexOrder = 1
+	IndexOrderDesc IndexOrder = -1
 )
 
+// Indexes returns a list of indexes for the given database and collection.
```
w84thesun, line 54: There is some inconsistency between `Indexes` and `IndexParams`. `Indexes` says to return a list of indexes but it returns a list of `IndexParams`. In its turn, `IndexParams` is meant to be parameters for index creation, not the index itself.
To avoid this confusion we should probably rename `Indexes` to `ListIndexParams` or 
rename `IndexParams` to `Index` and change the comments accordingly.

w84thesun (commented):
Great job! Added comment about `IndexParams` struct and `Indexes` function.

internal/handlers/pg/pgdb/indexes.go
rumyantseva: @w84thesun good point! I think I like the second approach more - to rename `IndexParams` to `Index`, as this is quite a high-level entity.

noisersup (approved):
Checked locally and looks good to me!

I'm only a bit worried about those changes from "index" to "name". The queryPlanner returns data directly from PostgreSQL, and afaik there's no field called `Name Cond`, which may result in wrong results on `resultPushdown` for example.

**EDIT** sorry approved by accident :sweat_smile:

internal/handlers/pg/msg_explain.go
```diff
@@ -115,7 +115,7 @@ func (h *Handler) MsgExplain(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 
 	// if the plan returned filter info or index info, it means that pushdown had been done
 	pushdown := queryPlanner.HasByPath(types.NewStaticPath("Plan", "Filter")) ||
-		queryPlanner.HasByPath(types.NewStaticPath("Plan", "Index Cond"))
+		queryPlanner.HasByPath(types.NewStaticPath("Plan", "Name Cond"))
```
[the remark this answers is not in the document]
rumyantseva: Omg :D That's my bad refactoring! Nice catch!

internal/handlers/pg/pgdb/database_metadata.go
```diff
@@ -104,7 +104,7 @@ func (m *metadata) ensure(ctx context.Context) (tableName string, created bool,
 		return "", false, lazyerrors.Error(err)
 	}
 
-	// Index to ensure that collection name is unique
+	// Name to ensure that collection name is unique
```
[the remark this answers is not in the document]
rumyantseva: Everything is fixed and ready for another review round.

w84thesun (approved):
Thanks for the fixes! 👍

noisersup (approved):
It's handy to get this data from metdata!

LGTM!

PR #2222, opened 2023-03-16
Add support for `$pull` array update operator

### LINKED ISSUE #826: Support `$pull` array update operator
https://www.mongodb.com/docs/manual/reference/operator/update/pull/

Only basic support is in scope, support of operators inside `$pull` (e.g. `{ $pull: { v: { $gte: 42 } } }`) is in a separate issue - https://github.com/FerretDB/FerretDB/issues/1765.

DoD:
- [x] Nested arrays (arrays of arrays) are not in the scope. 
- [x] Arrays of objects are in the scope
- [ ] <del>update user doc (supported operators) </del>

Compatibility tests for both valid results and errors are in the scope:
- [x] pull from existent array field
- [x] pull from non-existent field
- [x] pull from a non-array field
- [ ] negative test cases 
- [x] dot notation
- [x] null values in an array
- [x] arrays with duplicate values (should all duplicates be removed by `$pull` or only the first one?)
- [ ] think what other test cases might be needed

w84thesun:
# Description

Closes #826.

I didn't come up with any negative test cases. They might be encountered further when adding support for conditions. If you have some particular test cases, please suggest them.

## Readiness checklist


rumyantseva:
Wow, wow! Such a nice PR number!

internal/handlers/common/update_array_operators.go
```diff
@@ -414,3 +415,66 @@ func processPullAllArrayUpdateExpression(doc, update *types.Document) (bool, err
 
 	return changed, nil
 }
+
+// processPullArrayUpdateExpression changes document according to $pull array update operator.
+// If the document was changed it returns true.
+func processPullArrayUpdateExpression(doc *types.Document, update *types.Document) (bool, error) {
+	var changed bool
+
+	iter := update.Iterator()
```
rumyantseva, line 424: I think we still need `defer iter.Close()` even though in this particular case it doesn't hurt.

internal/handlers/common/update_array_operators.go
```diff
@@ -414,3 +415,66 @@ func processPullAllArrayUpdateExpression(doc, update *types.Document) (bool, err
...
+
+			return false, lazyerrors.Error(err)
+		}
+
+		path, err := types.NewPathFromString(key)
+		if err != nil {
+			return false, lazyerrors.Error(err)
+		}
+
+		// If the path does not exist, create a new array and set it.
```
rumyantseva, line 441: This is the part I don't understand: if the path doesn't exist, we create an empty array and then try to pull from it? But then we will not pull anything, and `changed` will be `false`, right? Then does it make sense to create this empty array at all? Can't we just return `false, nil` right here?

rumyantseva (commented):
Looks good to me overall, the PR satisfied the definition of done of the task.

Added a couple of questions/comments.

internal/handlers/common/update_array_operators.go
chilagrow: Yeah creating an array here but without setting changed = `true` is confusing. Given codecov says it has tested it, compat must also creates and empty array without setting changed = `true`.

chilagrow (approved):

internal/handlers/common/update_array_operators.go
w84thesun: Fixed it here and in a couple of other places.

internal/handlers/common/update_array_operators.go
w84thesun: I've added a comment that possibly could help to understand why that `SetByPath` call is needed.

internal/handlers/common/update_array_operators.go
```diff
@@ -414,3 +417,69 @@ func processPullAllArrayUpdateExpression(doc, update *types.Document) (bool, err
...
+		// If the path does not exist, try to set it.
+		// If it is not possible to set the path return error.
+		// This will deal with cases like $pull on a non-existing field or unset field.
+		if !doc.HasByPath(path) {
+			if err = doc.SetByPath(path, types.MakeArray(1)); err != nil {
+				return false, commonerrors.NewWriteErrorMsg(
+					commonerrors.ErrUnsuitableValueType,
+					err.Error(),
+				)
+			}
```
rumyantseva, line 453: Still, can we return `false, nil` somewhere here? :) If `HasByPath` gave us `false`, and `SetByPath` didn't give us error, do we need to navigate through the rest of the code?

rumyantseva (approved):
LGTM, but I still think a tiny optimization is possible (see the comment).

internal/handlers/common/update_array_operators.go
chilagrow: I think it's possible to set `continue` but I think `false, nil` would skip the next value of the iterator and that's not what we want.

chilagrow (approved):

internal/handlers/common/update_array_operators.go
rumyantseva: Ah! Yes, you are right!

PR #2232, opened 2023-03-17
Fix long overflow detection on `$mul`

### LINKED ISSUE #2057: Fix pushdowns for large double values

noisersup:
# Description

Closes #2057.

The current check for overflows doesn't work correctly for most long values larger than 1<<53, and creates false-negatives, by returning overflow errors on values that didn't overflow. That's because of lost precision on float64 multiplication.

For example, the current state doesn't fail on `9007199254740992 * 42`, but fails for `9007199254740993 * 42`, as at this point, the first number is converted to float64 unprecisely, which results in different result.

## Readiness checklist


internal/handlers/common/params.go
```diff
@@ -391,6 +374,26 @@ func multiplyNumbers(v1, v2 any) (any, error) {
 	}
 }
 
+// multiplyLongSafely returns the multiplication of two int64 values.
+// It handles int64 overflows, and returns errLongExceeded error on one.
+//
+// Please prefer to use multiplyNumbers instead of this function.
+func multiplyLongSafely(v1, v2 int64) (int64, error) {
+	switch {
+	case v1 == 0 || v2 == 0 || v1 == 1 || v2 == 1:
```
rumyantseva, line 383: I see why `v2 == 0` is here - to avoid division by zero later, but why do the other values are here too? Are those cases also special?

rumyantseva (commented):
Started reading the code (I haven't gone through `TestMultiplyLongSafely` yet). Asked one question.

internal/handlers/common/params.go
noisersup: We check for 1 and 0 for every value, because it's possible to multiply those with MinInt64. Without this check, function returns overflow on operations like `MinInt64 * 1` or `0 * MinInt64`, etc.

noisersup: Should I leave a comment explaining that? It was confusing either to me, and I've written this code 2 hours ago...

noisersup: although now I'm thinking about `-1` values either. It's not ready for review yet sorry

internal/handlers/common/params_test.go
```diff
@@ -0,0 +1,132 @@
...
+	for name, tc := range map[string]struct {
+		values []int64 // [v1,v2,v1+v2] or [v1,v2] if you don't want to check multiplication result
+
+		// overflow checks if at least one value returned errLongExceeded error.
+		overflow bool // defaults to false
+
+		// valuesRange allows to execute test for the numbers next to v1.
+		// if v1=42 and valuesRange=2, following numbers will be tested: [42,43,44]
+		// It shouldn't be used with multiplication result, but rather to test overflows on large ranges.
+		valuesRange int
```
noisersup, line 37: This one is useful if we want to check for overflow on large values. We haven't realized that we have a bug here so far, because we've used one or two single big values which were able to be presented as float64 precisely. While all other numbers around those values were loosing their precision.

chilagrow (commented):
Thanks for fixing the bug! The tests are a bit complicated, I think it's worth simplifying

internal/handlers/common/params_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMultiplyLongSafely(t *testing.T) {
+	t.Parallel()
+
+	for name, tc := range map[string]struct {
+		values []int64 // [v1,v2,v1+v2] or [v1,v2] if you don't want to check multiplication result
```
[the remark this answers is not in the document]
noisersup: I didn't want to create so many lines of codes just to place similar testcases under separate names, but yes we can do it

internal/handlers/common/params.go
```diff
@@ -391,6 +374,32 @@ func multiplyNumbers(v1, v2 any) (any, error) {
 	}
 }
 
+// multiplyLongSafely returns the multiplication of two int64 values.
+// It handles int64 overflows, and returns errLongExceeded error on one.
+//
+// Please prefer to use multiplyNumbers instead of this function.
```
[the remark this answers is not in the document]
noisersup: We had and it's covered [here](https://github.com/FerretDB/FerretDB/pull/2232/files#diff-a21b8c640c12ecfe3c572f59caf5486544b411403f77500597b373634b17b726R364) and [here](https://github.com/FerretDB/FerretDB/pull/2232/files#diff-a21b8c640c12ecfe3c572f59caf5486544b411403f77500597b373634b17b726R354)

internal/handlers/common/params_test.go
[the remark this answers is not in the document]
chilagrow: Just currently the logic of getting v1,v2 and expected is not so trivial. Sometimes, a bit repetitive way is easier to maintain, sorry for a picky comment :sweat:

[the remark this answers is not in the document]
noisersup: No worries! It doesn't sound picky :smile:

internal/handlers/common/params.go
rumyantseva, line 380: But when should we use this function if `multiplyNumbers` is preferred instead?

internal/handlers/common/params.go
rumyantseva: @noisersup I remember you said you want to try `-1` and some other cases, is it still actual? I see the tests that handle `-1`, but according to the logic in this function, `errLongExceeded` will be returned for them, right?

internal/handlers/common/params_test.go
```diff
@@ -0,0 +1,92 @@
...
+			err: errLongExceeded,
+		},
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			actualRes, err := multiplyLongSafely(tc.v1, tc.v2)
+			assert.Equal(t, tc.err, err)
+
```
rumyantseva, line 88:
```suggestion
```

internal/handlers/common/params_test.go
noisersup: Fixed, I picked the value from the range that might be rounded because of double precision

internal/handlers/common/params.go
noisersup: Not exactly, in the first `multiplyLongSafely` case we check only for `0` and `1`, as those are only values possible to multiply correctly with `MinInt64`.

All other values compared to any other value are handled after the switch case, and all overflows should be detected there.

internal/handlers/common/params.go
```diff
@@ -351,16 +351,8 @@ func multiplyNumbers(v1, v2 any) (any, error) {
 
 			return res, nil
 		case int64:
-			res := int64(v1) * v2
+			return multiplyLongSafely(int64(v1), v2)
```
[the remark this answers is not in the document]
noisersup: That's strange, for me it shows: "No Files covered by tests were changed"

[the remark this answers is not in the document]
rumyantseva: Probably something on CI then... Let's wait for CI to pass and let's see if it still persists.

internal/handlers/common/params.go
noisersup: I guess that only in `multiplyNumbers` as it's just a helper function for it

rumyantseva: Hmm, maybe we can rephrase the comment this way? Something like `Please always use multiplyNumbers as it calls multiplyLongSafely when needed.`

internal/handlers/common/params_test.go
```diff
@@ -0,0 +1,91 @@
...
+		"Zero": {
+			v1:       0,
+			v2:       1000,
+			expected: 0,
+		},
+		"One": {
+			v1:       42,
+			v2:       1,
+			expected: 42,
+		},
```
chilagrow, line 40: So if we remove https://github.com/FerretDB/FerretDB/pull/2232/files#diff-a21b8c640c12ecfe3c572f59caf5486544b411403f77500597b373634b17b726R385
at least some test fail.
```
		"Zero": {
			v1:       0,
			v2:       MinInt64,
			expected: MinInt64,
		},
		"One": {
			v1:       1,
			v2:       MinInt64,
			expected: MinInt64,
		},
```

noisersup: Yes, those will fail as `MinInt64` is handled explicitly

internal/handlers/common/params.go
noisersup: fixed

internal/handlers/common/params.go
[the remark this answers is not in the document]
rumyantseva: Yeah, that looks fixed

rumyantseva (approved):
LGTM

PR #2224, opened 2023-03-16
Collection deletion and metadata refactoring

rumyantseva:
# Description

Closes #1509.

This issue is a part of #1509, it consists of the metadata-related refactoring we discussed before.

## Readiness checklist

* [ ] I added/updated unit tests.
* [ ] I added/updated integration/compatibility tests.
* [ ] I added/updated comments and checked rendering.
* [x] I made spot refactorings.
* [ ] I updated user documentation.
* [ ] I ran `task all`, and it passed.
* [x] I ensured that PR title is good enough for the changelog.
* [x] (for maintainers only) I set Reviewers ([`@FerretDB/core`](https://github.com/orgs/FerretDB/teams/core)), Assignee, Labels, Project and project's Sprint fields.
* [x] I marked all done items in this checklist.

PR #2239, opened 2023-03-20
Change collection names mangling for PostgreSQL

### LINKED ISSUE #2210: Change collection names mangling for PostgreSQL
### What should be done?

After the introduction of the metadata table, the way we mangle collection names and use `formatCollectionName` is incorrect:
1. `formatCollectionName` should be an implementation detail for the metadata table. Functions like `DropCollection` should use the metadata table, not call this function themselves.
2. `formatCollectionName` only allows longer names than PostgreSQL but does not allow characters like `/`. It should.

We should:
1. Make all code except `metadata.ensure` use the metadata table.
2. Update name mangling to allow [all valid MongoDB characters](https://www.mongodb.com/docs/manual/reference/limits/#naming-restrictions) by replacing [invalid PostgreSQL characters](https://www.postgresql.org/docs/15/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS) with underscores.
3. Update documentation and diff tests.
4. Not change `testutil.CollectionName` since it should generate Tigris-compatible names. Add/update a PostgreSQL specific test for that.

### Where?

`pgdb`, PostgreSQL handler.

### Definition of Done

- unit tests added/updated;
- spot refactorings done;
- user documentation updated.

chilagrow:
# Description

Closes #2210.

Also updated collection name size limit to match with compat.

See change in diff test https://github.com/FerretDB/dance/pull/374
## Readiness checklist


internal/handlers/pg/pgdb/collections.go
```diff
@@ -32,7 +32,8 @@ import (
 )
 
 // validateCollectionNameRe validates collection names.
-var validateCollectionNameRe = regexp.MustCompile("^[a-zA-Z_-][a-zA-Z0-9_-]{0,119}$")
+// Empty collection name, names with `$` and `\x00` are not allowed.
+var validateCollectionNameRe = regexp.MustCompile("^[^$\x00]{1,235}$")
```
chilagrow, line 36: See https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Restriction-on-Collection-Names

> 235 bytes for sharded collections

This is where I got 235 from.

integration/basic_test.go
```diff
@@ -367,6 +367,33 @@ func TestCollectionName(t *testing.T) {
...
+			alt: "Invalid collection name: 'testcollectionname.\x00'",
+		},
+		"Dot": {
+			collection: "collection.name",
+		},
+		"Space": {
+			collection: " ",
+		},
+		"NonLatin": {
+			collection: "coleção",
```
rumyantseva, line 386: Those still look quite Latin to me, like extended Latin. Let's maybe add something Japanese, or Greek, or Cyrillic to make it look more non-latin (just to be absolutely sure how it works)?

internal/handlers/pg/pgdb/collections.go
```diff
@@ -174,16 +175,13 @@ func DropCollection(ctx context.Context, tx pgx.Tx, db, collection string) error
...
+	table, err := metadata.getTableName(ctx)
 	if err != nil {
 		return lazyerrors.Error(err)
 	}
-	if !slices.Contains(tables, table) {
-		return ErrTableNotExist
-	}
 
-	err = newMetadata(tx, db, collection).remove(ctx)
+	err = metadata.remove(ctx)
```
rumyantseva, line 184: I suggest removing metadata after we drop table, like in this https://github.com/FerretDB/FerretDB/pull/2224/files PR. It's not super critical, as we are still in the same transaction, but I think this approach is a bit more accurate.

internal/handlers/pg/pgdb/database_metadata.go
```diff
@@ -220,6 +220,8 @@ func (m *metadata) remove(ctx context.Context) error {
 
 // formatCollectionName returns collection name in form <shortened_name>_<name_hash>.
 // Changing this logic will break compatibility with existing databases.
```
rumyantseva, line 222: This line is not needed anymore (just like in https://github.com/FerretDB/FerretDB/pull/2224/files)

rumyantseva (approved):
Looks good to me overall, the changes in diff.md are covered by the tests.

I added a few minor comments, I leave it up to you to consider them or ignore them.

internal/handlers/pg/pgdb/database_metadata.go
chilagrow: Thanks :pray:

internal/handlers/pg/pgdb/collections.go
chilagrow: Updated it

integration/basic_test.go
chilagrow: Sure, let's try japanese because I know how to write collection :hugs:

rumyantseva (dismissed):
LGTM

AlekSi (commented):
Please ignore

AlekSi (commented):
> Update name mangling to allow [all valid MongoDB characters](https://www.mongodb.com/docs/manual/reference/limits/#naming-restrictions) by replacing [invalid PostgreSQL characters](https://www.postgresql.org/docs/15/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS) with underscores.

That part wasn't done. We actually should change `formatCollectionName` (and rename it).

> /\. ,"*<>:|?_e251b03d 

While PostgreSQL supports that, I think supporting those PostgreSQL table names will be problematic. Instead, we should mangle collection names into less-problematic names, nut just cut them.

internal/handlers/pg/pgdb/database_metadata.go
```diff
@@ -219,7 +219,8 @@ func (m *metadata) remove(ctx context.Context) error {
 }
 
 // formatCollectionName returns collection name in form <shortened_name>_<name_hash>.
-// Changing this logic will break compatibility with existing databases.
+//
+// Deprecated: this function usage is allowed for collection metadata creation only.
 func formatCollectionName(name string) string {
```
w84thesun, line 224: I suggest renaming it to `transformCollectionName` as `transform` seems to be a more neutral verb.

w84thesun (commented):
Great job! Let's change those special characters to underscore as @AlekSi mentioned.

Fashander (dismissed):

internal/handlers/pg/pgdb/collections.go
noisersup: Can we introduce a test case that checks for exactly 235 bytes? Or maybe update the one that contains 100 bytes.

noisersup (dismissed):
LGTM! I would suggest to add one test though.

internal/handlers/pg/pgdb/collections.go
chilagrow: Good point, I've replaced the test with 235.

noisersup (dismissed):
LGTM!

internal/handlers/pg/pgdb/database_metadata.go
AlekSi: FWIW, "mangling" is a term: https://en.wikipedia.org/wiki/Name_mangling. "transform" is a bit too generic, IMHO

internal/handlers/pg/pgdb/database_metadata.go
```diff
@@ -218,20 +225,24 @@ func (m *metadata) remove(ctx context.Context) error {
...
 }
 
-// formatCollectionName returns collection name in form <shortened_name>_<name_hash>.
-// Changing this logic will break compatibility with existing databases.
-func formatCollectionName(name string) string {
+// transform returns name in form <shortened_name>_<name_hash>.
+// It replaces special characters with `_`.
+//
+// Deprecated: this function usage is allowed for collection metadata creation only.
+func transform(name string) string {
```
AlekSi, line 232: Independently of "mangling"/"transform"/"format" discussion, that name is too short for a package of a size of pgdb. Transform what? Some string to some string?

`transformCollectionName` would be better. Another option could be `collectionNameToTableName`, but that's maybe a bit long

internal/handlers/pg/pgdb/database_metadata.go
chilagrow: :nerd_face:  Okay so it can be called `mangle`

internal/handlers/pg/pgdb/database_metadata.go
```diff
@@ -218,20 +225,24 @@ func (m *metadata) remove(ctx context.Context) error {
...
 	nameSymbolsLeft := maxTableNameLength - hash32.Size()*2 - 1
-	truncateTo := len(name)
+	truncateTo := len(transformed)
 
 	if truncateTo > nameSymbolsLeft {
 		truncateTo = nameSymbolsLeft
 	}
 
-	return name[:truncateTo] + "_" + fmt.Sprintf("%x", hash32.Sum([]byte{}))
+	return transformed[:truncateTo] + "_" + fmt.Sprintf("%x", hash32.Sum([]byte{}))
```
AlekSi, line 245:
```suggestion
	return transformed[:truncateTo] + "_" + fmt.Sprintf("%x", hash32.Sum(nil))
```

That's more idiomatic.

There is also a subtle problem with that line: `%x` omits leading zeros, and that could cause, in extreme cases, clashes. The valid syntax is `%08x`

internal/handlers/pg/pgdb/database_metadata.go
```diff
@@ -45,6 +46,12 @@ const (
 	maxIndexNameLength = 63
 )
 
+// specialCharacters are potential problematic characters that are replaced with `_`.
+// They are operators and special characters of pg and white space.
+// See https://www.postgresql.org/docs/15/sql-syntax-lexical.html#SQL-SYNTAX-OPERATORS
+// and https://www.postgresql.org/docs/15/sql-syntax-lexical.html#SQL-SYNTAX-SPECIAL-CHARS
+var specialCharacters = regexp.MustCompile("[-+*/<>=~!@#%^&|`?()\\[\\],;:. ]")
```
AlekSi, line 53: I think negating is a safer option there. Something like `[^a-z0-9_]`, but a special check is needed for the first symbol that should be `[a-z]`

chilagrow: That works as well. Let me do that.

internal/handlers/pg/pgdb/database_metadata.go
chilagrow: Let me update this to the valid syntax :+1:

internal/handlers/pg/pgdb/collections.go
AlekSi, line 36: After we validate collect name with regex, let's also check it with `utf8.ValidString`. Otherwise, we could have a lot of "fun" Unicode-related problems

internal/handlers/pg/pgdb/database_metadata.go
chilagrow: perhaps `collectionNameToTableName` is good for this one, even it's a bit long, it specifies the user case and we don't want anything else to be using this/

internal/handlers/pg/pgdb/database_metadata.go
chilagrow: Okay so I think I can match with `[^a-z][^a-z0-9_]*`

internal/handlers/pg/pgdb/collections.go
chilagrow: Added this

internal/handlers/pg/pgdb/database_metadata.go
```diff
@@ -45,6 +46,10 @@ const (
 	maxIndexNameLength = 63
 )
 
+// specialCharacters are potential problematic characters of pg table name
+// that are replaced with `_`.
+var specialCharacters = regexp.MustCompile("[^a-z][^a-z0-9_]*")
```
chilagrow, line 51: This replaces `+-/*<>=~!@#%^&|`?()[],;:.` with `__`. The first `[^a-z]` matches `+` and `[^a-z0-9_]*` matches the rest.
I can't figure out how to produce the same number of characters. But this is fine for mangling pg table name.  Please suggest if you have better approach :hugs:

internal/handlers/pg/pgdb/database_metadata.go
chilagrow: at the end I ended up using `collectionNameToTableName`

internal/handlers/pg/pgdb/database_metadata.go
```diff
@@ -218,20 +223,24 @@ func (m *metadata) remove(ctx context.Context) error {
...
-func formatCollectionName(name string) string {
+// collectionNameToTableName returns name in form <shortened_name>_<name_hash>.
+// It replaces special characters with `_`.
+//
+// Deprecated: this function usage is allowed for collection metadata creation only.
+func collectionNameToTableName(name string) string {
 	hash32 := fnv.New32a()
 	must.NotFail(hash32.Write([]byte(name)))
 
+	mangled := specialCharacters.ReplaceAllString(name, "_")
```
rumyantseva, line 234: Do we understand what the probability of having a collision is? The previous approach looked safe - the hashes themselves were pretty unique. If we replace all special characters in two different hashes, is it probable that we will have to equal strings as the result?

website/docs/diff.md
```diff
@@ -24,11 +24,9 @@ slug: /diff/
 8. Database and collection names restrictions:
    * name cannot start with the reserved prefix `_ferretdb_`;
    * database name must not include non-latin letters, spaces, dots, dollars;
-   * collection name must not include non-latin letters, spaces, dots or dollars;
-   * name must not start with a number;
+   * collection name must be valid UTF-8 characters;
```
rumyantseva, line 27: Is it different from how MongoDB behaves, don't they need a valid UTF-8 string as well?

rumyantseva (commented):
- The code looks good to me (with one question).
- Asked a question about some `diff.md` changes.

rumyantseva (approved):

internal/handlers/pg/pgdb/database_metadata.go
chilagrow: But hash is built on the `name`, so hash should be different even if the produce name is `__`. 

I just cannot make regular expression to work, could you suggest how I should build regular expression to avoid possible collision?

website/docs/diff.md
chilagrow: Yes this is different
mongo
```
test> db.createCollection("Ä°nanÃ§")
{ ok: 1 }
```

FerretDB
```
test> db.createCollection("Ä°nanÃ§")
MongoServerError: Invalid collection name: 'test.Ä°nanÃ§'
```

internal/handlers/pg/pgdb/database_metadata.go
rumyantseva: Well, maybe the current approach is already enough? I guess, ideally, for each possible character, we need a different symbol or a different sequence of symbols. What special characters `hash32` can reproduce?

website/docs/diff.md
rumyantseva: Oh, nice! Is there a diff test in `dance` for it?

internal/handlers/pg/pgdb/database_metadata.go
chilagrow: hash32 does not produce special characters. It produces hex.

website/docs/diff.md
chilagrow: No, I'm doubting if this is worth putting it to dance.

internal/handlers/pg/pgdb/database_metadata.go
rumyantseva: Ah! Got it! So, we need this for the prefix only! Then it's unique enough, yes (because we consider the hash  unique enough).

chilagrow: Yeah that logic has not change :hugs:

website/docs/diff.md
chilagrow: True, let me add it. :sweat_smile:

w84thesun (dismissed):
👍

rumyantseva (approved):
LGTM, a diff test for non-utf8 symbols is needed in dance diff tests.

website/docs/diff.md
chilagrow: That will be fixed :hugs:

integration/basic_test.go
```diff
@@ -380,9 +407,11 @@ func TestCollectionName(t *testing.T) {
 
 			assert.NoError(t, err)
 
-			names, err := collection.Database().ListCollectionNames(ctx, bson.D{})
+			newCollection := collection.Database().Collection(tc.collection)
+
+			// insert data to ensure collection can be updated.
+			_, err = newCollection.InsertOne(ctx, bson.D{})
 			require.NoError(t, err)
-			assert.Contains(t, names, tc.collection)
```
AlekSi, line 385: Why did we change that? Was there a problem with the old test?

AlekSi (approved):

rumyantseva (approved):

w84thesun (approved):

noisersup (approved):

integration/basic_test.go
chilagrow: Contains work fine. We changed how table name is mangled, I wanted to ensure that there is no error using mangled table name. I could have used `Find` but since collection was empty, I used Insert.

website/docs/diff.md
```diff
@@ -23,12 +23,11 @@ slug: /diff/
    * update operations producing `Infinity`, `-Infinity`, or `NaN` are not supported.
 8. Database and collection names restrictions:
    * name cannot start with the reserved prefix `_ferretdb_`;
-   * database name must not include non-latin letters, spaces, dots, dollars;
-   * collection name must not include non-latin letters, spaces, dots or dollars;
-   * name must not start with a number;
+   * database name must not include non-latin letters, spaces, dots;
```
chilagrow, line 26: dollar is not allowed in both ferretdb and mongodb so it's not diff. It's already tested https://github.com/FerretDB/FerretDB/blob/main/integration/basic_test.go#L418

integration/basic_test.go
```diff
@@ -380,9 +407,20 @@ func TestCollectionName(t *testing.T) {
 
 			assert.NoError(t, err)
 
+			// check collection name is in the list.
```
AlekSi, line 410:
```suggestion
			// check collection name is in the list
```
(sentence fragment - no dot at the end)

or

```suggestion
			// Check that collection name is in the list.
```
(full sentence with a dot at the end)

That's not important right now, but something to write down in the future.

AlekSi (approved):

Fashander (approved):

noisersup (approved):

PR #2244, opened 2023-03-20
Implement `createIndex` command

### LINKED ISSUE #1509: Implement `createIndexes` command

rumyantseva:
# Description

Closes #1509.

## Readiness checklist

### DIFF
--- a/internal/handlers/pg/msg_createindexes.go
+++ b/internal/handlers/pg/msg_createindexes.go
@@ -16,23 +16,79 @@ package pg
 
 import (
 	"context"
+	"errors"
+
+	"github.com/jackc/pgx/v4"
 
 	"github.com/FerretDB/FerretDB/internal/handlers/common"
+	"github.com/FerretDB/FerretDB/internal/handlers/pg/pgdb"
 	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/iterator"
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
 	"github.com/FerretDB/FerretDB/internal/util/must"
 	"github.com/FerretDB/FerretDB/internal/wire"
 )
 
 // MsgCreateIndexes implements HandlerInterface.
 func (h *Handler) MsgCreateIndexes(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
-	// TODO https://github.com/FerretDB/FerretDB/issues/78
+	dbPool, err := h.DBPool(ctx)
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
 
 	document, err := msg.Document()
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+
+	common.Ignored(document, h.L, "writeConcern", "commitQuorum", "comment")
+
+	command := document.Command()
+
+	db, err := common.GetRequiredParam[string](document, "$db")
 	if err != nil {
 		return nil, err
 	}
 
-	common.Ignored(document, h.L, "writeConcern", "commitQuorum", "comment")
+	collection, err := common.GetRequiredParam[string](document, command)
+	if err != nil {
+		return nil, err
+	}
+
+	idxArr, err := common.GetRequiredParam[*types.Array](document, "indexes")
+	if err != nil {
+		return nil, err
+	}
+
+	iter := idxArr.Iterator()
+	defer iter.Close()
+
+	err = dbPool.InTransactionRetry(ctx, func(tx pgx.Tx) error {
+		for {
+			_, _, err = iter.Next()
+
+			switch {
+			case err == nil:
+				// do nothing
+			case errors.Is(err, iterator.ErrIteratorDone):
+				// iterator is done, no more indexes to create
+				return nil
+			default:
+				return lazyerrors.Error(err)
+			}
+
+			index := pgdb.Index{
+				// TODO
+			}
+
+			if err = pgdb.CreateIndex(ctx, tx, db, collection, &index); err != nil {
+				return err
+			}
+		}
+	})
+	if err != nil {
+		return nil, err
+	}
 
 	var reply wire.OpMsg
 	must.NoError(reply.SetSections(wire.OpMsgSection{
--- a/internal/handlers/pg/pgdb/indexes.go
+++ b/internal/handlers/pg/pgdb/indexes.go
@@ -115,9 +115,8 @@ func Indexes(ctx context.Context, tx pgx.Tx, db, collection string) ([]Index, er
 	}
 }
 
-// createIndex creates a new index for the given params.
-// TODO This method will become exported in https://github.com/FerretDB/FerretDB/issues/1509.
-func createIndex(ctx context.Context, tx pgx.Tx, db, collection string, i *Index) error {
+// CreateIndex creates a new index for the given params.
+func CreateIndex(ctx context.Context, tx pgx.Tx, db, collection string, i *Index) error {
 	pgTable, pgIndex, err := newMetadata(tx, db, collection).setIndex(ctx, i.Name, i.Key, i.Unique)
 	if err != nil {
 		return err
--- a/internal/handlers/pg/pgdb/collections.go
+++ b/internal/handlers/pg/pgdb/collections.go
@@ -133,7 +133,7 @@ func CreateCollection(ctx context.Context, tx pgx.Tx, db, collection string) err
 		Unique: true,
 	}
 
-	if err := createIndex(ctx, tx, db, collection, indexParams); err != nil {
+	if err := CreateIndex(ctx, tx, db, collection, indexParams); err != nil {
 		return lazyerrors.Error(err)
 	}
 


Taskfile.yml
```diff
@@ -5,7 +5,7 @@ env:
   GORACE: halt_on_error=1,history_size=2
 
 vars:
-  INTEGRATIONTIME: 20m
+  INTEGRATIONTIME: 25m
```
rumyantseva, line 8: Same as in some other PR

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -55,43 +56,79 @@ func Indexes(ctx context.Context, tx pgx.Tx, db, collection string) ([]Index, er
...
 
 	return nil
 }
 
 // createPgIndexIfNotExists creates a new index for the given params if it does not exist.
-func createPgIndexIfNotExists(ctx context.Context, tx pgx.Tx, schema, table, index string, isUnique bool) error {
+func createPgIndexIfNotExists(ctx context.Context, tx pgx.Tx, schema, table, index string, fields IndexKey, isUnique bool) error {
+	if len(fields) == 0 {
+		return lazyerrors.Errorf("no fields for index")
+	}
```
chilagrow, line 87: To test and add non-lazy error https://github.com/FerretDB/FerretDB/issues/2311

internal/handlers/pg/pgdb/database_metadata.go
```diff
@@ -403,6 +421,47 @@ func (ms *metadataStorage) setIndex(ctx context.Context, index string, key Index
...
+func checkExistingIndex(existing *metadataIndex, new *metadataIndex) (bool, error) {
+	var indexNameMatch bool
+
+	if existing.Name == new.Name {
+		indexNameMatch = true
+	}
+
+	if len(existing.Key) != len(new.Key) {
+		if indexNameMatch {
+			return false, ErrIndexNameAlreadyExist
```
chilagrow, line 441: This needs testing https://github.com/FerretDB/FerretDB/issues/2311

internal/handlers/pg/pgdb/database_metadata.go
```diff
@@ -403,6 +421,47 @@ func (ms *metadataStorage) setIndex(ctx context.Context, index string, key Index
...
+		} else {
+			return false, nil
+		}
```
chilagrow, line 444: I think it can be simplified. Unless I'm reading it wrong. Applies to the rest of code as well
```suggestion
		}
		return false, nil
		
```

integration/indexes_compat_test.go
```diff
@@ -0,0 +1,382 @@
...
+		},
+		"single-index": {
+			models: []mongo.IndexModel{
+				{Keys: bson.D{{"v", -1}}},
+			},
+		},
+		"duplicate_id": {
+			models: []mongo.IndexModel{
+				{
+					Keys: bson.D{{"_id", 1}}, // this index is already created by default
```
chilagrow, line 85: I feel index crated by default was unique index, which is the same as ascending index?

chilagrow (commented):
Just suggesting adding todo issue number on where codecov reported no coverage.

Looking great

rumyantseva (commented):
Addressed comments/questions

internal/handlers/pg/msg_createindexes.go
```diff
@@ -43,3 +134,143 @@ func (h *Handler) MsgCreateIndexes(ctx context.Context, msg *wire.OpMsg) (*wire.
...
+
+		duplicateChecker[field] = struct{}{}
+
+		var orderParam int64
+
+		if orderParam, err = common.GetWholeNumberParam(order); err != nil {
+			return nil, commonerrors.NewCommandErrorMsgWithArgument(
+				commonerrors.ErrNotImplemented,
+				fmt.Sprintf("Index key value %q is not implemented yet", order),
+				"createIndexes",
```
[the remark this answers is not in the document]
chilagrow: ok for now this is fine as it is :)

internal/handlers/pg/pgdb/indexes.go
rumyantseva: Yes, it is marked as not tested because we handle this case on the handler level itself. I didn't want to "pushdown" all the validation to the pgdb level as in this case we will have to introduce even more new error variables here for each different validation rule that requires a different error code to be returned. If the problem is solved, could you please close the test? (or suggest what needs to be done otherwise)

chilagrow (approved):
Awesome work :+1:

w84thesun (approved):
That's a lot of work here! Changes look reasonable to me. I'd say that test case names could be in camel case but we can fix that later.

rumyantseva:
@w84thesun, thank you! I totally forgot about camelcase 😅 I'll fix it in the next task, then!

internal/handlers/pg/msg_createindexes.go
```diff
@@ -16,24 +16,116 @@ package pg
...
+
+	collection, err := common.GetRequiredParam[string](document, command)
+	if err != nil {
+		return nil, err
+	}
+
+	idxArr, err := common.GetRequiredParam[*types.Array](document, "indexes")
+	if err != nil {
+		return nil, err
+	}
```
noisersup, line 63: I'm not sure in which cases we can have different errors and in which not, so just saying that this one results in different error and error code after running: `db.runCommand({createIndexes: "values", indexes: "string"})`

```
Body diff:
--- res body
+++ proxy body
@@ -15,5 +15,5 @@
         },
-        "errmsg": "required parameter \"indexes\" has type string (expected *types.Array)",
-        "code": 2,
-        "codeName": "BadValue"
+        "errmsg": "BSON field 'createIndexes.indexes' is the wrong type 'string', expected type 'array'",
+        "code": 14,
+        "codeName": "TypeMismatch"
       },
```

internal/handlers/pg/msg_createindexes.go
```diff
@@ -16,24 +16,116 @@ package pg
...
+
+	if idxArr.Len() == 0 {
+		return nil, commonerrors.NewCommandErrorMsgWithArgument(
+			commonerrors.ErrBadValue,
+			"Must specify at least one index to create",
+			document.Command(),
+		)
+	}
```
noisersup, line 71: So, if I understand this one is covered by test, but codecov is just broken for now, right?

internal/handlers/pg/msg_createindexes.go
noisersup: I believe it will be a part of #2311 right?

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -55,43 +56,79 @@ func Indexes(ctx context.Context, tx pgx.Tx, db, collection string) ([]Index, er
...
+		case types.Ascending:
+			order = "ASC"
+		case types.Descending:
+			order = "DESC"
+		default:
+			return lazyerrors.Errorf("unknown sort order: %d", field.Order)
+		}
+
+		// It's important to sanitize field.Field data here, as it's a user-provided value.
+		fieldsDef[i] = fmt.Sprintf(`((_jsonb->%s)) %s`, quoteString(field.Field), order)
```
noisersup, line 111: Why instead of creating new function can't we just use SQL arguments here?

internal/handlers/pg/msg_createindexes.go
rumyantseva: @noisersup yes, correct, it's for #2311. For the current PR, we just want to return some error.

internal/handlers/pg/pgdb/indexes.go
rumyantseva: @noisersup could you suggest what exactly you mean? My idea with a separate function was to highlight how exactly we sanitize the user input. The `QuoteString` function used to be an exported function in the `pgx` package a while ago (no it's in the `internal` repo, so it's not really exported to the other repos), so my idea was to show that use case.

noisersup: Can't we just

```suggestion
		// It's important to sanitize field.Field data here, as it's a user-provided value.
		fieldsDef[i] = fmt.Sprintf(`((_jsonb->%s)) %s`, p.Next(), order)
		args := []any{field.Field}
```
So we get something like: `((_jsonb->$1))`?

internal/handlers/pg/msg_createindexes.go
rumyantseva: As codecov complains about every single condition, I think it's broken, yes. This case should be covered by test that is called `"empty"` where we provide a valid empty array of indexes.

internal/handlers/pg/pgdb/indexes.go
AlekSi: The root cause is that DDL statements like `CREATE INDEX` do not accept normal arguments

noisersup: That's interesting

noisersup: It seems that this function might be unsafe when specific options on pg are set [see](https://github.com/jackc/pgx/issues/868#issuecomment-725544647)

Can we verify if this affects us, and maybe add a check for those config variables?

rumyantseva: That's an interesting finding.

I have a test case `"dangerous-key"` that contains SQL injection. If sanitization is not set, the injection passes, and the test fails.  Do you think it's enough? Or should there be some extra checks?

rumyantseva: (upd: found where to add the check! thanks a lot @noisersup!)

rumyantseva: Ok, I added a check to `pool.go - pgPool. checkConnection`. I wanted to add is somewhere here too, but that would require us to keep `pgPool` here and in many other places (e.g. like `createCollection`), which will lead to having `pgPool` everywhere.

@noisersup @AlekSi does the current check look enough to you?

internal/handlers/pg/pgdb/pool.go
```diff
@@ -170,6 +170,10 @@ func (pgPool *Pool) checkConnection(ctx context.Context) error {
 			if setting != localeC && setting != localePOSIX && !isValidUTF8Locale(setting) {
 				return fmt.Errorf("pgdb.checkConnection: %q is %q", name, setting)
 			}
+		case "standard_conforming_strings": // To sanitize safely: https://github.com/jackc/pgx/issues/868#issuecomment-725544647
+			if setting != "on" {
+				return fmt.Errorf("pgdb.checkConnection: %q is %q, want %q", name, setting, "on")
+			}
```
AlekSi, line 176: 👍

noisersup (dismissed):
Thanks for applying changes. I've run postgresql with `-c 'standard_conforming_strings=off'`.
FerretDB seems to run, mongosh connection is initialized, but after any command we get an error, which is a good solution in my opinion! :+1: 
```
test> db.values.find()
MongoServerError: [msg_find.go:38 pg.(*Handler).MsgFind] [pg.go:130 pg.(*Handler).DBPool] pgdb.checkConnection: "standard_conforming_strings" is "off", want "on"
```

LGTM!

integration/indexes_compat_test.go
```diff
@@ -0,0 +1,388 @@
...
+					Keys:    bson.D{{"bar", -1}},
+					Options: new(options.IndexOptions).SetName("index-name"),
+				},
+			},
+			resultType:  emptyResult,
+			altErrorMsg: "One of the specified indexes already exists with a different key",
+		},
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
```
AlekSi, line 189: No blank line there…

integration/indexes_compat_test.go
```diff
@@ -0,0 +1,388 @@
...
+		},
+		"key-not-set": {
+			collectionName: "test",
+			resultType:     emptyResult,
+			skip:           "https://github.com/FerretDB/FerretDB/issues/2311",
+		},
+	} {
+		name, tc := name, tc
+
+		t.Run(name, func(t *testing.T) {
```
AlekSi, line 332: … but blank line there. Why?

internal/handlers/pg/msg_createindexes.go
```diff
@@ -16,24 +16,116 @@ package pg
...
+			indexDoc, ok := val.(*types.Document)
+			if !ok {
+				// TODO Add better validation and return proper error: https://github.com/FerretDB/FerretDB/issues/2311
+				return lazyerrors.Errorf("expected index document, got %T", val)
+			}
+
+			var index *pgdb.Index
+
+			index, err = processIndexOptions(indexDoc)
+			if err != nil {
```
AlekSi, line 100:
```suggestion
			if index, err = processIndexOptions(indexDoc); err != nil {
```

internal/handlers/pg/msg_createindexes.go
```diff
@@ -16,24 +16,116 @@ package pg
...
+
+			if err = pgdb.CreateIndexIfNotExists(ctx, tx, db, collection, index); err != nil {
+				return err
+			}
+		}
+	})
+
+	switch {
+	case err == nil:
+	// do nothing
```
AlekSi, line 112:
```suggestion
		// do nothing
```

internal/handlers/pg/msg_createindexes.go
```diff
@@ -43,3 +135,145 @@ func (h *Handler) MsgCreateIndexes(ctx context.Context, msg *wire.OpMsg) (*wire.
 
 	return &reply, nil
 }
+
+// processIndexOptions processes the given indexDoc and returns a pgdb.Index.
+func processIndexOptions(indexDoc *types.Document) (*pgdb.Index, error) {
+	var index pgdb.Index
+
+	iter := indexDoc.Iterator()
+
```
AlekSi, line 144: defer iter.Close()

internal/handlers/pg/msg_createindexes.go
```diff
@@ -43,3 +135,145 @@ func (h *Handler) MsgCreateIndexes(ctx context.Context, msg *wire.OpMsg) (*wire.
...
+				"createIndexes",
+			)
+		}
+	}
+}
+
+// processIndexKey processes the document containing the index key.
+func processIndexKey(keyDoc *types.Document) (pgdb.IndexKey, error) {
+	res := make(pgdb.IndexKey, 0, keyDoc.Len())
+	keyIter := keyDoc.Iterator()
```
AlekSi, line 218: defer keyIter.Close()

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -55,43 +56,81 @@ func Indexes(ctx context.Context, tx pgx.Tx, db, collection string) ([]Index, er
...
+	}
+
+	tx.Conn()
```
AlekSi, line 114: huh?

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -55,43 +56,81 @@ func Indexes(ctx context.Context, tx pgx.Tx, db, collection string) ([]Index, er
...
 	if _, err = tx.Exec(ctx, sql); err != nil {
 		return lazyerrors.Error(err)
 	}
 
 	return nil
 }
+
+// quoteString returns a string that is safe to use in SQL queries.
+//
+// Warning! Avoid using this function unless there is no other way.
```
AlekSi, line 128: Let's use the deprecation comment as we do in other places

AlekSi (dismissed):
Docs LGTM

internal/handlers/pg/pgdb/indexes.go
rumyantseva: oh :D

chilagrow (approved):

integration/indexes_compat_test.go
```diff
@@ -0,0 +1,388 @@
...
+		AddNonExistentCollection: true,
+	})
+	ctx, targetCollections, compatCollections := s.Ctx, s.TargetCollections, s.CompatCollections
+
+	for i := range targetCollections {
+		targetCollection := targetCollections[i]
+		compatCollection := compatCollections[i]
+
+		t.Run(targetCollection.Name(), func(t *testing.T) {
+			t.Helper()
```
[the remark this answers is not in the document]
chilagrow: keeping this one.

integration/indexes_compat_test.go
chilagrow: no blank everywhere now

internal/handlers/pg/msg_createindexes.go
chilagrow: added this

internal/handlers/pg/msg_createindexes.go
chilagrow: added this

internal/handlers/pg/msg_createindexes.go
chilagrow: space added

internal/handlers/pg/pgdb/indexes.go
chilagrow: It now has deprecated tag

integration/indexes_compat_test.go
chilagrow: removed blank line

integration/indexes_compat_test.go
```diff
@@ -0,0 +1,388 @@
...
+			targetRes := FetchAll(t, ctx, targetCur)
+			compatRes := FetchAll(t, ctx, compatCur)
+
+			assert.Equal(t, compatRes, targetRes)
+		})
+	}
+}
+
+func TestIndexesCreate(t *testing.T) {
+	t.Helper()
```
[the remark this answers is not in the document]
chilagrow: It has t.Parallel() instead of t.Helper().

integration/indexes_compat_test.go
```diff
@@ -0,0 +1,388 @@
...
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+
+	"github.com/FerretDB/FerretDB/integration/setup"
+	"github.com/FerretDB/FerretDB/integration/shareddata"
+)
+
+func TestIndexesList(t *testing.T) {
+	t.Helper()
```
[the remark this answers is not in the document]
chilagrow: yeah they are test, not helper. removed that and added t.parallel

chilagrow (approved):

AlekSi (approved):

AlekSi:
@rumyantseva please do a final look on recent commits and remove https://github.com/FerretDB/FerretDB/labels/do%20not%20merge label

PR #2245, opened 2023-03-21
Filter documents with iterators

### LINKED ISSUE #1733: Support `getMore` command

AlekSi:
# Description

Closes #1733.

## Readiness checklist

### DIFF
--- a/internal/clientconn/conninfo/conn_info.go
+++ b/internal/clientconn/conninfo/conn_info.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/FerretDB/FerretDB/internal/types"
 	"github.com/FerretDB/FerretDB/internal/util/debugbuild"
-	"github.com/FerretDB/FerretDB/internal/util/iterator"
 )
 
 // contextKey is a special type to represent context.WithValue keys a bit more safely.
@@ -98,7 +97,7 @@ func (connInfo *ConnInfo) Close() {
 
 // Cursor allows clients to iterate over a result set.
 type Cursor struct {
-	Iter   iterator.Interface[int, *types.Document]
+	Iter   types.DocumentsIterator
 	Filter *types.Document
 }
 
@@ -116,7 +115,7 @@ func (connInfo *ConnInfo) Cursor(id int64) *Cursor {
 }
 
 // StoreCursor stores cursor and return its ID.
-func (connInfo *ConnInfo) StoreCursor(iter iterator.Interface[int, *types.Document], filter *types.Document) int64 {
+func (connInfo *ConnInfo) StoreCursor(iter types.DocumentsIterator, filter *types.Document) int64 {
 	connInfo.rw.Lock()
 	defer connInfo.rw.Unlock()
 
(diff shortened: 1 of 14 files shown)


internal/clientconn/conninfo/conn_info.go
```diff
@@ -116,7 +114,7 @@ func (connInfo *ConnInfo) Cursor(id int64) *Cursor {
 }
 
 // StoreCursor stores cursor and return its ID.
-func (connInfo *ConnInfo) StoreCursor(iter iterator.Interface[int, *types.Document], filter *types.Document) int64 {
+func (connInfo *ConnInfo) StoreCursor(iter types.DocumentsIterator) int64 {
```
rumyantseva, line 117: @AlekSi do you understand why some lines are marked as not covered by tests? E.g. this function - the signature has changes, so it has changed in the callers too, but where are those callers?

internal/handlers/common/filter_iterator.go
```diff
@@ -0,0 +1,70 @@
...
+	iter   types.DocumentsIterator
+	filter *types.Document
+}
+
+// Next implements iterator.Interface. See FilterIterator for details.
+func (iter *filterIterator) Next() (int, *types.Document, error) {
+	for {
+		i, doc, err := iter.iter.Next()
+		if err != nil {
+			return 0, nil, lazyerrors.Error(err)
```
rumyantseva, line 48: So, even for the case when iterator is done, we return a lazy error? Should we follow this approach everywhere else as well?

internal/types/documents_iterator.go
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+import "github.com/FerretDB/FerretDB/internal/util/iterator"
+
+// DocumentsIterator represents an iterator over documents.
```
rumyantseva, line 19: Should we specify what `key` means in this case?

rumyantseva (commented):
I understand the idea behind DocumentsIterator and FilterIterator, and I'm ok with it. Overall, with all those iterators, the code looks more and more complicated, but it looks like those things are needed for us.

Asked a few questions.

internal/clientconn/conninfo/conn_info.go
AlekSi: There are none. This function will be used in future PRs.

internal/handlers/common/filter_iterator.go
AlekSi: > // Returned error could be (possibly wrapped) ErrIteratorDone or some fatal error
> // like (possibly wrapped) context.Canceled.

So the caller should be able to handle both. And there it allows us to simplify code

internal/types/documents_iterator.go
AlekSi: That's undefined in the type itself, and documented by users of it

rumyantseva (approved):

w84thesun (approved):

PR #2263, opened 2023-03-23
Extract dollar path field expression to types package

### LINKED ISSUE #2226: Extract dollar field path logic to types package
### What should be done?

Dollar field path logic (`$v`) is currently used by `$group` aggregator. Soon it will be used by `$unwind` aggregator. https://github.com/FerretDB/FerretDB/issues/1442

### Where?

https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/common/aggregations/group.go#L214

### Definition of Done

- [ ] integration/compatibility tests added/updated;
- [ ] Extract logic of getting field path to types package.
- [ ] Replace current implementation of `$group` dollar field logic with new implementation

chilagrow:
# Description

Closes #2226.

## Readiness checklist


internal/types/expression.go
```diff
@@ -0,0 +1,222 @@
...
+// find values that exist at suffix.
+// An array may return multiple values.
+// At each key of the path, it checks:
+//
+//	if the document has the key,
+//	if the array contains documents which have the key.
+//
+// It is different from `getDocumentsAtSuffix`, it does not find array item by
+// array dot notation `foo.0.bar`. It returns empty array [] because using index
+// such as `0` does not match using expression path.
```
chilagrow, line 177: Yet another way to find elements from path.

integration/shareddata/composites.go
```diff
@@ -73,6 +73,33 @@ var Composites = &Values[string]{
...
+	data: map[string]any{
+		"null":        nil,
+		"unset":       unset,
+		"array-empty": bson.A{},
+		"array-null":  bson.A{nil},
+	},
+}
+
+// ArrayAndDocuments contain composite values for tests.
+var ArrayAndDocuments = &Values[string]{
```
chilagrow, line 90: @noisersup you caught a bug :hugs: The commit relevant for this fix was https://github.com/FerretDB/FerretDB/pull/2263/commits/ecf2b70c7937f7711609fd350ae31656331f6629

internal/handlers/common/aggregations/group.go
```diff
@@ -205,31 +204,8 @@ type idAccumulator struct {
...
-	// so all `in` documents contain the same v.
-	v, err := in[0].GetByPath(path)
-	if err != nil {
-		return types.Null, nil
-	}
-
-	return v, nil
+func (a *idAccumulator) Accumulate(ctx context.Context, groupID any, in []*types.Document) (any, error) {
+	return groupID, nil
 }
```
noisersup, line 209: The code looks a bit confusing to me.

There's an `Accumulator` interface, it's usage is bit unclear, as it seems that all implementing structs doesn't accumulate any data.
After a while I realized it could be representation of MongoDB `$group`'s accumulators, like `$count`.
We have a nice map `accumulators` map that shows every supported accumulator, cool.

Afterwards, there's `idAccumulator`, which also implements the same `Accumulator` interface, but it doesn't exist as accumulator in mongodb docs.
It looks like some internal struct to handle `_id` fields, but it's under the same interface as `$count`, which is MongoDB functionality, right?

I understand logic more or less but at the first glance it was really confusing to understand what's responsible for what.

I assume that I could misunderstood something, so I don't have many ideas how to improve it, just wanted to leave it here, to
maybe consider future refactorings or not :)

noisersup (dismissed):
It looks like issue's DoD is done so LGTM, there are only linters to be fixed!

Apart from that I have some concerns (or just misunderstanding) about the already existing architecture of accumulators. I don't consider it as something blocking this particular PR, but feel free to discuss it :)

rumyantseva (approved):
The tests cover all the necessary cases, so the PR looks good. 

The implemented approach is similar to the other places where we work with paths, so it looks consistent with the other parts of the code.

I agree with Patryk's concerns, but maybe we just need more experience in writing some aggregation code.

internal/handlers/common/aggregations/group.go
chilagrow: Great feedback :hugs:  Let's not make `idAcumulator` be part of `Accumulator` implementation!

rumyantseva (approved):

noisersup (approved):
Thanks for changes! LGTM

PR #2246, opened 2023-03-21
Fix dot notation negative index errors

w84thesun:
# Description

Closes #2050.

## Readiness checklist

* [ ] I added/updated unit tests.
* [ ] I added/updated integration/compatibility tests.
* [ ] I added/updated comments and checked rendering.
* [ ] I made spot refactorings.
* [ ] I updated user documentation.
* [ ] I ran `task all`, and it passed.
* [ ] I ensured that PR title is good enough for the changelog.
* [ ] (for maintainers only) I set Reviewers ([`@FerretDB/core`](https://github.com/orgs/FerretDB/teams/core)), Assignee, Labels, Project and project's Sprint fields.
* [ ] I marked all done items in this checklist.

PR #2315, opened 2023-03-29
Fix failing test `TestAggregateCompatGroupDocDotNotation`

chilagrow:
# Description

See https://github.com/FerretDB/FerretDB/pull/2246#issuecomment-1487999359
and failing test https://github.com/FerretDB/FerretDB/actions/runs/4550452790/jobs/8023489900

`[42]` and `42` are the same sort order, so result in compat and target may be ordered whichever is first. This PR updates provider so that it produces unique sort order of result.
## Readiness checklist


AlekSi (changes requested):
I don't think we should change our test cases until the problem uncovered by them goes away.

AlekSi:
> [42] and 42 are the same sort order, so result in compat and target may be ordered whichever is first

That should be solved by always sorting by _id at last, and not allowing arrays for _id

chilagrow:
See the example https://github.com/FerretDB/FerretDB/pull/2246#issuecomment-1487999359, the result only has `_id` field and sorting is applied implicitly by aggregation test.

AlekSi:
> because _id is [42]

It should not be possible to set _id to [42]. From https://www.mongodb.com/docs/manual/core/document/#field-names:

> The field name _id is reserved for use as a primary key; its value must be unique in the collection, is immutable, and may be of any type _other than an array_.

If MongoDB breaks their own documentation there, let's consider that their mistake and don't repeat it.

chilagrow:
> 

That is unfortunately the case :cry:

AlekSi:
Could we set a different field in that test?

chilagrow:
If we set other field to the test, we will lose a test case capturing existing dot notation, `v.foo` is what we have now. The failing test and the `ArrayAndDocuments` were added at the same time to cover the case where `v.foo` exists in `array-document` and also `document` in this pr https://github.com/FerretDB/FerretDB/pull/2263. That's why I think updating provider is fine for this case

AlekSi:
Please see #2316

w84thesun (approved):
👍

noisersup (approved):

chilagrow:
Closing since this is not the way we want to handle this.

PR #2316, opened 2023-03-29
Draft

AlekSi:
# Description

Closes #{issue_number}.

## Readiness checklist

### DIFF
--- a/integration/aggregate_compat_test.go
+++ b/integration/aggregate_compat_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/primitive"
+	"golang.org/x/exp/slices"
 
 	"github.com/FerretDB/FerretDB/integration/setup"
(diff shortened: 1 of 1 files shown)


integration/aggregate_compat_test.go
```diff
@@ -656,9 +648,10 @@ func TestAggregateCompatGroupDocDotNotation(t *testing.T) {
 
 	testCases := map[string]aggregateStagesCompatTestCase{
 		"DocDotNotation": {
-			pipeline: bson.A{bson.D{{"$group", bson.D{
-				{"_id", "$v.foo"},
-			}}}},
+			pipeline: bson.A{
+				bson.D{{"$sort", bson.D{{"_id", 1}}}},
```
AlekSi, line 652: If our providers don't have _id with arrays (and they should not) and if our sorting is stable for scalar values, then sorting by _id before grouping should do the trick

chilagrow: Okay I see what you mean, I finally understood. The result of aggregation should in the order it was given so as long as input is sorted uniquely, we will always get deterministic result. Thanks for this!

PR #2309, opened 2023-03-28
Add basic `getMore` command

AlekSi:
# Description

Closes #1733.

## Readiness checklist

* [ ] I added/updated unit tests.
* [x] I added/updated integration/compatibility tests.
* [x] I added/updated comments and checked rendering.
* [x] I made spot refactorings.
* [ ] I updated user documentation.
* [x] I ran `task all`, and it passed.
* [x] I ensured that PR title is good enough for the changelog.
* [x] (for maintainers only) I set Reviewers ([`@FerretDB/core`](https://github.com/orgs/FerretDB/teams/core)), Assignee, Labels, Project and project's Sprint fields.
* [x] I marked all done items in this checklist.

PR #2319, opened 2023-03-29
Improve resource tracking

### LINKED ISSUE #2314: Use pprof's custom profilers and finalizers
> // Add adds the current execution stack to the profile, associated with value.
> // Add stores value in an internal map, so value must be suitable for use as
> // a map key and _will not be garbage collected_ until the corresponding
> // call to Remove. Add panics if the profile already contains a stack for value.

So pprof profiles do not play nicely with finalizers. Should we add some wrapper object? Or behave differently in debug build and not?

---

### What should be done?

When we use sharp-edged Finalizers, we store two copies of the stack: https://github.com/FerretDB/FerretDB/blob/7d318673dce4dc31cfefde0ae8490b5775a22911/internal/clientconn/conninfo/conn_info.go#L30-L82

In the example above, one is stored as a field, and another inside `connInfoProfiles`. But it is possible to keep only the latter and get a readable stack with https://pkg.go.dev/runtime/pprof#Profile.WriteTo (debug=1). We should do that.

### Where?

In all cases like the one above.

### Definition of Done

- only one stack is kept;
- unit tests added that ensure that stack trace on finalizer is human-readable.

AlekSi:
# Description

Closes #2309.
Closes #2314.

## Readiness checklist

### DIFF
--- a/internal/clientconn/conn.go
+++ b/internal/clientconn/conn.go
@@ -127,6 +127,11 @@ func newConn(opts *newConnOpts) (*conn, error) {
 //
 // The caller is responsible for closing the underlying net.Conn.
 func (c *conn) run(ctx context.Context) (err error) {
+	ctx, cancel := context.WithCancelCause(ctx)
+	defer func() {
+		cancel(lazyerrors.Errorf("run exits: %w", err))
+	}()
+
 	connInfo := conninfo.NewConnInfo()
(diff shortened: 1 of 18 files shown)


rumyantseva (approved):
I didn't check super attentively as there are a lot of changes, but I didn't see red flags at first sight.

Issue #2321, opened 2023-03-29
Replace all weird numbers with constant values


noisersup:
### What should be done?

Currently in integration tests we have a couple of values that look like this:
```
"double-7":                  1.79769e+307,
"double-max-overflow":       9.223372036854776833e+18,
"double-max-overflow-verge": 9.223372036854776832e+18,
```

It's difficult to understand what exactly they represent. 
For example, because of that reason we use `math.MaxInt64` in place of `9223372036854775807`, and `doubleMaxPrec` (which is defined with bitshift operator `1<<53 - 1`) in place of `9007199254740991.0`.

We should identify what exactly those values represent and replace them with calls to constants.

If it's possible - let's simplify creation of those constants by using bitshift operators or simple arithmetic, so they are "human-readable".

There's no need to replace values like this though, they look pretty straightforward (it's just a random larger integer, parsed to float64):
```
"double-4":                  float64(123456789)
```

### Where?

- shareddata package

### Definition of Done

- find all unclear numbers and identify them;
- create constants for those numbers, add comments for them;
- replace all numbers with created and/or existing constants;
- remove all TODO references to this issue.

PR #2324, opened 2023-03-29
Improve large numbers initialization in test data

noisersup:
# Description

Please also take a look at #2321 which should be done after merging this one.

## Readiness checklist


integration/shareddata/scalars.go
```diff
@@ -134,20 +138,22 @@ var Doubles = &Values[string]{
...
+		"double-prec-max":          doubleMaxPrec,
+		"double-prec-max-plus":     doubleMaxPrec + 1,
+		"double-prec-max-plus-two": doubleMaxPrec + 2,
+		"double-prec-max-minus":    doubleMaxPrec - 1,
 
 		// negative double big values ~ -(1<<61)
 		"double-neg-big":       -doubleBig,
-		"double-neg-big-plus":  -(doubleBig + 1),
-		"double-neg-big-minus": -(doubleBig - 1),
+		"double-neg-big-plus":  -doubleBig + 1,
```
chilagrow, line 149: Previously this was named `"double-neg-big-minus"` but rename is fine as long as tests don't fail :hugs:

integration/shareddata/scalars.go
```diff
@@ -134,20 +138,22 @@ var Doubles = &Values[string]{
...
+		"double-neg-big-minus": -doubleBig - 1,
 
 		// double min precision ~ -(1<<53 - 1)
-		"double-prec-min":       -(doubleMaxPrec - 1),
-		"double-prec-min-plus":  -(doubleMaxPrec - 1) + 1,
```
chilagrow, line 149: Where is `double-prec-min-plus`? It should `-doubleMaxPrec+2` but seems to be missing?

integration/shareddata/scalars.go
```diff
@@ -390,20 +396,22 @@ var Int64s = &Values[string]{
...
+		"int64-prec-max-minus":    int64(doubleMaxPrec) - 1,
 
 		// negative long big values ~ -(1<<61)
 		"int64-neg-big":       -longBig,
-		"int64-neg-big-plus":  -(longBig + 1),
-		"int64-neg-big-minus": -(longBig - 1),
-
-		// long representation of double max precision ~ -(1<<53-1)
-		"int64-prec-min":       -int64(doubleMaxPrec),
-		"int64-prec-min-plus":  -(int64(doubleMaxPrec) - 1) + 1,
```
chilagrow, line 405: This is missing, it should be `-int64(doubleMaxPrec)+2`

chilagrow (commented):
The shareddata is changed, some values seem missing. Perhaps there is a reason?

internal/util/must/must.go
```diff
@@ -35,3 +35,13 @@ func NoError(err error) {
 		panic(err)
 	}
 }
+
+// BeTrue panic if the b is not true.
+//
+// Use that function only for static initialization, test code, or statemants that
+// "can't" be true. When in doubt, don't.
```
AlekSi, line 42:
```suggestion
// "can't" be false. When in doubt, don't.
```

😆

integration/shareddata/scalars.go
rumyantseva: Renaming looks good to me, but if "plus" and "minus" are confusing, maybe we can use "left" and "right"... (to me, plus and minus are ok, as I'll take a look at the value anyway)

integration/shareddata/scalars.go
```diff
@@ -134,20 +138,22 @@ var Doubles = &Values[string]{
...
+		"double-neg-big-minus": -doubleBig - 1,
 
 		// double min precision ~ -(1<<53 - 1)
-		"double-prec-min":       -(doubleMaxPrec - 1),
-		"double-prec-min-plus":  -(doubleMaxPrec - 1) + 1,
-		"double-prec-min-minus": -(doubleMaxPrec - 1) - 1,
+		"double-prec-min":           -doubleMaxPrec,
+		"double-prec-min-plus":      -doubleMaxPrec + 1,
+		"double-prec-min-minus":     -doubleMaxPrec - 1,
+		"double-prec-min-minus-two": -doubleMaxPrec - 2,
```
rumyantseva, line 156: That's the one we didn't have before. I'm ok to have both `minus one` and `minus two`, but then let's also have `plus one` and `plus two` as `plus two` is something we have in the other places of this PR.

And if we have `minus two` here (if there is a reason for it), should we also have it in the other places?

rumyantseva (commented):
Great job clarifying those things! 

I left a few comments about the consistency of the cases.

As far as I understand, for negative cases, we should have `minus two`, `minus one`, the value itself, `plus one`. For positive cases we should have `minus one`, the value itself, `plus one`, `plus two`. But in some cases, it's not exactly like this.

integration/shareddata/scalars.go
noisersup: There's a `double-prec-min-plus` at line 154.

integration/shareddata/scalars.go
noisersup: I don't think that `+2` is necessary for `double-prec-min` cases. The point of doing a check for `-2` values is related to the fact that because of lost precision `double-prec-min`-1 == `double-prec-min`-2. The same applies for `double-prec-max` **+**`2`

integration/shareddata/scalars.go
noisersup: Please take a look, I believe it's relevant https://github.com/FerretDB/FerretDB/pull/2324/files#r1152969755

integration/shareddata/scalars.go
```diff
@@ -19,13 +19,17 @@ import (
...
+
 	"go.mongodb.org/mongo-driver/bson/primitive"
 )
 
 const (
-	doubleBig     = float64(1 << 61) // 2305843009213693952.0
-	longBig       = int64(1 << 61)   // 2305843009213693952
-	doubleMaxPrec = float64(1 << 53) // 9007199254740992.0
+	doubleMaxPrec = float64(1<<53 - 1) // 9007199254740991.0:    largest double values that could be represented as integer exactly
+	doubleBig     = float64(1 << 61)   // 2305843009213693952.0: larger then safe integer (doubleBig+1 == doubleBig)
```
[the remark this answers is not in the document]
noisersup: The largest safe integer is `1<<53`. The `1<<61` is just larger :)

integration/shareddata/scalars.go
rumyantseva: yeah, that sounds reasonable!

w84thesun (approved):
👍

integration/shareddata/scalars.go
chilagrow: I know the same variable name `double-prec-min-plus` exists. What I want to know is why the value of the variable has changed. It was  `-(doubleMaxPrec - 1) + 1` which is equivalent of `-doubleMaxPrec+2` right? Now it is `-doubleMaxPrec + 1` which is the different value.

chilagrow: Okay so this is for the same reason we don't need both +2 because it's tested by -1, and -2.

integration/shareddata/scalars.go
chilagrow: Ah I see, got it too

chilagrow (approved):

integration/shareddata/scalars.go
noisersup: The previous value was wrong.
`double-prec-min` is not `-(1<<53 -1) but just `-(1<<53)`

integration/shareddata/scalars.go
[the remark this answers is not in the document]
noisersup: done

rumyantseva (approved):

rumyantseva (approved):

PR #2335, opened 2023-03-31
Implement `$collStats` aggregation stage count field

### LINKED ISSUE #1416: Implement `$collStats` aggregation pipeline stage
https://www.mongodb.com/docs/manual/reference/operator/aggregation/collStats/

Without proper error handling, etc.

In the scope of this issue - only basic support:
- [x] storageStats - #2338
- [x] count -  #2338

In separate issues:
- #2342 - storageStats advanced fields
- #2341 - latencyStats and latencyStats.histograms
- #2341 - queryExecStats

Check for issue number in the code for unimplemented field.

chilagrow:
# Description

Closes #1416.

This implements `count` field only.

The main issue should be open until all `latencyStats`, `latencyStats.histograms`, `storageStats` and `queryExecStats` field are implemented in separate PRs.

## Readiness checklist


internal/handlers/common/aggregations/count.go
```diff
@@ -30,7 +30,7 @@ type count struct {
 }
 
 // newCount creates a new $count stage.
-func newCount(stage *types.Document) (Stage, error) {
+func newCount(stage *types.Document, fetcher Fetcher) (Stage, error) {
```
rumyantseva, line 33: Transfering fetcher to all the functions (even the ones where it's not needed) doesn't look very flexible.
Should we extend the `newX` functions with something more flexible? 

For example,

`newX (stage *types.Document, opts ...*stageOptions)`

and then `fetcher` could be one of such options that we pass only to the function that is needed?

Though, I'm not sure if it's possible to do it this way -- I see in the handler we iterate through `for i, d := range stagesDocs {` and then for each stage we call that `new` with fetcher... 

I think I need more time to think how it could be improved :)

rumyantseva (commented):
I left a comment (maybe we need options for stages), but I don't have a particular solution, so I will think if I can offer some exact approach.

internal/handlers/common/aggregations/count.go
chilagrow: Good point. for `newX(...)` let me see if I can think of a way to do it. :thinking: 

For `opts ...*stageOptions`, I thought of that approach, of passing values but we have to can't pass `pg dbPool` or `tigris dbPool` so we will have to pass an interface in it anyway for now. So I just have interface without wrapping in another structure.

rumyantseva:
See also #2338 that is based on this PR.

chilagrow:
Closing to favor https://github.com/FerretDB/FerretDB/pull/2338

PR #2359, opened 2023-04-04
Provide actual connection string example in `README.md`

rumyantseva:
# Description

I think we should use an example that works, so people don't need to think about what params are missing.

## Readiness checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ that don't have those problems.
 
 With that container running, you can:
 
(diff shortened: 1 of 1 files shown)


Fashander (approved):

rumyantseva:
After some additional testing, we clarified that the docs described things correctly (e.g. when the default PG connection string already contains username and password, the user don't have to set them in FerretDB connection string).

PR #2373, opened 2023-04-05
Update writing guide

Fashander:
# Description

## Readiness checklist


website/docs/contributing/writing-guide.md
```diff
@@ -5,11 +5,34 @@ draft: true
...
+---
+```
+
 ## Names and URLs
 
 Use `kebab-case-with-dashes` instead of `snake_case_with_underscores` or spaces
 for file names, directory names, and slugs because URL paths typically use dashes.
 
+Ensure that the file name/URL path matches the title of the page.
+For example, if the title of your page is "Getting Started", then the file name/URL path (`slug` in front matter) should also be "getting-started" to maintain consistency.
```
AlekSi, line 28: I think we should recommend omitting `slug:` be default, and say that it only for special cases like external links, backward compatibility, etc

AlekSi (commented):
I think we could also link to Docusaurus docs like https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#markdown-front-matter

w84thesun (approved):
👍

chilagrow (approved):

AlekSi (approved):

PR #2381, opened 2023-04-06
Introduce integration benchmarks

### LINKED ISSUE #2378: Introduce integration-level benchmarks
### What should be done?

Let's introduce benchmarks for a various queries to check if we don't regress in performance.

```[tasklist]
### Let's create 2 sets of tests
- [x] Inserting large amount of documents and querying them
- [ ] Inserting one large document, find and replace it #2379
```

- We shouldn't use any external data for that.
- All tests should compare time between 2 databases
- We should also create subsets that will run tests with and without pushdown
- We should compare benchmark results with previous one (benchstat)

### Where?

- `integration` package. (Consider using subpackage)

noisersup:
# Description

Closes #2378.

This PR introduces changes to setup functions and shareddata to make implementation of different benchmark cases easy. It also implements benchmarks for simple filtering. 

## Readiness checklist


noisersup:
copilot:all

integration/shareddata/shareddata.go
```diff
@@ -197,7 +199,43 @@ func (values *Values[idType]) IsCompatible(backend string) bool {
...
 
+// BenchmarkProvider is implemented by shared data sets that provide documents for benchmarks.
+type BenchmarkProvider interface {
+	// Hash returns expected hash of all provider documents.
+	Hash() string
+
+	// Docs returns iterator that returns all documents from provider.
+	// They should be always in deterministic order.
+	Docs() iterator.Interface[struct{}, bson.D]
+}
```
noisersup, line 210: We introduce another provider, as from `Provider` we require data to be stored in random order, and `BenchmarkProvider` needs to be deterministic.

We use iterator in place of slice, as benchmarks might include much more data than tests, and it shouldn't be store in memory.

integration/shareddata/benchmark.go
```diff
@@ -0,0 +1,49 @@
...
+package shareddata
+
+import "go.mongodb.org/mongo-driver/bson"
+
+// SimpleBenchmarkValues contain documents with various types of values.
+var SimpleBenchmarkValues BenchmarkValues = NewBenchmarkValues(
+	"Ha59TmC9nwTXqEob73HIwqWUcx2Lp7aengCtNsqjWK0=",
+	GenerateMixedValues(),
+)
+
```
noisersup, line 25: To make sure that generator produce exact same data every time in the same order, we calculate the checksum of all inserted documents to be sure that expected and actual checksum are the same.

Checksum is also used in benchmark test names. It's good for tooling like benchstat as after changing BenchmarkValues data, the results will not be compared with the old values.

integration/shareddata/shareddata.go
chilagrow: Does `Provider` needs to be random order? I think `Provider` being random is due to design or implementation. Because if provider can be deterministic, it's easier to maintain one implementation than having multiple.

integration/setup/setup.go
```diff
@@ -228,29 +276,74 @@ func setupCollection(tb testing.TB, ctx context.Context, client *mongo.Client, o
...
+			// We don't just write all the documents and Sum at the end, to not
+			// store all documents in memory.
+			currSum := hash.Sum(rawDoc)
+
+			// remove old checksum
+			hash.Reset()
+
+			// write literal document with previous checksum to calculate new checksum
+			if _, err = hash.Write(currSum); err != nil {
+				panic("Unexpected error: " + err.Error())
```
chilagrow, line 325: `require.NoError(tb.err)` is probably more suitable than `panic` in test setup

integration/setup/setup.go
```diff
@@ -228,29 +276,74 @@ func setupCollection(tb testing.TB, ctx context.Context, client *mongo.Client, o
...
 			}
-		})
+
+			insertDocs[i] = doc
+		}
```
chilagrow, line 329: Up to here is related to getting and verifying providers, shouldn't it be in shareddata rather than setup?

integration/shareddata/benchmark.go
```diff
@@ -0,0 +1,49 @@
...
+// limitations under the License.
+
+// Package shareddata provides data for tests and benchmarks.
+package shareddata
+
+import "go.mongodb.org/mongo-driver/bson"
+
+// SimpleBenchmarkValues contain documents with various types of values.
+var SimpleBenchmarkValues BenchmarkValues = NewBenchmarkValues(
+	"Ha59TmC9nwTXqEob73HIwqWUcx2Lp7aengCtNsqjWK0=",
```
chilagrow, line 22: I'm not sure what this hash is for. The `GenerateMixedValues()` generates values so hash should be different right?

chilagrow (commented):
Not finished reviewing it. I don't quite understand the approach yet.

integration/shareddata/benchmark.go
noisersup: ((discussed on meeting but leaving comments to write things on paper))

The hash here is the expected hash of all documents fetched from BenchmarkValues's iterator. There are two reasons behind the checksum:

1. The documents in BenchmarkProvider (which is the interface that is implemented by BenchmarkValues) must be exactly the same every time with the same order. That's why each provider has expected hash of all the documents. Setup function, while fetching the data from provider calculates the actual checksum and compares both expected and actual hash to be sure that we generate data properly.

2. When we compare results of the benchmarks with tooling like `benchstat`, test cases are compared with each other by testname. In the scenario, where we change the documents in provider, we don't want to compare the old results (on old data) with the new results (on new data). That's why we append the checksum of provider to the testname.

integration/setup/setup.go
noisersup: I wasn't 100% sure where to place the checksum but I've noticed that we already do some validation for the normal `Provider` interface inside the setup function.

The problem with extracting the validation logic to the BenchmarkValues, is the fact that we allow developer to introduce different structs that implement `BenchmarkProvider`, which would omit the checksum checking. 

I believe that validating the correct order of benchmark documents is one of the essential parts of the benchmark, and it seems to me that it should be enforced on every struct that implements BenchmarkProvider interface.


The other thing is that implementation would be slightly more complicated, as we would need to wrap iterator of BenchmarkValues into another iterator, that would calculate the checksum on-go. Also, BenchmarkValues would need to do some error handling because of that.

integration/shareddata/benchmark.go
chilagrow: Thanks for this. 
Comment about hash would be helpful, also my personal preference is to call `seed` instead of `generate` :hugs: but maybe like I said, it's personal :see_no_evil:

integration/shareddata/benchmark.go
```diff
@@ -0,0 +1,49 @@
...
+	GenerateMixedValues(),
+)
+
+// GenerateMixedValues returns generator that generates 400 documents with various types of values.
+func GenerateMixedValues() func() bson.D {
```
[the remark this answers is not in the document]
chilagrow: Just random thought, could this function return the seed and hash? so to have the return signature like `(hash string, getSeed func() bson.D)` :thinking: 

Why am I saying that is hash is very specific to this function, so it's a bit disconnected to have it outside it.

integration/setup/setup.go
chilagrow: I see it now. We read the values first time here so we don't want to do it a provider level

chilagrow (commented):
A few more comments thank for the walk through!

integration/shareddata/benchmark.go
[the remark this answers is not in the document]
noisersup: > Generate makes it sounds like it produces random value, what it does is it always produces the same deterministic value. We want to document this. I would rename it something like getSimpleSeed, getBenchmarkSeed or other

Generate function name comes from generator pattern. It doesn't generate a seed, it returns generated document. I'm going to rephrase the comment

> Also it doesn't need to be exported

That's right, thanks :)

> Why am I saying that is hash is very specific to this function, so it's a bit disconnected to have it outside it.

It's used at the setup level to compare it with actual hash, and at the test level to append it to the test name. While writing a benchmark you look at this data from the `BenchmarkProvider` interface perspective, which specifies `Hash()` function.

integration/shareddata/benchmark.go
noisersup: I believe that term seeding is much more commonly used as the starting point of pseudo random values chain. Here we want to make clear that data provided cannot be random.

Taskfile.yml
```diff
@@ -217,8 +217,44 @@ tasks:
...
 
+  bench-pg:
+    desc: "Run benchmarks for `pg` handler"
+    dir: integration
+    cmds:
+      - >
+        go test -bench='Benchmark.*' -run=^#
+        -benchtime={{.BENCHTIME}}
+        -target-backend=ferretdb-pg
+        -target-tls
```
AlekSi, line 228: Let's not benchmark TLS

Taskfile.yml
```diff
@@ -217,8 +217,44 @@ tasks:
         -target-url=mongodb://127.0.0.1:47017/
         -target-backend=mongodb
 
+  bench-pg:
+    desc: "Run benchmarks for `pg` handler"
+    dir: integration
+    cmds:
+      - >
+        go test -bench='Benchmark.*' -run=^#
```
AlekSi, line 225:
```suggestion
        go test -bench=. -run=XXX
```

That more idiomatic

Taskfile.yml
```diff
@@ -217,8 +217,44 @@ tasks:
...
+        -benchtime={{.BENCHTIME}}
+        -target-backend=ferretdb-pg
+        -target-tls
+        -postgresql-url=postgres://username@127.0.0.1:5432/ferretdb
```
AlekSi, line 229:
```suggestion
        -postgresql-url=postgres://username@127.0.0.1:5432/ferretdb
        | tee new-pg.txt
```

To actually generate that file

Taskfile.yml
```diff
@@ -217,8 +217,44 @@ tasks:
...
+      - >
+        go test -bench='Benchmark.*' -run=^#
+        -benchtime={{.BENCHTIME}}
+        -target-backend=mongodb
+        -target-url=mongodb://127.0.0.1:47017/
+      - ../bin/benchstat{{exeExt}} old-mongodb.txt new-mongodb.txt
+
   bench-short:
-    desc: "Benchmark for about 25 seconds (with default BENCHTIME)"
+    desc: "Benchmark internal package for about 25 seconds (with default BENCHTIME)"
```
AlekSi, line 257: Let's rename target to `bench-unit-short` instead

Taskfile.yml
```diff
@@ -217,8 +217,44 @@ tasks:
         -target-url=mongodb://127.0.0.1:47017/
         -target-backend=mongodb
 
+  bench-pg:
```
AlekSi, line 220: Let's move all those new targets to `integration/Taskfile.yml` because that file is already too large

integration/benchmarks_test.go
```diff
@@ -24,50 +25,33 @@ import (
...
+	}{
+		"String": {
+			filter: bson.D{{"v", "foo"}},
+		},
+		"DotNotation": {
+			filter: bson.D{{"v.42", "hello"}},
+		},
+	} {
+		b.Run(fmt.Sprint(name, "_", provider.Hash()), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
```
AlekSi, line 48: `b.ResetTimer()` before loop

integration/shareddata/benchmark.go
AlekSi: I agree with the second reason, but I'm not sure about the first one. Storing expected hashes seems to be overcomplicated. After all, hash changes are immediately noticeable anyway – you run benchmark twice without any changes, and you should get the same hashes

chilagrow: I'm forward with the idea of not storing hash here, and running benchmark twice to output the hash! It would be simpler

chilagrow (commented):
Looks good, I like the idea of not storing hash in the code!

integration/shareddata/benchmark.go
AlekSi: I'm not even sure we want to run the benchmark twice. As long as the data's hash is a part of the benchmark name, we will notice if it changes.

integration/shareddata/shareddata.go
AlekSi: `Provider` for integration tests returns documents in random order by design – to spot problems in our tests that depend on some specific order.

Let's keep those interfaces separate for now. We might merge them later.

chilagrow: true, the randomness can make tests more robust :+1:

integration/shareddata/iterator.go
```diff
@@ -0,0 +1,69 @@
...
+	resource.Track(iter, iter.token)
+
+	return iter
+}
+
+// valuesIterator iterates through bson.D documents generated by provided generator function.
+type valuesIterator struct {
+	generator func() bson.D
+	token     *resource.Token
+	m         sync.Mutex
```
AlekSi, line 45: Mutex should typically be on top, like a hat: https://dmitri.shuralyov.com/idiomatic-go/entries/2

If there is some reason for it being there (alignment, for example), that should be documented

integration/shareddata/benchmark.go
noisersup: > After all, hash changes are immediately noticeable anyway – you run benchmark twice without any changes, and you should get the same hashes

Not sure how is it noticeable. If the hash changed, the new test case with the new hash in name will be created. The benchstat will just notice that the new test case is created and won't compare it to anything. There's no place where we validate that hash is changed or not, and it would need to be checked manually.

integration/benchmarks_test.go
noisersup: That's purely to clear memory allocations? From the timer perspective we don't do any setup inside b.Run()

AlekSi: Yeah, you are right, there is no need for that

integration/shareddata/benchmark.go
AlekSi: So we will check that manually

chilagrow (dismissed):
Looks great!

w84thesun (approved):
LGTM

integration/setup/setup.go
```diff
@@ -228,29 +274,50 @@ func setupCollection(tb testing.TB, ctx context.Context, client *mongo.Client, o
...
+	for {
+		docs, err := iterator.ConsumeValuesN(iter, 10)
+		require.NoError(tb, err)
+
+		if len(docs) == 0 {
+			break
+		}
+
+		// convert []bson.D to []any, as mongodb requires.
+		var insertDocs []any = make([]any, len(docs))
```
AlekSi, line 306:
```suggestion
		insertDocs := make([]any, len(docs))
```

integration/shareddata/iterator.go
```diff
@@ -0,0 +1,113 @@
...
+	if iter.generator == nil {
+		return unused, nil, iterator.ErrIteratorDone
+	}
+
+	doc := iter.generator()
+	if doc == nil {
+		return unused, nil, iterator.ErrIteratorDone
+	}
+
+	rawDoc := []byte(fmt.Sprintf("%x", doc))
```
AlekSi, line 75: Huh? What does that produce? I wonder if could use something like `json.Marshal`

integration/shareddata/iterator.go
```diff
@@ -0,0 +1,113 @@
...
+	// store all documents in memory.
+	currSum := iter.hash.Sum(rawDoc)
+
+	// remove old checksum
+	iter.hash.Reset()
+
+	// write literal document with previous checksum to calculate new checksum
+	if _, err := iter.hash.Write(currSum); err != nil {
+		return unused, doc, err
+	}
```
AlekSi, line 89: That's a very strange way to hash documents _if_ it works. Are you building a blockchain there? :)

All we need to do here is to call `iter.hash.Write(rawDoc)`

integration/shareddata/iterator.go
```diff
@@ -0,0 +1,113 @@
...
+	defer iter.m.Unlock()
+
+	resource.Untrack(iter, iter.token)
+	iter.generator = nil
+}
+
+// Hash returns calculated checksum of all returned by valuesIterator documents.
+// It must be called after closing iterator, otherwise it returns the error.
+func (iter *valuesIterator) Hash() (string, error) {
+	if iter.generator != nil {
```
AlekSi, line 106: That read should be protected by mutex too

integration/shareddata/shareddata.go
```diff
@@ -197,7 +200,50 @@ func (values *Values[idType]) IsCompatible(backend string) bool {
...
+type BenchmarkValues struct {
+	// iter returns all documents in deterministic order.
+	iter *valuesIterator
+
+	// name represents the name of the benchmark values set.
+	name string
+}
+
+// Name implements BenchmarkProvider interface.
+func (b BenchmarkValues) Name() string {
```
AlekSi, line 229:
```suggestion
func (b *BenchmarkValues) Name() string {
```

integration/shareddata/shareddata.go
```diff
@@ -197,7 +200,50 @@ func (values *Values[idType]) IsCompatible(backend string) bool {
...
+	return b.name
+}
+
+// Hash implements BenchmarkProvider interface.
+// It returns actual hash of all documents produced by BenchmarkValues.
+// It will panic if iterator was not closed.
+func (b BenchmarkValues) Hash() string {
```
AlekSi, line 236:
```suggestion
func (b *BenchmarkValues) Hash() string {
```

integration/shareddata/shareddata.go
```diff
@@ -197,7 +200,50 @@ func (values *Values[idType]) IsCompatible(backend string) bool {
...
+	return must.NotFail(b.iter.Hash())
+}
+
+// Docs implements BenchmarkProvider interface.
+func (b BenchmarkValues) Docs() iterator.Interface[struct{}, bson.D] {
```
AlekSi, line 241:
```suggestion
func (b *BenchmarkValues) Docs() iterator.Interface[struct{}, bson.D] {
```

integration/shareddata/shareddata.go
```diff
@@ -197,7 +200,50 @@ func (values *Values[idType]) IsCompatible(backend string) bool {
...
+	return b.iter
+}
+
 // check interfaces
 var (
-	_ Provider = (*Values[string])(nil)
+	_ Provider          = (*Values[string])(nil)
+	_ BenchmarkProvider = (*BenchmarkValues)(nil)
```
AlekSi, line 248: With the current code both `*BenchmarkValues` and `BenchmarkValues` implement that interface. It is better to make more strict

AlekSi (commented):
Blockchain code looks weird

integration/shareddata/iterator.go
noisersup: I did it in a "blockchain" fashion, as it requires constant amount of memory, regardless from document amount. Writing all documents to the hash will result in storing all of them in memory, and as we use iterator here to save some space, I thought that doing that here is also reasonable.

AlekSi: > Writing all documents to the hash will result in storing all of them in memory

Why?.. I'm not sure why you think that, but that's how hashing is typically used: https://go.dev/play/p/5HzWihaoJUP `Write` can be (and normally should be) called multiple times and `Sum` accepts a slice only as performance optimizations in some use cases (to avoid allocation).

integration/shareddata/iterator.go
noisersup: done

integration/shareddata/iterator.go
```diff
@@ -0,0 +1,110 @@
...
+	if doc == nil {
+		return unused, nil, iterator.ErrIteratorDone
+	}
+
+	jsonDoc, err := json.Marshal(doc)
+	if err != nil {
+		return unused, nil, err
+	}
+
+	// write json representation of document to calculate hash later
```
AlekSi, line 80:
```suggestion
```

Because that comment is both obvious and technically incorrect: the hash is calculated after every block; `Sum` method may do nothing if no padding is needed

AlekSi (approved):

PR #2418, opened 2023-04-12
Validate `scale` param for `dbStats` and `collStats` correctly

### LINKED ISSUE #1346: Fix `SchemaStats` to return (somewhat) correct data
### What should be done?

We need to find a way to calculate the data needed by `SchemaStats`. The current query might not be correct. If we can't do it with one query but can with several – let's do that.

It also should work when we have concurrent creations and deletions of databases/schemas and collection/tables (we should receive some data but not PostgreSQL errors). See the test `TestCommandsAdministrationServerStatusStress`.

rumyantseva:
# Description

Closes #1346.

This PR is one of a few PRs for #1346. It is sent separately as it covers logic related to `scale` param validation and doesn't depend on the rest of the issue.

There is a minor refactoring: `errLongExceeded` error is split into `errLongExceededNegative` and `errLongExceededPositive` because, for the params like `skip` or `scale`, in case of negative overflow, we should still return a different error -- the value must be non-negative (for `skip`) or positive (for `scale`). 

See also https://github.com/FerretDB/dance/pull/407 for diff tests.

## Readiness checklist


integration/commands_administration_compat_test.go
```diff
@@ -0,0 +1,178 @@
...
+		altMessage string
+	}{
+		"scaleOne":      {scale: int32(1)},
+		"scaleBig":      {scale: int64(1000)},
+		"scaleMaxInt":   {scale: math.MaxInt64},
+		"scaleZero":     {scale: int32(0), resultType: emptyResult},
+		"scaleNegative": {scale: int32(-100), resultType: emptyResult},
+		"scaleFloat":    {scale: 2.8},
+		"scaleMinFloat": {scale: -math.MaxFloat64, resultType: emptyResult},
+		"scaleMaxFloat": {scale: math.MaxFloat64},
```
chilagrow, line 54: This test is unhappy in CI

```
2023-04-12T23:36:01.2508220Z     --- FAIL: TestCommandsAdministrationCompatDBStatsWithScale/scaleMaxFloat (0.02s)
2023-04-12T23:36:01.2508482Z         commands_administration_compat_test.go:136: 
2023-04-12T23:36:01.2509051Z             	Error Trace:	/home/runner/work/FerretDB/FerretDB/integration/commands_administration_compat_test.go:166
2023-04-12T23:36:01.2509311Z             	Error:      	Received unexpected error:
2023-04-12T23:36:01.2509609Z             	            	scale has to be > 0
2023-04-12T23:36:01.2510060Z             	Test:       	TestCommandsAdministrationCompatDBStatsWithScale/scaleMaxFloat
```

internal/handlers/common/scale.go
```diff
@@ -0,0 +1,86 @@
...
+	switch {
+	case errors.Is(err, errUnexpectedType):
+		if _, ok := value.(types.NullType); ok {
+			return 1, nil
+		}
+
+		return 0, commonerrors.NewCommandErrorMsgWithArgument(
+			commonerrors.ErrTypeMismatch,
+			fmt.Sprintf(
+				`BSON field '%s.scale' is the wrong type '%s', expected types '[long, int, decimal, double]'`,
```
chilagrow, line 58: So `command` can actually be not command, it can be `$collStats (stage)`. Which would make the formatting of `%s.scale` weird. I suggest passing `command` and another parameter something like `argument` or better name to store a value like `$collStats (stage)`.

internal/handlers/common/scale.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/FerretDB/FerretDB/internal/handlers/commonerrors"
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+)
+
+// GetScaleParam validates the given scale value for collStats and dbStats command.
+//
+// If the value is valid, it returns its int32 representation,
+// otherwise it returns a command error with the given command being mentioned.
+func GetScaleParam(command string, value any) (int32, error) {
```
chilagrow, line 31: We should try to use `NewCommandErrorMsgWithArgument` instead of `NewCommandError` where we can. It helps us collect stats :hugs:

chilagrow (commented):
I'm seeing that there will be other PRs, so the change I suggest can be done there too.

The failing one needs attention though :hugs:

integration/commands_administration_compat_test.go
rumyantseva: Oh, that's interesting :) It passes for me locally, maybe because of a different architecture.
I will remove this case for now (like I did with other cases), because the error message (and error code 0) that Mongo returns, doesn't make any sense 😅

internal/handlers/common/scale.go
rumyantseva: Mongo itself returns something like `$collStats.storageStats.scale` for this case, so let's do the same.

internal/handlers/common/scale.go
rumyantseva: Fixed

integration/commands_administration_compat_test.go
AlekSi: Please never remove test cases for found problems to make tests pass

rumyantseva: Ok, I'll add diff.md changes and dance tests then.

AlekSi: Or we could just skip them with an issue number

rumyantseva: But the current issue is _the_ issue number to have a proper fix :)

AlekSi: I believe we have an issue for architecture-specific float64 problems

AlekSi: #491. Please make it pass on CI and skip it on arm64. Search for `runtime.GOARCH == "arm64"`

rumyantseva: Well, this problem is also about Mongo returning 0 error code instead of `BadValue`.

AlekSi: That's not new too. See `getParameter` implementation, for example

Fashander (dismissed):

tools/checkswitch/testdata/testdata.go
```diff
@@ -16,9 +16,8 @@
 package testdata
 
 import (
-	"time"
```
w84thesun, line 19: I guess that was changed accidentally. 😃

w84thesun (approved):
LGTM, let's fix the import and merge this one!

tools/checkswitch/testdata/testdata.go
rumyantseva: Oh, that's my IDE :/ I turned off all the autoformatting but it keeps changing this file :/

website/docs/diff.md
```diff
@@ -29,6 +29,8 @@ slug: /diff/ # referenced in README.md and beacon
    * database name cannot contain capital letters;
 9. For Tigris, FerretDB requires Tigris schema validation for `create` command: validator must be set as `$tigrisSchemaString`.
     The value must be a JSON string representing JSON schema in [Tigris format](https://docs.tigrisdata.com/overview/schema).
+10. FerretDB offers the same validation rules for the `scale` parameter in both the `collStats` and `dbStats` commands.
+    If an invalid `scale` value is provided in the `dbStats` command, the same error codes will be triggered as with the `collStats` command.
```
noisersup, line 33: Should we start to document such differences like that every time when we spot them? It might be pretty common for us to use shared function for things that behave the same but return different code on mongo

noisersup (approved):
LGTM! One question about our process.

integration/commands_administration_compat_test.go
rumyantseva: Added a diff doc and a PR: https://github.com/FerretDB/dance/pull/407

Fashander (dismissed):

internal/handlers/common/params.go
```diff
@@ -363,11 +369,11 @@ func addNumbers(v1, v2 any) (any, error) {
 		case int64:
 			if v2 > 0 {
 				if int64(v1) > math.MaxInt64-v2 {
-					return nil, errLongExceeded
+					return nil, errLongExceededPositive
 				}
 			} else {
 				if int64(v1) < math.MinInt64-v2 {
-					return nil, errLongExceeded
+					return nil, errLongExceededPositive
```
chilagrow, line 376: Missed this first way around but I think it is `errLongExceededNegative`. But this doesn't really have any implication for addNumbers, since whichever error produce the same error message+code. Anyway something to address in upcoming PR.

internal/handlers/common/params.go
```diff
@@ -394,11 +400,11 @@ func addNumbers(v1, v2 any) (any, error) {
 		case int64:
 			if v2 > 0 {
 				if v1 > math.MaxInt64-v2 {
-					return nil, errLongExceeded
+					return nil, errLongExceededPositive
 				}
 			} else {
 				if v1 < math.MinInt64-v2 {
-					return nil, errLongExceeded
+					return nil, errLongExceededPositive
```
chilagrow, line 407: This too is  `errLongExceededNegative`, I think

chilagrow (approved):
looks good to me!

internal/handlers/common/params.go
rumyantseva: Nice catch! Let me fix it in this PR!

w84thesun (approved):
👍

noisersup (dismissed):
LGTM! Thanks for splitting the issue

chilagrow (approved):
awesome work!

internal/handlers/pg/msg_dbstats.go
```diff
@@ -36,17 +36,20 @@ func (h *Handler) MsgDBStats(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+	scale := int32(1)
 
-	scale, err = common.GetOptionalPositiveNumber(document, "scale")
-	if err != nil || scale == 0 {
-		scale = 1
+	var s any
+	if s, err = document.Get("scale"); err == nil {
+		if scale, err = common.GetScaleParam(command, s); err != nil {
+			return nil, err
+		}
```
rumyantseva, line 52: Covered in dance: https://github.com/FerretDB/dance/pull/407

w84thesun (approved):
👍

noisersup (approved):

chilagrow (approved):

internal/handlers/common/params.go
```diff
@@ -121,7 +121,8 @@ var (
 	errNotBinaryMask         = fmt.Errorf("not a binary mask")
 	errUnexpectedLeftOpType  = fmt.Errorf("unexpected left operand type")
 	errUnexpectedRightOpType = fmt.Errorf("unexpected right operand type")
-	errLongExceeded          = fmt.Errorf("long exceeded")
+	errLongExceededPositive  = fmt.Errorf("long exceeded - positive value")
+	errLongExceededNegative  = fmt.Errorf("long exceeded - negative value")
```
AlekSi, line 125: >  I added/updated comments and checked rendering.

> Before submitting a pull request, please make sure that:
> […]
> 2. Comments are added or updated for all new or changed code.
>    Please add missing comments for all (both exported and unexported)
>    new and changed top-level declarations (functions, types, etc).

Both points were ignored.

Updated possible error results are also not documented.

AlekSi (approved):
LGTM for docs, but please add and update comments in the next PR

internal/handlers/common/params.go
rumyantseva: @AlekSi could you please leave more specific feedback? What exactly needs to be documented here?
Sometimes when I leave comments (e.g. for structure fields) you say that they are obvious and not needed. What needs to be documented here?

rumyantseva: Should I document what each error in this list means?
Or should I document each function/method where these errors could be returned?
Or both? Or something else?

PR #2482, opened 2023-04-19
Add benchmark for replacing large document

### LINKED ISSUE #2378: Introduce integration-level benchmarks

noisersup:
# Description

Closes #2378.

## Readiness checklist


integration/Taskfile.yml
```diff
@@ -4,6 +4,7 @@ version: 3
 
 vars:
   BENCHTIME: 5s
+  BENCHNAME: '{{default "." .BENCHNAME}}'
```
noisersup, line 7: `env BENCHNAME=benchmarkname task bench-pg` to run specific benchmark

integration/shareddata/benchmark.go
```diff
@@ -15,14 +15,24 @@
...
+	"go.mongodb.org/mongo-driver/bson"
+)
 
 // MixedBenchmarkValues contain documents with various types of values.
 var MixedBenchmarkValues = BenchmarkValues{
 	name: "MixedBenchmarkValues",
 	iter: newValuesIterator(generateMixedValues()),
 }
 
+// LargeDocumentBenchmarkValues contains a signle large document with various types of values.
```
chilagrow, line 30:
```suggestion
// LargeDocumentBenchmarkValues contains a single large document with various types of values.
```

integration/shareddata/benchmark.go
```diff
@@ -48,3 +58,37 @@ func generateMixedValues() func() bson.D {
 
 	return gen
 }
+
+// generateLargeDocument returns generator that generates single deterministic document
+// with 200 elements that contain simple strings, doubles and objects.
```
chilagrow, line 63: Maybe not yet, I'm seeing `generateLargeDocument` and `generateLargeDocument` are very similar. It uses the same values to produce the documents, the subtle difference is the field key, `generateLargeDocument`  produces `{_id: "0", "v":<<val1>>}` and `generateLargeDocument` produces  `{_id: "0", "0":<<val1>>, "1":<<val2>>,...}`. When we add more generator we can think of merging those two functions and just have one with argument like field length and document length. But okay for now.

integration/benchmarks_test.go
```diff
@@ -55,3 +55,31 @@ func BenchmarkQuery(b *testing.B) {
...
+	filter := bson.D{{"_id", 0}}
+	runsCount := 1
+
+	b.Run(provider.Hash(), func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			var doc bson.D
+			err := coll.FindOne(ctx, filter).Decode(&doc)
+			require.NoError(b, err)
+
+			doc[runsCount].Value = i * 11111
```
chilagrow, line 76: Not quite sure what is `runsCount`. It will always be 1 right? Because `runsCount++` is outside for loop.

What is the significance of `doc["1"].Value`? :thinking: Is it the first non _id field value?

integration/shareddata/benchmark.go
```diff
@@ -48,3 +58,37 @@ func generateMixedValues() func() bson.D {
...
+	gen := func() bson.D {
+		if done {
+			return nil
+		}
+
+		done = true
+
+		return doc
+	}
+
```
w84thesun, line 92: What is the purpose of that part of the code?

w84thesun (commented):
Nice one! Asked a question about part of `generateLargeDocument`.

integration/benchmarks_test.go
noisersup: > Not quite sure what is `runsCount`. It will always be 1 right?

testing.Benchmark will run the benchmark multiple times, each time with increased b.N number based on the execution time. We count those runs and use them to update next and next fields inside the document, that's because we want to make sure that we actually replace the data (to satisfy `require.Equal(b, int64(1), updateRes.ModifiedCount)`)

integration/shareddata/benchmark.go
noisersup: To stop iterator (i.e. make it return `errIteratorDone`) the generator must to return `nil`. It may look weird on a single document generator, but it makes sense when generator returns 400 of documents.

integration/shareddata/benchmark.go
noisersup: Good point! I believe we could extract generator into separate interface and made them more general with some init functions like the one from your example!

integration/benchmarks_test.go
chilagrow: I see, you are using that to generate random value to modify the document.

chilagrow (approved):
Looks good to me!

rumyantseva (approved):
LGTM

PR #2484, opened 2023-04-19
Implement projection fields assignment

### LINKED ISSUE #2446: Support field projections assignment
### What should be done?

We should support fields assignment with projection as shown in the example below.

```js
db.values.find({}, { v: '123'})
[
  { _id: 'datetime', v: '123' },
  { _id: 'int64-max', v: '123' },
  { _id: 'double-2', v: '123' },
  { _id: 'string-whole', v: '123' },
  { _id: 'binary', v: '123' },
  { _id: 'regex', v: '123' },
  { _id: 'int32-zero', v: '123' },
  { _id: 'double-max', v: '123' },
]
```

### Where?

internal/handlers/common/projection.go

### Definition of Done

- all handlers updated;
- unit tests added/updated;
- spot refactorings done;
- user documentation updated;
- something else?

w84thesun:
# Description

Closes #2446.

## Readiness checklist


integration/query_projection_compat_test.go
```diff
@@ -100,6 +102,28 @@ func TestQueryProjectionCompat(t *testing.T) {
...
+		},
+		"Assign2FieldsIncludeID": {
+			filter:     bson.D{},
+			projection: bson.D{{"_id", true}, {"foo", nil}, {"bar", "qux"}},
+		},
+		"Assign1FieldExcludeID": {
+			filter:      bson.D{},
+			projection:  bson.D{{"_id", false}, {"foo", primitive.Regex{Pattern: "^fo"}}},
+			skipIDCheck: true,
+		},
```
chilagrow, line 126: Just noticed we don't have test case for testing exclusion and inclusion both. I think below two should error but give different error message. Could you add them please?
```
"Exclusion": {
			filter:     bson.D{},
			projection: bson.D{{"foo", false}, {"bar", true}},
		},
```

```
"Inclusion": {
			filter:     bson.D{},
			projection: bson.D{{"foo", true}, {"bar", false}},
		},
```

chilagrow (commented):
Just a single comment while looking at tests :hugs:

integration/query_projection_compat_test.go
chilagrow: Perfect thanks!

noisersup (approved):
:+1:

PR #2485, opened 2023-04-20
Add a YAML formatter

### LINKED ISSUE #1301: Use YAML formatter
### What should be done?

We already have formatters for the Go source and Markdown files. Let's add another one for all YAML files.

Preferably, it should support our existing style as closely as possible. Preferably, it should be written in Go; a Node.js tool inside a Docker container is the second best option.

### Where?

https://github.com/FerretDB/FerretDB/blob/main/Taskfile.yml

wqhhust:
# Description

Closes #1301 .

## Readiness checklist


Taskfile.yml
```diff
@@ -146,6 +146,13 @@ tasks:
       - bin/goimports{{exeExt}} -format-only -local=github.com/FerretDB/FerretDB -w .
       - bin/gofumpt{{exeExt}} -w .
 
+  fmt-yaml:
+    desc: "Format yaml"
+    cmds:
+      - find . -name "*.yml" -exec chmod o+w {} \;
+      - docker run --rm -v $(pwd):/work tmknom/prettier --write --parser=yaml "./**/*.yml"
```
rumyantseva, line 153: I like how this formatter works! And I like that you added a docker-compose configuration for it, so it can be used the same way how other tasks in this file.

For this command itself, let's also use `docker compose` with the configuration you provided, not just `docker`?

rumyantseva (changes requested):
The formatter looks good to me, but I suggest we use `docker compose` for the task, so it's consistent with the other tasks and uses the same dedicated configuration. 

I also changed the title of the PR, so it contains a brief description of changes for the changelog.

Taskfile.yml
wqhhust: updated to use docker compose.

Taskfile.yml
```diff
@@ -153,6 +153,13 @@ tasks:
       - bin/goimports{{exeExt}} -format-only -local=github.com/FerretDB/FerretDB -w .
       - bin/gofumpt{{exeExt}} -w .
 
+  fmt-yaml:
+    desc: "Format yaml"
+    cmds:
+      - find . -name "*.yml" -exec chmod o+w {} \;
+      - docker compose run --rm -v $(pwd):/work prettier --write --parser=yaml "./**/*.yml"
+      - find . -name "*.yml" -exec chmod o-w {} \;
```
rumyantseva, line 161: Thank you a lot for fixing this so quickly!

My last concern is about those `find` commands. We try to make Taskfile more or less OS-independent, so it can be called on Linux, Mac and Windows. With `find` I'm not sure if it would work on Windows the same way.

I understand your idea to set/unset writing permission, but I tried running it locally without those `find` commands, and still, the formatter was able to modify my files. Are there other cases when such permissions must be set?

wqhhust: Thanks for your feedback. I am running docker under ubuntu, and without using find I got the following error.
image

On which env did you test successfully even without using find? on Window?

wqhhust: How about write a go program to change the permission?

rumyantseva: Thanks a lot for the explanation! I don't think we need a separate go program, but we can use some power of `Taskfile`!

On Mac, Windows, and Linux I tried something like

```
      - >
        {{if eq OS "linux"}}find . -name "*.yml" -exec chmod o+w {} \;{{end}}
      - docker compose run --rm -v $(pwd):/work prettier --write --parser=yaml "./**/*.yml"
      - >
        {{if eq OS "linux"}}find . -name "*.yml" -exec chmod o-w {} \;{{end}}
```

and it worked! Only linux needs permissions to be changed (windows and mac work as is). And the `{{if eq OS "linux"}}...{{end}}` worked well on my ubuntu. 

So, I approve your PR, feel free to either include these statements, or alternatively I can include them in a separate PR.

rumyantseva (approved):
Thanks a lot for contributing!

w84thesun (approved):
LGTM!

chilagrow (approved):
Thanks for your contribution!

PR #2488, opened 2023-04-20
Document `createIndexes`, `listIndexes`, and `dropIndexes` commands

### LINKED ISSUE #2271: Document `createIndexes` command
### What should be done?

Things to mention:

- `createIndexes` called for a non-existent collection will create the collection, and the given indexes;
- `createIndexes` called for a non-existent field creates index for this field;
- attempt to call `createIndexes` for an index that already exists (when the name and the key are equal) will return index name but will not create a duplicate index.
- attempt to call `createIndexes` for same name and different keys or same keys but different names will return an error.

Fashander:
# Description

Closes #2271.

## Readiness checklist


website/docs/indexes.md
```diff
@@ -0,0 +1,126 @@
...
+  },
+  ok: 1
+}
+```
+
+## How to drop indexes
+
+You can also drop all the indexes or a particular index in a specified collection, except the default index (`_id`).
+
+FerretDB supports the use of `dropIndex()` and  `dropIndexes() command.
```
chilagrow, line 101:
```suggestion
FerretDB supports the use of `dropIndex()` and `dropIndexes()` command.
```

website/docs/indexes.md
```diff
@@ -0,0 +1,126 @@
+---
+sidebar_position: 7
+---
+
+# Indexes
+
+Indexes are essential in improving query performance by enabling fast retrieval of relevant documents when querying large collections.
+
+Indexes in FerretDB are created based on a specific field(s) within documents.
+When an index is created, this creates a data structure that maps between the index values and the document locations, making it possible to retrieve documents more quickly based on those fields.
```
chilagrow, line 10: `this creates a data structure that maps between the index values and the document locations`
This sounds a bit vague and specific at the same time, it doesn't give much information to the users though.

website/docs/indexes.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+Suppose you want to find all products with a price greater than `$900`.
+Without an index, each query will need to scan each document in the collection to find the ones that match, which can be resource intensive.
+
+However, if there’s an index on the `price` field, the query only scans two documents where their prices are greater than $900, instead of all four.
+
+## How to create indexes
+
+Use the create use the createIndexes() command to create an index on a collection.
```
chilagrow, line 30: Could we also use quotes in other places where we refer a function please?
```suggestion
Use the create use the `createIndexes()` command to create an index on a collection.
```

website/docs/indexes.md
```diff
@@ -0,0 +1,126 @@
...
+Below is an example of a compound index that uses `price` and `category` fields from the `products` collection as the index key.
+
+```js
+db.products.createIndex({ price: 1, category: 1 })
+```
+
+:::note
+
+* If the createIndex() command is called for a non-existent collection, it will create the collection and its given indexes.
+* If the createIndex() command is called for a non-existent field, only an index for the field is created.
```
chilagrow, line 59: This sounds a bit strange. Does this mean that it creates an index but does not create a field right?

website/docs/indexes.md
```diff
@@ -0,0 +1,126 @@
...
+* if you attempt to create an index with the same name and key as an existing index, the system will not create a duplicate index.
```
chilagrow, line 60:
```suggestion
* If you attempt to create an index with the same name and key as an existing index, the system will not create a duplicate index.
```

website/docs/indexes.md
chilagrow: @Fashander any thoughts about this comment?

website/docs/indexes.md
```diff
@@ -0,0 +1,126 @@
...
+  },
+  ok: 1
+}
+```
+
+## How to drop indexes
+
+You can also drop all the indexes or a particular index in a specified collection, except the default index (`_id`).
+
+FerretDB supports the use of `dropIndex()` and `dropIndexes() command.
```
chilagrow, line 101: This formatting need to be fixed .
```suggestion
FerretDB supports the use of `dropIndex()` and `dropIndexes()` command.
```

website/docs/indexes.md
Fashander: I've updated the sentence to be a bit clearer. Although maybe some graphical representation will be better in explaining it

website/docs/indexes.md
```diff
@@ -0,0 +1,126 @@
...
+  },
+  ok: 1
+}
+```
+
+## How to drop indexes
+
+You can also drop all the indexes or a particular index in a specified collection, except the default index (`_id`).
+
+FerretDB supports the use of `dropIndex()` and `dropIndexes()` command.
```
rumyantseva, line 101:
```suggestion
FerretDB supports the use of`dropIndexes` command.
```

There is only one command - `dropIndexes`. The rest is either wrappers or handlers of particular clients or dirvers.

website/docs/indexes.md
```diff
@@ -0,0 +1,126 @@
...
+
+Use the `dropIndex()` command to drop a particular index from a collection.
```
rumyantseva, line 103:
```suggestion
For example, use the mongosh `dropIndex()` method to drop a particular index from a collection.
```

website/docs/indexes.md
```diff
@@ -0,0 +1,126 @@
...
+For the same example above, you can rewrite it as:
+
+```js
+db.products.dropIndex( { "price" : 1 } )
+```
+
+Use the `dropIndexes()` command to remove one or more indexes from the collection.
+
+You can remove all the indexes from the collection besides the default index (`_id`).
+The following command removes all indexes from the collection.
```
rumyantseva, line 120: I don't think so :) If it does so, it's a bug on our side. One should specify the indexes or set `*` to remove all indexes.

rumyantseva (changes requested):
Suggested a few things and asked a few questions.

There is one thing to clarify: commands and methods/wrappers/handers are not the same.
Commands are on the server (FerretDB) side, and methods/wrappers/handers are on the client/driver side.
There could be a few methods to call the same command. For example, both `dropIndex()` and `dropIndexes()` methods call `dropIndexes()` command.

I find this mixture of methods and commands a bit inconsistent, but probably this is how we use in the other parts of our docs too...

website/docs/indexes.md
Fashander: Maybe there's a bug here because the commands drops all the indexes, although I think it works just as mentioned here: https://www.mongodb.com/docs/manual/reference/method/db.collection.dropIndexes/#mongodb-method-db.collection.dropIndexes

```
> db.products.getIndexes()
[
  { v: 2, key: { _id: 1 }, name: '_id_' },
  { v: 2, key: { price: 1 }, name: 'price_1' },
  { v: 2, key: { price: -1 }, name: 'price_-1' }
]
> db.products.dropIndexes()
{
  nIndexesWas: 3,
  msg: 'non-_id indexes dropped for collection',
  ok: 1
}
> db.products.getIndexes()
[ { v: 2, key: { _id: 1 }, name: '_id_' } ]
```

rumyantseva: @Fashander, what we mix is what we call "command" (server side) and "mongosh method" (client side). The code you show it a mongosh method that calls the command under the hood. The question is what we document. I believe we should document the command (server side), and then we should document how `*` works.

Fashander: I've made a few changes to this. PTAL

chilagrow (approved):

Fashander:
@AlekSi PTAL

AlekSi (approved):

PR #2496, opened 2023-04-21
New version notification uses `update_available` from telemetry

### LINKED ISSUE #1751: Test notifying about newer versions
### What should be done?

We should test if we display a prompt about newer version in `startupWarning` and prometheus metrics. We should check if it's shown if:
- the local version is not equal do latest;
- telemetry is enabled;
- it works for both tigris and pg handler.

Useful links:
- https://github.com/FerretDB/FerretDB/pull/1738#discussion_r1061407184
- https://github.com/FerretzB/FerretDB/issues/1443

### Where?

Maybe [here](https://github.com/FerretDB/FerretDB/blob/3e28163606eedc26d5a4dd9e1213671577d672b1/integration/commands_diagnostic_test.go#L149-L156), but it can be unit/compat or integration test, though please take a note that MongoDB doesn't display such an information.

### Definition of Done

- unit/compat or integration tests for both pg and tigris handlers added.

chilagrow:
# Description

Closes #1751.

## Readiness checklist


internal/util/state/state.go
```diff
@@ -21,14 +21,14 @@ import (
...
 	"github.com/FerretDB/FerretDB/internal/util/must"
 )
 
 // State represents FerretDB process state.
 type State struct {
-	UUID      string `json:"uuid"`
-	Telemetry *bool  `json:"telemetry,omitempty"` // nil for undecided
+	UUID              string `json:"uuid"`
+	Telemetry         *bool  `json:"telemetry,omitempty"` // nil for undecided
+	IsUpdateAvailable bool   `json:"update_available"`
```
chilagrow, line 31: Got this from beacon https://github.com/FerretDB/beacon/blob/main/internal/handler.go#L58, maybe not correct.

internal/util/telemetry/reporter.go
```diff
@@ -61,7 +61,8 @@ type request struct {
 
 // response represents telemetry response.
 type response struct {
-	LatestVersion string `json:"latest_version"`
+	LatestVersion   string `json:"latest_version"`
+	UpdateAvailable bool   `json:"update_available"`
```
chilagrow, line 65: From beacon https://github.com/FerretDB/beacon/blob/main/internal/handler.go#L58.

w84thesun (approved):
LGTM 👍

internal/util/telemetry/reporter_test.go
```diff
@@ -88,3 +93,62 @@ func TestNewReporterLock(t *testing.T) {
...
+				L:             zap.L(),
+				ReportTimeout: 1 * time.Minute,
+			}
+
+			r, err := NewReporter(&opts)
+			assert.NoError(t, err)
+
+			r.report(testutil.Ctx(t))
+
+			s := r.P.Get()
```
rumyantseva, line 148: This is the part I don't understand. I see `r.report` is called above, and the info on wether the update is available is identified as a part of `r.report`. Beacon returns `updateAvailable` only if `POST` was requested - https://github.com/FerretDB/beacon/blob/main/internal/handler.go#L156. With that, I don't understand what do we expect with this `r.P.Get()` as Beacon won't return `updateAvailable` for it.

internal/util/telemetry/reporter_test.go
```diff
@@ -88,3 +93,62 @@ func TestNewReporterLock(t *testing.T) {
...
+		},
+	} {
+		name, tc := name, tc
+
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			// use httptest.NewServer so http request on ts.URL returns tc.reporterResponse body
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusCreated)
```
rumyantseva, line 126: This is not a very accurate mock of becaon as beacon have a different logic for `GET` and `POST` methods - https://github.com/FerretDB/beacon/blob/e965bc7609649e31cca83783f95f32861afb38ac/internal/handler.go#L134 (the responses are very different for `GET` and `POST`.

rumyantseva (changes requested):
The implementation looks good to me, but I don't see that the test checks the necessary logic.

internal/util/telemetry/reporter_test.go
chilagrow: `r.report()` either calls `POST` or it doesn't call telemetry at all. `GET` is not called by `r.report` I guess it's used by something else. When `POST is called, it updates the state of the provider by `r.P.Update` if any update is available.

To check if the state is updated by `r.report()`, we use provider to get value `r.P.Get()`.

When beacon doesn't call `POST` indeed there won't be any update available.

Added comments in the code, let me know if that clarifies :hugs:

internal/util/telemetry/reporter_test.go
chilagrow: `r.report()` makes http `POST` request, so it does not call `GET` even though on API it's available. I added a check for request method type though :hugs:

internal/util/telemetry/reporter_test.go
chilagrow: Meanwhile, I can update the reporter to return the state `(r *Reporter) report(ctx context.Context) *state.State` which removes the need to call provider from test.

rumyantseva: Ah, ok! I think I now understand :)

rumyantseva (approved):
LGTM

w84thesun (approved):
:shipit:

internal/util/telemetry/reporter_test.go
noisersup: This one looks resolved, so I'm closing it

noisersup (approved):
LGTM!

AlekSi:
> I ensured that PR title is good enough for the changelog.

Please check what we have in the changelog now: https://github.com/FerretDB/FerretDB/blob/main/CHANGELOG.md

chilagrow:
> > I ensured that PR title is good enough for the changelog.
> 
> Please check what we have in the changelog now: https://github.com/FerretDB/FerretDB/blob/main/CHANGELOG.md

It can be better :sweat:

PR #2525, opened 2023-04-27
Add identifying Arch Linux version in `hostInfo` command

### LINKED ISSUE #2500: Improve how we identify Linux version in `hostInfo` command
### What should be done?

In `hostInfo` we identify Linux' release by checking `/etc/os-release` file and parsing the version from it: https://github.com/FerretDB/FerretDB/blob/b1cc4e8b1138805a9d16fafb4ff880dc73d4e6eb/internal/handlers/common/msg_hostinfo.go#L47

However, not each Linux distro contains the version information in this file. As the result, this check https://github.com/FerretDB/FerretDB/blob/b1cc4e8b1138805a9d16fafb4ff880dc73d4e6eb/integration/commands_diagnostic_test.go#L227 will fail for such distros.

Let's investigate what's possible for such distros - is there a way to get the info about the version, should we set some "default value", or should we allow an empty version to be set (and then we need to fix the test).

Many thanks to @raeidish for reporting this problem with `Arch Linux` (see also https://gist.github.com/natefoo/814c5bf936922dad97ff).

kropidlowsky:
# Description

Closes #2500.

Add identifying Arch Linux version in `hostInfo` command.
Most of Linux distributions contain `/etc/os-release` - the file used by FerretDB for `hostInfo` command.
However, Arch has `/etc/arch-release` instead of it. To keep functionality for most of Linux distributions the special Arch condition has been added.

## Readiness checklist

### DIFF
--- a/internal/handlers/common/msg_hostinfo.go
+++ b/internal/handlers/common/msg_hostinfo.go
@@ -40,7 +40,10 @@ func MsgHostInfo(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
 	if runtime.GOOS == "linux" {
 		file, err := os.Open("/etc/os-release")
 		if err != nil {
-			return nil, lazyerrors.Error(err)
+			file, err = os.Open("/etc/arch-release")
+			if err != nil {
+				return nil, lazyerrors.Error(err)
+			}
 		}
 		defer file.Close()
 


rumyantseva (approved):
Looks good to me, thanks a lot for contributing!

w84thesun (approved):
Nice one!

PR #2532, opened 2023-04-29
Perform `insertMany` in a single transaction

### LINKED ISSUE #2510: Perform `insertMany` in a single transaction
### What should be done?

Currently we perform insertion of multiple documents by calling `insertDocument` function multiple times. The function creates transaction for each document.

Let's try to perform all of the inserts in a single transaction. That might remove some overhead related to the transaction and potentially decrease the insert time.

Merging everything into single transaction will break compatibility, as on error - PostgreSQL will "undo" all previous inserts, while MongoDB won't. In such scenarios we should perform insertions once again, but in separate transactions.

Let's start by introducing simple insertion benchmark in separate PR. It will allow us to check if there's any performance gain after second PR with the actual change.

## Where
`pgdb` package

## DoD
- [x] Benchmark for `insertMany` created
- [x] Inserts of many documents are performed in a single transaction
- [x] If any of documents failed on insert, insertion is performed once again in separate transactions
- [x] Benchmarks results are compared.

raeidish:
# Description

Closes #2510 

## Readiness checklist


internal/handlers/pg/msg_insert.go
```diff
@@ -95,6 +95,33 @@ func (h *Handler) MsgInsert(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+
+	dbPool.InTransaction(ctx, func(tx pgx.Tx) error {
+        var err error
+        for i := 0; i < docs.Len(); i++ {
+            doc := must.NotFail(docs.Get(i))
+            err = insertDocument(ctx, dbPool, qp, doc)
+            var we *commonerrors.WriteErrors
+            switch {
+            case err == nil:
+                inserted++
```
rumyantseva, line 110: In principle, we don't need to calculate it: as it's a transition, we insert all or nothing.

raeidish: Yes that is true, i will discard it in the single transaction run

noisersup:
Is there anything I can help with, or is this one ready for review?

AlekSi:
(more specifically, we are not sure due to PR title 😄)

raeidish:
Aha! Sorry I’m a bit busy at the moment, the pr is ready for review. There was one suggestion where I would revert the changes to find and modify and update calls. This is because of the update function might return incorrect error messages

noisersup (approved):
Looks great! I've tested it locally and there benchmark results have improved.

Issue #2295, opened 2023-03-27
Data is not updated in PostgreSQL from first attempt but second attempt works fine


pavel-odintsov:
### Versions
```
ferretdb --version
version: v0.9.4-rc.2-24-gb7bd32f
commit: b7bd32f20537bd8c92313236e0f96171d9c3a2a2
branch: unknown
dirty: true
package: deb
debugBuild: false
```

### What did you do?

I've tried using FastNetMon with FerretDB integration and I've tried adding new network from Go app:
```
sudo fcli set main networks_list 252.0.0.1/24
```

It clearly succeeded and I have FerretDB debug log:
```
INFO        listener        clientconn/listener.go:305        Connection started        {"conn": "127.0.0.1:41526 -> 127.0.0.1:27017"}
INFO        listener        clientconn/listener.go:305        Connection started        {"conn": "127.0.0.1:41524 -> 127.0.0.1:27017"}
INFO        listener        clientconn/listener.go:305        Connection started        {"conn": "127.0.0.1:41536 -> 127.0.0.1:27017"}
INFO        pgdb        v4@v4.18.1/conn.go:354        Exec        {"sql": "begin", "args": [], "time": "257.314µs", "commandTag": "BEGIN", "pid": 123863}
INFO        pgdb        v4@v4.18.1/conn.go:354        Query        {"pid": 123863, "sql": "SELECT EXISTS ( SELECT 1 FROM information_schema.columns WHERE table_schema = $1 AND table_name = $2 )", "args": ["fastnetmon","_ferretdb_database_metadata"], "time": "606.367µs", "rowCount": 1}
INFO        pgdb        v4@v4.18.1/conn.go:354        Query        {"time": "598.683µs", "rowCount": 1, "pid": 123863, "sql": "SELECT _jsonb  FROM \"fastnetmon\".\"_ferretdb_database_metadata\" WHERE _jsonb->$1 @> $2", "args": ["_id","\"configuration\""]}
INFO        pgdb        v4@v4.18.1/conn.go:354        Query        {"pid": 123863, "sql": "SELECT _jsonb  FROM \"fastnetmon\".\"configuration_b2c3d7b9\"", "args": [], "time": "8.95503ms", "rowCount": 2}
INFO        pgdb        v4@v4.18.1/conn.go:354        Exec        {"sql": "commit", "args": [], "time": "422.125µs", "commandTag": "COMMIT", "pid": 123863}
[...]

PR #2537, opened 2023-05-01
Refactor integration benchmarks

AlekSi:
# Description

Closes #{issue_number}.

## Readiness checklist


integration/shareddata/benchmarks.go
```diff
@@ -0,0 +1,107 @@
...
+			{"id", int32(i)},
+			{"v", values[i%l]},
+		}
+		i++
+		return doc
+	}
+})
+
+// BenchmarkLargeDocuments provides a single large document with fields of various types.
+var BenchmarkLargeDocuments = newGeneratorBenchmarkProvider("LargeDocuments", 123, func(n int) generatorFunc {
```
AlekSi, line 57: We should make this one much better. For example, we could generate something similar to https://github.com/FerretDB/FerretDB/issues/2295#issuecomment-1488308844

integration/shareddata/benchmark_provider.go
```diff
@@ -0,0 +1,133 @@
...
+
+// generatorFunc is a function that returns the next generated bson.D document, or nil.
+//
+// The order of documents returned must be deterministic.
+type generatorFunc func() bson.D
+
+// newGeneratorFunc returns a new generatorFunc that return n documents.
+//
+// All returned functions should be independent from each other, but return the same documents in the same order.
+type newGeneratorFunc func(n int) generatorFunc
```
noisersup, line 87: `newGeneratorFunc` sounds more like an actual function and less than type, I would suggest to call it `generatorFuncConstructor` or something like that.

integration/shareddata/benchmark_provider.go
```diff
@@ -0,0 +1,133 @@
...
+	Name() string
+
+	// NewIterator returns a new iterator for documents in the same order.
+	NewIterator() iterator.Interface[struct{}, bson.D]
+}
+
+// hashBenchmarkProvider checks that BenchmarkProvider's NewIterator methods returns a new iterator
+// for the same documents in the same order,
+// and returns a hash of those documents that could be used as a part of benchmark name.
+func hashBenchmarkProvider(bp BenchmarkProvider) string {
```
noisersup, line 42: So we moved from calculating hash of exact documents that are inserted to the collection, to calculating hash before inserting them. I see some potential disadvantages like for example the generator might be dependent on unix timestamp in seconds, so it might pass the `hashBenchmarkProvider` comparison but afterwards insert different documents, although it's really unusual situation, and this change will allow us to have a test name before inserting documents so I like it :)

noisersup (commented):
Some comments before working on that, nothing important

integration/Taskfile.yml
```diff
@@ -3,9 +3,10 @@
 version: 3
 
 vars:
+  COUNT: 10
   BENCHTIME: 5s
-  BENCHNAME: '{{default "." .BENCHNAME}}'
-  RACEFLAG: -race={{ne OS "windows"}}
+  BENCHNAME: .
```
noisersup, line 8: Why are we removing option to provide benchmark name? It's useful for benchmark development.

AlekSi: That's the default value, it could be overwritten. The old value did essentially the same, but in a more complicated way

integration/shareddata/benchmark_provider.go
AlekSi: that's one of the reason we iterate twice and compare results

integration/Taskfile.yml
AlekSi: I think `task target VAR=value` should work

noisersup: It works, thanks!

chilagrow (approved):

AlekSi (commented):
LGTM

Issue #2345, opened 2023-04-03
Do not check error messages in compatibility tests


chilagrow:
### What should be done?

We should update all our compatibility tests to _not_ compare error messages. `altMessage` and other similar things should not be present in compat tests. Additionally, let's add comments to `testXXXCompat` helpers where we compare error types, codes, and names, explicitly mentioning that we don't compare messages (i.e., we did not forget).

### Where?

Compat tests.

### Definition of Done

- integration/compatibility tests added/updated;
- spot refactorings done.

AlekSi:
Compat tests don't check error messages by design – that will make them much more dependent on the data, but we introduced compat tests to depend on the data less, not more.

Instead of checking error messages, let's add comments explaining that we don't do that explicitly.

PR #2546, opened 2023-05-02
command name can be lowercase

chilagrow:
# Description

Closes #{issue_number}.

## Readiness checklist


w84thesun (dismissed):
LGTM!

integration/commands_administration_compat_test.go
```diff
@@ -85,11 +84,8 @@ func TestCommandsAdministrationCompatCollStatsWithScale(t *testing.T) {
 				targetErr = UnsetRaw(t, targetErr)
 				compatErr = UnsetRaw(t, compatErr)
 
-				// TODO https://github.com/FerretDB/FerretDB/issues/2545
 				if tc.altMessage != "" {
-					var expectedErr mongo.CommandError
-					require.ErrorAs(t, compatErr, &expectedErr)
-					AssertEqualAltError(t, expectedErr, tc.altMessage, targetErr)
+					AssertMatchesCommandError(t, compatErr, targetErr)
```
AlekSi, line 88: Note that `tc.altMessage` is not used there.

If then intention was to do a part of #2345 too, the I think the whole `if-else` could be replaced by a single call to `AssertMatchesCommandError`

integration/helpers.go
```diff
@@ -222,26 +221,38 @@ func AssertMatchesCommandError(t testing.TB, expected, actual error) {
...
 	t.Helper()
 
-	var aErr, eErr mongo.WriteException
+	a, okA := actual.(mongo.WriteException)   //nolint:errorlint // do not inspect error chain
+	e, okE := expected.(mongo.WriteException) //nolint:errorlint // do not inspect error chain
 
-	if ok := errors.As(actual, &aErr); !ok || len(aErr.WriteErrors) != 1 {
+	if !okA && !okE {
+		// driver sent an error which is not WriteException, check errors are the same.
 		assert.Equal(t, expected, actual)
```
AlekSi, line 234: That would happily allow `AssertMatchesWriteError(t, io.EOF, io.EOF)`, but that contradicts the function description

integration/helpers.go
```diff
@@ -222,26 +221,38 @@ func AssertMatchesCommandError(t testing.TB, expected, actual error) {
...
 		return
 	}
 
-	if ok := errors.As(expected, &eErr); !ok || len(eErr.WriteErrors) != 1 {
-		assert.Equal(t, expected, actual)
-		return
-	}
+	require.Truef(t, okA, "actual is %T, not mongo.WriteException", a)
+	require.Truef(t, len(a.WriteErrors) == 1, "actual is %v, expected one mongo.WriteError", a.WriteErrors)
```
AlekSi, line 239: `require.Len`

AlekSi (changes requested):

integration/commands_administration_compat_test.go
chilagrow: Replaced if-else with `AssertMatchesCommandError`. I also removed `tc.altMessage` as well since it was no longer used. So this issue also closes https://github.com/FerretDB/FerretDB/issues/2345

integration/helpers.go
chilagrow: FYI, currently I'm seeing error from the driver that _update needs to have at least one field and cannot be empty_. I guess we are testing that in few tests.

chilagrow: Example https://github.com/FerretDB/FerretDB/actions/runs/4955317748/jobs/8864615428

rumyantseva (dismissed):
LGTM

w84thesun (dismissed):

noisersup (dismissed):
Looks good to me! Are we planning to introduce comparison of messages in the future? To stay independent from data we can use regexes or replace the values with placeholders before comparison

integration/helpers.go
AlekSi: `AssertMatchesWriteError` should check that both errors are `WriteError`. It should not consider other error types to be acceptable. That line of code is not correct

chilagrow: Updated to assert only for `WriteError`. For `ReplaceEmptyDocument` test, given the error comes from mongo go driver https://github.com/mongodb/mongo-go-driver/blob/master/mongo/mongo.go#L158, not sure if we want to test driver errors. But for now, I'm checking that specific error message for target and compat before calling `WriteError`. 
https://github.com/FerretDB/FerretDB/blob/bab38aeab4cf2919ee0b633d773a9c68698e2cfc/integration/update_compat_test.go#L435

integration/helpers.go
```diff
@@ -222,26 +221,33 @@ func AssertMatchesCommandError(t testing.TB, expected, actual error) {
...
+	eErr := e.WriteErrors[0]
 
-	assert.Equal(t, eErr.WriteErrors[0].Code, aErr.WriteErrors[0].Code)
+	aErr.Raw = nil
+	eErr.Raw = nil
+
+	actualMessage := aErr.Message
+	aErr.Message = eErr.Message
+
+	if !AssertEqualWriteError(t, eErr, aErr) {
```
rumyantseva, line 248: So, `AssertEqualWriteError` is similar to `AssertMatchesWriteError`, but requires the same error message? So, "matches" means the same code but possible different messages? Are we fine with the naming of these functions? Should `AssertMatchesWriteError` be called something like `AssertEqualWriteErrorCodes` or something like this for more consistency?

w84thesun (dismissed):
LGTM! Nice catch with `multi` flag!

rumyantseva (approved):
LGTM. The checks look reasonable! 

I'm not sure about the naming (left a suggestion about naming), but I don't really like my suggestion as the functions have different signatures (different error types).

integration/helpers.go
chilagrow: Just looking at `WriteError` https://github.com/mongodb/mongo-go-driver/blob/master/mongo/errors.go#L302, we don't ignore `Index` and `Details` so it compares them too. :hugs:

chilagrow: AssertEqualWriteError -> ignore Raw, AssertMatchesCommandError -> ignore Raw and Message.

noisersup (approved):

PR #2553, opened 2023-05-02
Improve handling of `os-release` files

### LINKED ISSUE #2500: Improve how we identify Linux version in `hostInfo` command

kropidlowsky:
# Description

Closes #2500 .

Add  in parseOSRelease support for:
- unquoting values
- dealing with more than one `=` in values - there should be only one delimiter per line

## Readiness checklist

### DIFF
--- a/internal/handlers/common/parse_osrelease.go
+++ b/internal/handlers/common/parse_osrelease.go
@@ -39,11 +39,12 @@ func parseOSRelease(r io.Reader) (string, string, error) {
 		value := strings.Join(str[1:], "")
 		unquotedValue, err := strconv.Unquote(value)
 
-		if errors.Is(err, nil) {
+		switch {
+		case errors.Is(err, nil):
 			configParams[str[0]] = unquotedValue
-		} else if errors.Is(err, strconv.ErrSyntax) {
+		case errors.Is(err, strconv.ErrSyntax):
 			configParams[str[0]] = value
-		} else {
+		default:
 			return "", "", lazyerrors.Error(err)
 		}
 	}
--- a/internal/handlers/common/parse_osrelease.go
+++ b/internal/handlers/common/parse_osrelease.go
@@ -16,6 +16,7 @@ package common
 
 import (
 	"bufio"
+	"errors"
 	"io"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


internal/handlers/common/parse_osrelease.go
```diff
@@ -34,7 +36,16 @@ func parseOSRelease(r io.Reader) (string, string, error) {
...
+		value := strings.Join(str[1:], "")
+		unquotedValue, err := strconv.Unquote(value)
+
+		if errors.Is(err, nil) {
+			configParams[str[0]] = unquotedValue
+		} else if errors.Is(err, strconv.ErrSyntax) {
+			configParams[str[0]] = value
+		} else {
+			return "", "", lazyerrors.Error(err)
+		}
```
w84thesun, line 48: That part will look more clear with `switch`, I think.
```suggestion
		switch {
		case errors.Is(err, nil):
			configParams[str[0]] = unquotedValue
		case errors.Is(err, strconv.ErrSyntax):
			configParams[str[0]] = value
		default:
			return "", "", lazyerrors.Error(err)
		}
```

w84thesun (commented):
Thank you for your contribution! Please take a look at the small suggestion I propose.

w84thesun (approved):
LGTM!

AlekSi (approved):

PR #2577, opened 2023-05-08
Do panic in integration tests if connection can't be established

rumyantseva:
# Description

In `integration/setup/listener.go` in the `tb.Cleanup` call we wait for a message from `done` channel. If the connection with the client wasn't established, this message will be never received, as listener is still runing. As a result, the tests are hanging until timeout. This PR fixes this problem by panicking if the connection can't be established.

For example, this problem occurred in this https://github.com/FerretDB/FerretDB/pull/2540 PR, and it was hard to catch the problem with failing integration tests.

## Readiness checklist


integration/setup/client.go
```diff
@@ -106,12 +106,14 @@ func setupClient(tb testing.TB, ctx context.Context, uri string) *mongo.Client {
 	defer trace.StartRegion(ctx, "setupClient").End()
 
 	client, err := makeClient(ctx, uri)
-	require.NoError(tb, err, "URI: %s", uri)
+	if err != nil {
+		return nil, err
+	}
```
AlekSi, line 111: I don't think we should assert on some errors and return others. The simplest solution might be panic

rumyantseva: Ok, if panic is enough, let's panic.

AlekSi: Actually, there is a simpler fix

w84thesun (approved):
Nice catch!

AlekSi (approved):

noisersup (approved):
Nice one!

PR #2578, opened 2023-05-08
Fix exponential backoff minimum duration

### LINKED ISSUE #1720: Use exponential backoff with jitter
### What should be done?

We should use exponential backoff with jitter in places where we wait for something.

See https://aws.amazon.com/ru/blogs/architecture/exponential-backoff-and-jitter/

### Where?

Places that use `ctxutil.Sleep`, `time.Sleep`, `context.WithTimeout`, etc.

### Definition of Done

- no new external dependency;
- unit test for math added;
- spot refactorings done.

noisersup:
# Description

Closes #1720.

## Readiness checklist


internal/util/ctxutil/ctxutil.go
```diff
@@ -74,7 +74,9 @@ func DurationWithJitter(cap time.Duration, retry int64) time.Duration {
 
 	maxMilliseconds := float64(base.Milliseconds()) * math.Pow(2, float64(retry))
 	capMilliseconds := float64(cap.Milliseconds())
-	lowestValue := int64(math.Min(capMilliseconds, maxMilliseconds))
 
-	return time.Duration(rand.Int63n(lowestValue)) * time.Millisecond
+	lowestValue := int64(math.Min(capMilliseconds, maxMilliseconds)) - 100
+
+	// Math/rand is good enough because we don't need the randomness to be cryptographically secure.
+	return time.Duration(rand.Int63n(lowestValue)+100) * time.Millisecond
```
AlekSi, line 81: I think that would panic with `cap = 50 * time.Millisecond`, for example. Please add a test

noisersup: Should we add a comment that provided cap cannot be smaller than `100ms`, or should we for example decrease min to 0ms if the cap is smaller than 100ms?

noisersup: Also, do we even need minimum jitter?

internal/util/ctxutil/ctxutil.go
```diff
@@ -61,10 +61,10 @@ func SleepWithJitter(ctx context.Context, d time.Duration, attempts int64) {
 	<-sleepCtx.Done()
 }
 
-// DurationWithJitter returns an exponential backoff duration based on retry with random jitter.
-// The maximum sleep is the cap. The minimum duration is at least 100 milliseconds.
+// DurationWithJitter returns an exponential backoff duration based on retry with random "full jitter".
+// https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
 //
-// Math/rand is good enough because we don't need the randomness to be cryptographically secure.
+// The maximum sleep is the cap. The minimum duration is at least 100 milliseconds.
```
rumyantseva, line 67: I think 100 milliseconds as the minimum duration is too much. I'd do something like 3-5 milliseconds, and then if it didn't help, we increase it within the next retry.

internal/util/ctxutil/ctxutil.go
noisersup: For now I've updated description of function to mention expected parameters, added custom panic message similarly to the one that we already have for negative `retry`, and added test case that validates if it actually panics.

internal/util/ctxutil/ctxutil.go
noisersup: Done

internal/util/ctxutil/ctxutil.go
```diff
@@ -61,20 +62,28 @@ func SleepWithJitter(ctx context.Context, d time.Duration, attempts int64) {
...
-// The maximum sleep is the cap. The minimum duration is at least 100 milliseconds.
+// DurationWithJitter returns an exponential backoff duration based on retry with random "full jitter".
+// https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
 //
-// Math/rand is good enough because we don't need the randomness to be cryptographically secure.
+// The maximum sleep is the cap. The minimum sleep is at least 3 milliseconds.
+// Provided cap must be larger than minimum sleep, and retry number must be a positive number.
 func DurationWithJitter(cap time.Duration, retry int64) time.Duration {
 	const base = time.Millisecond * 100
+	const minMilliseconds = 3
```
chilagrow, line 72: It's a bit weird to see `base` as Duration and `minMilliseconds` as int. Could we use same unit here? :hugs:

chilagrow (commented):
Just tiny comment :hugs: nice work!

internal/util/ctxutil/ctxutil.go
```diff
@@ -61,20 +62,28 @@ func SleepWithJitter(ctx context.Context, d time.Duration, attempts int64) {
...
+	if cap.Milliseconds() <= minMilliseconds {
+		panic(fmt.Sprintf("cap must be larger than min sleep (%dms)", minMilliseconds))
+	}
+
 	maxMilliseconds := float64(base.Milliseconds()) * math.Pow(2, float64(retry))
 	capMilliseconds := float64(cap.Milliseconds())
-	lowestValue := int64(math.Min(capMilliseconds, maxMilliseconds))
 
-	return time.Duration(rand.Int63n(lowestValue)) * time.Millisecond
+	lowestValue := int64(math.Min(capMilliseconds, maxMilliseconds)) - minMilliseconds
```
rumyantseva, line 85: I think I am a bit lost with calculation :)

As far as I understand, we use this one:
```
temp = min(cap, base * 2**attempt)
sleep = temp/2 + random_between(0, temp/2)
```

so, in our case `temp = min(capMilliseconds, maxMilliseconds)`
then what is the rest of the calculation about?

noisersup: The calculation you mentioned is the Equal Jitter

We use Full Jitter approach:
```
sleep = random_between(0, min(cap, base * 2 ** attempt))
```

internal/util/ctxutil/ctxutil.go
noisersup: done

chilagrow (dismissed):

noisersup:
## TODO
- [ ] rename variables
- [ ] generate graph from the duration test results

noisersup:
image

Didn't know how exactly should I model the data so I've callected all returned backoff durations and counted how often they were returned. The timeouts are distributed pretty equally to me.

internal/util/ctxutil/ctxutil.go
```diff
@@ -61,20 +62,32 @@ func SleepWithJitter(ctx context.Context, d time.Duration, attempts int64) {
...
+// https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
 //
-// Math/rand is good enough because we don't need the randomness to be cryptographically secure.
-func DurationWithJitter(cap time.Duration, retry int64) time.Duration {
-	const base = time.Millisecond * 100
+// The maximum sleep is the cap. The minimum sleep is at least 3 milliseconds.
+// Provided cap must be larger than minimum sleep, and attempt number must be a positive number.
+func DurationWithJitter(cap time.Duration, attempt int64) time.Duration {
+	const base = 100      // ms
+	const minDuration = 3 // ms
```
noisersup, line 72: Simplified both values to integers, as doing calculations on durations creates pretty difficult code to read

internal/util/ctxutil/ctxutil.go
```diff
@@ -61,20 +62,32 @@ func SleepWithJitter(ctx context.Context, d time.Duration, attempts int64) {
...
-	lowestValue := int64(math.Min(capMilliseconds, maxMilliseconds))
+	// calculate base backoff based on base duration and amount of attempts
+	backoff := float64(base * math.Pow(2, float64(attempt)))
+	// cap is a max limit of possible durations returned
+	maxDuration := int64(math.Min(float64(capDuration), backoff))
+
+	sleep := rand.Int63n(maxDuration-minDuration) + minDuration
 
-	return time.Duration(rand.Int63n(lowestValue)) * time.Millisecond
+	// Math/rand is good enough because we don't need the randomness to be cryptographically secure.
```
w84thesun, line 91: Should we move that comment one line above?

w84thesun (commented):
Overall looks good, added a small comment.

internal/util/ctxutil/ctxutil.go
chilagrow: What is the reason for `capDuration` instead of `cap`? all `base`, `minDuration`, `capDuration` `maxDuration` are same unit, so it's a bit odd to have different naming.

This is honestly just super minor.

chilagrow (approved):

internal/util/ctxutil/ctxutil.go
noisersup: I wanted to store cap in milliseconds in variable, as it's used multiple times in code and calling `cap.Milliseconds()` inside `math.Min` just makes i more difficult to read, we have also `maxDuration` and `minDuration`, what could we do instead?

internal/util/ctxutil/ctxutil.go
noisersup: You're right, fixed

chilagrow (dismissed):

internal/util/ctxutil/ctxutil.go
chilagrow: yeah that's fine :hugs:

w84thesun (approved):
:shipit:

AlekSi:
> Didn't know how exactly should I model the data

Well, we want the graph of exponential backoff to looks like an exponent. Like in the original article. I'm not sure how to read the current graph.

noisersup:
> Well, we want the graph of exponential backoff to looks like an exponent. Like in the original article. I'm not sure how to read the > current graph.

Are you talking about dot graph or the linear graph from the article?

noisersup:
The exponent graph on their article shows impact of amount of competing clients on the amount of calls, which requires implementation of clients that will reconnect and make their connection successful or not.

rumyantseva (approved):
I'm still curious to see the distribution, but the approach looks good to me.

In principle, we can call a few loops that call `DurationWithJitter` in parallel (with retry being increased in each loop) and save results. That will give some distribution similar to the idea of parallel connections.

rumyantseva (approved):

noisersup:
image

@AlekSi New chart based on the results of new test. It shows total amount of calls for simulation with n clients.

AlekSi:
On a call with Elena, I just released that I misread graphs in the original article, and that was the source of confusion. I will send a PR explaining what I meant.

PR #2620, opened 2023-05-12
Use `ExtractParameters` in handlers

### LINKED ISSUE #2465: Refactor how we process handlers params
### What should be done?

The `internal/handlers/common` package contains code responsible for processing and validating various handlers' parameters. This code includes both generic functions such as `GetRequiredParam` and `GetOptionalParam`, as well as parameter-specific functions such as `GetSkipParam` and `GetLimitParam`. However, there are a few issues with this code.

Firstly, the **errors** returned by these functions are not adequately documented. It is unclear what these errors represent and when they should be returned. Additionally, the functions themselves lack documentation about the potential errors they may return.

Secondly, the **functions** are spread out across the entire package, making it challenging to locate and understand them.

To address these issues, significant refactoring is required. In this issue, let's create a separate package `commonparams` dedicated solely to parameter processing. This would consolidate all related code in one place and enable better documentation of both the functions and the errors they may return.

**For the scope of this issue**, the goal is to learn about parameter processing issues and send a pull request with better structuring, documentation, and possibly other improvements. 

### Where?

https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/common/params.go
https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/common/limit.go
https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/common/scale.go
https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/common/skip.go

Check other `Get*Param` functions too.

### Definition of Done

- all handlers updated;
- unit tests added/updated if needed;
- integration/compatibility tests added/updated if needed;
- spot refactorings done;
- significant refactoring are not done.

w84thesun:
# Description

Closes #2465.

In this PR I changed the `delete`, `distinct`, `find`, `findAndModify`, `insert` and `update` handlers to extract parameters with the `ExtractParam` function.

`explain` has to be implemented separately as its behaviour is slightly different from other handlers.

I added support for the following new tags:
   - `positiveNumber` - provided value must be of types [int, long, double] and greater than 0,
     double values would be rounded to long;
   - `wholePositiveNumber` - provided value must be of types [int, long] and greater than 0;
   - `numericBool` - provided value must be of types [bool, int, long, double] and would be converted to bool;
   - `zeroOrOneAsBool` - provided value must be of types [int, long, double] with possible values `0` or `1`.

_Please suggest better naming for `numericBool` and `zeroOrOneAsBool`._

Those are needed to choose from different numeric parameter functions like `GetBoolOptionalParam` or `getWholeParamStrict`. We can remove them if we will decide to unify the way we process parameters.

Some of the error messages were unified to form `BSON <handler>.<parameter> has wrong ...`

## Readiness checklist


internal/handlers/commonparams/extract_params.go
```diff
@@ -37,10 +40,30 @@ import (
...
+//   - `ignored` - field is ignored;
+//   - `numericBool` - field is of type `bool` but can be parsed from numeric types;
+//   - `strict` - fields of numeric types are parsed strictly;
+//   - `positive` - fields of numeric types must be positive;
+//   - `numericAsBool` - fields of numeric types that must have as values only `0` or `1`.
 //
 // Collection field processed in a special way. For the commands that require collection name
 // it is extracted from the command name.
+// If the field could have different types (e.g. `*types.Document` and `*types.Array`) then
+// the field must be of type `any`.
```
AlekSi, line 52: In that particular case it could (and probably should) be `types.CompositeTypeInterface`

internal/handlers/commonparams/extract_params.go
```diff
@@ -37,10 +40,30 @@ import (
 //   - `opt` - field is optional;
 //   - `non-default` - field is unimplemented for non-default values;
 //   - `unimplemented` - field is not implemented yet;
-//   - `ignored` - field is ignored.
+//   - `ignored` - field is ignored;
+//   - `numericBool` - field is of type `bool` but can be parsed from numeric types;
+//   - `strict` - fields of numeric types are parsed strictly;
+//   - `positive` - fields of numeric types must be positive;
+//   - `numericAsBool` - fields of numeric types that must have as values only `0` or `1`.
```
AlekSi, line 47: Some of those tags are hard to understand. What is the default value for `opt`? Why `non-default` is not implemented? (I guess it should be `unimplemented-non-default`, but then what is the default value?) `strict` can't be defined as "strict" (you can't define X in terms of X itself). Etc

AlekSi (commented):
Please re-request review when tests pass

internal/handlers/commonparams/extract_params.go
w84thesun: Some of the parameters we don't process right now, like `count.hint` could be either `string` or `*types.Document`. That's why I suggested using `any`.

AlekSi: I think we should remove everything that is not being used: writing code "just in case" might we fine when design is frozen, but it is not frozen yet, and then we will have more code to update

w84thesun: But if we remove everything that is not used we will return errors like 'field X is not expected' for the fields that could be used in some command. If we ok with that then I can remove all unused fields.

AlekSi: Then it is being used, but indirectly… Let's wrap it up so there are no regressions (like we had with `count`). We will return to it later

internal/handlers/commonparams/extract_params.go
w84thesun: I rewrote the descriptions to explain what those tags should be used for.

internal/handlers/commonparams/extract_params.go
```diff
@@ -211,8 +268,87 @@ func setStructField(elem *reflect.Value, i int, command, key string, val any) er
...
+			if !ok {
+				return commonerrors.NewCommandErrorMsgWithArgument(
+					commonerrors.ErrTypeMismatch,
+					fmt.Sprintf(
+						`BSON field '%s.%s' is the wrong type '%s', expected type '%s'`,
+						command, key, AliasFromType(val), AliasFromType(fv.Interface()),
+					),
+					command,
+				)
+			}
```
chilagrow, line 339: In the future we may find this logic may be too strict. Array may contain elements that isn't document. But we don't have example parameter for now so it's probably ok as it is for now.

internal/handlers/common/delete.go
```diff
@@ -15,147 +15,59 @@
...
 
-	collectionParam, err := document.Get(document.Command())
-	if err != nil {
-		return nil, err
-	}
-
-	var ok bool
-	if collection, ok = collectionParam.(string); !ok {
+	err := commonparams.ExtractParams(document, "delete", &params, l)
+	if errors.Is(err, commonparams.ErrFieldNotPopulated) {
```
chilagrow, line 60: `"BSON field 'delete.deletes.limit' is missing but a required field"` is too specific error message since there are other field not optional. Missing field may not be `limit` but it might be `q` filter.

internal/handlers/common/update_params.go
```diff
@@ -15,132 +15,63 @@
...
 
-// UpdatesParams represents parameters for update command.
+// UpdatesParams represents parameters for the update command.
 type UpdatesParams struct {
-	DB         string
-	Collection string
-	Updates    []UpdateParams
+	DB         string         `ferretdb:"$db"`
+	Collection string         `ferretdb:"collection"`
+	Updates    []UpdateParams `ferretdb:"updates"`
```
chilagrow, line 28: Should `Updates` be `opt` parameter?

internal/handlers/commonparams/params.go
```diff
@@ -74,59 +74,134 @@ func GetWholeNumberParam(value any) (int64, error) {
...
-	case errors.Is(err, ErrLongExceededPositive):
-		return math.MaxInt64, nil
+	return whole, nil
+}
 
-	case errors.Is(err, ErrLongExceededNegative):
-		return 0, commonerrors.NewCommandError(
-			commonerrors.ErrValueNegative,
-			fmt.Errorf("BSON field '%s' value must be >= 0, actual value '%d'", param, int(math.Ceil(value.(float64)))),
+// GetBoolOptionalParam returns doc's bool value for key.
```
chilagrow, line 182:
```suggestion
// GetBoolOptionalParam returns bool value of v.
```

internal/handlers/common/distinct.go
```diff
@@ -30,68 +29,33 @@ import (
...
-	Ignored(document, l, ignoredFields...)
-
 	var dp DistinctParams
 
-	if dp.DB, err = GetRequiredParam[string](document, "$db"); err != nil {
-		return nil, err
-	}
-
-	collectionParam, err := document.Get(document.Command())
+	err := commonparams.ExtractParams(document, "distinct", &dp, l)
```
chilagrow, line 47: I think you want to catch `ErrFieldNotPopulated` for missing `key`

internal/handlers/commonparams/extract_params.go
```diff
@@ -29,18 +29,42 @@ import (
...
-//   - `ignored` - field is ignored.
+//   - `opt` - field is optional, the field value would not be set if it's not present in the document;
+//   - `unimplemented-non-default` - error would be returned if non-default value for the field is provided;
+//   - `unimplemented` - error would be returned if the value is present in the document;
+//   - `ignored` - field is ignored and would not be set, but it would be logged;
+//   - `positiveNumber` - provided value must be of types [int, long, double] and greater than 0,
+//     double values would be rounded to long;
+//   - `wholePositiveNumber` - provided value must be of types [int, long] and greater than 0;
+//   - `numericBool` - provided value must be of types [bool, int, long, double] and would be converted to bool;
+//   - `numericAsBool` - provided value must be of types [int, long, double] with possible values `0` or `1`.
```
chilagrow, line 48: `zeroOrOneAsBool` might be more specific :hugs:

chilagrow (commented):
Great work :hugs: My comments are mainly just catching corner cases.

internal/handlers/common/update_params.go
```diff
@@ -15,132 +15,63 @@
...
-	Update  *types.Document
-	Comment string
-	Multi   bool
-	Upsert  bool
+	// TODO: https://github.com/FerretDB/FerretDB/issues/2627
+	// get comment from query, e.g. db.collection.UpdateOne({"_id":"string", "$comment: "test"},{$set:{"v":"foo""}})
+	Filter *types.Document `ferretdb:"q,opt"`
+	Update *types.Document `ferretdb:"u,opt"`
+	Multi  bool            `ferretdb:"multi,opt"`
+	Upsert bool            `ferretdb:"upsert,opt"`
```
rumyantseva, line 46: Isn't his one `numericAsBool`?

internal/handlers/commonparams/extract_params.go
```diff
@@ -29,18 +29,42 @@ import (
...
 )
 
+// ErrFieldNotPopulated is returned when a required field is not populated.
+var ErrFieldNotPopulated = errors.New("required field is not populated")
+
 // ExtractParams fill passed value structure with parameters from the document.
 // If the passed value is not a pointer to the structure it panics.
 // Parameters are extracted by the field name or by the `ferretdb` tag.
 //
 // Possible tags:
```
rumyantseva, line 39: What I don't see in this list is numbers that will be rounded to X if too big value is set. As far as I remember, this is how `scale` in `collStats`/`dbStats` work. But I guess we don't move `scale` yet, right?

internal/handlers/commonparams/extract_params.go
```diff
@@ -29,18 +29,42 @@ import (
...
-//   - `unimplemented` - field is not implemented yet;
-//   - `ignored` - field is ignored.
+//   - `opt` - field is optional, the field value would not be set if it's not present in the document;
+//   - `unimplemented-non-default` - error would be returned if non-default value for the field is provided;
+//   - `unimplemented` - error would be returned if the value is present in the document;
+//   - `ignored` - field is ignored and would not be set, but it would be logged;
+//   - `positiveNumber` - provided value must be of types [int, long, double] and greater than 0,
+//     double values would be rounded to long;
+//   - `wholePositiveNumber` - provided value must be of types [int, long] and greater than 0;
+//   - `numericBool` - provided value must be of types [bool, int, long, double] and would be converted to bool;
```
rumyantseva, line 47: Are there examples where `numericBool` and `numericAsBool` used? I don't see them being provided as tags.

rumyantseva (commented):
Started reviewing (not a full review yet). 

New tags seem useful!

rumyantseva (commented):
Started reviewing (not a full review yet). 

New tags seem useful!

internal/handlers/common/update_params.go
w84thesun: No, it’s a required parameter.

chilagrow: Ah okie so we need to handle `ErrFieldNotPopulated` for missing field. :hugs:

w84thesun: Make it handled inside `checkAllRequiredFIeldsPopulated`.

internal/handlers/common/distinct.go
w84thesun: Make it handled inside checkAllRequiredFIeldsPopulated.

internal/handlers/common/update_params.go
w84thesun: I don't see it's tested for those values. In `findAndModify` it's just bool.

internal/handlers/commonparams/extract_params.go
w84thesun: Yes, we move only `delete`, `distinct`, `find`, `findAndModify`, `insert` and `update`.

internal/handlers/commonparams/extract_params.go
w84thesun: Make it handled inside `checkAllRequiredFIeldsPopulated`.

internal/handlers/common/update_params.go
chilagrow: It's not tested but turns out it is `numericBool`.

```
db.testdata.runCommand({"findAndModify":"testdata","update":{"$set":{}}, "upsert":-50})
{
  lastErrorObject: {
    n: 1,
    updatedExisting: false,
    upserted: ObjectId("6461df27425e9da653295bf9")
  },
  value: null,
  ok: 1
}
```

internal/handlers/commonparams/extract_params.go
```diff
@@ -267,8 +405,17 @@ func checkAllRequiredFieldsPopulated(v *reflect.Value, command string, keys []st
...
+		if key == "-" {
+			continue
+		}
+
 		if !slices.Contains(keys, key) {
-			return lazyerrors.Errorf("required field %q is not populated", key)
+			return commonerrors.NewCommandErrorMsgWithArgument(
+				commonerrors.ErrMissingField,
+				fmt.Sprintf("BSON field '%s.%s' is missing but a required field", command, key),
+				"limit",
```
chilagrow, line 417:
```suggestion
				command,
```

internal/handlers/common/update_params.go
chilagrow: Yeah looks good

internal/handlers/commonparams/params.go
```diff
@@ -74,59 +74,134 @@ func GetWholeNumberParam(value any) (int64, error) {
...
 
-	case errors.Is(err, ErrLongExceededNegative):
-		return 0, commonerrors.NewCommandError(
-			commonerrors.ErrValueNegative,
-			fmt.Errorf("BSON field '%s' value must be >= 0, actual value '%d'", param, int(math.Ceil(value.(float64)))),
+// GetBoolOptionalParam returns bool value of v.
+// Non-zero double, long, and int values return true.
+// Zero values for those types, as well as nulls and missing fields, return false.
+// Other types return a protocol error.
+func GetBoolOptionalParam(v any, key string) (bool, error) {
```
rumyantseva, line 186: `getOptionalPositiveNumber(key string, value any)` has the signature key-value, this one has the opposite signature - `value - key`. Let's make them the same, so it's consistent?

rumyantseva (approved):
Ok, I like the tests, and the returned errors look reasonable to me.
I see a bit of inconsistency in function names, but I think we can fix them later when we meet this inconsistency somewhere.

internal/handlers/commonparams/extract_params.go
AlekSi: That being said, the original review comment still stands – that example is bad, the one in your response is much better. The comment should be updated.

internal/handlers/commonparams/extract_params.go
AlekSi: > `opt` - field is optional, the field value would not be set if it's not present in the document;

Ok, so if the field is required, will it be set if it is not in the document? That's pretty much the opposite of being required, no?

Based on that documentation, I don't know how to use that function

PR #2631, opened 2023-05-15
Use separate codebase for aggregation `$project` and query `projection`

### LINKED ISSUE #2430: Support dot notation in query projections
### What should be done?

We should add support for dot notation in query projections.

See relevant test cases with this issue number.

The tricky part is that we have to create an empty document when the given path could not be reached.

```js
db.projections.insertMany([
 {
    a: { b: { c: { d: 123 }, e: 13 }, f: 14 },
    g: 15
  },
  {
    a: { b: { c: [ 1, 2 ], e: 13 }, f: 14 },
    g: 15
  },
  {
    a: { b: 12 }
  },
  {
    a: { b: { c: 12 } } 
  }
]);


db.projections.find({}, {'_id': 0,'a.b.c.d': 0})
[
  { a: { b: { c: {}, e: 13 }, f: 14 }, g: 15 },
  {
    a: { b: { c: [ 1, 2 ], e: 13 }, f: 14 },
    g: 15
  },
  { a: { b: 12 } },
  { a: { b: { c: 12 } } }
]


 db.projections.find({}, {'_id': 0,'a.b.c.d': 1})
[
  {
    a: { b: { c: { d: 123 } } }
  },
  { a: { b: { c: [] } } },
  { a: {} },
  { a: { b: {} } }
]

db.projections.find({}, {'_id': 0,'a.b.c': 0})
[
  { a: { b: { e: 13 }, f: 14 }, g: 15 },
  { a: { b: { e: 13 }, f: 14 }, g: 15 },
  { a: { b: 12 } },
  { a: { b: {} } }
]


db.projections.find({}, {'_id': 0,'a.b.c': 1})
[
  {
    a: { b: { c: { d: 123 } } }
  },
  {
    a: { b: { c: [ 1, 2 ] } }
  },
  { a: {} },
  { a: { b: { c: 12 } } }
]

```

### Where?

internal/handlers/common/projection.go

### Definition of Done

- all handlers updated;
- unit tests added/updated;
- spot refactorings done;
- user documentation updated;
- something else?

chilagrow:
# Description

Closes #2430.

`project` in aggregation and `projection` in query is different. We do not use common code.

## Readiness checklist

### DIFF
--- a/internal/handlers/common/aggregations/projection.go
+++ b/internal/handlers/common/aggregations/projection.go
@@ -1,0 +1,482 @@
+// Copyright 2021 FerretDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package aggregations
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/FerretDB/FerretDB/internal/handlers/commonerrors"
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/iterator"
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+	"github.com/FerretDB/FerretDB/internal/util/must"
+)
+
+var errProjectionEmpty = errors.New("projection is empty")
+
+// ValidateProjection check projection document.
+// Document fields could be either included or excluded but not both.
+// Exception is for the _id field that could be included or excluded.
+func ValidateProjection(projection *types.Document) (*types.Document, bool, error) {
+	validated := types.MakeDocument(0)
+
+	if projection.Len() == 0 {
+		return nil, false, errProjectionEmpty
+	}
+
+	var projectionVal *bool
+
+	iter := projection.Iterator()
+	defer iter.Close()
+
+	for {
+		key, value, err := iter.Next()
+		if errors.Is(err, iterator.ErrIteratorDone) {
+			break
+		}
+
+		if err != nil {
+			return nil, false, lazyerrors.Error(err)
+		}
+
+		if strings.Contains(key, "$") {
+			return nil, false, commonerrors.NewCommandErrorMsg(
+				commonerrors.ErrNotImplemented,
+				fmt.Sprintf("projection operator $ is not supported in %s", key),
+			)
+		}
+
+		var result bool
+
+		switch value := value.(type) {
+		case *types.Document:
+			return nil, false, commonerrors.NewCommandErrorMsg(
+				commonerrors.ErrNotImplemented,
+				fmt.Sprintf("projection expression %s is not supported", types.FormatAnyValue(value)),
+			)
+		case *types.Array, string, types.Binary, types.ObjectID,
+			time.Time, types.NullType, types.Regex, types.Timestamp: // all this types are treated as new fields value
+			result = true
+
+			validated.Set(key, value)
+		case float64, int32, int64:
+			// projection treats 0 as false and any other value as true
+			comparison := types.Compare(value, int32(0))
+
+			if comparison != types.Equal {
+				result = true
+			}
+
+			// set the value with boolean result to omit type assertion when we will apply projection
+			validated.Set(key, result)
+		case bool:
+			result = value
+
+			// set the value with boolean result to omit type assertion when we will apply projection
+			validated.Set(key, result)
+		default:
+			return nil, false, lazyerrors.Errorf("unsupported operation %s %value (%T)", key, value, value)
+		}
+
+		if projection.Len() == 1 && key == "_id" {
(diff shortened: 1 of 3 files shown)


w84thesun (approved):
LGTM!

internal/handlers/common/aggregations/projection.go
```diff
@@ -0,0 +1,482 @@
...
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+	"github.com/FerretDB/FerretDB/internal/util/must"
+)
+
+var errProjectionEmpty = errors.New("projection is empty")
+
+// ValidateProjection check projection document.
+// Document fields could be either included or excluded but not both.
+// Exception is for the _id field that could be included or excluded.
+func ValidateProjection(projection *types.Document) (*types.Document, bool, error) {
```
rumyantseva, line 35: Let's describe possible errors and when they can be returned

internal/handlers/common/aggregations/projection_iterator.go
```diff
@@ -0,0 +1,83 @@
...
+
+// ProjectionIterator returns an iterator that projects documents returned by the underlying iterator.
+// It will be added to the given closer.
+//
+// Next method returns the next projected document.
+//
+// Close method closes the underlying iterator.
+func ProjectionIterator(iter types.DocumentsIterator, closer *iterator.MultiCloser, projection *types.Document) (types.DocumentsIterator, error) { //nolint:lll // for readability
+	projectionValidated, inclusion, err := ValidateProjection(projection)
+	if errors.Is(err, errProjectionEmpty) {
```
rumyantseva, line 33: Minor comment: we use `switch-case` in other place likes this -- smth like

```switch {
case err == nil:
// ...
case errors.Is(...)
//
default:
// lazyerror
}
```

I suggest we follow the same approach everywhere, so the code looks more consistent and it's easier to read it.

internal/handlers/common/aggregations/projection.go
```diff
@@ -0,0 +1,482 @@
...
+	"time"
+
+	"github.com/FerretDB/FerretDB/internal/handlers/commonerrors"
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/iterator"
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+	"github.com/FerretDB/FerretDB/internal/util/must"
+)
+
+var errProjectionEmpty = errors.New("projection is empty")
```
rumyantseva, line 30: Let's document this error - why we need it, how to use it, etc

internal/handlers/common/aggregations/projection_iterator.go
chilagrow: I personally prefer current implementation because `err == nil` looks less natural. Looking the codebase, we have mixture but it's sensible to decide which way we can do more consistently. @w84thesun, @noisersup, @AlekSi? :hugs:

w84thesun: My personal preference would be to process errors in `switch` for more than 1 case in a way:
```
{
result, err := function()
if err != nil {
    switch {
    case errors.Is():
        // ...
    case errors.Is():
       // ...
    default:
      // ...
    }
}

// normal flow
}

Comment (rumyantseva) [reply internal/handlers/common/aggregations/projection_iterator.go:33]: It seems that the number of prefered options will be equal to the number of people in the team :D

File: internal/handlers/common/aggregations/projection.go (hunk 1 above)
Comment (chilagrow) [reply internal/handlers/common/aggregations/projection.go:42]: Added list of errors :hugs:

File: internal/handlers/common/aggregations/projection_iterator.go (hunk 2 above)
Comment (chilagrow) [reply internal/handlers/common/aggregations/projection_iterator.go:33]: Okay this one might not be easy to make it consistent. I'm realising I'm seeing more variety :see_no_evil:

File: internal/handlers/common/aggregations/projection.go (hunk 3 above)
Comment (chilagrow) [reply internal/handlers/common/aggregations/projection.go:31]: Added document string.

**Review by rumyantseva — APPROVED** [review]
As we agreed, I approve because this code doesn't make things worse - we continue working the same way, just getting ready to handle aggregation separately. Also, now we have a separate issue to continue testing aggregation.

**Review by w84thesun — APPROVED** [review]

File: internal/handlers/common/aggregations/projection.go [hunk 4]
```diff
@@ -0,0 +1,510 @@
...
+
+// ValidateProjection check projection document.
+// Document fields could be either included or excluded but not both.
+// Exception is for the _id field that could be included or excluded.
+//
+// Errors:
+//   - `ErrProjectionExIn` when there is exclusion in inclusion projection;
+//   - `ErrProjectionInEx` when there is inclusion in exclusion projection;
+//   - `ErrNotImplemented` when there is unimplemented projection operators and expressions;
+//   - `errProjectionEmpty` when projection document is empty.
```
AlekSi, line 41: :warning: I don't think it is correct for the same exported function to return both protocol-error errors like `CommandError` with `ErrProjectionExIn` code and internal errors like `errProjectionEmpty`. Not to mention that documentation mixes protocol error codes and Go-level error values.

internal/handlers/common/aggregations/projection.go
```diff
@@ -0,0 +1,510 @@
...
+		key, value, err := iter.Next()
+		if errors.Is(err, iterator.ErrIteratorDone) {
+			break
+		}
+
+		if err != nil {
+			return nil, false, lazyerrors.Error(err)
+		}
+
+		if strings.Contains(key, "$") {
```
AlekSi, line 65: I think that should be `strings.HasPrefix`

PR #2634, opened 2023-05-15
Refactor `gitBinaryMaskParam` function

### LINKED ISSUE #2465: Refactor how we process handlers params

w84thesun:
# Description

Closes #2465.

## Readiness checklist


internal/handlers/common/params.go
```diff
@@ -186,7 +204,14 @@ func GetSkipStageParam(value any) (int64, error) {
 
 // getBinaryMaskParam matches value type, returning bit mask and error if match failed.
 // Possible values are: position array ([1,3,5] == 010101), whole number value and types.Binary value.
-func getBinaryMaskParam(mask any) (uint64, error) {
+//
+// Returned errors:
+//   - ErrBadValue if positional argument is not a number;
```
chilagrow, line 209:
```suggestion
//   - ErrBadValue if positional argument is not int32;
```

internal/handlers/common/params.go
```diff
@@ -186,7 +204,14 @@ func GetSkipStageParam(value any) (int64, error) {
...
+//   - ErrBadValue if positional argument is negative;
+//   - ErrBadValue if bitmask argument is not a whole number;
```
chilagrow, line 211:
```suggestion
```
I think we are not returning `ErrBadValue` for non-whole number. We return `ErrFailedToParse`

chilagrow (commented):
Great doc :muscle:

Just a couple of minor comments :hugs:

internal/handlers/common/params.go
```diff
@@ -26,8 +26,11 @@ import (
...
 
-// GetRequiredParam returns doc's value for key
-// or protocol error for missing key or invalid type.
+// GetRequiredParam returns doc's value for key.
+//
+// Returned errors:
+//   - ErrBadValue if the document value is missing;
+//   - ErrBadValue if the document value is not of the expected type.
 func GetRequiredParam[T types.Type](doc *types.Document, key string) (T, error) {
 	var zero T
```
AlekSi, line 35: `defer contract.EnsureError(err, ErrBadValue)`

internal/handlers/common/params.go
w84thesun: Fixed. It could be `[int, long, double]`.

internal/handlers/common/params.go
w84thesun: Removed that line.

internal/handlers/commonparams/params.go
```diff
@@ -76,8 +72,11 @@ func GetWholeNumberParam(value any) (int64, error) {
 
 // getWholeParamStrict validates the given value for find and count commands.
 //
-// If the value is valid, it returns its int64 representation,
-// otherwise it returns a command error with the given command being mentioned.
+// # If the value is valid, it returns its int64 representation
+//
+// Command errors:
+//   - ErrTypeMismatch if the value is not of types [int, long, double, null];
```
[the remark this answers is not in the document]
w84thesun: It return ErrTypeMismatch if the value is `[int, long, double, null]`, so it seems to be correct statement 😄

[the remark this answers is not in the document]
chilagrow: Just want to double check that I'm not misleading this :eyes:  
https://github.com/FerretDB/FerretDB/blob/9c440cc2d533508e619528cd1d605d5079daad7e/internal/handlers/commonparams/params.go#L84-L87

internal/handlers/common/params.go
w84thesun: It seems that I have to specify that it's not "errors" but error types. We can check that those errors have `CommandError` type but there is no way to check the error code now. We could possibly extend the `contract` package with `EnsureCommandError` and `EnsureWriteError`. What do you think?

AlekSi: That's what I tried to do – and that's surprisingly not easy.

Let's only update documentation for now. `ErrBadValue` is not an error and also not a error type

rumyantseva (approved):

internal/handlers/common/params.go
chilagrow: I like this. I think this is the practice we can expand for new code bases :hugs:

chilagrow (approved):

AlekSi:
On hold until after 1.2.0 release

rumyantseva (approved):
LGTM!

PR #2637, opened 2023-05-15
Fix index creation on nested fields

### LINKED ISSUE #2606: Indexes are created incorrectly for nested fields
### FerretDB version

1.1.0

### Backend

PostgreSQL

### Environment

- OS: MacOS
- Deployment: `task run`

### What did you do?

```
db.c.createIndex({"foo.bar": 1})
```

Index is created for a nested field foo -> bar

### What did you see instead?

Index is created for a field "foo.bar" which is not correct, also such fields with dots are not possible at all.

wqhhust:
# Description

Closes #2606.

## Readiness checklist


rumyantseva (commented):
The change itself looks reasonable! 
We need to write tests to ensure that everything works correctly.

We have tests for indexes in `internal/handlers/pg/pgdb/indexes_test.go`. One option would be to extend `TestCreateIndexIfNotExists` test to cover nested fields too. A similar approach to the current one - create an index for a nested field and compare created index definition with expected one.

And also we can add one test case in `TestDropIndexes` to ensure that it's possible to drop indexes for nested fields.

Do you any help with it?

wqhhust:
added the test

internal/handlers/pg/pgdb/indexes_test.go
```diff
@@ -39,11 +39,11 @@ func TestCreateIndexIfNotExists(t *testing.T) {
 	collectionName := testutil.CollectionName(t)
 	setupDatabase(ctx, t, pool, databaseName)
 
-	indexName := "test"
+	indexName := "test_nested_field"
 	err := pool.InTransaction(ctx, func(tx pgx.Tx) error {
 		idx := Index{
 			Name: indexName,
-			Key:  []IndexKeyPair{{Field: "foo", Order: types.Ascending}, {Field: "bar", Order: types.Descending}},
+			Key:  []IndexKeyPair{{Field: "foo.bar", Order: types.Ascending}},
```
rumyantseva, line 46: I like the test, and as now we have two tests with similar logic, it seems that it's time to introduce table tests here! ;)
So, let's create a map of test cases. It seems that we only need a couple of fields - index and expected definition. And then we can add a test cases without nested fields, a test case with one nested field, and maybe a test case with two nested fields (like `v.foo.bar`) just in case.

wqhhust:
fixed. 
image

internal/handlers/pg/pgdb/indexes_test.go
```diff
@@ -39,32 +39,69 @@ func TestCreateIndexIfNotExists(t *testing.T) {
...
-			Key:  []IndexKeyPair{{Field: "foo", Order: types.Ascending}, {Field: "bar", Order: types.Descending}},
-		}
-		return CreateIndexIfNotExists(ctx, tx, databaseName, collectionName, &idx)
-	})
-	require.NoError(t, err)
-
-	tableName := collectionNameToTableName(collectionName)
-	pgIndexName := indexNameToPgIndexName(collectionName, indexName)
+	for name, tc := range map[string]struct {
+		expecteDefinition string // the expected index definition in postgresql
```
rumyantseva, line 43:
```suggestion
		expectedDefinition string // the expected index definition in postgresql
```

a tiny typo

internal/handlers/pg/pgdb/indexes_test.go
```diff
@@ -39,32 +39,69 @@ func TestCreateIndexIfNotExists(t *testing.T) {
...
+			index: Index{
+				Name: "foo_dot_bar_dot_c",
+				Key: []IndexKeyPair{
+					{Field: "foo.bar.c", Order: types.Ascending},
+				},
+			},
+			expecteDefinition: "((((_jsonb -> 'foo'::text) -> 'bar'::text) -> 'c'::text))",
+		},
+	} {
+		// https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721#how-to-solve-this
```
rumyantseva, line 75:
```suggestion
```

I'm not sure we should mention this example here, as it's quite common for our codebase to have parallel subtests.

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -204,8 +204,15 @@ func createPgIndexIfNotExists(ctx context.Context, tx pgx.Tx, schema, table, ind
...
-		// It's important to sanitize field.Field data here, as it's a user-provided value.
-		fieldsDef[i] = fmt.Sprintf(`((_jsonb->%s)) %s`, quoteString(field.Field), order)
+		// if the key is foo.bar, then need to modify it to foo -> bar
+		var transformedParts []string
+		fs := strings.Split(field.Field, ".")
+
+		for _, f := range fs {
+			// It's important to sanitize field.Field data here, as it's a user-provided value.
+			transformedParts = append(transformedParts, quoteString(f))
+		}
```
rumyantseva, line 214:
```suggestion
		fs := strings.Split(field.Field, ".")
		transformedParts := make([]string, len(fs))

		for i, f := range fs {
			// It's important to sanitize field.Field data here, as it's a user-provided value.
			transformedParts[i] = quoteString(f)
		}
```

Minor suggestion. I think this approach is more consistent with the rest of the codebase. When we know the exact length of the slice, we allocate memory when creating it.

rumyantseva (commented):
Great, now we have table tests!

I left a couple of relatively minor comments, but overall looks good to me!

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -204,8 +204,15 @@ func createPgIndexIfNotExists(ctx context.Context, tx pgx.Tx, schema, table, ind
 			return lazyerrors.Errorf("unknown sort order: %d", field.Order)
 		}
 
-		// It's important to sanitize field.Field data here, as it's a user-provided value.
-		fieldsDef[i] = fmt.Sprintf(`((_jsonb->%s)) %s`, quoteString(field.Field), order)
+		// if the key is foo.bar, then need to modify it to foo -> bar
+		fs := strings.Split(field.Field, ".")
+		transformedParts := make([]string, len(fs))
+
+		for i, f := range fs {
```
chilagrow, line 211: Let's just `j` here, `i` is used in the outer loop and it's easier to follow if they don't share the same loop variable name :pray:

chilagrow (commented):
Great work, just one tiny comment!

chilagrow (approved):
Looks good to me! Thanks for the contribution.

rumyantseva (approved):
LGTM

PR #2653, opened 2023-05-17
Fix reporting of updates availability

### LINKED ISSUE #2624: Information about available updates is reported incorrectly
### What should be done?

https://github.com/FerretDB/FerretDB/blob/f98dc956e704778d50f8c2caf511f36115e787b9/internal/util/telemetry/reporter.go#L290-L300

Suppose an update is no longer available (for example, it was yanked). In that case, information about that is not updated in the state. As a result, FerretDB always reports that an update is available.

And the current tests can't cover that problem because they are table-driven; they should test scenarios instead.

Let's update tests first, then fix the issue.

### Where?

https://github.com/FerretDB/FerretDB/tree/main/internal/util/telemetry

### Definition of Done

- bug fixed;
- unit tests added/updated;
- spot refactorings done.

rumyantseva:
# Description

Closes #2624.

## Readiness checklist

### DIFF
--- a/internal/util/telemetry/reporter_test.go
+++ b/internal/util/telemetry/reporter_test.go
@@ -96,7 +96,7 @@ func TestNewReporterLock(t *testing.T) {
 
 func TestReporterReport(t *testing.T) {
 	t.Parallel()
-
+	// TODO
 	for name, tc := range map[string]struct {
 		f                *Flag
 		telemetryReponse string


internal/util/telemetry/reporter_test.go
```diff
@@ -97,72 +97,106 @@ func TestNewReporterLock(t *testing.T) {
...
+		assert.Equal(t, "0.4.0", s.LatestVersion)
+
+		// Disable telemetry and call the telemetry server again.
+		require.NoError(t, provider.Update(func(s *state.State) { s.Telemetry = pointer.ToBool(false) }))
+		r.report(testutil.Ctx(t))
+
+		// Expect no update to be available, and the latest version to be empty.
+		s = r.P.Get()
+		assert.False(t, s.IsUpdateAvailable)
+		assert.Equal(t, "", s.LatestVersion)
```
rumyantseva, line 174: @AlekSi please let me know if I understood this correctly: if telemetry is disabled, we reset the info about update availability and the latest version.

chilagrow (approved):
Looks good!

internal/util/state/state.go
```diff
@@ -32,7 +32,7 @@ type State struct {
 	// never persisted
 	TelemetryLocked   bool      `json:"-"`
 	Start             time.Time `json:"-"`
-	LatestVersion     string    `json:"-"` // empty if IsUpdateAvailable is false
+	LatestVersion     string    `json:"-"` // empty if telemetry is disabled; latest version reported by telemetry otherwise
```
AlekSi, line 35: That first part should not be true anymore. We should not set `LatestVersion` to the empty string if telemetry is disabled because there is no need to – we don't check it's value if it is empty or not without checking `UpdateAvailable` first

internal/util/state/state.go
```diff
@@ -32,7 +32,7 @@ type State struct {
...
 	HandlerVersion    string    `json:"-"` // may be empty
 	IsUpdateAvailable bool      `json:"-"`
```
AlekSi, line 37:
```suggestion
	UpdateAvailable bool      `json:"-"`
```

The old name was used to avoid a clash with a function that this PR removes

internal/util/telemetry/reporter.go
```diff
@@ -239,8 +239,18 @@ func makeRequest(s *state.State, m *connmetrics.ConnMetrics) *request {
 // with update available and latest version if any update is available.
 func (r *Reporter) report(ctx context.Context) {
 	s := r.P.Get()
+
 	if s.Telemetry != nil && !*s.Telemetry {
 		r.L.Debug("Telemetry is disabled, skipping reporting.")
+
+		// If telemetry is disabled, we can't say anything about update availability or latest version.
+		if err := r.P.Update(func(s *state.State) {
+			s.LatestVersion = ""
```
AlekSi, line 248: No need to do that

internal/util/telemetry/reporter.go
```diff
@@ -239,8 +239,18 @@ func makeRequest(s *state.State, m *connmetrics.ConnMetrics) *request {
 // with update available and latest version if any update is available.
 func (r *Reporter) report(ctx context.Context) {
 	s := r.P.Get()
+
 	if s.Telemetry != nil && !*s.Telemetry {
 		r.L.Debug("Telemetry is disabled, skipping reporting.")
+
+		// If telemetry is disabled, we can't say anything about update availability or latest version.
```
AlekSi, line 246: Well, we can say what the last known latest version was. We just don't use it without checking `UpdateAvailable`

internal/util/telemetry/reporter_test.go
AlekSi: There is no need to reset the info about the latest version anymore. We did that when there was only that field and packed two things into it. Now there is a separate field for that.

rumyantseva: fixed

internal/util/state/state.go
rumyantseva: fixed

internal/util/state/state.go
rumyantseva: fixed

internal/util/telemetry/reporter.go
rumyantseva: fixed

internal/util/telemetry/reporter.go
```diff
@@ -239,8 +239,17 @@ func makeRequest(s *state.State, m *connmetrics.ConnMetrics) *request {
 // with update available and latest version if any update is available.
 func (r *Reporter) report(ctx context.Context) {
 	s := r.P.Get()
+
 	if s.Telemetry != nil && !*s.Telemetry {
 		r.L.Debug("Telemetry is disabled, skipping reporting.")
+
+		// If telemetry is disabled, we can't say anything about update availability, but we keep latest version info.
+		if err := r.P.Update(func(s *state.State) {
+			s.UpdateAvailable = false
```
AlekSi, line 248: We should do when telemetry is disabled, when one day passes and send a report or not

AlekSi (changes requested):

noisersup (dismissed):
Great work! I like the helper functions for enabling/disabling telemetry.

chilagrow (dismissed):

w84thesun (approved):
LGTM

PR #2656, opened 2023-05-17
Implement metadata storage

### LINKED ISSUE #2387: Support basic SQLite backend
We should add support for SQLite to allow using FerretDB without any external database process.

w84thesun:
# Description

Closes #2387.

## Readiness checklist


internal/backends/collection.go
```diff
@@ -59,7 +60,10 @@ type InsertParams struct {
 }
 
 // InsertResult represents the results of Collection.Insert method.
-type InsertResult struct{}
+type InsertResult struct {
+	Errors        *commonerrors.WriteErrors
```
AlekSi, line 64: I'm not sure at all about returning protocol-level errors straight from the backend implementation. That's barely okay for now, but let's not expand it.

internal/handlers/common/insert.go
```diff
@@ -28,6 +30,8 @@ type InsertParams struct {
 	Collection string       `ferretdb:"collection"`
 	Ordered    bool         `ferretdb:"ordered,opt"`
 
+	DocSlice []*types.Document `ferretdb:"-"`
```
AlekSi, line 33: We must not make our existing handlers worse.

Let's pass `*types.Array` to the `Insert` instead of iterator for now

internal/backends/collection.go
w84thesun: We can replace it with `[]error` and wrap it with `*commonerrors.WriteErrors` at the handler level.

internal/handlers/common/insert.go
w84thesun: Fixed.

AlekSi (approved):

Issue #1432, opened 2022-11-16
Implement `$unset` aggregation pipeline stage


rumyantseva:
### What should be done?

https://www.mongodb.com/docs/manual/reference/operator/aggregation/unset/

We should add support for that aggregation stage. It could be done in three steps / PRs:

* [ ] single field;
* [ ] multiple fields (array);
* [ ] dot notation support for both.

### Where?

* https://github.com/FerretDB/FerretDB/tree/main/internal/handlers/common/aggregations/stages
* https://github.com/FerretDB/FerretDB/tree/main/integration

### Definition of Done

- all handlers updated;
- integration/compatibility tests added/updated;
- spot refactorings done.

rumyantseva:
@shibasisp thank you! I assigned the issue. If you have questions regarding anything, please feel free to ask them here or on Slack.

PR #2676, opened 2023-05-21
[#1432] Implement  pipeline stage

shibasisp:
# Description

Closes #{1432}.

Implemented `$unset` aggregation pipeline. Currently this works with:
- [] single field
- [] multiple fields (array);
- [] dot notation support for both.

## Readiness checklist


integration/aggregate_documents_compat_test.go
```diff
@@ -1562,3 +1562,47 @@ func TestAggregateCompatProject(t *testing.T) {
...
 	testAggregateStagesCompat(t, testCases)
 }
+
+func TestAggregateCompatUnset(t *testing.T) {
+	t.Parallel()
+
+	testCases := map[string]aggregateStagesCompatTestCase{
+		"InvalidType": {
+			pipeline: bson.A{
+				bson.D{{"$unset", "invalid"}},
```
rumyantseva, line 1572: Why is is called `InvalidType`? This example looks like a valid `$unset` to me, where we unset a field called `invalid`. If it's about an invalid type, then maybe we can use something like  `bson.D{{"$unset",  false}},` (then the type `bool` is really invalid. 

Let's also add one more test where wrong types are set in an array. For example, something like `{$unset: ["v", int32(42), false]}`

integration/aggregate_documents_compat_test.go
```diff
@@ -1562,3 +1562,47 @@ func TestAggregateCompatProject(t *testing.T) {
...
+			pipeline: bson.A{
+				bson.D{{"$unset", "v"}},
+			},
+		},
+		"UnsetID": {
+			pipeline: bson.A{
+				bson.D{{"$unset", "_id"}},
+			},
+		},
+		"Unset2Fields": {
```
rumyantseva, line 1586: I suggest also adding a test case where we unset both `_id` and `v` fields. So, in the results, we will see some empty docs for some providers.

integration/aggregate_documents_compat_test.go
```diff
@@ -1562,3 +1562,47 @@ func TestAggregateCompatProject(t *testing.T) {
...
+				bson.D{{"$unset", "invalid"}},
+			},
+			resultType: emptyResult,
+		},
+		"Unset1Field": {
+			pipeline: bson.A{
+				bson.D{{"$unset", "v"}},
+			},
+		},
+		"UnsetID": {
```
rumyantseva, line 1581: This test case is a bit tricky: because we unset the field `_id`, we need to specify how we want to sort the results. Otherwise, our compat tests might fail because target and compat systems will return results that are sorted in a different way (or where sorting might vary). To do this, I suggest sorting first and then doing unsets. 

So the pipeline will be something like this:
```go
pipeline: bson.A{
	bson.D{{"$sort", bson.D{{"_id", 1}}}},
	bson.D{{"$unset", "_id"}},
},
```

internal/handlers/common/aggregations/stages/unset.go
```diff
@@ -0,0 +1,99 @@
...
+		defer fieldIter.Close()
+		for {
+			_, field, err := fieldIter.Next()
+			if err != nil {
+				if errors.Is(err, iterator.ErrIteratorDone) {
+					break
+				}
+				return nil, lazyerrors.Error(err)
+			}
+			fieldStr := field.(string)
```
rumyantseva, line 66: If fields are set something like `[1, 2, 3]`, it will panic. I think we need a more accurate check here by returning an error if type is not valid.

rumyantseva (changes requested):
Thank you very much for contributing! 

I see the PR is a draft, but the review was requested, so I leave some comments.

I like the implementation, but I suggest making some additions in tests (see my comments), so we can reveal some problems or errors that need special handling.

Could you also run `task lint` please? We have some specific rules to format code, it helps us keep the code consistent.

AlekSi:
@shibasisp Any news? Anything we could help with?

shibasisp:
Apologies for the delay. @AlekSi I will be addressing the comments by this weekend.

AlekSi (changes requested):
It does not compile

rumyantseva (dismissed):
LGTM. Thanks for contributing!

internal/handlers/common/aggregations/stages/unset.go
```diff
@@ -0,0 +1,119 @@
...
+
+	return &unset{
+		field: fieldsToUnset,
+	}, nil
+}
+
+// Process implements Stage interface.
+func (u *unset) Process(_ context.Context,
+	iter types.DocumentsIterator, closer *iterator.MultiCloser,
+) (types.DocumentsIterator, error) {
```
AlekSi, line 107:
```suggestion
func (u *unset) Process(_ context.Context, iter types.DocumentsIterator, closer *iterator.MultiCloser) (types.DocumentsIterator, error) { //nolint:lll // for readability
```

We don't wrap lines like that

integration/aggregate_documents_compat_test.go
```diff
@@ -145,7 +145,6 @@ func testAggregateStagesCompatWithProviders(t *testing.T, providers shareddata.P
 					}
 				})
 			}
-
```
AlekSi, line 148: Let's keep that newline

AlekSi (approved):

internal/handlers/common/aggregations/stages/unset.go
```diff
@@ -0,0 +1,117 @@
...
+	field *types.Document
+}
+
+// newUnset validates unset document and creates a new $unset stage.
+func newUnset(stage *types.Document) (aggregations.Stage, error) {
+	fields, err := stage.Get("$unset")
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+	fieldsToUnset := must.NotFail(types.NewDocument())
```
noisersup, line 45:
```suggestion
        // fieldsToUnset contains keys with `false` values. They are used to specify projection exclusion later.
	fieldsToUnset := must.NotFail(types.NewDocument())
```

internal/handlers/common/aggregations/stages/unset.go
```diff
@@ -0,0 +1,117 @@
...
+
+	switch fields := fields.(type) {
+	case *types.Array:
+		if fields.Len() == 0 {
+			return nil, commonerrors.NewCommandErrorMsgWithArgument(
+				commonerrors.ErrStageUnsetNoPath,
+				"$unset specification must be a string or an array with at least one field",
+				"$unset (stage)",
+			)
+		}
```
noisersup, line 55: It would be nice to add a couple of testcases to cover those errors if possible.

noisersup (approved):
Thanks for your contribution! The logic here looks good to me. I would only suggest to add a comment as it's not clear at the first glance that `ProjectionDocument` would use exclusion and not inclusion for sure.

Also we might add 3 test cases for $unset field to make a code coverage slightly better:
- empty array
- empty string
- array with empty string

internal/handlers/common/aggregations/stages/unset.go
rumyantseva: I agree with @noisersup, for this line we can add a test case with an empty array as `$unset` param.

internal/handlers/common/aggregations/stages/unset.go
```diff
@@ -0,0 +1,117 @@
...
+					"$unset specification must be a string or an array containing only string values",
+					"$unset (stage)",
+				)
+			}
+
+			fieldsToUnset.Set(fieldStr, false)
+		}
+	case string:
+		if fields == "" {
+			return nil, commonerrors.NewCommandErrorMsgWithArgument(
```
rumyantseva, line 83: And for this line, we can add a test case with an empty string as `$unset` param.

rumyantseva (commented):
Let's add the last couple of tests to cover error codes fully!

rumyantseva (commented):
Let's add the last couple of tests to cover error codes fully!

chilagrow (approved):
Great, thanks for your amazing contribution :hugs:

noisersup (approved):

PR #2713, opened 2023-05-25
Fix path collisions for multiple update operators

### LINKED ISSUE #2539: Handle path collisions for multiple update operators
### Versions

commit.txt: cf0328f6dbb715471908563b62f9ab7c50e77a7d
branch.txt: main
version.txt: v1.0.0-83-gcf0328f6

### What did you do?

See repro in #2538.

### How can we reproduce it?

See above.

Successful update operation.

### What did you see instead?

```
"Updating the path 'a.0.c' would create a conflict at 'a.0.c'"
```

### Environment

```markdown
- OS: Ubuntu
- Architecture: x86_64
- Version: 22.10
- Go: go1.19.3 linux/amd64
- Git: 2.37.2
- Deployment: Docker
- Deployment details: 23.0.3
```

noisersup:
# Description

Closes #2539.

## Readiness checklist


internal/handlers/common/update_test.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/FerretDB/FerretDB/internal/types"
+)
+
+func TestPathValidator(t *testing.T) {
```
rumyantseva, line 26: Hmm, what integration tests cover this logic? It's nice to have unit tests, but we need the logic to be proved by integration tests.

internal/handlers/common/update.go
```diff
@@ -899,10 +899,69 @@ func newUpdateError(code commonerrors.ErrorCode, msg, command string) error {
...
+type pathValidator struct {
+	paths map[string]*pathValidator
+	last  bool
+	root  bool
+}
+
+// validate takes a slice of path keys and checks if the path
+// conflicts with previously validated path. If it's not, then
+// the path is saved in pathValidator.
+func (tree *pathValidator) validate(keys []string) bool {
```
chilagrow, line 930: Does this function work if we have the same key on the path? `{foo: {foo: {foo:}}}` and trying to insert `foo.foo.bar` for example?

internal/handlers/common/update.go
```diff
@@ -899,10 +899,69 @@ func newUpdateError(code commonerrors.ErrorCode, msg, command string) error {
...
+	}
+
+	return tree.paths[key].validate(keys)
+}
+
 // validateOperatorKeys returns error if any key contains empty path or
 // the same path prefix exists in other key or other document.
 func validateOperatorKeys(command string, docs ...*types.Document) error {
-	seen := map[string]struct{}{}
+	seen := newPathValidator()
```
chilagrow, line 964: The variable and func very different naming. We need more appropriate naming here :hugs:

internal/handlers/common/update.go
```diff
@@ -899,10 +899,69 @@ func newUpdateError(code commonerrors.ErrorCode, msg, command string) error {
...
+	key, keys := keys[0], keys[1:]
+
+	if node, ok := tree.paths[key]; ok {
+		// if path exists in validator and it's the last element, conflict is detected
+		if node.last {
+			return false
+		}
+
+		// if there's no more keys in current path,
+		// but child path exists, conflict is detected
```
chilagrow, line 946: I think we should use similar terminology as the code. Does _child path exists_ mean node exists right?

chilagrow (commented):
I don't quite understand the implementation, it seems to work though :see_no_evil:

internal/handlers/common/update.go
noisersup: `foo.foo.foo` does not conflict with `foo.foo.bar`. Both of them conflicts with `foo.foo` though. That's how validator should work here.

chilagrow: I initially didn't understand that the `pathValidator` was created for each node, so this one is clear now!

integration/findandmodify_compat_test.go
```diff
@@ -190,19 +190,21 @@ func TestFindAndModifyCompatUpdate(t *testing.T) {
 			},
 			skipForTigris: "schema validation would fail",
 		},
-		"InvalidOperator": {
```
chilagrow, line 193: Why were these tests removed? If they were duplicate of other tests, it makes sense.

integration/findandmodify_compat_test.go
```diff
@@ -190,19 +190,21 @@ func TestFindAndModifyCompatUpdate(t *testing.T) {
...
+
+		"MultipleOperatorsDotNotation": {
 			command: bson.D{
-				{"query", bson.D{{"_id", bson.D{{"$exists", false}}}}},
-				{"update", bson.D{{"$invalid", "non-existent-field"}}},
+				{"query", bson.D{{"_id", "array-documents-two-fields"}}},
```
chilagrow, line 196: Is it on purpose to match only `array-documents-two-fields`? It's fine if we want to specifically test these cases.

chilagrow (commented):
Looks great, just minor comments

rumyantseva (commented):
A couple of comments about tests.

integration/findandmodify_compat_test.go
noisersup: Yeah we want to test specifically this document

integration/findandmodify_compat_test.go
```diff
@@ -228,6 +230,67 @@ func TestFindAndModifyCompatUpdate(t *testing.T) {
...
+				{"update", bson.D{
+					{"$set", bson.D{{"v.0.field", 4}}},
+					{"$inc", bson.D{{"v", 4}}},
+				}},
+			},
+		},
+	}
+
+	testFindAndModifyCompat(t, testCases)
```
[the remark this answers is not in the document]
noisersup: thanks!

rumyantseva (approved):
LGTM

internal/handlers/common/update.go
```diff
@@ -899,10 +899,73 @@ func newUpdateError(code commonerrors.ErrorCode, msg, command string) error {
...
+// pathValidator stores validated paths and checks for any conflicts
+// between them. The conflict is when one path points to the part of
+// another path.
+//
+// Examples of conflicts:
+// `foo.bar, foo.bar`
+// `foo.bar, foo.bar.x`
+//
+// This is not a conflict:
+// `foo.bar.x, foo.bar.y`.
```
AlekSi, line 920: FWIW, all those examples do not show a need for a tree structure – that's just string prefixes. Is it worth complicating our code there?

noisersup: Checking just for string prefixes might be simpler solution, although there might be some other issues to handle like the fact that `v` is prefix for `vfoo` (without the dot)

AlekSi: Well, not literally, but by segments. Our paths already store segments in a slice.

noisersup: That creates O(n2) time complexity, but I guess we don't expect many paths so it doesn't need to scale. I'm going to try this out then

AlekSi: Human cognition scales much worse :)

chilagrow (dismissed):

internal/handlers/common/update.go
```diff
@@ -902,11 +902,11 @@ func newUpdateError(code commonerrors.ErrorCode, msg, command string) error {
 // validateOperatorKeys returns error if any key contains empty path or
 // the same path prefix exists in other key or other document.
 func validateOperatorKeys(command string, docs ...*types.Document) error {
-	seen := map[string]struct{}{}
+	paths := []*types.Path{}
```
rumyantseva, line 905: I like the var name `seen` more because it showed what was the var about. To me `visistedPath` would be a better name because it has some semantics. But it's rather a minor comment.

internal/handlers/common/update.go
```diff
@@ -918,23 +918,56 @@ func validateOperatorKeys(command string, docs ...*types.Document) error {
...
 		}
 	}
 
 	return nil
 }
 
+// checkSlicePrefix returns true if prefix slice is the beginning of the target slice.
+// The example of slice prefix: target = ["a","b","c"] prefix = ["a","b"];
+// For empty prefix slice it returns false, if target slice is not empty.
+func checkSlicePrefix[T comparable](target, prefix []T) bool {
```
rumyantseva, line 953: Why do we need generics here?

rumyantseva (commented):
Looks good to me overall, but I suggest a minor refactoring to have fewer uncovered lines of code.

internal/handlers/common/update.go
noisersup: As it was easy to make this function more general I did it, but if we prefer strings here I can change that

rumyantseva (approved):
LGTM

rumyantseva (dismissed):

internal/handlers/common/update.go
```diff
@@ -918,23 +918,47 @@ func validateOperatorKeys(command string, docs ...*types.Document) error {
...
+			visistedPaths = append(visistedPaths, &nextPath)
 		}
 	}
 
 	return nil
 }
 
+// checkSlicePrefix returns true if oneaslice is the beginning of another slice.
+// The example of slice prefix: arr1 = ["a","b","c"] arr2 = ["a","b"];
+// For empty prefix slice it returns false, if target slice is not empty.
```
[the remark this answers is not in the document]
noisersup: sounds more clear thanks:)

rumyantseva (dismissed):

w84thesun (dismissed):
👍

internal/handlers/common/update.go
```diff
@@ -902,11 +902,11 @@ func newUpdateError(code commonerrors.ErrorCode, msg, command string) error {
 // validateOperatorKeys returns error if any key contains empty path or
 // the same path prefix exists in other key or other document.
 func validateOperatorKeys(command string, docs ...*types.Document) error {
-	seen := map[string]struct{}{}
+	visistedPaths := []*types.Path{}
```
chilagrow, line 905:
```suggestion
	visitedPaths := []*types.Path{}
```

internal/handlers/common/update.go
```diff
@@ -918,23 +918,47 @@ func validateOperatorKeys(command string, docs ...*types.Document) error {
...
+// If both slices are empty, it returns true. If one slice is empty and another slice is not, it returns false.
+func checkSlicePrefix(arr1, arr2 []string) bool {
+	target, prefix := arr1, arr2
+
+	if len(target) < len(prefix) {
+		target, prefix = prefix, target
+	}
+
+	if len(prefix) == 0 {
+		return len(target) == 0
```
chilagrow, line 950: Codecov thinks this is not tested, could this be tested?

noisersup: Hmm not really (with integration tests), as path cannot be empty (empty string would result in slice with 1 item).

internal/handlers/common/update.go
noisersup: Nice catch!

w84thesun (approved):

rumyantseva (approved):

internal/handlers/common/update.go
chilagrow: Okie, I though it might work with `.` path. But might not.

chilagrow (approved):

PR #2714, opened 2023-05-25
Implement `envtool version` command

### LINKED ISSUE #2684: Implement `envtool version` command
### What should be done?

We should implement a new `envtool` command that is very similar to `shell read` subcommand, but reads a fixed file build/version/version.txt and omits the leading `v` in the output. It should exit with non-zero exit code on any error.

See #2685 for the reason why we need that command.

### Where?

https://github.com/FerretDB/FerretDB/tree/main/cmd/envtool

### Definition of Done

`go run ./cmd/envtool version` produces something like `1.2.0-1-g6170f970a`.

jeremyphua:
# Description

Closes #2684.

## Readiness checklist

### DIFF
--- a/cmd/envtool/envtool.go
+++ b/cmd/envtool/envtool.go
@@ -364,8 +364,8 @@ func read(w io.Writer, paths ...string) error {
 }
 
 // printVersion will print out version omitting leading v
-func printVersion(w io.Writer) error {
-	b, err := os.ReadFile(versionFile)
+func printVersion(w io.Writer, file string) error {
+	b, err := os.ReadFile(file)
 	if err != nil {
 		return err
 	}
@@ -425,7 +425,7 @@ func main() {
 	case "shell read <path>":
 		err = read(os.Stdout, cli.Shell.Read.Paths...)
 	case "version":
-		err = printVersion(os.Stdout)
+		err = printVersion(os.Stdout, versionFile)
 	default:
 		err = fmt.Errorf("unknown command: %s", cmd)
 	}
--- a/cmd/envtool/envtool.go
+++ b/cmd/envtool/envtool.go
@@ -54,6 +54,7 @@ var (
 	errorTemplate = template.Must(template.New("error").Option("missingkey=error").Parse(string(errorTemplateB)))
 )
 
+// versionFile contains version information with leading v.
 const versionFile = "build/version/version.txt"
 
 // waitForPort waits for the given port to be available until ctx is done.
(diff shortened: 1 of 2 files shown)


cmd/envtool/envtool.go
```diff
@@ -361,6 +363,18 @@ func read(w io.Writer, paths ...string) error {
 	return nil
 }
 
+// printVersion will print out version omitting leading v
```
rumyantseva, line 366: The linter complains about one tiny thing: the comments should be ended with `.`

cmd/envtool/envtool.go
```diff
@@ -54,6 +54,8 @@ var (
 	errorTemplate = template.Must(template.New("error").Option("missingkey=error").Parse(string(errorTemplateB)))
 )
 
+const versionFile = "build/version/version.txt"
```
rumyantseva, line 57: We usually document all unexported consts too, so let's add some description?

For example:

```suggestion
// versionFile contains version information with leading v.
const versionFile = "build/version/version.txt"
```

cmd/envtool/envtool.go
```diff
@@ -361,6 +363,18 @@ func read(w io.Writer, paths ...string) error {
...
 }
 
+// printVersion will print out version omitting leading v
+func printVersion(w io.Writer, file string) error {
+	b, err := os.ReadFile(file)
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(w, string(b)[1:])
```
rumyantseva, line 373: I suggest before trimming the data, we check that the string in the file has prefix `v`, and trim it only in this case. 
Currently, our versions always have the prefix `v`, but overall it looks like a nice extra check. E.g. if the prefix is `v` we trim it and return the rest, otherwise we return data as is.

rumyantseva (commented):
Thank you for clarifying the task and contributing! This approach looks good to me, it's consistent with the other commands we have in `envtool`. 

I tried it on Mac and Windows, and it works.

I left a couple of minor comments.

cmd/envtool/envtool.go
jeremyphua: Done!

cmd/envtool/envtool.go
jeremyphua: Done!

cmd/envtool/envtool.go
jeremyphua: Done!

AlekSi (approved):

AlekSi:
Thank you!

rumyantseva (approved):
Thanks a lot!

PR #2728, opened 2023-05-29
Remove `sjson.GetTypeOfValue` helper

noisersup:
# Description

It does basically the same thing as `commonparams.AliasFromType`.

## Readiness checklist


AlekSi (changes requested):
They _happen_ to do the same thing.

`sjson` layer is responsible for preserving [BSON-type](https://bsonspec.org/spec.html) information in storage. Currently, we store type information in the special schema field. For example, for "64-bit binary floating point" value, we store a string "double". That string could be replaced with a number or even a bitmask as long as we preserve backward compatibility for client applications (and not even that if we bump a major version).

Type aliases are [fixed](https://www.mongodb.com/docs/manual/reference/bson-types/) and cannot be changed without breaking compatibility with MongoDB.

What we need to do instead is:
* make this difference clear from code doc comments;
* check that we don't use a wrong function somewhere (spoiler: we do at least in once place)

internal/handlers/pg/pgdb/query.go
```diff
@@ -299,11 +300,11 @@ func prepareWhereClause(p *Placeholder, sqlFilters *types.Document) (string, []a
 						// type not supported for pushdown
 
 					case float64, bool, int32, int64:
-						filters = append(filters, fmt.Sprintf(sql, p.Next(), p.Next(), sjson.GetTypeOfValue(v)))
+						filters = append(filters, fmt.Sprintf(sql, p.Next(), p.Next(), commonparams.AliasFromType(v)))
```
rumyantseva, line 303: This is the part I don't understand. I see why it's safe to use `commonparams.AliasFromType` to show type when we return errors (we just return it, we don't need to compare etc), but in this place, I'm not sure if I understand the use case. Is it safe?

internal/handlers/pg/pgdb/query.go
```diff
@@ -299,11 +300,11 @@ func prepareWhereClause(p *Placeholder, sqlFilters *types.Document) (string, []a
...
 						args = append(args, rootKey, v)
 
 					case string, types.ObjectID, time.Time:
-						filters = append(filters, fmt.Sprintf(sql, p.Next(), p.Next(), sjson.GetTypeOfValue(v)))
+						filters = append(filters, fmt.Sprintf(sql, p.Next(), p.Next(), commonparams.AliasFromType(v)))
```
chilagrow, line 307: Our pg supports less types than bson does (`commonparams.AliasFromType`), so for this usage we want stricter `sjson.GetTypeOfValue` I think.

AlekSi (approved):

AlekSi (approved):

rumyantseva (approved):
👌

PR #2730, opened 2023-05-29
Replace document slices with iterators

### LINKED ISSUE #2723: Use iterators in `distinct` command
### What should be done?

`distinct` command should use iterators for documents instead of slices similarly to `count`, `find` and other commands.

### Where?

- `handlers` package.

### Definition of Done
- commands like `distinct`, `count`, and `find` doesn't preload documents to slices, but operate on iterators.
- unit tests updated;
- spot refactorings done;

noisersup:
# Description

Closes #2723.

## Readiness checklist


AlekSi (commented):
If we already made changes in `count`, and they work, let's keep them

internal/handlers/pg/msg_count.go
```diff
@@ -50,26 +51,40 @@ func (h *Handler) MsgCount(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, e
...
 
 	var resDocs []*types.Document
 	err = dbPool.InTransaction(ctx, func(tx pgx.Tx) error {
-		resDocs, err = fetchAndFilterDocs(ctx, &fetchParams{tx, &qp, h.DisableFilterPushdown})
-		return err
+		var iter types.DocumentsIterator
+
+		iter, _, err = pgdb.QueryDocuments(ctx, tx, &qp)
+		if err != nil {
+			return err
```
rumyantseva, line 64: Should this one be a lazy error?

internal/handlers/pg/msg_distinct.go
```diff
@@ -47,23 +48,52 @@ func (h *Handler) MsgDistinct(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg
...
-		return err
-	})
+		var iter types.DocumentsIterator
 
-	if err != nil {
-		return nil, err
-	}
+		iter, _, err = pgdb.QueryDocuments(ctx, tx, &qp)
+		if err != nil {
+			return err
```
rumyantseva, line 64: Should this one be a lazy error?

internal/handlers/pg/msg_distinct.go
```diff
@@ -47,23 +48,52 @@ func (h *Handler) MsgDistinct(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg
...
+		}
+
+		closer := iterator.NewMultiCloser(iter)
+		defer closer.Close()
+
+		iter = common.FilterIterator(iter, closer, qp.Filter)
+
+		iter, err = common.DistinctIterator(iter, closer, dp.Key)
+		if err != nil {
+			return err
```
rumyantseva, line 74: Same here, looks like a potential lazy error

internal/handlers/pg/msg_distinct.go
```diff
@@ -47,23 +48,52 @@ func (h *Handler) MsgDistinct(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg
...
+		}
+
+		closer := iterator.NewMultiCloser(iter)
+		defer closer.Close()
+
+		iter = common.FilterIterator(iter, closer, qp.Filter)
+
+		iter, err = common.DistinctIterator(iter, closer, dp.Key)
```
rumyantseva, line 72: why `closer`, and not `iter`?

rumyantseva (commented):
Asked a few questions.

internal/handlers/pg/msg_count.go
noisersup: `QueryDocuments` already returns lazyerror

internal/handlers/pg/msg_distinct.go
noisersup: Same as above

internal/handlers/common/distinct_iterator.go
```diff
@@ -0,0 +1,131 @@
...
+	"errors"
+	"sync/atomic"
+
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/iterator"
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+	"github.com/FerretDB/FerretDB/internal/util/must"
+)
+
+// DistinctIterator returns an iterator that returns a single document containing
```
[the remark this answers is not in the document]
noisersup: > If we always return a single document as a result, why do we need to return an iterator at all? 

We do the same for `count_iterator`. Not sure what's the exact architectural decision behind that, but I assume that we aim to wrap iterators in iterators to make every "stage" of query processing more consistent(?)

> (so, each new value is returned when Next() is called)

It would still require to store those distinct values in memory, as they need to be sorted.

internal/handlers/pg/msg_count.go
rumyantseva: Nested lazy errors help to trace the chain of exact error calls, so it might make sense to return them everywhere where they are needed. Then in case they happen, it will be easier to debug the chain of errors.

internal/handlers/common/distinct_iterator.go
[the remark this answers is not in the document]
rumyantseva: @AlekSi could you explain the reason, please? It seems that I don't really understand the idea of having iterators for these cases.

internal/handlers/common/distinct_iterator.go
```diff
@@ -0,0 +1,131 @@
...
+// the distinct values from documents on the specified key.
+// It will be added to the given closer.
+//
+// Next method return the distinct value of all documents in iterator.
+//
+// Close method closes the underlying iterator.
+func DistinctIterator(iter types.DocumentsIterator, closer *iterator.MultiCloser, key string) (types.DocumentsIterator, error) {
+	res := &distinctIterator{
+		iter: iter,
+	}
```
chilagrow, line 37: I think we can check path error, then initialize `distinctIterator`, so this can be moved to line L43, right before closer. I find having closer and iterator together is easier to read

internal/handlers/common/distinct_iterator.go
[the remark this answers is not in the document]
chilagrow: Unlike filter, count accumulate the values from multiple documents to a single one. Distinct on the other hand _could_ return each distinct values as a document, but this is quite awkward. For simplicity, returning it in single document is easier.

Let's assume we have input to distinct like `[{v: 1}, {v: [1, 23]}, {v:true}, {v: 23}, {v:true}]`

current impl for calling Next() returns like this:
```
Next() -> `{'v': bson.Array[1, 23, true]}
Next() -> iterator.ErrIteratorDone
```
if we want to return each value, we would wrap it in document something like below? Also it may complicate array values such as `[1, 23]` as we need to keep in memory for current index of a document and array currently being worked on. 
```
Next() -> `{'v':1}
Next() -> `{'v': 23}
Next() -> `{'v': true}
Next() -> iterator.ErrIteratorDone
```
At the end, we want to return `bson.Array[1, 23, true]` to wire. :hugs:

internal/handlers/common/distinct_iterator.go
```diff
@@ -0,0 +1,131 @@
...
+				if !distinct.Contains(val) {
+					distinct.Append(val)
+				}
+			}
+		}
+	}
+
+	SortArray(distinct, types.Ascending)
+
+	return unused, must.NotFail(types.NewDocument(iter.path.Suffix(), distinct)), nil
```
chilagrow, line 120: Here we really prefer to return `(struct{}, *types.Array, error)` but that's not what interface wants. Could you document in doc string of this function that we create document and use suffix key?

internal/handlers/common/distinct_iterator.go
```diff
@@ -0,0 +1,131 @@
...
+	iter types.DocumentsIterator
+	path types.Path
+	done atomic.Bool
+}
+
+// Next implements Iterator Interface.
+// The first call returns the array of distinct documents from underlyin iterator,
+// subsequent calls return ErrIteratorDone.
+// If iterator contains no document, it returns empty array.
+func (iter *distinctIterator) Next() (struct{}, *types.Document, error) {
```
chilagrow, line 61: Example would be nice to demonstrate what it returns :hugs:

chilagrow: Also distinct is special about handling array index dot notation, I think it ignores it, right?
And if path contains array, it fetches elements in the array, right? Could you document them please?

chilagrow (commented):
Just few comments :hugs:

internal/handlers/pg/msg_count.go
AlekSi: Otherwise, that would be a broken chain :)

internal/handlers/common/distinct_iterator.go
```diff
@@ -0,0 +1,131 @@
...
+	var unused struct{}
+
+	distinct := types.MakeArray(0)
+
+	done := iter.done.Swap(true)
```
AlekSi, line 66: #2759

internal/handlers/common/distinct_iterator.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// Next implements Iterator Interface.
+// The first call returns the array of distinct documents from underlyin iterator,
+// subsequent calls return ErrIteratorDone.
+// If iterator contains no document, it returns empty array.
+func (iter *distinctIterator) Next() (struct{}, *types.Document, error) {
+	var unused struct{}
+
+	distinct := types.MakeArray(0)
```
AlekSi, line 64: I think the `types.Array` is not the best structure for representing a set. Is it because of the custom logic in `types.Array.Contains`? I'm not sure

internal/handlers/common/distinct_iterator.go
AlekSi: See also #2759. The "constructor" should be documented; methods of unexported type should refer to it.

internal/handlers/common/distinct_iterator.go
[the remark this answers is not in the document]
AlekSi: `CountIterator` is being used in the `$count` stage, and all aggregation pipeline stages return an iterator over documents.

But here, we make an iterator for the `distinct` command, not for some aggregation pipeline stage that should fit a specific interface. I think returning just `*types.Array` makes sense. That issue was about _accepting_ the iterator instead of the slice, not about _returning_ the iterator.

internal/handlers/tigris/msg_distinct.go
```diff
@@ -42,18 +43,28 @@ func (h *Handler) MsgDistinct(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg
...
+
 	qp := tigrisdb.QueryParams{
 		DB:         dp.DB,
 		Collection: dp.Collection,
-		Filter:     dp.Filter,
 	}
 
-	resDocs, err := fetchAndFilterDocs(ctx, &fetchParams{dbPool, &qp, h.DisableFilterPushdown})
+	if !h.DisableFilterPushdown {
+		qp.Filter = filter
```
rumyantseva, line 54: `filter` variable is used only once, right? isn't it possible to do something like `qp.Filter = dp.Filter`? I don't really understand why we need a new variable there.

rumyantseva (approved):
LGTM

internal/handlers/tigris/msg_distinct.go
noisersup: It's used twice: here to setup queryparam filter, and on the line 65 to apply filters after fetching documents. We could remove this variable and get value from `dp.Filter` in both places, although it seemed to me a bit unclear as it's easy to confuse `qp.Filter` with `dp.Filter`.

w84thesun (approved):

internal/handlers/common/distinct_iterator.go
chilagrow: Good to know, because I had the same experience with the map key in 
 $group` https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/common/aggregations/stages/group.go#L245 :see_no_evil:

chilagrow (approved):
Great work!

AlekSi (approved):

Issue #226, opened 2022-01-05
Support field extraction into separate columns


AlekSi:
It should be possible to extract common document fields like `_id` into separate PostgreSQL columns. That would make documents smaller (because those field names will not be repeated in every document) and may also make indexes faster. That should also allow us to support more PostgreSQL-like backends and their sharding.

Should be done after #179.

Let's do it only during collection creation and don't move data live.

```[tasklist]
### Tasks
- [ ] #3252
```

PR #2731, opened 2023-05-29
Refactor SQLite handler

AlekSi:
# Description

It might be easier to review a result than changes:
* https://github.com/AlekSi/FerretDB/tree/metadata/internal/backends
* https://github.com/AlekSi/FerretDB/tree/metadata/internal/handlers/sqlite

## Readiness checklist


internal/backends/sqlite/collection.go
```diff
@@ -32,304 +30,146 @@ import (
...
-		if err != nil {
-			// TODO: check error
-			tx.Rollback()
-		}
-	}()
+	// TODO https://github.com/FerretDB/FerretDB/issues/2750
 
-	var inserted int64
+	db := c.r.DatabaseGetExisting(ctx, c.dbName)
+	query := fmt.Sprintf(`INSERT INTO %q (_ferretdb_sjson) VALUES (?)`, c.r.CollectionToTable(c.name))
```
chilagrow, line 79: We are changing the column name here, I believe sqlite is still experimental so backwards compatibility isn't an issue I suppose.

internal/backends/sqlite/collection.go
```diff
@@ -32,304 +30,146 @@ import (
...
 	}
 
-	var committed bool
+	tableName := c.r.CollectionToTable(c.name)
 
-	defer func() {
-		if committed {
-			return
-		}
+	query := fmt.Sprintf(`UPDATE %q SET _ferretdb_sjson = ? WHERE json_extract(_ferretdb_sjson, '$._id') = ?`, tableName)
```
chilagrow, line 122: This is great :hugs: , only tableName now needs `%q`

internal/handlers/sqlite/msg_insert.go
```diff
@@ -41,28 +40,31 @@ func (h *Handler) MsgInsert(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
-	if len(res.Errors) > 0 {
-		var errs *commonerrors.WriteErrors
-
-		for i := 0; i < len(res.Errors); i++ {
-			errs.Append(err, int32(i))
-		}
-
-		replyDoc = errs.Document()
-	}
+	// TODO
```
chilagrow, line 58: I suppose this is okay like this for now?

chilagrow (approved):
Looks good to me, just a couple of question

internal/backends/sqlite/collection.go
AlekSi: Yeah, and now we use a reserved prefix so we could implement #226 more easily

internal/handlers/sqlite/msg_insert.go
AlekSi: Added the issue link

internal/backends/sqlite/collection.go
chilagrow: Yeah that's good direction :+1:

internal/backends/sqlite/collection.go
chilagrow: That's convenient for this!

chilagrow (approved):
:rocket:

rumyantseva (approved):
I mostly checked structure and what is in which package. The approach looks reasonable!

I need an explanation about packages inside `sqlite` handler: why `pool` is inside `metadata` and what is considered as `metadata` in geneal.

AlekSi:
That's exactly the problem I mentioned on DS calls. Basically, `metadata` and `pool` could be in the same package – except that they will be able to access each other's unexported methods and fields, and that would be a mess.

PR #2814, opened 2023-06-07
Implement `createIndexes` for unique indexes

### LINKED ISSUE #2045: Implement `createIndexes` for unique indexes
### What should be done?

Support `unique` option for `createIndexes` command.

In the pgdb package, the `CreateIndexIfNotExists` function already accepts `pgdb.Index` struct and supports `unique` option. The only thing that needs to be done is implementation of the logic on the handler level.

### Where

`MsgCreateIndexes`

### Definition of done

- [x]  Implement the creation of unique indexes in the handler.
- [x] Compat tests: check that after creating a unique index `listIndexes` returns the same results for target and compat.
- [x] Compat tests: check that an attempt to insert a duplicate value in a field with a unique index will return the correct error. Bonus points: create a unique index with a few fields and check the same.
- [x] Unit tests: it's possible to check that creating a unique index causes the correct index definition to be set, see `TestCreateIndexIfNotExists`.
- [x] Check TODO in the code.
- [ ] Update supported commands documentation.

w84thesun:
# Description

Closes #2045.

## Readiness checklist


internal/handlers/pg/msg_createindexes.go
```diff
@@ -122,6 +122,13 @@ func (h *Handler) MsgCreateIndexes(ctx context.Context, msg *wire.OpMsg) (*wire.
 			"One of the specified indexes already exists with a different key",
 			document.Command(),
 		)
+	case errors.Is(err, pgdb.ErrUniqueViolation):
+		return nil, commonerrors.NewCommandErrorMsgWithArgument(
+			commonerrors.ErrDuplicateKeyInsert,
+			"Index build failed",
+			document.Command(),
+		)
```
chilagrow, line 130: Could you add integration test for checking error message please. Example https://github.com/FerretDB/FerretDB/blob/main/integration/indexes_test.go#L176 :hugs:

internal/handlers/pg/msg_insert.go
```diff
@@ -153,14 +152,11 @@ func insertDocument(ctx context.Context, tx pgx.Tx, qp *pgdb.QueryParams, doc an
...
-		// TODO Extend message for non-_id unique indexes in https://github.com/FerretDB/FerretDB/issues/2045
-		idMasrshaled := must.NotFail(json.Marshal(must.NotFail(d.Get("_id"))))
-
 		return commonerrors.NewWriteErrorMsg(
-			commonerrors.ErrDuplicateKey,
+			commonerrors.ErrDuplicateKeyInsert,
 			fmt.Sprintf(
-				`E11000 duplicate key error collection: %s.%s index: _id_ dup key: { _id: %s }`,
-				qp.DB, qp.Collection, idMasrshaled,
+				`E11000 duplicate key error collection: %s.%s`,
```
chilagrow, line 158: Also could you add test for this error message please.

internal/handlers/pg/pgdb/database_metadata.go
```diff
@@ -467,6 +467,11 @@ func (ms *metadataStorage) setIndex(ctx context.Context, index string, key Index
 	pgTable = metadata.table
 	pgIndex = indexNameToPgIndexName(ms.collection, index)
 
+	switch unique.(type) {
+	case nil:
```
chilagrow, line 471: Where does this `nil` come from? It's not bson type, I'm curious about this special nil handling :thinking:

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -29,7 +32,7 @@ import (
 type Index struct {
 	Name   string
 	Key    IndexKey
-	Unique bool
+	Unique any
```
chilagrow, line 35: Why is it no longer bool? Also what does `unique = types.NullType{}` represent in our metaIndex? A comment / doc string is appreciated here :hugs: 

If we want to specify unique, not unique and unset, then perhaps `*bool` is more suitable :thinking:

integration/indexes_compat_test.go
```diff
@@ -751,3 +751,124 @@ func TestDropIndexesCommandCompat(t *testing.T) {
...
+	t.Parallel()
+
+	for name, tc := range map[string]struct { //nolint:vet // for readability
+		models    []mongo.IndexModel // required, index to create
+		insertDoc bson.D             // required, document to insert for uniqueness check
+		new       bool               // optional, insert new document before check uniqueness
+
+		skip string // optional, skip test with a specified reason
+	}{
+		"ExistingFieldIndex": {
```
chilagrow, line 767: Could you test the case with setting unique to `_id` please.

integration/insert_compat_test.go
```diff
@@ -124,18 +125,20 @@ func testInsertCompat(t *testing.T, testCases map[string]insertCompatTestCase) {
...
 						if targetErr != nil {
-							// Skip inserts that could not be performed due to Tigris schema validation.
-							var e mongo.BulkWriteException
-							if errors.As(targetErr, &e) && e.HasErrorCode(documentValidationFailureCode) {
-								setup.SkipForTigrisWithReason(t, targetErr.Error())
-							}
+							switch targetErr := targetErr.(type) { //nolint:errorlint // don't inspect error chain.
+							case mongo.WriteException:
+								// Skip inserts that could not be performed due to Tigris schema validation.
+								if targetErr.HasErrorCode(documentValidationFailureCode) {
```
chilagrow, line 131: I thought previously it threw `mongo.BulkWriteException` but now it seems to be working with `mongo.WriteException`, not quite sure why. Maybe it was using errors.As and it unwrapped `BulkWriteException` previously.

integration/indexes_compat_test.go
```diff
@@ -751,3 +751,124 @@ func TestDropIndexesCommandCompat(t *testing.T) {
...
+			models: []mongo.IndexModel{
+				{
+					Keys:    bson.D{{"v", 1}},
+					Options: new(options.IndexOptions).SetUnique(true),
```
chilagrow, line 771: Could you add a test for non boolean value of  unique please? I'm not sure if it returns error or succeeds, you may be able to add a test case to this in case of error https://github.com/FerretDB/FerretDB/blob/main/integration/indexes_test.go#L176

chilagrow (commented):
Great work so far! I have a few suggestions :hugs:

integration/indexes_compat_test.go
w84thesun: Added.

integration/indexes_compat_test.go
w84thesun: Added [here](https://github.com/FerretDB/FerretDB/pull/2814/files#diff-b2e4c964b500457dfe583f5f7f7138a8871544e5338b53bbc0d8a02721a9e46bR375).

internal/handlers/pg/msg_createindexes.go
w84thesun: I can't reproduce it now. It was probably about creating an index for a not existing field, but we now have a separate [issue](https://github.com/FerretDB/FerretDB/issues/2830).

noisersup (dismissed):
Looks great!

internal/handlers/pg/pgdb/indexes.go
chilagrow: I see it now that unset `unique` checks that `_id` cannot be unique.

internal/handlers/pg/pgdb/database_metadata.go
chilagrow: ok `nil` is no longer in the code base.

chilagrow:
Is this case handled? The data is from task `env-data`. If not maybe this is something we might need to handle in other issue. 

```
test> db.ArrayInt32s.createIndex({v: 1}, {unique: false})
v_1
test> db.ArrayInt32s.createIndex({v: 1}, {unique: true})
MongoServerError: An existing index has the same name as the requested index. When index names are not specified, they are auto generated and can cause conflicts. Please refer to our documentation. Requested index: { v: 2, unique: true, key: { v: 1 }, name: "v_1" }, existing index: { v: 2, key: { v: 1 }, name: "v_1" }
```

chilagrow (commented):
So many edge cases with unique index creation :open_mouth:

My main comments are just testing error messages, others are not so important.

internal/handlers/pg/msg_createindexes.go
chilagrow: Maybe this is related to below? I just used `Doubles` collection as it is from `task env-data` and it somehow does not create index on MongoDB. It's strange I don't understand mongodb what is is checking :thinking: 
```
test> db.Doubles.createIndex({v: 1}, {unique: true})
MongoServerError: Index build failed: 3c540fe6-c241-4c3f-a0e7-c0dc2c0af62a: Collection test.Doubles ( 6fceefe4-8aab-457a-afc5-f3696e9585b4 ) :: caused by :: E11000 duplicate key error collection: test.Doubles index: v_1 dup key: { v: -9.223372036854776e+18 }

```

integration/indexes_compat_test.go
w84thesun: Added a test to `TestCreateIndexesInvalidSpec`.

integration/indexes_compat_test.go
w84thesun: Added a test to `TestCreateIndexesInvalidSpec`.

internal/handlers/pg/msg_createindexes.go
w84thesun: We have a test for creating a unique index for an existing field. I'm not sure that it depends on the data in a collection. I can't come up with a test case that would trigger a unique violation in PG.  I created an issue for adding this test case.

chilagrow: This one is just totally strange, but yeah let's not make it part of this issue, it looks I don't understand how mongodb works :see_no_evil:

integration/indexes_compat_test.go
chilagrow: Perfect, thanks!

integration/indexes_compat_test.go
chilagrow: Great thanks!

chilagrow (dismissed):
Great work!

noisersup (approved):
LGTM (There are some merge conflicts to resolve)

internal/handlers/pg/pgdb/indexes.go
```diff
@@ -218,11 +226,22 @@ func createPgIndexIfNotExists(ctx context.Context, tx pgx.Tx, schema, table, ind
 	sql := `CREATE` + unique + ` INDEX IF NOT EXISTS ` + pgx.Identifier{index}.Sanitize() +
 		` ON ` + pgx.Identifier{schema, table}.Sanitize() + ` (` + strings.Join(fieldsDef, `, `) + `)`
 
-	if _, err = tx.Exec(ctx, sql); err != nil {
+	_, err = tx.Exec(ctx, sql)
```
rumyantseva, line 229: I see a bit of inconsistency: above you use `if err :=` and here you don't use it.
But it's not critical for me.

internal/handlers/pg/pgdb/database_metadata.go
```diff
@@ -396,11 +401,19 @@ func metadataToDocument(metadata *metadata) *types.Document {
 			keyDoc.Set(pair.Field, int32(pair.Order)) // order is set as int32 to be sjson-marshaled correctly
 		}
 
+		// If unique is nil, it wasn't passed with index definition.
```
rumyantseva, line 404: I'm not sure about this approach -- if we should set it all, or we just shouldn't store `unique` field in this case, or we should set it as `false`. But I understand that it could cause some difference in how FerretDB responds vs how MongoDB responds (e.g. FerretDB responds with `false`, and MongoDB with `nil`), so I'm ok to have it like this for now.

rumyantseva (approved):
LGTM overall. I have a couple of minor points, but I suggest we leave them for now to get the task done.

Issue #2692, opened 2023-05-23
Implement `envtool tests shard` command


AlekSi:
### What should be done?

We want to make CI faster by splitting integration tests into multiple parallel GitHub Actions jobs. Currently, we split them by handler (PostgreSQL, SQLite, etc), but we could also split tests of one handler. For example, we could run `go test -run=^(TestOne|BenchmarkTwo)$` in one job, `go test -run=^(ExampleThree|FuzzFour)$` in another, etc. But we don't want to update our CI configurations every time we add or remove a top-level integration test.

Let's implement `envtool tests shard` command. It would output the value of the `-run` flag that we would then pass to `go test`. For example:

```
$ envtool tests shard -index=0 -total=2

^(TestOne|BenchmarkTwo)$

$ envtool tests shard -index=1 -total=2

^(ExampleThree|FuzzFour)$
```

It then could be used as `go test -run=$(envtool tests shard -index=0 -total=2)`.

A list of test functions could be received using `go test -list=.` command. `envtool tests shard` should call it in the `integartion` directory, split output into lines, sort them, then iterate over them and split them into multiple shards.

This task is only about implementing that command and testing it manually. CI integration is not in the scope.

### Where?

- https://github.com/FerretDB/FerretDB/tree/main/cmd/envtool

### Definition of Done

- `envtool tests shard -index=0 -total=2` command works;
- there are some basic tests that actually run `go test -list=.` and shard results in place.

AlekSi:
@kropidlowsky PTAL

kropidlowsky:
@AlekSi Could you please assign it to me?

PR #2816, opened 2023-06-07
Add tests naming section to `CONTRIBUTING.md`

rumyantseva:
# Description

Closes FerretDB/engineering#66.

- Inspired by #2708 and the brainstorming we had after that.
- Feel free to share your thoughts about the items that look questionable.
- Please note that our current naming doesn't always follow the provided guidelines, so after choosing the final guidelines we will have separate PRs or issues to refactor the tests to meet the guidelines. 
- This is the first version of the guidelines, we can adjust it later based on our needs.
- I set the label "do not merge", so everyone can take a look.

## Readiness checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -233,16 +233,16 @@ Also, we should use driver methods as much as possible instead of testing comman
 
 #### Integration tests naming guidelines
 
-1. Test files should be named after the command they test. 
-   For example, `find_test.go` for testing the find command. 
+1. Test files should be named after the command they test.
+   For example, `find_test.go` for testing the find command.
    Test files belonging to the same commands group are organized into folders named after this group,
    such as `aggregation`, `operations`, `diagnostic`, etc.
 2. If the file contains compatibility tests, add the `_compat` suffix. For example, `find_compat_test.go`.
-3. Test names should include the name of the command being tested. 
+3. Test names should include the name of the command being tested.
    For instance, `TestDistinct` for testing the distinct command.
-4. Compatibility tests should have `Compat` in the name, following the command. 
+4. Compatibility tests should have `Compat` in the name, following the command.
    For example, `TestDistinctCompat`.
-5. Test names should be descriptive and provide information about the functionality or condition being tested. 
+5. Test names should be descriptive and provide information about the functionality or condition being tested.
    If the test is checking for a specific error scenario, include the error scenario in the name.
 6. Keep test names concise, avoiding overly cryptic names. Use abbreviations when appropriate.
 7. Avoid including test data in the name to maintain clarity and prevent excessively long names.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -231,6 +231,24 @@ Writing separate tests might be much better than making a single test that check
 
 Also, we should use driver methods as much as possible instead of testing commands directly via `RunCommand`.
 
+#### Integration tests naming guidelines
+
+1. Test files should be named after the command they test. 
+   For example, `find_test.go` for testing the find command. 
+   Test files belonging to the same commands group are organized into folders named after this group,
+   such as `aggregation`, `operations`, `diagnostic`, etc.
+2. If the file contains compatibility tests, add the `_compat` suffix. For example, `find_compat_test.go`.
+3. Test names should include the name of the command being tested. 
+   For instance, `TestDistinct` for testing the distinct command.
+4. Compatibility tests should have `Compat` in the name, following the command. 
+   For example, `TestDistinctCompat`.
+5. Test names should be descriptive and provide information about the functionality or condition being tested. 
+   If the test is checking for a specific error scenario, include the error scenario in the name.
+6. Keep test names concise, avoiding overly cryptic names. Use abbreviations when appropriate.
+7. Avoid including test data in the name to maintain clarity and prevent excessively long names.
+8. Test case names should follow TitleCase capitalization style.
+9. Test and subtest names should not exceed 64 characters.
+
 ### Submitting code changes
 
 #### Before submitting PR
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -246,7 +246,7 @@ Also, we should use driver methods as much as possible instead of testing comman
    If the test is checking for a specific error scenario, include the error scenario in the name.
 6. Keep test names concise, avoiding overly cryptic names. Use abbreviations when appropriate.
 7. Avoid including test data in the name to maintain clarity and prevent excessively long names.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -231,6 +231,24 @@ Writing separate tests might be much better than making a single test that check
...
+3. Test names should include the name of the command being tested.
+   For instance, `TestDistinct` for testing the distinct command.
+4. Compatibility tests should have `Compat` in the name, following the command.
+   For example, `TestDistinctCompat`.
+5. Test names should be descriptive and provide information about the functionality or condition being tested.
+   If the test is checking for a specific error scenario, include the error scenario in the name.
+6. Keep test names concise, avoiding overly cryptic names. Use abbreviations when appropriate.
+7. Avoid including test data in the name to maintain clarity and prevent excessively long names.
+8. Test case names should follow `TitleCase` capitalization style.
+9. Test and subtest names should not exceed 64 characters.
```
chilagrow, line 250: In addition, 64 characters rule also adds `pg`, `sqlite` etc suffix, so it would be a bit shorter. :see_no_evil:

CONTRIBUTING.md
```diff
@@ -231,6 +231,24 @@ Writing separate tests might be much better than making a single test that check
 
 Also, we should use driver methods as much as possible instead of testing commands directly via `RunCommand`.
 
+#### Integration tests naming guidelines
+
+1. Test files should be named after the command they test.
+   For example, `find_test.go` for testing the find command.
+   Test files belonging to the same commands group are organized into folders named after this group,
+   such as `aggregation`, `operations`, `diagnostic`, etc.
```
chilagrow, line 239: When we add a new folder, do we also need to modify [Taskfile](https://github.com/FerretDB/FerretDB/blob/main/Taskfile.yml#L189) so that newly tests in folder are run? :thinking: 

I think this is a good organisation, but might add complications to adding test, and we want adding tests to be easy.

CONTRIBUTING.md
```diff
@@ -231,6 +231,24 @@ Writing separate tests might be much better than making a single test that check
 
 Also, we should use driver methods as much as possible instead of testing commands directly via `RunCommand`.
 
+#### Integration tests naming guidelines
+
+1. Test files should be named after the command they test.
+   For example, `find_test.go` for testing the find command.
+   Test files belonging to the same commands group are organized into folders named after this group,
```
chilagrow, line 238: This is a question to everyone, do we want a test file for each operator such as `find_bitwiseanyset_compat_test.go` or are we happy grouping them in a same file with _similar_ operators like now? Not sure how we've grouped them now but it makes sense.

chilagrow (commented):
Great one! Let me know if anyone has suggestion or input!

CONTRIBUTING.md
```diff
@@ -231,6 +231,24 @@ Writing separate tests might be much better than making a single test that check
...
+   such as `aggregation`, `operations`, `diagnostic`, etc.
+2. If the file contains compatibility tests, add the `_compat` suffix. For example, `find_compat_test.go`.
+3. Test names should include the name of the command being tested.
+   For instance, `TestDistinct` for testing the distinct command.
```
chilagrow, line 242: Finally, for `RunCommand` do we have a preference? :thinking: 

Currently we have `TestFindAndModifyCommand`, `TestCommandsFindAndModify`, `TestFindAndModifyRunCommand`, or just `TestFindAndModify` because findAndModify can only be run as `RunCommand`.

CONTRIBUTING.md
rumyantseva: Good point! Overall, there won't be a lot of folders and new folders won't be added too often, but I agree the process of adding new tests/folders should remain simple.

CONTRIBUTING.md
rumyantseva: To be honest, I don't like the current approach of having everything in the same file because it's hard to navigate through the files that have 1000+ lines of code. If the files for different operators are separate, it's easier to go through them, scroll through the tests for a particular operator, etc.

CONTRIBUTING.md
rumyantseva: Yeah, let's document this one too!

chilagrow: But I don't know which one to use. Up to you to suggest, I'm always in for shortest name :see_no_evil:

CONTRIBUTING.md
chilagrow: Yeah agree. If we all agree let's document that too :hugs:

CONTRIBUTING.md
chilagrow: This may also have impact on recent work about running integration tests in shard https://github.com/FerretDB/FerretDB/issues/2692

Anyway, I think we need to make sure everyone is onboard, if so then we need to make sure we have tooling in place to do it simply. :hugs:

CONTRIBUTING.md
AlekSi: Yeah, `find`, `update,` `aggregate` will be huge without additional level of splitting

CONTRIBUTING.md
AlekSi: That should not be a problem with [`./...`](https://pkg.go.dev/cmd/go#hdr-Package_lists_and_patterns) as long as test names are unique. Envtool should be able to check that

CONTRIBUTING.md
```diff
@@ -231,6 +231,24 @@ Writing separate tests might be much better than making a single test that check
...
+4. Compatibility tests should have `Compat` in the name, following the command.
+   For example, `TestDistinctCompat`.
+5. Test names should be descriptive and provide information about the functionality or condition being tested.
+   If the test is checking for a specific error scenario, include the error scenario in the name.
+6. Keep test names concise, avoiding overly cryptic names. Use abbreviations when appropriate.
+7. Avoid including test data in the name to maintain clarity and prevent excessively long names.
+8. Test case names should follow `TitleCase` capitalization style.
```
AlekSi, line 249: Some points, including that one, should clarify the difference between top-level test functions and subtests. For example, `TestFind/Doubles` is a valid (full) test name, but not a valid test function name

CONTRIBUTING.md
AlekSi: I think this is about the test name _segments_

CONTRIBUTING.md
AlekSi: I think we should specifically mention that no spaces, dashes or underscores should be used anywhere

w84thesun (commented):
Should we close this one as we already added some naming guidelines?

PR #2829, opened 2023-06-13
Handle `$type` aggregation operator errors properly

### LINKED ISSUE #2678: Implement `$type` aggregation operator
### What should be done?

https://www.mongodb.com/docs/manual/reference/operator/aggregation/type/

Currently we have a stub for `$type` operator so it just needs an implementation.
The `commonparam` package offers the `AliasFromType` function which can be used for the operator either.

#2635 might be helpful, there's a basic implementation there and couple of test cases.

As `$type` differentiate between unset and Null, we should probably modify `expression.Evaluate` to return nil and not `Null` for non-existent paths. Some changes in code will be needed to handle function calls in a different way.

As we currently don't support standard operators anywhere, let's add basic operators support in the `$project` stage. Other stages will be covered in #2679 after implementing `$type`.

### DoD
- `$type` operator is supported
- recursive operator calls in `$type` are supported (for example `{$type:{$type:42}}`)
- dot notation is supported and tested properly (arrays, and array indexes should be also tested)
- compat tests implemented.

noisersup:
# Description

Closes #2678.

## Readiness checklist


internal/handlers/common/aggregations/operators/operators_errors.go
```diff
@@ -0,0 +1,72 @@
...
+package operators
+
+// operatorErrorCode represents the type of error.
+type operatorErrorCode uint
+
+const (
+	_ operatorErrorCode = iota
+
+	// ErrArgsInvalidLen indicates that operator have invalid amount of arguments.
+	ErrArgsInvalidLen // Expression takes different amount of arguments.
```
chilagrow, line 25: What is the comment `// Expression takes different amount of arguments` about? It seems slightly differently worded version of `// ErrArgsInvalidLen indicates that operator have invalid amount of arguments.` :smile:

internal/handlers/common/aggregations/operators/type.go
```diff
@@ -56,28 +56,38 @@ func (t *typeOp) Process(doc *types.Document) (any, error) {
...
 
 			if typeParam, err = operator.Process(doc); err != nil {
-				// TODO https://github.com/FerretDB/FerretDB/issues/2678
-				return nil, lazyerrors.Error(err)
+				var opErr OperatorError
+				if !errors.As(err, &opErr) {
+					return nil, lazyerrors.Error(err)
+				}
+
+				return nil, err
```
chilagrow, line 79: This is not tested, please add a test when you remove todo comment.

internal/handlers/common/aggregations/stages/projection/projection.go
```diff
@@ -577,31 +577,50 @@ func setBySourceOrder(key string, val any, source, projected *types.Document) {
...
-	case errors.Is(err, operators.ErrNotImplemented):
+	case operators.ErrNotImplemented:
 		return commonerrors.NewCommandErrorMsgWithArgument(
 			commonerrors.ErrEmptySubProject,
 			"Invalid $project :: caused by :: An empty sub-projection is not a valid value."+
 				" Found empty object at path",
 			"$project (stage)",
 		)
-	case errors.Is(err, operators.ErrWrongType):
+	case operators.ErrArgsInvalidLen:
```
chilagrow, line 611: For each command error added, could you test with integration test for the error message too please.

See how it's done for rest of aggregation $project https://github.com/FerretDB/FerretDB/blob/main/integration/aggregate_documents_test.go#L71

internal/handlers/common/aggregations/stages/projection/projection.go
```diff
@@ -577,31 +577,50 @@ func setBySourceOrder(key string, val any, source, projected *types.Document) {
 // processOperatorError takes internal error related to operator evaluation and
 // returns proper CommandError that can be returned by $process aggregation stage.
 func processOperatorError(err error) error {
```
chilagrow, line 579: Could you fix go doc here too please? We don't have `$process` aggregation stage.

chilagrow: Also could you add list of errors like https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/common/aggregations/stages/projection/projection.go#L49

I think it also wise to add top level exported function `ProjectDocument`

internal/handlers/common/aggregations/operators/operators_errors.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// ErrWrongType indicates that operator field is not a document.
+	ErrWrongType // Invalid type of operator field (expected document)
+
+	// ErrEmptyField indicates that operator field does not specify any operator.
```
chilagrow, line 30:
```suggestion
	// ErrEmptyField indicates that operator field is empty.
```

Suggesting this because `operator field does not specify any operator` seems to indicate ErrNoOperator

internal/handlers/common/aggregations/operators/operators.go
```diff
@@ -98,19 +86,41 @@ func NewOperator(doc any) (Operator, error) {
...
+		return nil, newOperatorError(
+			ErrNotImplemented,
+			fmt.Errorf("The operator %s is not implemented yet", operator),
+		)
+	case !supported && !unsupported:
+		return nil, newOperatorError(
+			ErrInvalidExpression,
+			fmt.Errorf("Unrecognized expression '%s'", operator),
+		)
+	default:
```
chilagrow, line 121: Minor suggestion of simplifying. I think we don't need panic, if it's in `supported` let's return new operator. This sway it reduces untestable code by integration test. 
```
case supported:
     return newOperator(operatorDoc)
case !unsupported:
     ErrInvalidExpression
default:
     ErrNotImplemented
```

internal/handlers/common/aggregations/operators/operators.go
```diff
@@ -120,3 +130,158 @@ var Operators = map[string]newOperatorFunc{
 	"$type": newType,
 	// please keep sorted alphabetically
 }
+
+// unsupportedOperators maps all unsupported yet operators.
+var unsupportedOperators = map[string]struct{}{
```
chilagrow, line 135: Great work listing them all!

chilagrow (commented):
Let's add integration tests for newly added error messages :hugs:

internal/handlers/common/aggregations/stages/projection/projection.go
chilagrow: How about documenting returned error on exported function `ProjectDocument`?

internal/handlers/common/aggregations/operators/type.go
chilagrow: Is there a reason why this is not tested? It returns `OperatorError` so it should be tested right?

internal/handlers/common/aggregations/operators/operators.go
chilagrow: You liked it but there is no change in the code? :thinking:

integration/aggregate_documents_compat_test.go
```diff
@@ -1690,6 +1697,20 @@ func TestAggregateCompatProject(t *testing.T) {
 				bson.D{{"$project", bson.D{{"type", bson.D{{"$type", true}}}}}},
 			},
 		},
+		"ProjectEmpty": {
```
chilagrow, line 1700: This is already tested here https://github.com/FerretDB/FerretDB/blob/main/integration/aggregate_documents_test.go#L81

So no need to add compat test.

integration/aggregate_documents_compat_test.go
```diff
@@ -1626,14 +1626,21 @@ func TestAggregateCompatProject(t *testing.T) {
...
-			skip: "https://github.com/FerretDB/FerretDB/issues/2678",
+			resultType: emptyResult,
 		},
 		"TypeRecursiveInvalid": {
 			pipeline: bson.A{
 				bson.D{{"$sort", bson.D{{"_id", -1}}}},
 				bson.D{{"$project", bson.D{{"type", bson.D{{"$type", bson.D{{"v", "$v"}}}}}}}},
 			},
 		},
+		"TypeRecursiveArrayInvalid": {
```
chilagrow, line 1637: This test is the same as https://github.com/FerretDB/FerretDB/pull/2829/files#diff-8e3642cbc845cda87e6c35924b1170d8932453298b2c4552d662d6a9a04037a4R303

Any reason why we want to duplicate it in compat and integration test? :thinking:

internal/handlers/common/aggregations/operators/operators.go
```diff
@@ -67,12 +50,17 @@ type Operator interface {
 func NewOperator(doc any) (Operator, error) {
 	operatorDoc, ok := doc.(*types.Document)
 	if !ok {
-		// TODO: https://github.com/FerretDB/FerretDB/pull/2789
-		return nil, ErrWrongType
+		return nil, newOperatorError(
+			ErrWrongType,
+			"Invalid type of operator field (expected document)",
+		)
```
chilagrow, line 56: Could you add test for this please :pray:

chilagrow (commented):
Please see the previous comments, some where not applied, so I think there was some reason that I don't know :thinking: 

My most important comments are about untested code

internal/handlers/common/aggregations/operators/operators.go
noisersup: There is a change in code, but maybe I misunderstood the comment.

The old code:
```switch {
	case supported && unsupported:
		panic(fmt.Sprintf("operator %q is in both `operators` and `unsupportedOperators`", operator))
	case supported && !unsupported:
		return newOperator(operatorDoc)
	case !supported && unsupported:
		return nil, newOperatorError(
			ErrNotImplemented,
			fmt.Errorf("The operator %s is not implemented yet", operator),
		)
	case !supported && !unsupported:
		return nil, newOperatorError(
			ErrInvalidExpression,
			fmt.Errorf("Unrecognized expression '%s'", operator),
		)
	default:
		panic("Unreachable")
	}
```

The new code:
```
	switch {
	case supported && unsupported:
		panic(fmt.Sprintf("operator %q is in both `operators` and `unsupportedOperators`", operator))
	case supported && !unsupported:
		return newOperator(operatorDoc)
	case !supported && unsupported:
		return nil, newOperatorError(
			ErrNotImplemented,
			fmt.Sprintf("The operator %s is not implemented yet", operator),
		)
	default:
		return nil, newOperatorError(
			ErrInvalidExpression,
			fmt.Sprintf("Unrecognized expression '%s'", operator),
		)
	}

```

I've replaced the unreachable panic and simplified the switch a bit.

noisersup: Oh, maybe you were talking about the panic about operator that is both supported and unsupported? We have similar logic in [stages](https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/common/aggregations/stages/stages.go#L95-L115). I think it's worth to catch if operator is removed from `unsupportedOperators` list after introducing support for it. If we think otherwise then we can get rid off the second panic :)

internal/handlers/common/aggregations/operators/type.go
noisersup: I'm sure that after adding some recursive test cases, code coverage showed this line as covered. I believe that it's not covered right now, as I've moved error checking into the validation (it was necessary to work correctly with empty collections).

noisersup: Nice catch, it looks like all "Recursive" tests I've created weren't recursive :)

integration/aggregate_documents_compat_test.go
noisersup: This test is different now. It should be recursive

internal/handlers/common/aggregations/operators/operators.go
noisersup: As I wanted to move most of the parsing logic to the `NewOperator`, this line is unreachable for now, as `NewOperator` is called from type switches that already check for `*types.Document`.

internal/handlers/common/aggregations/operators/operators.go
chilagrow: Ah yeah never mind, I didn't explain it well enough. Let's leave it and at some point we may or may not :hugs: I was just asking why not have supported case, unsupported case and default case for not implemented, ignoring `supported && unsupported`,

internal/handlers/common/aggregations/stages/projection/projection_iterator.go
```diff
@@ -59,8 +59,9 @@ func (iter *projectionIterator) Next() (struct{}, *types.Document, error) {
 	}
 
 	projected, err := ProjectDocument(doc, iter.projection, iter.inclusion)
-	if err != nil {
-		return unused, nil, lazyerrors.Error(err)
+
+	if err = processOperatorError(err); err != nil {
```
chilagrow, line 63: This line is not needed, it's unreachable because `ProjectDocument` already returns command error. :hugs:

chilagrow (approved):
Looks good, I left comments unresolved for super minor issues. Great work!

internal/handlers/common/aggregations/stages/projection/projection_iterator.go
noisersup: nice catch! Fixed

chilagrow (approved):
:rocket:

w84thesun (approved):
LGTM

PR #2868, opened 2023-06-20
Validate database and collection names for SQLite handler

### LINKED ISSUE #2749: Validate database and collection names
### What should be done?

The `sqlite` handler should validate that database and collection names are valid according to general FerretDB rules.

The `sqlite` backend should implement additional validation for SQLite-specific rules.

We also should change `CollectionToTable` method. Currently, it does two things:
* converts collection name to valid table name on collection creation;
* converts existing collection name to table name.

The second thing should be done by querying the metadata table.

The manglign scheme should be improved too.

### Where?

* https://github.com/FerretDB/FerretDB/tree/main/internal/backends/sqlite
* https://github.com/FerretDB/FerretDB/tree/main/internal/handlers/sqlite

### Definition of Done

TBD
- `sqlite` handler updated;
- unit tests added/updated;
- integration/compatibility tests checked;
- spot refactorings done;

noisersup:
# Description

Closes #2749.

This PR is a part of the issue, database validation will be handled in second PR.
 
## TODO
- [x] Verify `insert` and `update` test cases 

```
[user@nixpc:~/FerretDB]$ task test-integration-sqlite TEST_RUN='TestCollectionName'
task: [test-integration-sqlite] go test -count=1 -run='TestCollectionName' -timeout=35m -race=true -tags=ferretdb_debug,ferretdb_tigris,ferretdb_hana -shuffle=on -coverpkg=../... -coverprofile=integration-sqlite.txt . -target-backend=ferretdb-sqlite -target-tls -compat-url='mongodb://username:password@127.0.0.1:47018/?tls=true&tlsCertificateKeyFile=../build/certs/client.pem&tlsCaFile=../build/certs/rootCA-cert.pem' -disable-filter-pushdown

ok      github.com/FerretDB/FerretDB/integration        1.453s  coverage: 40.0% of statements in ../...
```

## Readiness checklist


AlekSi:
CleanShot 2023-06-26 at 09 59 32@2x

PR is not an issue

integration/basic_test.go
```diff
@@ -405,6 +405,9 @@ func TestCollectionName(t *testing.T) {
 		"Capital": {
 			collection: "A",
 		},
+		"sqlite": {
+			collection: "sqlite",
+		},
```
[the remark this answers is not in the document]
rumyantseva: What do you mean by "`sqlite_` is reserved prefix"? Wasn't `_ferretdb` a reserved prefix?

internal/backends/sqlite/metadata/registry.go
```diff
@@ -17,22 +17,35 @@ package metadata
...
+	"github.com/FerretDB/FerretDB/internal/util/must"
 )
 
-// metadataTableName is a SQLite table name where FerretDB metadata is stored.
-const metadataTableName = "_ferretdb_collections"
+const (
+	// Reserved prefix for database and collection names.
+	reservedPrefix = "_ferretdb_"
+
+	// Reserved prefix for SQLite table name.
```
rumyantseva, line 40: Should we also describe how it's used? It's not straightforward why we need to reserve another prefix in addition to `_ferretdb_`

integration/basic_test.go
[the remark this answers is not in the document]
noisersup: `_ferretdb` is our reserved prefix, but `sqlite_` is SQLite reserved prefix.

https://www.sqlite.org/lang_createtable.html
> Table names that begin with "sqlite_" are reserved for internal use. It is an error to attempt to create a table with a name that starts with "sqlite_".

internal/backends/sqlite/metadata/registry.go
```diff
@@ -58,11 +71,23 @@ func (r *Registry) Close() {
 	r.p.Close()
 }
 
-// CollectionToTable converts FerretDB collection name to SQLite table name.
-func (r *Registry) CollectionToTable(collectionName string) string {
-	// TODO https://github.com/FerretDB/FerretDB/issues/2749
-	h := sha1.Sum([]byte(collectionName))
-	return hex.EncodeToString(h[:])
+// collectionToTable converts FerretDB collection name to SQLite table name.
+// It returns error if the name is incorrect for SQLite backend.
```
rumyantseva, line 75: Let's document which error (like we do in other places).

internal/backends/sqlite/metadata/registry.go
```diff
@@ -58,11 +71,23 @@ func (r *Registry) Close() {
...
+	}
+
+	hash32 := fnv.New32a()
+	must.NotFail(hash32.Write([]byte(collectionName)))
+
+	// SQLite table cannot start with _sqlite prefix
+	if strings.HasPrefix(strings.ToLower(collectionName), reservedTablePrefix) {
+		collectionName = "_" + collectionName
+	}
+	// TODO case insensitive
```
rumyantseva, line 88: The hashes for `collection` and `COLLECTION` will be different, right? So, it seems that there is not much to handle here (but we might want to add tests, yes).

internal/backends/sqlite/metadata/registry.go
```diff
@@ -169,6 +198,33 @@ func (r *Registry) CollectionCreate(ctx context.Context, dbName string, collecti
 	return true, nil
 }
 
+// CollectionGet returns table name associated with provided collection.
```
rumyantseva, line 201: Why `CollectionGet`? Shouldn't it be something like `GetTableName`? 

Also, the function is not fully documented, e.g. it doesn't document cases when the database doesn't exist.

internal/backends/sqlite/metadata/registry.go
```diff
@@ -169,6 +198,33 @@ func (r *Registry) CollectionCreate(ctx context.Context, dbName string, collecti
...
+	query := fmt.Sprintf("SELECT table_name FROM %q WHERE name = ?", metadataTableName)
+
+	rows, err := db.QueryContext(ctx, query, collectionName)
+	if err != nil {
+		return "", lazyerrors.Error(err)
+	}
+	defer rows.Close()
+
+	if !rows.Next() {
+		return "", lazyerrors.New("no collection found")
```
rumyantseva, line 217: I guess it's not a lazy error but some other error? Maybe the easiest way is to run some test that already queries/modifies a non-existent collection to see how this should reflect it.

internal/backends/sqlite/metadata/registry.go
```diff
@@ -179,7 +235,10 @@ func (r *Registry) CollectionDrop(ctx context.Context, dbName string, collection
 		return false, nil
 	}
 
-	tableName := r.CollectionToTable(collectionName)
+	tableName, err := r.CollectionGet(ctx, dbName, collectionName)
+	if err != nil {
+		return false, nil
```
rumyantseva, line 240: It's not enough to just return nil in this case, e.g. if the query failed because `rows.Scan` returned an error, we should have a lazy error here.

internal/handlers/sqlite/msg_create.go
```diff
@@ -108,3 +119,29 @@ func (h *Handler) MsgCreate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 		return nil, lazyerrors.Error(err)
 	}
 }
+
+// validateCollectionNameRe validates collection names.
+// Empty collection name, names with `$` and `\x00` are not allowed.
+// Collection names that start with `.` are also not allowed.
+var validateCollectionNameRe = regexp.MustCompile("^[^.$\x00][^$\x00]{0,234}$")
```
rumyantseva, line 126: That's an interesting place to define a new variable. Do we have a convention where new variables should be defined? E.g. you didn't define it inside a function, but on a package level, on the other hand, you defined it in the middle of the file. If it's used inside a particular function only, maybe we don't need it on a package level? Or if we need it on a package level (why?) it should be on the top of the file where it's used?

internal/handlers/sqlite/msg_create.go
```diff
@@ -108,3 +119,29 @@ func (h *Handler) MsgCreate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+
+// createCollection validates the collection name and creates collection.
```
rumyantseva, line 128: Let's document errors this function can return.

internal/handlers/sqlite/sqlite.go
```diff
@@ -30,6 +32,14 @@ import (
 	"github.com/FerretDB/FerretDB/internal/util/state"
 )
 
+var (
```
rumyantseva, line 35: Why do we want to differentiate those two errors?

rumyantseva (commented):
Left some questions

integration/basic_test.go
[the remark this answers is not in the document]
rumyantseva: So, we want to test that even if `sqlite` is set for the collection name, we still can use it? Should it be `sqlite_`, not `sqlite` then?

internal/backends/sqlite/metadata/registry.go
noisersup: They will be different, although it's not possible to test in a reasonable way. We have a lot of possibilities like fOo Foo foO... Just want to be sure that we don't expect any hash collision here, or should we create some validation just to be sure

internal/backends/sqlite/metadata/registry.go
chilagrow, line 240: I think we should return `return false, nil` only case where collection doesn't exist. Other case related to DB should return lazy error.

internal/handlers/sqlite/msg_create.go
```diff
@@ -96,6 +99,14 @@ func (h *Handler) MsgCreate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 
 		return &reply, nil
 
+	case errors.Is(err, ErrInvalidCollectionName):
+		msg := fmt.Sprintf("Invalid collection name: '%s.%s'", dbName, collectionName)
+		return nil, commonerrors.NewCommandErrorMsg(commonerrors.ErrInvalidNamespace, msg)
```
chilagrow, line 104: Could you use `NewCommandErrorMsgWithArgument` please :hugs:

internal/handlers/sqlite/msg_create.go
```diff
@@ -96,6 +99,14 @@ func (h *Handler) MsgCreate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+
+	case errors.Is(err, ErrCollectionStartsWithDot):
+		msg := fmt.Sprintf("Collection names cannot start with '.': %s", collectionName)
+		return nil, commonerrors.NewCommandErrorMsg(commonerrors.ErrInvalidNamespace, msg)
```
chilagrow, line 108: Also could you use NewCommandErrorMsgWithArgument please!

internal/handlers/sqlite/msg_create.go
```diff
@@ -108,3 +119,29 @@ func (h *Handler) MsgCreate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 		return nil, lazyerrors.Error(err)
 	}
 }
+
+// validateCollectionNameRe validates collection names.
+// Empty collection name, names with `$` and `\x00` are not allowed.
+// Collection names that start with `.` are also not allowed.
```
chilagrow, line 125: Collection name starting with `.` is checked by `strings.HasPrefix(collectionName, ".")` so we can simply this right?

Also let's mention valid collection name length :hugs:

internal/backends/sqlite/database.go
```diff
@@ -71,6 +72,9 @@ func (db *database) ListCollections(ctx context.Context, params *backends.ListCo
 func (db *database) CreateCollection(ctx context.Context, params *backends.CreateCollectionParams) error {
 	created, err := db.r.CollectionCreate(ctx, db.name, params.Name)
 	if err != nil {
+		if errors.Is(err, metadata.ErrInvalidCollectionName) {
+			return err
+		}
```
chilagrow, line 77: Not sure if this is need? :thinking:

internal/backends/sqlite/metadata/registry.go
rumyantseva: Ok, if sqlite table names are case insensitive, what I can suggest is to "unify" the way we call them, e.g. by lower-casing the names, so then in the meta data we store all the table names lower-cased (for simplicity). That's for the scope of this task. Collisions are not for the scope of this task. However, if we can create a unique index from the field `table_name`, let's do it, and we will be safe for now.

chilagrow (commented):
Some comment :hugs:

integration/basic_test.go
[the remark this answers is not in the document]
AlekSi: Yes, it should be `sqlite_` or `sqlite_something`. `sqlite` itself is a valid table name.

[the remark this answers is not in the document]
noisersup: As we append "_<checksum>", it still creates `sqlite_`

internal/backends/sqlite/metadata/registry.go
```diff
@@ -17,22 +17,35 @@ package metadata
...
+
+	// Reserved prefix for SQLite table name.
+	reservedTablePrefix = "sqlite"
+
+	// SQLite table name where FerretDB metadata is stored.
+	metadataTableName = reservedPrefix + "collections"
+)
+
+// ErrInvalidCollectionName indicates that a collection didn't pass name checks.
+var ErrInvalidCollectionName = fmt.Errorf("invalid FerretDB collection name")
```
AlekSi, line 48: Let's not introduce multiple error-handling ways and use `backends.Error` together with `backends.ErrorCodeIs`. Specifically, we already have `backends.ErrorCodeCollectionNameIsInvalid`

internal/backends/sqlite/metadata/registry.go
noisersup: As `Registry` struct provides functions for Database and Collection operators all of the Collection related functions start with "`Collection`": `CollectionList`, `CollectionCreate`. I thought that using `CollectionGet` would match the convention.

internal/backends/sqlite/metadata/registry.go
noisersup: Nice catch!

internal/handlers/sqlite/sqlite.go
noisersup: That's how MongoDB works. They return different error message for those cases.

internal/handlers/sqlite/msg_create.go
noisersup: It's left here just to show what's the regex of proper collection name

internal/handlers/sqlite/sqlite.go
rumyantseva: Well, I believe MongoDB doesn't return "invalid FerretDB collection name" error message :) For simplicity, I'd suggest we just return "invalid FerretDB collection name"  in both cases and not differentiate it. We use the same `ErrCollectionStartsWithDot` error code in both error messages anyway. Otherwise, we might want to have a separate error message for a case when both MongoDB and FerretDB consider the collection name too long etc. For simplicity, for now, I'd just keep the same error message not to introduce unclarity or technical debt. Otherwise, if you want proper error messages, please create a separate issue and describe there how they should work.

w84thesun (commented):
Not much to add really.

internal/handlers/sqlite/sqlite.go
noisersup: To simplify error handling and stay consistent with MongoDB handling I've used backends.Error as sugested [here](https://github.com/FerretDB/FerretDB/pull/2868#discussion_r1243560985). WDYT?

internal/backends/sqlite/collection.go
```diff
@@ -53,7 +53,10 @@ func (c *collection) Query(ctx context.Context, params *backends.QueryParams) (*
 		}, nil
 	}
 
-	tableName := c.r.CollectionToTable(c.name)
+	tableName, err := c.r.CollectionGet(ctx, c.dbName, c.name)
+	if err != nil {
+		return nil, lazyerrors.Error(err)
```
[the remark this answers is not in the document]
noisersup: > what happens if the table does not exist, I don't think lazy error is appropriate here.

This issue doesn't seem to be about error handling for non-existing collections. On the main branch we already return lazyerror for non-existing collections

> Also CollectionGet returns empty string if database does not exist, is that expected?

We do the same thing for `CollectionList` and `CollectionDrop`.

internal/backends/sqlite/collection.go
```diff
@@ -117,7 +128,10 @@ func (c *collection) Update(ctx context.Context, params *backends.UpdateParams)
 		return nil, lazyerrors.Errorf("no database %q", c.dbName)
 	}
 
-	tableName := c.r.CollectionToTable(c.name)
+	tableName, err := c.r.CollectionGet(ctx, c.dbName, c.name)
+	if err != nil {
+		return nil, lazyerrors.Error(err)
```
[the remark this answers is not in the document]
noisersup: https://github.com/FerretDB/FerretDB/pull/2868#discussion_r1244241565

internal/handlers/sqlite/sqlite.go
rumyantseva: Yeah, sounds good to me!

internal/backends/sqlite/metadata/registry.go
```diff
@@ -58,11 +70,24 @@ func (r *Registry) Close() {
...
+	if strings.HasPrefix(collectionName, reservedPrefix) {
+		return "", backends.NewError(backends.ErrorCodeCollectionNameIsInvalid, nil)
+	}
+
+	hash32 := fnv.New32a()
+	must.NotFail(hash32.Write([]byte(collectionName)))
+
+	collectionName = strings.ToLower(collectionName)
+
+	// SQLite table cannot start with _sqlite prefix
```
chilagrow, line 85: This comment is not the same as what it does. :thinking:

internal/backends/sqlite/metadata/registry.go
```diff
@@ -17,22 +17,34 @@ package metadata
...
 
-// metadataTableName is a SQLite table name where FerretDB metadata is stored.
-const metadataTableName = "_ferretdb_collections"
+const (
+	// Reserved prefix for database and collection names.
+	reservedPrefix = "_ferretdb_"
+
+	// Reserved prefix for SQLite table name.
+	// https://www.sqlite.org/lang_createtable.html
+	reservedTablePrefix = "sqlite"
```
chilagrow, line 43: From above link I think it should be `sqlite_`

internal/backends/sqlite/collection.go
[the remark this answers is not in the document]
chilagrow: My question was not about what main branch does, it's about what mongodb does... Maybe we already have test for that?

We should create an issue if we are not handling it right though. :thinking:

internal/backends/error.go
```diff
@@ -36,6 +36,7 @@ const (
 	ErrorCodeCollectionDoesNotExist
 	ErrorCodeCollectionAlreadyExists
 	ErrorCodeCollectionNameIsInvalid
+	ErrorCodeCollectionStartsWithDot
```
AlekSi, line 39: That's not a backend error – it is not generated by any backend

internal/handlers/sqlite/msg_create.go
```diff
@@ -96,15 +97,43 @@ func (h *Handler) MsgCreate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
 
 	default:
 		return nil, lazyerrors.Error(err)
 	}
 }
+
+// createCollection validates the collection name and creates collection.
+// It returns ErrInvalidCollectionName when the collection name is invalid.
+// It returns ErrCollectionStartsWithDot if collection starts with a dot.
+func createCollection(ctx context.Context, db backends.Database, collectionName string) error {
```
AlekSi, line 120: Please move that function back into `MsgCreate`

internal/backends/sqlite/metadata/registry.go
noisersup: That's because we always append `_<hash>` to the end of collection name, so if someone passes sqlite as collection name it will result in `sqlite_...`. Although to stay more consistent with docs I replaced this with `sqlite_` as you suggested and now the check is done against the table name and not collection name.

integration/basic_test.go
```diff
@@ -524,7 +528,7 @@ func TestDatabaseName(t *testing.T) {
 		ctx, collection := setup.Setup(t)
 
 		err := collection.Database().Client().Database("").CreateCollection(ctx, collection.Name())
-		expectedErr := driver.InvalidOperationError(driver.InvalidOperationError{MissingField: "Database"})
+		expectedErr := errors.New("database name cannot be empty")
```
noisersup, line 531: Merged from #2934 to make tests pass

internal/backends/sqlite/metadata/registry.go
```diff
@@ -58,11 +70,22 @@ func (r *Registry) Close() {
...
 }
 
-// CollectionToTable converts FerretDB collection name to SQLite table name.
-func (r *Registry) CollectionToTable(collectionName string) string {
-	// TODO https://github.com/FerretDB/FerretDB/issues/2749
-	h := sha1.Sum([]byte(collectionName))
-	return hex.EncodeToString(h[:])
+// collectionToTable converts FerretDB collection name to SQLite table name.
+// It returns ErrInvalidCollectionName error if the name is incorrect for SQLite backend.
+func collectionToTable(collectionName string) (string, error) {
```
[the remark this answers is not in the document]
noisersup: Nice catch

w84thesun (dismissed):
LGTM!

internal/handlers/sqlite/msg_create.go
```diff
@@ -78,6 +90,22 @@ func (h *Handler) MsgCreate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+	}
+
+	if !validateCollectionNameRe.MatchString(collectionName) ||
+		!utf8.ValidString(collectionName) {
+		msg := fmt.Sprintf("Invalid collection name: '%s.%s'", dbName, collectionName)
+		return nil, commonerrors.NewCommandErrorMsgWithArgument(commonerrors.ErrInvalidNamespace, msg, "create")
+	}
+
+	if strings.HasPrefix(collectionName, reservedPrefix) {
+		msg := fmt.Sprintf("Invalid collection name (%q is a reserved prefix): '%s.%s'", reservedPrefix, dbName, collectionName)
```
chilagrow, line 105: You might want to return the error message the same as other cases `Invalid collection name: '%s.%s'`. When we do run dance diff test for sqlite, we would want this test to pass

https://github.com/FerretDB/dance/blob/main/tests/diff/names_test.go#L48

chilagrow (dismissed):
Looks great

internal/backends/sqlite/metadata/registry.go
chilagrow: I like the way you made the change :+1:

internal/backends/sqlite/metadata/registry.go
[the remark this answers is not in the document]
chilagrow: What has happened to the case insensitivity?

internal/backends/sqlite/metadata/registry.go
```diff
@@ -169,6 +198,33 @@ func (r *Registry) CollectionCreate(ctx context.Context, dbName string, collecti
 	return true, nil
 }
 
+// CollectionGet returns table name associated with provided collection.
+func (r *Registry) CollectionGet(ctx context.Context, dbName string, collectionName string) (string, error) {
+	db := r.p.GetExisting(ctx, dbName)
+	if db == nil {
+		return "", nil
```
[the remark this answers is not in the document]
chilagrow: When CI reports coverage, we may want to replace these with panic, but for now lets keep it

internal/backends/sqlite/metadata/registry.go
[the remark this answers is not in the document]
chilagrow: outdated, now returns not exist error

internal/backends/sqlite/metadata/registry.go
AlekSi: * `CollectionList` returns a list of collections;
* `CollectionCreate` creates a collection;
* `CollectionDrop` drops collection;
* `CollectionGet`… does not get a collection.

internal/backends/sqlite/metadata/registry.go
```diff
@@ -139,13 +161,14 @@ func (r *Registry) CollectionList(ctx context.Context, dbName string) ([]string,
 //
 // Returned boolean value indicates whether the collection was created.
 // If collection already exists, (false, nil) is returned.
+// If collection name doesn't pass backend level validation it returns ErrInvalidCollectionName.
```
AlekSi, line 164: But it does not.

We also talked about how documentation for returned errors is fragile (because you have to update comments for all transitive callers) and how we will use contracts instead. Please don't document returned errors and error codes in comments.

internal/backends/sqlite/metadata/registry.go
```diff
@@ -169,6 +192,38 @@ func (r *Registry) CollectionCreate(ctx context.Context, dbName string, collecti
...
+
+	rows, err := db.QueryContext(ctx, query, collectionName)
+	if err != nil {
+		return "", lazyerrors.Error(err)
+	}
+	defer rows.Close()
+
+	if !rows.Next() {
+		return "", backends.NewError(
+			backends.ErrorCodeCollectionDoesNotExist,
```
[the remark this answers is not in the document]
noisersup: I think I don't quiet understand your comment. Do you mean that this error is not handled anywhere and just returned as lazyerror, or you mean that it shouldn't be returned here?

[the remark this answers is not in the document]
AlekSi: Both, I guess. Why we return that error instead of some other error? What would change if we return `fmt.Errorf("boom")`?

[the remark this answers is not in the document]
noisersup: This errorcode is returned as it's used by `registry.CollectionDrop`, which returns nil if collection does not exist.

w84thesun (approved):
🚀

internal/handlers/sqlite/msg_create.go
chilagrow: Any comment?

internal/backends/sqlite/metadata/registry.go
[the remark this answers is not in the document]
AlekSi: It is reachable if database exists, but collection does not:
```
db.lala.find()
MongoServerError: [msg_find.go:63 sqlite.(*Handler).MsgFind] [collection.go:58 sqlite.(*collection).Query] ErrorCodeCollectionDoesNotExist: collection "lala" does not exist
```

internal/backends/sqlite/metadata/registry.go
```diff
@@ -169,6 +185,38 @@ func (r *Registry) CollectionCreate(ctx context.Context, dbName string, collecti
...
+// If database does not exist, no error is returned.
+func (r *Registry) GetTableName(ctx context.Context, dbName string, collectionName string) (string, error) {
+	db := r.p.GetExisting(ctx, dbName)
+	if db == nil {
+		return "", nil
+	}
+
+	query := fmt.Sprintf("SELECT table_name FROM %q WHERE name = ?", metadataTableName)
+
+	rows, err := db.QueryContext(ctx, query, collectionName)
```
AlekSi, line 199: That should be `QueryRowContext`

internal/backends/sqlite/metadata/registry.go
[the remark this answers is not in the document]
AlekSi: Ok, but why did you ignore the three other cases I mentioned? All of them wrap that error with lazy error to basically hide it from `ErrorCodeIs`

AlekSi (commented):
There are problems with error handling, and then there is a regression: https://github.com/FerretDB/FerretDB/pull/2868/files#r1247574472

internal/handlers/sqlite/msg_create.go
noisersup: Sorry, missed this one, yeah makes sense

internal/backends/sqlite/metadata/registry.go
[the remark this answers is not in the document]
noisersup: Does it look good right now?

w84thesun (approved):
👍

rumyantseva (approved):
LGTM

rumyantseva (approved):

internal/backends/collection.go
```diff
@@ -115,7 +115,7 @@ type UpdateResult struct {
 // Update updates documents in collection.
 func (cc *collectionContract) Update(ctx context.Context, params *UpdateParams) (res *UpdateResult, err error) {
 	defer observability.FuncCall(ctx)()
-	defer checkError(err)
+	defer checkError(err, ErrorCodeCollectionDoesNotExist)
```
AlekSi, line 118: Why? I don't see where we handle that error, and updating non-existing collections is possible in MongoDB

noisersup: From my understanding `sqlite` msg_update handles non-existent collections. So when this function is called we expect that collection already exists.

About the error code: we create contracts for backend codes only if they are handled somewhere?

AlekSi: I mean, we should not create error codes if they are ignored. The same thing with `Query` for non-existing databases or collections – primary users (`find`, `aggregate`, etc) just return empty results, so why would we make it return an error just to be ignored?

noisersup: For now we are able to easily see where and how functions like `Query` are used, although when we'll start to add more and more functionality the `Query` function might be used in many places, but someone might make a mistake and forget to ensure that collection exists beforewards. In results one would need to spend some time on debugging empty `Query` results. If we return the `lazyerror` that we expect to not be reached we can easily see what went wrong and how we can fix it.

If our current design assumes that we don't return such errors then let's just return empty results, although I'd suggest to consider this later

chilagrow (dismissed):

AlekSi:
Let's wait until after v1.5.0

w84thesun (approved):
:shipit:

rumyantseva (approved):
LGTM!

rumyantseva (dismissed):
LGTM.

internal/handlers/sqlite/msg_dropdatabase.go
```diff
@@ -59,6 +59,8 @@ func (h *Handler) MsgDropDatabase(ctx context.Context, msg *wire.OpMsg) (*wire.O
 	switch {
 	case err == nil:
 		res.Set("dropped", dbName)
+	case backends.ErrorCodeIs(err, backends.ErrorCodeDatabaseNameIsInvalid):
+		// nothing?
```
noisersup, line 63: I guess we can decide based on what MongoDB does for dropping invalid names, right?

AlekSi: Yeah, a test would be needed there, but that's not super important

PR #2878, opened 2023-06-21
Test `getMore` integration test using one connection pool

### LINKED ISSUE #1807: Investigate `TestGetMoreErrors` sessions errors
I think our tests should have better control over connections. `*mongo.Client` uses a pool of connections; be default we don't know if two subsequent commands use the same connection or not.

It should be possible to:
* create one client that uses MongoDB URI query parameters to fix connection pool size to 1, ensuring that all commands run with this client use the same connection;
* create a separate client like the first one, ensuring that different clients use different connections.

That might require quite some refactoring of our tests.

---

### What should be done?

In #1733 was introduced test `TestGetMoreErrors`. To test parameters other than cursor id we need to call `find` first to get the correct cursor id. The problem is that from time to time MongoDB will return an error `(Location50738) Cannot run getMore on cursor 5531067849097530169, which was created in session ..., in session ... `. Adding those tests would make MongoDB integration test flaky.

We should either investigate why MongoDB returns that error or a way to enable those test without failing other tests.

### Where?

Search for this issue ID in the code.

### Definition of Done
- unit tests added;
- integration tests added;
- compatibility tests added;
- spot refactorings were done;
- user documentation updated;

chilagrow:
# Description

Closes #1807.

## Readiness checklist


integration/query_test.go
```diff
@@ -950,7 +950,11 @@ func TestQueryBatchSize(t *testing.T) {
 
 func TestQueryCommandGetMore(t *testing.T) {
 	t.Parallel()
-	ctx, collection := setup.Setup(t)
+	s := setup.SetupWithOpts(t, &setup.SetupOpts{
+		ClientOptions: options.Client().SetMaxPoolSize(1),
```
AlekSi, line 954: At the very least, we should also set `SetMinPoolSize` to prevent the number of active connections from dropping to zero. Please also check other options.

integration/setup/client.go
```diff
@@ -77,10 +77,8 @@ func mongoDBURI(tb testing.TB, opts *mongoDBURIOpts) string {
 	return u.String()
 }
 
-// makeClient returns new client for the given working MongoDB URI.
-func makeClient(ctx context.Context, uri string) (*mongo.Client, error) {
-	clientOpts := options.Client().ApplyURI(uri)
-
+// makeClient returns new client for the given MongoDB client options.
+func makeClient(ctx context.Context, clientOpts *options.ClientOptions) (*mongo.Client, error) {
```
AlekSi, line 81: Please pass all options via URI's query parameters

integration/setup/setup.go
```diff
@@ -86,6 +86,9 @@ type SetupOpts struct {
 
 	// Benchmark data provider. If empty, collection is not created.
 	BenchmarkProvider shareddata.BenchmarkProvider
+
+	// ClientOptions overwrites the options set in MongoDB URI when the same option is set.
+	ClientOptions *options.ClientOptions
```
AlekSi, line 91:
```suggestion
	ExtraOptions url.Values
```

AlekSi:
> Do we want to test this only for FerretDB?

First of all, we want to understand how MongoDB works. We can't do that if we don't run tests for it.

> As far as I know, MongoDB does not provide information related to uniquely identifying a connection for commands.

That's why we fix `*mongo.Client`'s pool size to 1 – it is easy to uniquely identify a connection if there is only one of them.

>> create a separate client like the first one, ensuring that different clients use different connections.
>
> Also this can be implemented for FerretDB, but I didn't find equivalent check for MongoDB.

Not sure I understand why we can't create a separate `*mongo.Client` instance for MongoDB. See `TestCommandsDiagnosticWhatsMyURI` for example.

> If we want to ensure FerretDB does use the same connection,

It will use the same connection if there is only one. And if there is only one, if we configure `*mongo.Client` to use the only one.

integration/commands_diagnostic_test.go
```diff
@@ -406,3 +409,130 @@ func TestCommandsDiagnosticWhatsMyURI(t *testing.T) {
...
+	s := setup.SetupWithOpts(t, &setup.SetupOpts{
+		ExtraOptions: q1,
+	})
+
+	collection1 := s.Collection
+	databaseName := s.Collection.Database().Name()
+	collectionName := s.Collection.Name()
+
+	t.Run("SameClient", func(t *testing.T) {
+		setup.SkipExceptMongoDB(t, "https://github.com/FerretDB/FerretDB/issues/2906")
```
chilagrow, line 435: Fails on FerretDB, see https://github.com/FerretDB/FerretDB/actions/runs/5373306151/jobs/9747542079?pr=2878

integration/query_test.go
chilagrow: Good idea, applied in the tests.

integration/setup/client.go
chilagrow: Using URI now :+1:

integration/commands_diagnostic_test.go
AlekSi: All MongoDB URI query parameters are driver / client options, not MongoDB / FerretDB server options. `maxPoolSize`, `minPoolSize` and few other options speicify how many connections the client opens and when. Those options are not visible for the server.

integration/commands_diagnostic_test.go
```diff
@@ -406,3 +409,133 @@ func TestCommandsDiagnosticWhatsMyURI(t *testing.T) {
...
+
+		t.Parallel()
+
+		num := runtime.GOMAXPROCS(-1) * 10
```
AlekSi, line 439: Please use `teststress.Stress`

integration/query_test.go
```diff
@@ -950,7 +951,19 @@ func TestQueryBatchSize(t *testing.T) {
...
+
+	q := url.Values{}
+
+	// set 1 to ensure only one pool exists duration of the test,
+	// which forces a client to use a single connection pool
+	q.Set("maxPoolSize", "1")
+	q.Set("minPoolSize", "1")
+
+	s := setup.SetupWithOpts(t, &setup.SetupOpts{
+		ExtraOptions: q,
```
AlekSi, line 963:
```suggestion
		ExtraOptions: url.Values{
			"minPoolSize": []string{"1"},
			"maxPoolSize": []string{"1"},
		}
```

or something like that

integration/setup/client.go
```diff
@@ -77,10 +77,24 @@ func mongoDBURI(tb testing.TB, opts *mongoDBURIOpts) string {
 	return u.String()
 }
 
-// makeClient returns new client for the given working MongoDB URI.
-func makeClient(ctx context.Context, uri string) (*mongo.Client, error) {
-	clientOpts := options.Client().ApplyURI(uri)
+// makeClient returns new client for the given MongoDB URI and extra options.
+// The extraOpts that comes later overwrites value for the query key.
+func makeClient(ctx context.Context, uri string, extraOpts url.Values) (*mongo.Client, error) {
```
AlekSi, line 82:
```suggestion
func makeClient(ctx context.Context, uri string) (*mongo.Client, error) {
```

Options should already be in the `uri`

integration/setup/client.go
```diff
@@ -103,15 +117,15 @@ func makeClient(ctx context.Context, uri string) (*mongo.Client, error) {
 //
 // If the connection can't be established, it panics,
 // as it doesn't make sense to proceed with other tests if we couldn't connect in one of them.
-func setupClient(tb testing.TB, ctx context.Context, uri string) *mongo.Client {
+func setupClient(tb testing.TB, ctx context.Context, uri string, extraOpts url.Values) *mongo.Client {
```
AlekSi, line 120:
```suggestion
func setupClient(tb testing.TB, ctx context.Context, uri string) *mongo.Client {
```

AlekSi (changes requested):

integration/query_test.go
```diff
@@ -1179,15 +1190,48 @@ func TestQueryCommandGetMore(t *testing.T) {
 				Message: "BSON field 'getMore.collection' is missing but a required field",
 			},
 		},
+		"UnsetAllBatchSize": {
```
chilagrow, line 1193: These tests address this comment https://github.com/FerretDB/FerretDB/pull/2864#discussion_r1243195284

integration/query_test.go
```diff
@@ -1179,15 +1190,48 @@ func TestQueryCommandGetMore(t *testing.T) {
...
 			}
 
-			// TODO: https://github.com/FerretDB/FerretDB/issues/1807
-			// Do not run tests in parallel, MongoDB throws error that session and cursor do not match.
+			// Do not run subtests in t.Parallel() to reduce the occurrence
+			// of session error. Even the same connection is used for the test
+			// by setting `minPoolSize` and `maxPoolSize` to 1, on rare occasion
+			// it returns error due to different session calling `find` and `getMore`.
+			// Supporting session would help us understand fix it
+			// https://github.com/FerretDB/FerretDB/issues/153.
```
chilagrow, line 1233: So far this hasn't happened as long as sub tests don't run in parallel. But just a warning

integration/commands_diagnostic_test.go
chilagrow: This was totally my fault, I forgot to add pool size to newly created listeners :sweat:

integration/commands_diagnostic_test.go
chilagrow: Made minor change to export goroutine numbers.

noisersup (dismissed):
Looks great!

integration/query_test.go
```diff
@@ -1280,3 +1324,114 @@ func TestQueryCommandGetMore(t *testing.T) {
...
+	collection1 := s.Collection
+	databaseName := s.Collection.Database().Name()
+	collectionName := s.Collection.Name()
+
+	docs := generateDocuments(0, 5)
+	_, err := collection1.InsertMany(s.Ctx, docs)
+	require.NoError(t, err)
+
+	t.Run("SameClient", func(t *testing.T) {
+		// Do not run subtests in t.Parallel() to reduce the occurrence
```
AlekSi, line 1349: That should eliminate those occurrences, not just reduce them

integration/query_test.go
```diff
@@ -1280,3 +1324,114 @@ func TestQueryCommandGetMore(t *testing.T) {
...
+		}).Decode(&res)
+		require.NoError(t, err)
+	})
+
+	t.Run("DifferentClient", func(t *testing.T) {
+		// error returned from using different clients are session related error,
+		// hence currently FerretDB does not return an error
+		setup.SkipExceptMongoDB(t, "https://github.com/FerretDB/FerretDB/issues/153")
+
+		t.Parallel()
```
AlekSi, line 1390: It also can't be a parallel test – it will be broken once we add another parallel subtest

integration/setup/client.go
```diff
@@ -33,26 +33,43 @@ import (
 
 // mongoDBURIOpts represents mongoDBURI's options.
 type mongoDBURIOpts struct {
+	baseURI        string
```
AlekSi, line 36: I will like we are going back and forth there.

`mongoDBURI` returns a MongoDB URI for the given options. Those options are created only by the in-process listener, where no URI exists and we need to create one.

When URI exists (for example, when the target system is not in-process FerretDB), we just use it with extra query options for specific tests if they need them.

With this change, `mongoDBURI` could be used to construct MongoDB URI from many different options, making it much harder to understand.

chilagrow: Okie, so mongoDBURI should not be modified because it's for in-process listener and parameters should stay vanilla. Once mongoDBURI is built, then I can use that uri to add extra query options then.

integration/query_test.go
chilagrow: Yep thanks

integration/query_test.go
chilagrow: In practice, I'm not 100% sure that it eliminates them. But that is the hope yes.

integration/commands_diagnostic_test.go
```diff
@@ -404,3 +406,106 @@ func TestCommandsDiagnosticWhatsMyURI(t *testing.T) {
 		assert.NotEqual(t, ports[0], ports[1])
 	}
 }
+
+// TestCommandWhatsMyURIConnection tests that a client uses the same connection for
+// all commands run by the client, and different clients use different connection.
+// The port number is used to validate if the same connection is used or not,
+// since unique port number is assigned to each client connection in the test setup.
```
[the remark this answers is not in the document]
chilagrow: Yes true thanks, purely for me to check our integration setup is working.... Let me update the comment.

integration/query_test.go
```diff
@@ -1280,3 +1324,114 @@ func TestQueryCommandGetMore(t *testing.T) {
...
+		err = collection1.Database().RunCommand(s.Ctx, bson.D{
+			{"getMore", cursorID},
+			{"collection", collection1.Name()},
+		}).Decode(&res)
+		require.NoError(t, err)
+	})
+
+	t.Run("DifferentClient", func(t *testing.T) {
+		// error returned from using different clients are session related error,
+		// hence currently FerretDB does not return an error
```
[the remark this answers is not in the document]
chilagrow: Last I tested, FerretDB does not return error. I could return not implemented error in place though.

integration/setup/client.go
AlekSi: We could probably rename that function to make it more clear

integration/commands_diagnostic_test.go
```diff
@@ -404,3 +406,106 @@ func TestCommandsDiagnosticWhatsMyURI(t *testing.T) {
...
+func TestCommandWhatsMyURIConnection(t *testing.T) {
+	t.Parallel()
+
+	// set minPoolSize and maxPoolSize 1 to ensure only one pool exists duration of the test,
+	// which forces a client to use a single connection pool
```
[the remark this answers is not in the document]
chilagrow: Thanks let me update the comment!

integration/query_test.go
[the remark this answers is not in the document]
AlekSi: https://github.com/FerretDB/FerretDB/blob/5b1e6b61547e7e644e3c023694baa2adff21afde/internal/handlers/common/getmore.go#L89-L90

https://github.com/FerretDB/FerretDB/blob/5b1e6b61547e7e644e3c023694baa2adff21afde/internal/handlers/common/getmore.go#L111

If there is a bug, we should fix it. In any case, that test should check that some error is returned

integration/setup/client.go
chilagrow: Any preference on the name?  `InProcessFerretDBURI`?

integration/query_test.go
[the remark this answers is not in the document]
chilagrow: Perfect, let me check.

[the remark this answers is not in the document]
AlekSi: Ah… Oh… Well, yeah, I get it now… Yeah, okat :)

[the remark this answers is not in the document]
chilagrow: > Ah… Oh… Well, yeah, I get it now… Yeah, okay :)

What did I miss?? :thinking:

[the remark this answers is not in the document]
AlekSi: In that case, the username, database, and collection are all the same. In theory, even without sessions, we could check that the _connection_ (client's port) is the same. But that would probably break more things. So yeah, let's keep it.

integration/query_test.go
AlekSi: The only case when it will not be fixed is if the client closes and re-opens that single connection. If it does, we could use it as we want in all those tests. With `maxIdleTimeMS=0` (which is a default, but I think we could pass that explicitly to make things a bit more clear), that should not happen.

chilagrow: Thanks for looking into this. Great.

integration/setup/client.go
AlekSi: I don't know :)

integration/setup/setup.go
```diff
@@ -133,13 +137,31 @@ func SetupWithOpts(tb testing.TB, opts *SetupOpts) *SetupResult {
...
+			for _, v := range vs {
+				q.Set(k, v)
+			}
+		}
+
+		u.RawQuery = q.Encode()
+		uri = u.String()
+
+		// If ExtraOptions is set for in-process FerretDB, two clients are created
+		// on purpose. setupListener creates a client to check in-process FerretDB
```
AlekSi, line 162: No, there is no purpose now :)

AlekSi (approved):

integration/setup/setup.go
AlekSi: @chilagrow PTAL at my changes

chilagrow: Thank you for refactoring that part of the code :+1:

noisersup (approved):
:+1:

PR #2880, opened 2023-06-21
Add basic tests for iterators

AlekSi:
# Description

Closes #{issue_number}.

## Readiness checklist

### DIFF
--- a/integration/commands_administration_test.go
+++ b/integration/commands_administration_test.go
@@ -1071,6 +1071,8 @@ func TestCommandsAdministrationServerStatusFreeMonitoring(t *testing.T) {
 }
 
 func TestCommandsAdministrationServerStatusStress(t *testing.T) {
+	// TODO rewrite using teststress.Stress
+
 	setup.SkipForTigrisWithReason(t, "https://github.com/FerretDB/FerretDB/issues/1507")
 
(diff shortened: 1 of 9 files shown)


integration/commands_administration_test.go
```diff
@@ -1071,6 +1071,8 @@ func TestCommandsAdministrationServerStatusFreeMonitoring(t *testing.T) {
 }
 
 func TestCommandsAdministrationServerStatusStress(t *testing.T) {
+	// TODO rewrite using teststress.Stress
```
rumyantseva, line 1074: Wow wow! :)

AlekSi: wow? Ok, but that PR is neither approved nor rejected, and I'm not sure what to do with that comment

rumyantseva: No, no I'm still reading :) I like the idea, do we need an issue number for these new TODOs?

rumyantseva (approved):
I like the idea of moving this separate `Stress` logic into a separate function! One suggestion - maybe in our test naming convention we should mention that there are those `Stress` tests and how they should be named. Could be in the same PR where the existing tests will use the new `Stress` function.

noisersup (approved):
Makes sense!

PR #2882, opened 2023-06-21
Add validating max nested document/array depth

### LINKED ISSUE #1639: Validate maximum nested document/array depth
### What should be done?

Just like MongoDB, we should disallow documents and arrays that exceed the maximum nested object depth.

For now, let's do that for incoming documents only in `bson` and/or `wire` packages, or just in validation in `types` package.

DoD:
- add a simple compat test (check that FerretDB returns an error when MongoDB does, the error doesn't need to be the same)
- @noisersup: create a follow-up issue to make the errors equal and to cover all the leftovers if any (e.g. dance tests if needed)

kropidlowsky:
# Description

Closes #1639.

Add validating max nested document/array depth to align it with MongoDB requirement.

## Readiness checklist


internal/bson/document.go
```diff
@@ -196,9 +201,13 @@ func (doc *Document) ReadFrom(r *bufio.Reader) error {
 			fields = append(fields, field{key: key, value: value})
 
 		case tagArray:
-			// TODO check max nesting https://github.com/FerretDB/FerretDB/issues/1639
-
 			var v arrayType
+
+			doc.nesting = +1
```
rumyantseva, line 206: What does this expression mean? `doc.nesting ` is set to `1`? Why?

integration/helpers.go
```diff
@@ -437,3 +438,34 @@ func generateDocuments(startID, endID int32) bson.A {
 
 	return docs
 }
+
+// CreateNestedDocument creates a mock nested Bson document. The nesting level is based on integer parameter.
+func CreateNestedDocument(n int) bson.M {
```
noisersup, line 443: It seems to me that we test for nested documents but don't test documents nested in arrays, nested in documents, ...; Can we modify this function to produce this type of nesting?

noisersup (changes requested):

kropidlowsky:
Hi,

Sorry for not responding - I made  the reaction to the last comment to let you know I do remember.
I could not continue, because I got quite busy at work recently. I wanted to go on with it around now.

I saw some updates on this PR + original issue from @noisersup .
Please let me know if there is anything what I can make here or to take another issue :blush:

integration/nested_document_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"go.mongodb.org/mongo-driver/bson"
+)
+
+func TestCreateNestedDocument(t *testing.T) {
+	t.Parallel()
+
+	t.Run("0", func(t *testing.T) {
+		embdyDoc := bson.D{{"v", nil}}
```
AlekSi, line 28: What is "embdy"? Typo for "empty"? But then it is not empty?..

I think we could just use `expected` and `actual` :)

internal/bson/bson.go
```diff
@@ -59,7 +59,7 @@ import (
 type bsontype interface {
 	bsontype() // seal for go-sumtype
 
-	ReadFrom(*bufio.Reader) error
+	ReadFrom(*bufio.Reader, int) error
```
AlekSi, line 62:
```suggestion
	// good explanatory comment
	readNested(*bufio.Reader, int) error
	ReadFrom(*bufio.Reader) error
```

I think we should not leak that detail into the exported API. `wire` package should not change. `bson_test.go` probably too

integration/nested_document_test.go
```diff
@@ -0,0 +1,44 @@
...
+package integration
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.mongodb.org/mongo-driver/bson"
+)
+
+func TestCreateNestedDocument(t *testing.T) {
```
rumyantseva, line 24: I like that we test it! :)

internal/bson/array.go
```diff
@@ -29,9 +29,13 @@ type arrayType types.Array
 func (a *arrayType) bsontype() {}
 
 // ReadFrom implements bsontype interface.
-func (a *arrayType) ReadFrom(r *bufio.Reader) error {
+func (a *arrayType) ReadFrom(r *bufio.Reader, nesting int) error {
+	if nesting > maxNesting {
+		return lazyerrors.Errorf("bson.Document.ReadFrom (over nested document. Max supported nesting: %d)", maxNesting)
```
rumyantseva, line 34: Let's maybe also print `nesting` too?
Also, the message might be confusing for someone who sees it for the first time.
Maybe something like "the document has the depth X, max supported depth is Y"?

Also, this is not `bson.Document` but rather `bson.Array` I guess.

internal/bson/bson.go
rumyantseva, line 62: Well, this `int` without any explanation looks a bit confusing here. Also, it doesn't feel like it really belongs to this interface. If I have a simple type, I don't have any depth.

I understand why you introduced it - you need to know what is the current depth.
But to me, it rather looks like there is one more interface here, that specifies that nested nature - only for arrays and documents. Then you can keep this `ReadFrom` as it is, and you don't need to have calls like `doc.ReadFrom(bufr, 0)`.


Something like 

```
type bsonnested interface {
    ReadFrom(*bufio.Reader, int) error
```

and only Array and Document should implement it.

And then the original ReadFrom will be the one that calls this "nested" ReadFrom. 
This way, we hide that internal information about nestedness from the initial callers of `ReadFrom`.

If my idea is not clear, just ping me for a quick pairing session.

internal/bson/document.go
```diff
@@ -135,7 +135,11 @@ func (doc *Document) Values() []any {
 }
 
 // ReadFrom implements bsontype interface.
-func (doc *Document) ReadFrom(r *bufio.Reader) error {
+func (doc *Document) ReadFrom(r *bufio.Reader, nesting int) error {
+	if nesting > maxNesting {
+		return lazyerrors.Errorf("bson.Document.ReadFrom (over nested document. Max supported nesting: %d)", maxNesting)
```
rumyantseva, line 140: Same as above (maybe a bit more clarity for this error message).

rumyantseva (commented):
I understand why the interface changed this way, but I suggest keeping it as is and introducing a new, more internal, interface. This way, we can hide implementation details. 

Let me know if it's unclear, or you disagree or something :)

integration/basic_test.go
```diff
@@ -604,6 +604,53 @@ func TestDebugError(t *testing.T) {
...
+		},
+		"10ok": {
+			doc: CreateNestedDocument(10),
+		},
+		"100ok": {
+			doc: CreateNestedDocument(100),
+		},
+		"179ok": {
+			doc: CreateNestedDocument(179),
+		},
```
chilagrow, line 625: Currently we only test nesting limit for document, could you test also array as well? Is something like `[{doc:[{doc:[...]}]}]`  not possible? :thinking:

internal/bson/bson.go
rumyantseva: yeah, sounds good to me too!

AlekSi: There is also no need to export those methods

internal/bson/array.go
noisersup: As we don't proceed the document after exceeding the maxDepth, we don't know the actual depth of provided exceeding document

integration/basic_test.go
noisersup: The only type that can be inserted is the document, any other type like array is inside provided document, right?

AlekSi: FWIW, I think it is not a problem to support arrays inside arrays there, even if they are rejected at a later layer (just like duplicate fields, for example)

internal/bson/document.go
```diff
@@ -136,6 +136,18 @@ func (doc *Document) Values() []any {
 
 // ReadFrom implements bsontype interface.
 func (doc *Document) ReadFrom(r *bufio.Reader) error {
+	return doc.readNested(r, 0)
+}
+
+// readNested, similarly to ReadFrom, takes raw bytes from reader
+// and unmarshal them to the Document.
+// It also takes the nesting value (if called from different bsontype),
+// and checks if the document doesn't exceed the max nesting allowed.
```
chilagrow, line 145: I think this comment can be simplified. We check nesting for document and array, that's it right?

integration/basic_test.go
chilagrow: Yeah that's right. Let's add a simple test case for nested array then :+1:

internal/bson/document.go
```diff
@@ -136,6 +136,18 @@ func (doc *Document) Values() []any {
...
+func (doc *Document) readNested(r *bufio.Reader, nesting int) error {
+	if nesting > maxNesting {
+		return fmt.Errorf("bson.Document.ReadFrom (document has exceeded the max supported nesting: %d)", maxNesting)
```
rumyantseva, line 148:
```suggestion
		return fmt.Errorf("bson.Document.readNested: document has exceeded the max supported nesting: %d", maxNesting)
```

internal/bson/array.go
```diff
@@ -30,8 +30,20 @@ func (a *arrayType) bsontype() {}
...
+	return a.readNested(r, 0)
+}
+
+// readNested, similarly to ReadFrom, takes raw bytes from reader
+// and unmarshal them to the arrayType.
+// It also takes the nesting value (if called from different bsontype),
+// and checks if the document doesn't exceed the max nesting allowed.
+func (a *arrayType) readNested(r *bufio.Reader, nesting int) error {
+	if nesting > maxNesting {
+		return lazyerrors.Errorf("bson.Array.ReadFrom (document has exceeded the max supported nesting: %d)", maxNesting)
```
rumyantseva, line 42:
```suggestion
		return lazyerrors.Errorf("bson.Array.readNested: document has exceeded the max supported nesting: %d", maxNesting)
```

rumyantseva (approved):
I like the new approach: it's much simpler and we didn't modify the common interface.

integration/basic_test.go
noisersup: How do we want to test nested arrays in integration tests? They'll behave differently on Mongodb

AlekSi: By not testing it with integration tests, I guess :)

noisersup: Well, to create an unit test we might need to provide binary representation as an input, and I'm not sure if that's what we do

chilagrow: Maybe I'm misunderstanding how nesting validation works. But if we have document nested in array nested in document etc, only 179 level of nesting is valid right? This is allowed by both so I feel that is testable by integration test.

As for pure nested array, I _thought_ that is validated after we check for 179 level of nesting (maybe I'm wrong). If that's that case, we can test nesting array of 197 and check that it returns error in integration test?

chilagrow: Actually testing array is already done and I didn't understand `createNestedDocument` function :see_no_evil:

chilagrow (approved):
:rocket:

rumyantseva (approved):
LGTM

AlekSi (approved):

AlekSi:
@rumyantseva please resolve your conversations

AlekSi:
@noisersup that PR requires your approval 😁

PR #2922, opened 2023-06-27
Convert SQLite directory to URI

### LINKED ISSUE #2753: Convert SQLite URI to directory
### What should be done?

SQLite backend should accept not a directory or file name, but SQLite URI (see https://www.sqlite.org/uri.html and https://www.sqlite.org/c3ref/open.html#urifilenamesinsqlite3open) with the following changes:
* it should not contain a data file name, only a directory name (because we support multiple databases / files in the same directory);
* the path should end with `/` (to highlight that this is a directory);
* authority could be only omitted or empty; it can't be "localhost" or any other non-empty value.

(the query parameters should still be allowed)

Valid examples:
* `file:./` (current directory);
* `file:./dir/` (subdirectory `dir` in the current directory);
* `file:dir/?mode=ro` (the same as above with a query parameter);
* `file:/dir/` (absolute directory path);
* `file:///dir/` (the same as above with empty authority).

Backend should correctly add database file name to that URI even when query parameters are present.

### Where?

SQLite backend and handler.

### Definition of Done

* [ ] document that

- all handlers updated;
- unit tests added/updated;
- spot refactorings done;
- something else?

w84thesun:
# Description

Closes #2753.

## Readiness checklist


internal/backends/sqlite/backend.go
```diff
@@ -29,37 +31,79 @@ import (
...
 	}
 
 	return backends.BackendContract(&backend{
 		r:   r,
-		dir: params.Dir,
+		uri: uri,
 	}), nil
 }
 
+// validateURI checks given URI value and returns parsed URI.
```
chilagrow, line 61:
```suggestion
// validateURI checks given URI value and returns parsed URL.
```

internal/backends/sqlite/backend.go
```diff
@@ -29,37 +31,79 @@ import (
...
+	}
+
+	dir := uri.Path
+
+	if uri.Opaque != "" {
+		dir = uri.Opaque
+	}
+
+	if !strings.HasSuffix(dir, "/") {
+		return nil, lazyerrors.Errorf("backend URI should be a directory: %q", value)
```
chilagrow, line 90:
```suggestion
		return nil, lazyerrors.Errorf("backend URI should be a directory ending with '/': %q", value)
```

internal/backends/sqlite/backend.go
```diff
@@ -29,37 +31,79 @@ import (
...
+	}
+
+	if uri.Host != "" {
+		return nil, lazyerrors.Errorf("backend URI should not contain host: %q", value)
+	}
+
+	dir := uri.Path
+
+	if uri.Opaque != "" {
+		dir = uri.Opaque
```
chilagrow, line 86: What condition is it not sufficient to check `uri.Path` contains `/` but must check `uri.Opaque`? Is it something to do with query parameter? :thinking:

internal/backends/sqlite/backend_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewBackend(t *testing.T) {
+	t.Parallel()
+
+	err := os.MkdirAll("tmp/dir", os.ModePerm)
```
chilagrow, line 28: Is the reason not using `t.TempDir()` because we want to use current dir `.` notation? :thinking:

chilagrow (commented):
Looks good, just some tiny comments

internal/backends/sqlite/backend.go
w84thesun: It’s how `url.URL` works. Local paths would be in the `Opaque` field.

internal/backends/sqlite/backend_test.go
w84thesun: Yes, exactly 😀

internal/backends/sqlite/backend.go
chilagrow: You would need to update the test too :see_no_evil:

chilagrow (commented):
Looks good, let's update unit test to pass.

chilagrow (dismissed):

noisersup (dismissed):
Looks good to me!

internal/backends/sqlite/metadata/pool/pool.go
```diff
@@ -50,17 +51,17 @@ type Pool struct {
 }
 
 // New creates a pool for SQLite databases in the given directory.
-func New(dir string, l *zap.Logger) (*Pool, error) {
+func New(uri *url.URL, l *zap.Logger) (*Pool, error) {
 	// TODO accept URI with a directory name, not just directory name
 	// https://github.com/FerretDB/FerretDB/issues/2753
```
AlekSi, line 56: I guess that should be done in that PR

internal/backends/sqlite/metadata/pool/pool.go
```diff
@@ -50,17 +51,17 @@ type Pool struct {
...
 
-	matches, err := filepath.Glob(filepath.Join(dir, "*"+filenameExtension))
+	matches, err := filepath.Glob(filepath.Join(uri.Path, "*"+filenameExtension))
 	if err != nil {
 		return nil, lazyerrors.Error(err)
 	}
 
 	p := &Pool{
-		dir:   dir,
+		dir:   uri.Path,
```
AlekSi, line 64: 🛑 And after that, we lose query parameters that we should pass to `openDB` below, as well in `GetOrCreate` via `databasePath` and maybe somewhere else

internal/backends/sqlite/backend.go
```diff
@@ -29,37 +31,79 @@ import (
...
 	return backends.BackendContract(&backend{
 		r:   r,
-		dir: params.Dir,
+		uri: uri,
 	}), nil
 }
 
+// validateURI checks given URI value and returns parsed URL.
+// URI should contain 'file' scheme and point to an existing directory.
+// Path should end with '/'. Authority should be empty.
```
AlekSi, line 63: Authority could also be absent: empty – `file://<empty>/`, absent – `file:/`. See https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax

internal/backends/sqlite/backend_test.go
```diff
@@ -0,0 +1,153 @@
...
+	_, err = os.Create("tmp/file")
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		err := os.Remove("tmp/file")
+		require.NoError(t, err)
+		err = os.Remove("tmp/dir")
+		require.NoError(t, err)
+		err = os.Remove("tmp")
+		require.NoError(t, err)
```
AlekSi, line 40: `os.RemoveAll`

AlekSi (changes requested):

cmd/ferretdb/main.go
```diff
@@ -104,7 +104,7 @@ var pgFlags struct {
 //
 // See main_sqlite.go.
 var sqliteFlags struct {
-	SQLiteURI string `name:"sqlite-uri" default:"." help:"Directory path or 'file' URI for 'sqlite' handler."`
+	SQLiteURI string `default:"." help:"'file' scheme URI with directory path for 'sqlite' handler." name:"sqlite-uri" `
```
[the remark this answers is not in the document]
w84thesun: I changed the order because of that:
```
Error: cmd/ferretdb/main.go:107:19: tag is not aligned , should be: default:"file://." help:"'file' scheme URI with directory path for 'sqlite' handler." name:"sqlite-uri" (tagalign)
	SQLiteURI string `name:"sqlite-uri" default:"file://." help:"'file' scheme URI with directory path for 'sqlite' handler." `
```
Should I add a `nolint` here?

internal/backends/sqlite/backend_test.go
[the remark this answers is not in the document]
w84thesun: Fixed

internal/backends/sqlite/metadata/pool/pool.go
w84thesun: Missed that, sorry.

internal/backends/sqlite/metadata/pool/pool.go
w84thesun: Saved `uri` instead of `dir`.

internal/backends/sqlite/backend.go
w84thesun: Mentioned that it could be absent.

internal/backends/sqlite/backend_test.go
w84thesun: Fixed.

cmd/ferretdb/main.go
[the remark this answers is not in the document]
AlekSi: https://golangci-lint.run/usage/linters/#tagalign

chilagrow (dismissed):

internal/backends/sqlite/metadata/pool/pool.go
AlekSi: That does not change the fact that openDB accepts matched files (without query parameters) in one place and the database path (also without query parameters) in another.

PR #2934, opened 2023-06-28
Fix integration tests after bumping deps

noisersup:
# Description

Removed the test as it tests driver and not the database.

## Readiness checklist

### DIFF
--- a/integration/basic_test.go
+++ b/integration/basic_test.go
@@ -15,7 +15,6 @@
 package integration
 
 import (
-	"errors"
 	"fmt"
 	"math"
 	"strings"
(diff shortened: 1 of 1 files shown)


integration/basic_test.go
```diff
@@ -524,7 +524,7 @@ func TestDatabaseName(t *testing.T) {
 		ctx, collection := setup.Setup(t)
 
 		err := collection.Database().Client().Database("").CreateCollection(ctx, collection.Name())
-		expectedErr := driver.InvalidOperationError(driver.InvalidOperationError{MissingField: "Database"})
+		expectedErr := errors.New("database name cannot be empty")
```
AlekSi, line 527: That error is returned by the driver, not FerretDB or MongoDB. We should just delete that test.

noisersup: That's right

AlekSi: We could, in theory, write a test that uses `RunCommand` to do `insert`… But let's not do that now.

AlekSi (approved):

rumyantseva (approved):
Thanks for the quick fix! It answers the exact question I was about to ask :)

PR #2939, opened 2023-06-29
Implement `killCursors` command

### LINKED ISSUE #1514: Implement `killCursors` command
### What should be done?

* https://www.mongodb.com/docs/manual/reference/command/killCursors/#mongodb-dbcommand-dbcmd.killCursors

AlekSi:
# Description

Closes #1514.

## Readiness checklist


internal/handlers/common/killcursors.go
```diff
@@ -0,0 +1,112 @@
...
+			if errors.Is(err, iterator.ErrIteratorDone) {
+				break
+			}
+
+			return nil, lazyerrors.Error(err)
+		}
+
+		id, ok := v.(int64)
+		if !ok {
+			return nil, commonerrors.NewCommandErrorMsgWithArgument(
```
chilagrow, line 79: If cursors array contains a valid cursor and an invalid cursor like `bson.A{Long(1234),"foo"}`, I *feel* the command shouldn't kill the first cursor. :thinking:

chilagrow (commented):
Just one small comment :hugs:

CONTRIBUTING.md
```diff
@@ -248,6 +248,13 @@ Writing separate tests might be much better than making a single test that check
 
 Also, we should use driver methods as much as possible instead of testing commands directly via `RunCommand`.
 
+We have a lot of existing helpers to convert the driver's `bson.D` values to our `*types.Document` values,
+to compare them, etc.
+In most cases, they should be used instead of (deprecated) `bson.D.Map()`,
+`bson.Unmarshal`, decoding into a struct with `bson` tags, comparing fields one-by-one, etc.
+The bar for adding new helpers is very high.
+Please check all existing ones.
```
rumyantseva, line 256: 👍

rumyantseva (approved):
LGTM

internal/handlers/common/killcursors.go
```diff
@@ -0,0 +1,112 @@
+// Copyright 2021 FerretDB Inc.
```
quasilyte, line 1: What do you think about adding a comment or two about the copyright headers in the contributions guide?

There is more than one way to do the copyrights:
1. always use the same date (the first copyright's date)
2. a new file gets the current year's value in the header
3. increase the date annually, in every file (in which case a transient inconsistency is OK as it will be fixed later by the tool)

The Go project used the second (2) approach, so if you see a mismatching year in a new file, it's something that can be reported on the review. It doesn't look like it's the case here  (2021 is used everywhere), but it should be described somewhere, so we never get any confusion like this in the future.

(Or maybe you have a tool that checks the headers that ensures the year 2021, in which case never mind.)

internal/handlers/common/killcursors.go
AlekSi: Added a test for that also demonstrates the usage of XFail

internal/handlers/common/killcursors.go
AlekSi: We tend not to document things that are enforced by the linter.

We also can't describe something that could be interpreted by someone somewhere as legal advice without saying "THIS IS NOT A LEGAL ADVICE" in all caps and with much more words than "a comment or two".

internal/handlers/common/killcursors.go
chilagrow: Great, very fun of this helper function!

integration/query_test.go
```diff
@@ -834,3 +838,640 @@ func TestQueryCommandSingleBatch(t *testing.T) {
 		})
 	}
 }
+
+func TestQueryBatchSize(t *testing.T) {
```
chilagrow, line 842: Feel free to remove all changes in `query_test.go` file. They were moved and renamed to `getmore_test.go` :hugs:

AlekSi: Done

chilagrow (approved):
This is good to go by me!

rumyantseva (approved):
LGTM

PR #2953, opened 2023-07-03
Add integration tests for `maxTimeMS` in `find`, `aggregate` and `getMore`

### LINKED ISSUE #1808: Support `find`'s / `aggregate`'s / `getMore`'s `maxTimeMS` parameter
### What should be done?

As we don't support backend-side cursors for now just return from `getMore` when timeout is exceeded potentially removing cursor from `connInfo`.

See also:
https://www.mongodb.com/docs/manual/reference/method/cursor.maxTimeMS/
https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.rst#cursors
```[tasklist]
### Tasks
- [ ] https://github.com/FerretDB/FerretDB/issues/2983
- [ ] https://github.com/FerretDB/FerretDB/issues/2963
- [ ] https://github.com/FerretDB/FerretDB/issues/2984
```

chilagrow:
# Description

Closes #1808.

In this PR, maxTimeMS is not yet supported. So do not close the issue. It validates maxTimeMS parameter and runs integration test, some tests are only for MongoDB.

## Readiness checklist


integration/getmore_test.go
```diff
@@ -701,3 +703,379 @@ func TestGetMoreCommandConnection(t *testing.T) {
...
+		assert.NotZero(t, cursorID)
+
+		err = collection.Database().RunCommand(ctx, bson.D{
+			{"getMore", cursorID},
+			{"collection", collection.Name()},
+			{"batchSize", 2000},
+			{"maxTimeMS", 1},
+		}).Decode(&res)
+
+		AssertEqualAltCommandError(
```
chilagrow, line 1070: > maxTimeMS cannot be set when getMore is called on a non-tailable cursor.

from https://www.mongodb.com/docs/manual/reference/command/getMore/#command-fields

integration/aggregate_documents_compat_test.go
```diff
@@ -285,16 +284,13 @@ func TestAggregateCommandCompat(t *testing.T) {
 			},
 			resultType: emptyResult,
 		},
-		"MaxTimeMSNegative": {
```
chilagrow, line 288: Moved to `TestAggregateCommandMaxTimeMSErrors/NegativeLong`

integration/aggregate_documents_compat_test.go
```diff
@@ -228,8 +228,7 @@ func testAggregateCommandCompat(t *testing.T, testCases map[string]aggregateComm
 					return
 				}
 				require.NoError(t, compatErr, "compat error; target returned no error")
-
-				AssertEqualDocumentsSlice(t, compatRes, targetRes)
+				require.Equal(t, compatRes, targetRes)
```
rumyantseva, line 231: So, the result is a document, not a slice, right? Should we use `AssertEqualDocuments`?

internal/handlers/common/getmore.go
```diff
@@ -81,7 +83,56 @@ func GetMore(ctx context.Context, msg *wire.OpMsg, registry *cursor.Registry) (*
 	}
 
 	// TODO maxTimeMS https://github.com/FerretDB/FerretDB/issues/1808
-	// TODO comment
```
rumyantseva, line 84: I think this todo was to handle `comment` param (e.g. we have it ignored in the other parts of the code). So, as I don't see `comment` being handled here, this TODO seems to be actual (we might need an issue for it).

internal/handlers/pg/msg_aggregate.go
```diff
@@ -91,10 +93,64 @@ func (h *Handler) MsgAggregate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
 		v = int64(0)
 	}
 
-	maxTimeMS, err := commonparams.GetValidatedNumberParamWithMinValue(document.Command(), "maxTimeMS", v, 0)
+	// cannot use other existing commonparams function, they return different error codes
+	maxTimeMS, err := commonparams.GetWholeNumberParam(v)
 	if err != nil {
-		// unreachable for MongoDB GO driver, it validates maxTimeMS parameter
-		return nil, lazyerrors.Error(err)
+		switch {
```
rumyantseva, line 99: Is this switch-case the same as in `internal/handlers/common/getmore.go`? If so, maybe we can move the common code to some function and call it? (Though, it feels like the opposite of what we did a while ago...)

rumyantseva (commented):
The code looks good to me overall, I left a couple of tiny questions/comments to clarify a couple of details.

internal/handlers/pg/msg_aggregate.go
chilagrow: Unfortunately they are not the same unfortunately, their only difference of error code is upon `commonparams.ErrLongExceededNegative` returns `ErrBadValue` for `getMore` while `ErrLongExceededNegative` for `aggregate`.

integration/aggregate_documents_compat_test.go
chilagrow: Yes they are documents. Don't we used `AssertEqualDocuments` when response contains special bson that cannot be compared with `Equal`? I thought testify package assertions were preferred when it works :thinking:

rumyantseva: I'm not sure anymore 😅 
@AlekSi what is preferred to check responses - to use `assert.Equal` as much as possible or to use `AssertEqualDocuments` as much as possible and why?

internal/handlers/pg/msg_aggregate.go
rumyantseva: Ah! Well, yes, then we have to keep both! Thanks for the explanation.

noisersup (dismissed):
LGTM!

internal/handlers/common/getmore.go
chilagrow: Added issue link, see around L137

integration/aggregate_documents_compat_test.go
chilagrow: I really don't have strong opinion on this, let's just use `AssertEqualDocuments` as suggested. See the update :hugs:

rumyantseva (dismissed):
LGTM

noisersup (dismissed):
Good job!

noisersup (approved):

rumyantseva (approved):
LGTM

PR #2965, opened 2023-07-04
Implement ping command for SQLite

### LINKED ISSUE #2756: Implement `ping` command for SQLite
### What should be done?

`ping` command should actually connect to the backend.

Authentication is not in the scope.

### Where?

- SQLite handler.

### Definition of Done

- SQLite handler updated;
- spot refactorings done.

noisersup:
# Description

Closes #2756.

## Readiness checklist

### DIFF
--- a/internal/backends/database.go
+++ b/internal/backends/database.go
@@ -36,6 +36,7 @@ type Database interface {
 	Close()
 
 	Collection(string) Collection
+	Ping() error
 	ListCollections(context.Context, *ListCollectionsParams) (*ListCollectionsResult, error)
 	CreateCollection(context.Context, *CreateCollectionParams) error
 	DropCollection(context.Context, *DropCollectionParams) error
--- a/internal/handlers/sqlite/msg_ping.go
+++ b/internal/handlers/sqlite/msg_ping.go
@@ -25,6 +25,10 @@ import (
 // MsgPing implements HandlerInterface.
 func (h *Handler) MsgPing(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
 	var reply wire.OpMsg
+
+	db := h.b.Database("")
+	defer db.Close()
+
 	must.NoError(reply.SetSections(wire.OpMsgSection{
 		Documents: []*types.Document{must.NotFail(types.NewDocument(
 			"ok", float64(1),


internal/backends/sqlite/database.go
```diff
@@ -95,6 +95,26 @@ func (db *database) DropCollection(ctx context.Context, params *backends.DropCol
...
+		if err != nil {
+			return lazyerrors.Error(err)
+		}
+
+		defer db.r.DatabaseDrop(ctx, db.name)
+	}
+
+	return sqlDB.PingContext(ctx)
+}
+
```
noisersup, line 117: I decided to create `Ping` as a part of database interface to use native backends ping commands in place of "implementing" pings in our way, feel free to discuss this approach.

AlekSi: I don't think we should do that. On the one hand, our interface is "fat" enough to allow efficient implementations – for example, `Query` is very flexible in what it pushes down and not. On the other hand, it should be "thin" enough to avoid extra implementation. For example, `ping` command handle could just call `ListDatabases` – that would check that database exists but does not require each backend to implement one extra command.

noisersup: We cannot guarantee if `ListDatabases` implementation for specific backend does actually initialize any connection. Even in SQLite `ListDatabases`, doesn't query anything from sqlite module, but access existing `Pool` struct with stored `dbs` in it.

noisersup: I wonder what do we expect from ping? My understanding for now is that we want to ensure that it's possible to have any connectivity with the backend. With Postgres it's easy as this connectivity is just connection over socket/network, but SQLite? What do we want to ensure?

internal/backends/sqlite/database.go
```diff
@@ -95,6 +95,26 @@ func (db *database) DropCollection(ctx context.Context, params *backends.DropCol
...
+	// (there might be some fs permission errors)
+	if sqlDB == nil {
+		var err error
+
+		sqlDB, err = db.r.DatabaseGetOrCreate(ctx, db.name)
+		if err != nil {
+			return lazyerrors.Error(err)
+		}
+
+		defer db.r.DatabaseDrop(ctx, db.name)
```
rumyantseva, line 112: Here I have my favorite question - what if someone worked with a database with this name in another request (e.g. inserted some data)? In this case, we will drop it. 

I understand why you do it - there are not many commands that would cause interaction with the database, but this particular approach doesn't really look safe. What do you think?

integration/basic_test.go
```diff
@@ -629,3 +629,37 @@ func TestDebugError(t *testing.T) {
...
+	})
+
+	t.Run("NonExistentDB", func(t *testing.T) {
+		expectedDatabases, err := db.Client().ListDatabases(ctx, bson.D{})
+		require.NoError(t, err)
+
+		dbName := "NonExistentDatabase"
+		for _, database := range expectedDatabases.Databases {
+			require.NotEqual(t, dbName, database.Name)
+		}
```
rumyantseva, line 654: Would `require.Contains(t, expectedDatabases.Databases, dbName)` work instead of this loop?

integration/basic_test.go
```diff
@@ -629,3 +629,37 @@ func TestDebugError(t *testing.T) {
...
+func TestPing(t *testing.T) {
+	// Shouldn't be ran in parallel, as NonExistentDB test requires
+	// the current state of database list, which can be changed by other tests.
+
+	ctx, collection := setup.Setup(t)
+	db := collection.Database()
+
+	t.Run("Multiple", func(t *testing.T) {
+		for i := 0; i < 5; i++ {
+			err := db.RunCommand(ctx, bson.D{{"ping", int32(1)}}).Err()
```
rumyantseva, line 642: Don't we want to test the response too, not only the error?

rumyantseva (commented):
The idea looks good to me, left a few comments/questions.

internal/backends/sqlite/database.go
chilagrow: Is there transaction equivalent which ensures rollback? Rollback on transaction is more solid than defer drop. :hugs:

integration/basic_test.go
```diff
@@ -629,3 +629,37 @@ func TestDebugError(t *testing.T) {
 		require.NoError(t, db.Client().Ping(ctx, nil), "other errors should not close connection")
 	})
 }
+
+func TestPing(t *testing.T) {
+	// Shouldn't be ran in parallel, as NonExistentDB test requires
+	// the current state of database list, which can be changed by other tests.
+
+	ctx, collection := setup.Setup(t)
+	db := collection.Database()
```
chilagrow, line 638: What is the expected response if ping fails, for example if DB is not accepting connection.

chilagrow (commented):
Just few things to think about :hugs:

integration/basic_test.go
rumyantseva: @chilagrow yes, exactly! Thanks for clarification :)

internal/backends/database.go
```diff
@@ -136,6 +138,18 @@ func (dbc *databaseContract) DropCollection(ctx context.Context, params *DropCol
 	return
 }
 
+// Ping uses native backend method to check if connection with database can be established.
+//
+// If database does not exist; it should be created automatically to check connection,
+// and dropped afterwards.
+func (dbc *databaseContract) Ping(ctx context.Context) (err error) {
+	defer observability.FuncCall(ctx)()
+	defer checkError(err)
```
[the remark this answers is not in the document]
noisersup: Nice catch! It totally looks like a bug.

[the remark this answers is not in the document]
AlekSi: I just wanted it to be higher in the function body (as it is basically a part of the documentation comment) and broke it during that process 😬

integration/basic_test.go
```diff
@@ -629,3 +629,37 @@ func TestDebugError(t *testing.T) {
 		require.NoError(t, db.Client().Ping(ctx, nil), "other errors should not close connection")
 	})
 }
+
+func TestPing(t *testing.T) {
+	// Shouldn't be ran in parallel, as NonExistentDB test requires
+	// the current state of database list, which can be changed by other tests.
```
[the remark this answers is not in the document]
noisersup: NonExistentDB test uses the `ListDatabases` command to ensure that no database is created during ping (which might be possible on implementation), to ensure that we take the databases list before and after ping.

integration/basic_test.go
noisersup: I believe the expected results are different for each backend.

MongoDB doesn't have any additional backend so they probably return `{ok:1}` or the request is just not delivered to the service. 

In our case - apart from "client -> FerretDB" connection we also want to make sure that FerretDB has connection with selected backend.

As different backend implementation might return different errors I guess the error message should be backend dependent. WDYT?

integration/basic_test.go
noisersup: It wouldn't work as `expectedDatabases.Databases()` returns the slice of structs that specify more than the `Name` field. The only way here is to access the name by the struct ".Name` field

integration/basic_test.go
```diff
@@ -629,3 +629,37 @@ func TestDebugError(t *testing.T) {
...
+		}
+
+		err = db.Client().Database(dbName).RunCommand(ctx, bson.D{{"ping", int32(1)}}).Err()
+		require.NoError(t, err)
+
+		// Ensure that we don't create database on ping
+		actualDatabases, err := db.Client().ListDatabases(ctx, bson.D{})
+		require.NoError(t, err)
+
+		assert.EqualValues(t, expectedDatabases, actualDatabases)
```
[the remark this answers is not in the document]
noisersup: If only want to check that `NonExistentDatabase` then yes, I guess it's sufficient. However there's a chance that our implementation uses some constant name to do ping. Although, as we decided to not implement `Ping` function for each backend but have one general implementation, I guess this test is too much.

integration/basic_test.go
rumyantseva: ok, then let's keep the manual check!

noisersup: Actually, I realized that we can just use `ListDatabases` filter for that:
```
		expectedDatabases, err := db.Client().ListDatabases(ctx, bson.D{{"name", dbName}})
		require.NoError(t, err)
		require.Empty(t, expectedDatabases.Databases)
```

integration/basic_test.go
```diff
@@ -629,3 +629,48 @@ func TestDebugError(t *testing.T) {
...
+
+		expectedDatabases, err := db.Client().ListDatabases(ctx, bson.D{{"name", dbName}})
+		require.NoError(t, err)
+		require.Empty(t, expectedDatabases.Databases)
+
+		res := db.Client().Database(dbName).RunCommand(ctx, bson.D{{"ping", int32(1)}})
+
+		var actualRes bson.D
+		err = res.Decode(&actualRes)
+		require.NoError(t, err)
```
AlekSi, line 668: If we don't check `actualErr`, we could just check `res.Err()`.

Or we just lost the comparison with `expectedRes`? I think we should check that `ok` is not `1` (`0`, absent, etc).

integration/basic_test.go
AlekSi: As long as it does not include `ok: 1`, we are good, but we should check that

integration/basic_test.go
```diff
@@ -629,3 +629,48 @@ func TestDebugError(t *testing.T) {
...
+
+	ctx, collection := setup.Setup(t)
+	db := collection.Database()
+
+	expectedRes := bson.D{{"ok", float64(1)}}
+
+	t.Run("Multiple", func(t *testing.T) {
+		t.Parallel()
+
+		for i := 0; i < 5; i++ {
```
AlekSi, line 644: What do we want to achieve there?

internal/handlers/sqlite/msg_ping.go
```diff
@@ -17,14 +17,38 @@ package sqlite
...
+
+	dbName, err := common.GetRequiredParam[string](document, "$db")
+	if err != nil {
+		return nil, err
+	}
+
+	db := h.b.Database(dbName)
+	defer db.Close()
+
+	// Collection name cannot be empty, so this is safe to use.
```
AlekSi, line 44: Uh, but any collection should be safe to use, no?

internal/backends/sqlite/database.go
AlekSi: Those are good points – ping should not cause any observable side-effects, and creating even a temporary database or collection is observable by a parallel client

internal/backends/database.go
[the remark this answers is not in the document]
AlekSi: Looked into it again – facepalm me, good job @quasilyte

AlekSi (approved):
LGTM overall to avoid blocking that PR. We are good as long as we don't create databases/collections and introduce changes in the contract (that should be completely reworked anyway…). I also think we miss a check for `ok` in the test, but that's minor

integration/basic_test.go
chilagrow: Thanks for clarification :+1:

As a user of ping, if ping doesn't return `{ok :1}` I would expect there is some issue with connection or setup. Can this be tested?

chilagrow (commented):
Just one comment about if we can to test failure case for ping.

rumyantseva (dismissed):
LGTM

integration/basic_test.go
noisersup: Not really, I'm not sure how can we simulate the lack of connection with SQLite.

AlekSi: For example, by using an existing non-writable directory

noisersup: We probably won't have any error here. Should ping check for that?

AlekSi: > We probably won't have any error here

Why not try that?

noisersup: Do we even have any Setup that changes sqlite dbs location?

integration/basic_test.go
noisersup: That pings are not closing connection and it's possible to invoke them multiple times.

internal/handlers/sqlite/msg_ping.go
noisersup: Yeah, good point

integration/basic_test.go
chilagrow: Yes, but it's a constant location at the moment https://github.com/FerretDB/FerretDB/blob/main/integration/setup/startup.go#L50

noisersup: This variable is used by newHandlerFunc, and setup functions connect to already existing handlers.

chilagrow: Creating the handler is part of the test setup too. What is the challenge here? :thinking:

noisersup: Creating the handler is part of Setup function, which is called only once from `main_test.go`. From my understanding there's no way of easily modifying the SQLite db directory for one specific test/test case.

chilagrow: I did quick experiment and found out even if we have 0x000 permission, it's not possible for query to fail, so ping will not fail on this.
https://github.com/FerretDB/FerretDB/pull/2997

chilagrow (dismissed):

rumyantseva (approved):
LGTM

chilagrow (approved):
LGTM

integration/basic_test.go
noisersup: done

PR #2988, opened 2023-07-05
Implement SQLite metadata transactions

### LINKED ISSUE #2747: Implement metadata caching
### What should be done?

- Add unique indexes for metadata `table` and `table_name` 
- Use ~transactions~ lock for collection creation and drop

### Where?

See references to this issue.

### Definition of Done

TBD

- SQLite backend updated;
- spot refactorings done;

noisersup:
# Description

Closes #2747.

This PR is a first part of work in the issue.

## Readiness checklist


internal/backends/sqlite/metadata/pool/pool.go
```diff
@@ -244,3 +244,34 @@ func (p *Pool) Drop(ctx context.Context, name string) bool {
...
 	return true
 }
+
+// InTransaction wraps the given function f in a SQLite transaction.
+//
+// If f returns an error, the transaction is rolled back.
+//
+// Passed context will be used for transaction.
+// Context cancellation DOES rollback the transaction.
+func InTransaction(ctx context.Context, db *sql.DB, f func(*sql.Tx) error) (err error) {
```
rumyantseva, line 254: So, for no `InTransaction` itself would be enough? We don't need the one with retries yet?

internal/backends/sqlite/metadata/registry.go
```diff
@@ -147,22 +147,25 @@ func (r *Registry) CollectionCreate(ctx context.Context, dbName string, collecti
...
-	query := fmt.Sprintf("CREATE TABLE %q (_ferretdb_sjson TEXT)", tableName)
-	if _, err = db.ExecContext(ctx, query); err != nil {
-		var e *sqlite.Error
-		if errors.As(err, &e) && e.Code() == sqlitelib.SQLITE_ERROR {
-			return false, nil
+	err = pool.InTransaction(ctx, db, func(tx *sql.Tx) error {
+		query := fmt.Sprintf("CREATE TABLE %q (_ferretdb_sjson TEXT)", tableName)
+		if _, err = tx.ExecContext(ctx, query); err != nil {
+			var e *sqlite.Error
+			if errors.As(err, &e) && e.Code() == sqlitelib.SQLITE_ERROR {
```
rumyantseva, line 154: I wonder if in this case we have the same problem as with pg: if an error happened during the transaction, the next query won't be executed. Is it possible to test/reproduce this to ensure that `insert` will be called, and not ignored?

rumyantseva (commented):
The implementation looks clear to me, just one question.

internal/backends/sqlite/metadata/pool/pool.go
```diff
@@ -244,3 +244,34 @@ func (p *Pool) Drop(ctx context.Context, name string) bool {
 
 	return true
 }
+
+// InTransaction wraps the given function f in a SQLite transaction.
+//
+// If f returns an error, the transaction is rolled back.
+//
+// Passed context will be used for transaction.
+// Context cancellation DOES rollback the transaction.
```
chilagrow, line 253: Can you add a test for cancelling context please.

internal/backends/sqlite/metadata/pool/pool.go
```diff
@@ -244,3 +244,34 @@ func (p *Pool) Drop(ctx context.Context, name string) bool {
...
+	var tx *sql.Tx
+
+	if tx, err = db.BeginTx(ctx, nil); err != nil {
+		err = lazyerrors.Error(err)
+		return
+	}
+
+	if err = f(tx); err != nil {
+		err = lazyerrors.Error(err)
+		_ = tx.Rollback()
```
chilagrow, line 264: If rollback fails, it is logged right? I don't know where we configure logging for rollback on sqlite :see_no_evil:

internal/backends/sqlite/metadata/pool/pool.go
chilagrow: We need unit test for this though! Let's at least cover basic scenarios for `f` producing error.

internal/backends/sqlite/metadata/pool/pool.go
```diff
@@ -244,3 +244,34 @@ func (p *Pool) Drop(ctx context.Context, name string) bool {
...
+// Context cancellation DOES rollback the transaction.
+func InTransaction(ctx context.Context, db *sql.DB, f func(*sql.Tx) error) (err error) {
+	var tx *sql.Tx
+
+	if tx, err = db.BeginTx(ctx, nil); err != nil {
+		err = lazyerrors.Error(err)
+		return
+	}
+
+	if err = f(tx); err != nil {
```
chilagrow, line 262: If `f(tx)` panics, is the transaction also rolled back? I think test would be helpful here too!

internal/backends/sqlite/metadata/registry.go
```diff
@@ -147,22 +147,25 @@ func (r *Registry) CollectionCreate(ctx context.Context, dbName string, collecti
...
+				return nil
```
chilagrow, line 155: Could add a comment  that `SQLITE_ERROR` is returned when a table exist? It's generic error code so a bit hard to figure out what is trying to capture. :hugs:

chilagrow (commented):
Let's add tests :hugs:

internal/backends/sqlite/metadata/pool/pool.go
AlekSi, line 254: If that function is needed at all (and I'm not sure about it), please:
1. Move it to where it is used.
2. Unexport it.
3. Make it handle the logic described in the large comment inside `InTransactionKeep`.

internal/backends/sqlite/metadata/registry.go
AlekSi: What is that problem?

internal/backends/sqlite/metadata/pool/pool.go
AlekSi: Why do we need retries?

AlekSi (commented):
Overall, it looks like we copy the bad design of pgdb without copying fixes for known bugs (see `InTransactionKeep`). This way, we are going to make an inferior copy of pgdb. We pretty much should do the opposite.

internal/backends/sqlite/metadata/registry.go
rumyantseva: Because if the second query in the transaction is not executed, the result will be different than what we want.

internal/backends/sqlite/metadata/pool/pool.go
rumyantseva: If a transaction fails because of a concurrent query, we might want to retry it.
I'm asking if we need it because I don't see if we can work with concurrent queries yet.

internal/backends/sqlite/metadata/registry.go
noisersup: Why do we need to execute the second query in this case? we return `false,nil` if there's the `SQLITE_ERROR` intentionally.

internal/backends/sqlite/metadata/pool/pool.go
noisersup: test added

internal/backends/sqlite/metadata/registry.go
noisersup: done

internal/backends/sqlite/metadata/pool/pool.go
noisersup: Afaik we don't log rollbacks in any place of our code. The error here might be returned here if transaction was already commited or rolled back, and we don't expect such situations.

noisersup:
> Overall, it looks like we copy the bad design of pgdb

I've asked team on DS if we like the `InTransaction` approach, and everybody liked it. I understand that we don't want to copy `pgdb` bad practices to the new package, but is there anything wrong with `InTransaction`?

internal/backends/sqlite/metadata/pool/pool.go
noisersup: Why is it not needed? It seems to me much better than implementing rollbacks for each new function. Implementing the transaction boilerplate in every function might create some places where we don't do rollback that should be done.

noisersup: Please take a look at `TestInTransactionRollback`, I believe it covers everything from the comment inside `InTransactionKeep`.

internal/backends/sqlite/metadata/pool/pool.go
chilagrow: Postgres logs failed rollback, but not on go application level. If there is network issue, having log would be helpful. I _think_ sqlite _probably_ by default has similar logging enabled for things such as failed roll back. But I'm just not sure.

internal/backends/sqlite/metadata/pool/pool.go
chilagrow: Thanks, added suggestion for simplifying tests :hugs:

internal/backends/sqlite/metadata/registry.go
chilagrow: Great, very clear!

internal/backends/sqlite/metadata/registry_test.go
```diff
@@ -14,8 +14,185 @@
...
+		db, err := r.DatabaseGetOrCreate(ctx, "RollbackOnPanic")
+		require.NoError(t, err)
+
+		require.Panics(t, func() {
+			_ = inTransaction(ctx, db, func(tx *sql.Tx) error {
+				_, err = tx.ExecContext(ctx, "CREATE TABLE test (foo TEXT)")
+				require.NoError(t, err)
+
+				var rows *sql.Rows
+				rows, err = tx.QueryContext(ctx, "SELECT name FROM sqlite_schema WHERE type='table'")
```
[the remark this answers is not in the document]
chilagrow: I think you misunderstood what I wrote. See `WHERE name='test'` part, then we only have to check `require.True(t,rows.Next())` once right?
```suggestion
				rows, err = tx.QueryContext(ctx, "SELECT name FROM sqlite_schema WHERE type='test'")
```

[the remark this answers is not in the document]
noisersup: Ah yes, you're right. Seems much more simple :)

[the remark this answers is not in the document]
chilagrow: And second time like you do are doing for making sure that it returns false :hugs:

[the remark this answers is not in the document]
chilagrow: Ops I had typo above, it's `name='test'` not `type='test'` :see_no_evil:

AlekSi:
On hold for now

AlekSi:
#3146

internal/backends/sqlite/metadata/pool/pool.go
AlekSi: (FWIW: panic in `f` would leak transaction)

PR #3002, opened 2023-07-06
Remove Tigris from tests

### LINKED ISSUE #2253: Remove `documentValidationFailureCode` from integration test.
### What should be done?

`documentValidationFailureCode` is used to skip tigris test in compat test. We want to remove this. But this might not be simple.

For example, an operator that update the value like `bson.D{{"$set", bson.D{{"v", int32(42)}}}}` would pass on Int32s provider on tigris but fail on all other providers, because tigris schema won't be the right one for other ones.
We need to think of a setup to only run tigris test on specific provider. It kind of feels like we may need to set tigris schema per test case.

### Where?

[`documentValidationFailureCode`](https://github.com/FerretDB/FerretDB/blob/5705db7778f875c0a33b17203816672602e9b1ef/integration/helpers.go#L38)

### Definition of Done

- spot refactorings done;

AlekSi:
# Description

That's all Folks!

Closes #2253.

## Readiness checklist


rumyantseva (approved):
:shipit:

PR #3009, opened 2023-07-07
Allow `$set` operator to update `_id`

### LINKED ISSUE #2803: Error updating a document passing its same id with the $set operator
We probably need to move that validation closer to the actual update.

The original issue is below.

---

### FerretDB version

1.2.1

### Backend

PostgreSQL

### Environment

- OS: Mac OS
- Deployment: Docker using ghcr.io/ferretdb/all-in-one:1.2.1 image
- Deployment details: Docker desktop Version 4.20.0 (109717)

### What did you do?

With MongoDB it is possibile to update a document by passing the id, but this gives an error with FerretDB.

For example, here is a mongo shell session against MongoDB 6

```shell
foo [direct: primary] test> db.coll.insertOne({_id:1})
foo [direct: primary] test> db.coll.updateOne({_id:1},{$set:{_id:1,a:1}})
{
  acknowledged: true,
  insertedId: null,
  matchedCount: 1,
  modifiedCount: 1,
  upsertedCount: 0
}

foo [direct: primary] test> db.coll.find()
[ { _id: 1, a: 1 } ]
foo [direct: primary] test> db.coll.updateOne({_id:1},{$set:{_id:2,a:1}})
MongoServerError: Performing an update on the path '_id' would modify the immutable field '_id'
foo [direct: primary] test> db.coll.updateOne({_id:1},{$set:{_id:1,a:1}})
{
  acknowledged: true,
  insertedId: null,
  matchedCount: 1,
  modifiedCount: 0,
  upsertedCount: 0
}
```

MongoDB raises an error if one tries to update the immutable _id, but it allows one to execute the script if the _id is the same.

With FerretDB the behavior is different. As you can see, FerretDB raises an error even when the _id is the same:

```shell
test> db.test.insertOne({_id:1})
test> db.test.updateOne({_id:1},{$set:{_id:1,a:1}})
MongoServerError: Performing an update on the path '_id' would modify the immutable field '_id'
```

We expect no error when updating a document using a command like `db.coll.updateOne({_id:1},{$set:{_id:1,a:1}})` 

### What did you see instead?

The following error is raised: 

```
MongoServerError: Performing an update on the path '_id' would modify the immutable field '_id'
```

chilagrow:
# Description

Closes #2803.

Scope changed, because of https://github.com/FerretDB/FerretDB/issues/3013 we decided to remove validation.

## Readiness checklist

### DIFF
--- a/integration/findandmodify_compat_test.go
+++ b/integration/findandmodify_compat_test.go
@@ -333,9 +333,20 @@ func TestFindAndModifyCompatUpdateSet(t *testing.T) {
 		},
 		"UpdateID": {
 			command: bson.D{
-				{"update", bson.D{{"$set", bson.D{{"_id", "non-existent"}}}}},
+				{"update", bson.D{{"$set", bson.D{{"_id", "int32"}}}}},
 			},
-			resultType: emptyResult, // _id must be an immutable field
+		},
+		"UpdateExistingID": {
+			command: bson.D{
+				{"query", bson.D{{"_id", "int32"}}},
+				{"update", bson.D{{"$set", bson.D{{"_id", "int32-1"}}}}},
+			},
+		},
+		"UpdateSameID": {
+			command: bson.D{
+				{"query", bson.D{{"_id", "int32"}}},
+				{"update", bson.D{{"$set", bson.D{{"_id", "int32"}}}}},
+			},
 		},
 	}
 
@@ -643,6 +654,27 @@ func TestFindAndModifyCompatUpsertSet(t *testing.T) {
 				{"update", bson.D{{"$set", bson.D{{"_id", "double"}}}}},
 			},
 			resultType: emptyResult, // _id must be an immutable field
+		},
+		"UpsertIDNoQuery": {
+			command: bson.D{
+				{"upsert", true},
+				{"update", bson.D{{"$set", bson.D{{"_id", "int32"}, {"v", int32(2)}}}}},
+			},
+		},
+		"UpsertExistingID": {
+			command: bson.D{
+				{"query", bson.D{{"_id", "int32"}}},
+				{"upsert", true},
+				{"update", bson.D{{"$set", bson.D{{"_id", "int32-1"}, {"v", int32(2)}}}}},
+			},
+			resultType: emptyResult, // _id must be an immutable field
+		},
+		"UpsertSameID": {
+			command: bson.D{
+				{"query", bson.D{{"_id", "int32"}}},
+				{"upsert", true},
+				{"update", bson.D{{"$set", bson.D{{"_id", "int32"}, {"v", int32(2)}}}}},
+			},
 		},
 	}
 
--- a/integration/findandmodify_compat_test.go
+++ b/integration/findandmodify_compat_test.go
@@ -667,7 +667,7 @@ func TestFindAndModifyCompatUpsertSet(t *testing.T) {
 				{"upsert", true},
 				{"update", bson.D{{"$set", bson.D{{"_id", "int32-1"}, {"v", int32(2)}}}}},
 			},
-			resultType: emptyResult, // _id must be an immutable field
+			resultType: emptyResult,
 		},
 		"UpsertSameID": {
 			command: bson.D{
--- a/integration/findandmodify_compat_test.go
+++ b/integration/findandmodify_compat_test.go
@@ -331,7 +331,7 @@ func TestFindAndModifyCompatUpdateSet(t *testing.T) {
 				{"update", bson.D{{"$set", bson.D{{"v", "foo"}}}}},
 			},
 		},
-		"UpdateID": {
+		"UpdateIDNoQuery": {
 			command: bson.D{
 				{"update", bson.D{{"$set", bson.D{{"_id", "int32"}}}}},
 			},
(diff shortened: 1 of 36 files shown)


integration/findandmodify_compat_test.go
```diff
@@ -333,9 +333,20 @@ func TestFindAndModifyCompatUpdateSet(t *testing.T) {
 		},
 		"UpdateID": {
 			command: bson.D{
-				{"update", bson.D{{"$set", bson.D{{"_id", "non-existent"}}}}},
```
chilagrow, line 336: Use `int32` instead of ``non-existent` because behaviour is different if collection contains `int23` or not.

rumyantseva: Isn't it a different case?

integration/findandmodify_compat_test.go
```diff
@@ -333,9 +333,20 @@ func TestFindAndModifyCompatUpdateSet(t *testing.T) {
 		},
 		"UpdateID": {
```
rumyantseva, line 334: Should this one be called `UpsertIDNoQuery` like the one below?

integration/findandmodify_compat_test.go
chilagrow: What do you mean by that? 

When compat test runs on `Scalars` or `Int32` collection, it contains _id `int32` and this query does not have filter, so it does not return error. This was discovered just running compat test.
See also comment in the code 
> // when filter is not set and if collection contains ID to set, it does not error

When compat test runs on other collections, it doesn't contain `int32` so behaviour is the same as `non-existent` id and returns error.

internal/handlers/common/findandmodify.go
```diff
@@ -196,11 +196,11 @@ func PrepareDocumentForUpsert(docs []*types.Document, params *FindAndModifyParam
 // prepareDocumentForInsert creates an insert document from the parameter.
 // When inserting new document we must check that `_id` is present, so we must extract `_id`
 // from query or generate a new one.
-func prepareDocumentForInsert(params *FindAndModifyParams) (*types.Document, error) {
+func prepareDocumentForInsert(docs []*types.Document, params *FindAndModifyParams) (*types.Document, error) {
```
rumyantseva, line 199: `prepareDocumentForInsert` used to be a function that "made" a document based on incoming params. Now it also receieves a slice of documents that are not really used, but just passed to some other functions... Is there a way we can redesign the code to avoid this param?

integration/findandmodify_compat_test.go
rumyantseva: Ok, got it! Then why for `SetImmutableID` below there was `non-existent` used and not `int32`?

chilagrow: Yes we can use `int32` or `non-existent`. FYI that test uses `ArrayDocuments` provider and the purpose of that test is check error message, so it's more obvious to use `non-existent` to ensuring test returns expected error message. For example, if that test changes provider from `ArrayDocuments` to `Int32` it will fail but using `non-existent` will pass test for any provider.

For compat test, we want to use input that discovers as much cases as possible, hence using int32.

internal/handlers/common/update.go
```diff
@@ -37,7 +37,8 @@ import (
 // To validate update document, must call ValidateUpdateOperators before calling UpdateDocument.
 // UpdateDocument returns CommandError for findAndModify case-insensitive command name,
 // WriteError for other commands.
-func UpdateDocument(command string, doc, update *types.Document) (bool, error) {
+// TODO https://github.com/FerretDB/FerretDB/issues/3013
```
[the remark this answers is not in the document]
chilagrow: Yes fully aware, I will handle reorganising this code now. If it gets hard to review I'll do that in separate PR :+1:

internal/handlers/common/findandmodify.go
chilagrow: Actually, most codes with prepare naming here are actually no prepare and they need documents from the DB. See `PrepareDocumentForUpsert` for example. I think it was always updating the document instead of preparing. I'll think on this

integration/findandmodify_compat_test.go
chilagrow: Wise idea :+1:

internal/handlers/common/update.go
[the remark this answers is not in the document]
rumyantseva: @chilagrow let's discuss it at the daily sync

internal/handlers/common/update.go
```diff
@@ -37,7 +37,8 @@ import (
...
+func UpdateDocument(command string, doc, update, filter *types.Document, ids []any) (bool, error) {
```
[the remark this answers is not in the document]
chilagrow: Scoped changed and we are just not gonna validate immutable _id for now.

rumyantseva (approved):
Based on the number of things that need to be done for refactoring, I think the current simple solution is the best choice to solve the exact blocker quickly!

AlekSi (approved):
:shipit:

ujibang:
should be reopened, incomplete fix see https://github.com/FerretDB/FerretDB/issues/2803#issuecomment-1658114246

PR #3031, opened 2023-07-11
Fix error checking in backend contracts

AlekSi:
# Description

`defer` was used incorrectly.

## Readiness checklist

### DIFF
--- a/internal/backends/backend.go
+++ b/internal/backends/backend.go
@@ -90,12 +90,13 @@ type DatabaseInfo struct {
 }
 
 // ListDatabases returns a Database instance for given parameters.
-func (bc *backendContract) ListDatabases(ctx context.Context, params *ListDatabasesParams) (res *ListDatabasesResult, err error) {
+func (bc *backendContract) ListDatabases(ctx context.Context, params *ListDatabasesParams) (*ListDatabasesResult, error) {
 	defer observability.FuncCall(ctx)()
-	defer checkError(err)
-	res, err = bc.b.ListDatabases(ctx, params)
 
-	return
+	res, err := bc.b.ListDatabases(ctx, params)
+	checkError(err)
+
+	return res, err
 }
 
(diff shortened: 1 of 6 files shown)


noisersup (approved):

internal/backends/backend.go
```diff
@@ -90,12 +90,13 @@ type DatabaseInfo struct {
...
 // ListDatabases returns a Database instance for given parameters.
-func (bc *backendContract) ListDatabases(ctx context.Context, params *ListDatabasesParams) (res *ListDatabasesResult, err error) {
+func (bc *backendContract) ListDatabases(ctx context.Context, params *ListDatabasesParams) (*ListDatabasesResult, error) {
 	defer observability.FuncCall(ctx)()
-	defer checkError(err)
-	res, err = bc.b.ListDatabases(ctx, params)
 
-	return
+	res, err := bc.b.ListDatabases(ctx, params)
+	checkError(err)
```
chilagrow, line 97: When we have multiple functions that returns error, I think it's nice to be able to make use of defer like this.
```
var err error
defer func(err error){ checkError(err)}(err)

var res any
if res, err = A(); err !=nil{
    return nil,err
}

if res, err = B(); err !=nil{
    return nil,err
}
```

Alternatively we could check error on each function, but I think it looks a bit busy.
```
res, err := A()
checkError(err)
if err !=nil{ 
    return nil,err
}

res, err = B()
checkError(err)
if err !=nil{
    return nil,err
}
```
What's our preference? :hugs:

chilagrow (approved):

internal/backends/backend.go
AlekSi: I don't know, that depends on the context. In this case, that situation is not possible – we wrap exactly one method call with an error check

chilagrow: Okie, for our simple case what we have is good!

chilagrow (approved):

PR #3087, opened 2023-07-20
Use one implementation for finding path values

### LINKED ISSUE #2348: Merge `aggregation.getPathValue()` and `common.getDocumentsAtSuffix()`
### What should be done?

We introduced `getPathValue` handle array with index dot notation differently from `getDocumentsAtSuffix`. We should standardise them into one function with options

**types.GetByPath(...)**
We initially had one way to find by path with dot notation, it can only return one value because array is not iterated.
https://github.com/FerretDB/FerretDB/blob/9d75840bea854c84b766c9a8de0266439a394934/internal/types/document.go#L313
We introduced two more functions to iterates on array. They are different from `GetByPath(...)` because they can return multiple values with exception on `getPathValue(...)` with `IgnoreArrays` flag which doesn't look into array.

**types.getPathValue(...)**

https://github.com/FerretDB/FerretDB/blob/9536babb87ab2873de3383c684b837d30db49a75/internal/handlers/common/aggregations/expression.go#L199
used by `$group` stage, `$sum` accumulator and `$type` operator
`IgnoreArrays` flag used by `$unwind` stage. It disable array notations like `foo.bar` to access array items like `foo: [{bar:42}]`.

**types.getDocumentsAtSuffix(...)**
https://github.com/FerretDB/FerretDB/blob/c6eccb44f5838bf98f485ef2e5fb4e793c26dbca/internal/handlers/common/filter.go#L194
used by `distinct` and filtering documents.

**Differences**

|docs|path|`GetByPath(`|`getPathValue(`|`IgnoreArrays` |`getDocumentsAtSuffix(`|
|--|--|--|--|--|--|
|`{}`|`foo.bar`|`Err`|`[]`|`[]`|`[]`|
|`{foo: {bar: 1}}`|`foo.bar`|`1`|`[1]`|`[1]`|`[{bar: 1}]`|
|`{}`|`foo.0.bar`|`Err`|`[]`|`[]`|`[]`|
|`{foo: [{bar: 1}]}`|`foo.0.bar`|`1`|`[]`|`[]`|`[{bar: 1}]`|
|`{foo: [{bar: 1}, {bar: 2}]}`|`foo.bar`|`Err`|`[1,2]`|`[]`|`[{bar: 1}, {bar: 2}]`|

### Where?

https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/common/aggregations/expression.go
https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/common/filter.go
### Definition of Done

- spot refactorings done;
- add unit tests for the new merged function, so we can reduce documentation and cover those in unit tests
- Merge `getPathValue` and `getDocumentsAtSuffix` and move it to somewhere accessible, in a separate package
- New function returns minimal response, something like `(suffix string, []any)` with `"bar", [1, 2]` for the last example
- Make it configurable for using *array notation* (i.e. `foo.bar`) and *array index notation* (i.e. `foo.0.bar`)
- Make new function work in `filter.go`, will require refactoring
- Make new function work in `expression.go`
- Don't touch `types.GetByPath(...)`, it will be too large for this issue. But welcome to make new function usable for it

chilagrow:
# Description

Closes #2348.

I created a new package `commonpath`.
## Readiness checklist


AlekSi:
(please review, but don't merge yet)

internal/handlers/commonpath/path_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	emptyDoc := new(types.Document)
+	docDoc := must.NotFail(types.NewDocument("foo", must.NotFail(types.NewDocument("bar", 1))))
+	docArrayOne := must.NotFail(types.NewDocument("foo", must.NotFail(types.NewArray(
+		must.NotFail(types.NewDocument("bar", 1)),
+	))))
+	docArrayTwo := must.NotFail(types.NewDocument("foo", must.NotFail(types.NewArray(
+		must.NotFail(types.NewDocument("bar", 1)),
+		must.NotFail(types.NewDocument("bar", 2)),
+	))))
```
rumyantseva, line 37: So, we only consider cases when the array contains a document?

For example, a case where `doc` is `{"foo": [1, 2, 3]}` and `path` is `foo.1` is not included (and as far as I can see will return an empty result). Should we highlight that `FindValues` only finds array's elements that are documents but not scalars?

rumyantseva (commented):
Based on my past experience with path-related code, the implementation looks clear to me. Asked a couple of questions where I'm not sure.

internal/handlers/commonpath/path_test.go
rumyantseva: It looks to me that the previous code was able to work with scalars, see this line: https://github.com/FerretDB/FerretDB/pull/3087/files#diff-3925922352e709447e9f2cf1c69b321f762441108abe68c4ffacded239bc8c40L315. But maybe I misunderstand something.

internal/handlers/common/distinct.go
```diff
@@ -102,26 +102,18 @@ func FilterDistinctValues(iter types.DocumentsIterator, key string) (*types.Arra
...
+			return nil, lazyerrors.Error(err)
 		}
 
-		for _, doc := range docsAtSuffix {
-			val, err := doc.Get(suffix)
-			if err != nil {
-				continue
-			}
+		// vals contains all values exist at the suffix of the path
+		vals, err := commonpath.FindValues(doc, path, commonpath.FindValuesOpts{})
```
noisersup, line 111: Small suggestion: I'd suggest to allow using nil for opts that will set default values (at least that's what I think we do in most of options in the code)
```suggestion
		vals, err := commonpath.FindValues(doc, path, nil)
```

noisersup (approved):
Overall it seems that it handles existing logic and it's great that we merge all of those functions into the single one with options

internal/handlers/commonpath/path.go
```diff
@@ -0,0 +1,109 @@
...
+	IgnoreArrayElement bool
+}
+
+// FindValues goes through each key of the path iteratively on doc to find values
+// at the suffix of the path. At each key of the path, it checks:
+//   - if the document has the key;
+//   - if the array contains an element at an index where index equals to the key, and;
+//   - if the array contains one or more documents which have the key.
+//
+// It returns a slice of values and an empty array is returned if no value was found.
```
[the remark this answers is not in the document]
rumyantseva: @chilagrow was this change pushed? :)

internal/handlers/commonpath/path_test.go
```diff
@@ -0,0 +1,107 @@
...
+			path: types.NewStaticPath("foo", "0", "bar"),
+			opts: &FindValuesOpts{FindArrayIndex: false},
+			res:  []any{},
+		},
+		"ArrayIndexScalar": {
+			doc:  arrayScalarThree,
+			path: types.NewStaticPath("foo", "1"),
+			opts: &FindValuesOpts{FindArrayIndex: true},
+			res:  []any{1},
+		},
```
rumyantseva, line 78: Let's also add the one with `FindArrayIndex: false`?

```
 {
			doc:  arrayScalarThree,
			path: types.NewStaticPath("foo"),
			opts: &FindValuesOpts{FindArrayIndex: false},
			res:  []any{0, 1, 2},
		},
```

internal/handlers/commonpath/path.go
```diff
@@ -0,0 +1,117 @@
...
+// FindValuesOpts sets options for FindValues.
+type FindValuesOpts struct {
+	// FindArrayIndex gets an element at the specified index of an array.
+	// Using path `v.0` and `v` is an array, it returns 0-th index element of the array.
+	// If `v` is not an array, FindArrayIndex has no impact.
+	FindArrayIndex bool
+	// SearchArray searches all document of an array to find documents that contains path key.
+	// Using path `v.foo` and `v` is an array, it returns all document which has key `foo`.
+	// If `v` is not an array, SearchArray has no impact.
+	SearchArray bool
```
rumyantseva, line 35: Should this one be called `SearchInArray` or `IterateThroughArray`? (But for this second option I actually liked the previous one with `Ignore` because we were really ignoring things in this case)

In this case, we don't aim to search for an array, we rather iterate through it.

And the doc is something like "If `SearchInArray` is set, the search will iterate through the whole array to find matches in any of its elements".

internal/handlers/commonpath/path.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/FerretDB/FerretDB/internal/util/iterator"
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+)
+
+// FindValuesOpts sets options for FindValues.
+type FindValuesOpts struct {
+	// FindArrayIndex gets an element at the specified index of an array.
+	// Using path `v.0` and `v` is an array, it returns 0-th index element of the array.
+	// If `v` is not an array, FindArrayIndex has no impact.
+	FindArrayIndex bool
```
rumyantseva, line 31: Maybe we can call it `StrictArrayIndexes` and the documentation is like "if `StrictArrayIndexes` is set, search in array will expect the exact index to be set in the given path."

But it's just a suggestion, feel free to ignore if you like your field name more.

internal/handlers/commonpath/path.go
```diff
@@ -0,0 +1,117 @@
...
+							continue
+						}
+
+						values = append(values, v)
+
+						continue
+					}
+				}
+
+				if !opts.SearchArray {
```
rumyantseva, line 81: Hmm, if `FindArrayIndex` is set as true and  `SearchArray` is set as true, the result is not really clear to me as we will search by index first, and if we didn't find the index, we will also iterate through the whole array, right? Is there a use case for it, or should this combination be prohibited?

rumyantseva (commented):
Looks good to me overall, asked a couple of more questions. 
In principle, I don't mind merging it as is. We have everything covered with integration tests, and we can always improve the documentation or naming later if we have a clear idea of how things should be called.

internal/handlers/commonpath/path.go
chilagrow: This combination should be allowed. It's used by filter as far as I know. Also it might be used by distinct too.

internal/handlers/commonpath/path.go
chilagrow: Renamed this to `SearchInArray`. For ignoring other one, I'm not totally sure to have one option that enables array search and another option disables index search. :thinking:

internal/handlers/commonpath/path_test.go
chilagrow: I added `Array` test case and `ArrayIndexScalarFindArrayIndexFalse` possibly one of them covers it? Did you want to search path `foo` or set `FindArrayIndex` false :hugs:

internal/handlers/commonpath/path.go
chilagrow: I kept it as it is, only one element can exist at an index so bit odd to call it `StrictArrayIndexes`. Also if `SearchInArray` option is set, it iterates the array which might be counter intuitive to use the wording `strict` :thinking:

internal/handlers/commonpath/path_test.go
rumyantseva: Well, I just wanted to see more scalars, but as now we call the code that works with the path from many places, this unit test is not needed, integration will cover it, so all god :)

rumyantseva (dismissed):
LGTM. I don't think it is worth discussing naming things for now (we already discuss some), and the code itself solves the necessary things!

noisersup (dismissed):

internal/handlers/commonpath/commonpath.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+// FindValuesOpts sets options for FindValues.
+type FindValuesOpts struct {
+	// SearchInArray searches by iterating through the whole array to find documents that contains path key.
+	// Using path `v.foo` and `v` is an array, it returns all document which has key `foo`.
+	// If `v` is not an array, SearchInArray has no impact.
+	SearchInArray bool
+	// FindArrayIndex gets an element at the specified index of an array.
+	// Using path `v.0` and `v` is an array, it returns 0-th index element of the array.
```
AlekSi, line 34: … but isn't that normal behavior of types.Path / dot notation? I think that comment should explain how it behaves when this option is off.

internal/handlers/commonpath/commonpath.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+)
+
+// FindValuesOpts sets options for FindValues.
+type FindValuesOpts struct {
+	// SearchInArray searches by iterating through the whole array to find documents that contains path key.
+	// Using path `v.foo` and `v` is an array, it returns all document which has key `foo`.
+	// If `v` is not an array, SearchInArray has no impact.
+	SearchInArray bool
+	// FindArrayIndex gets an element at the specified index of an array.
```
AlekSi, line 33: Option does not get anything

internal/handlers/commonpath/commonpath.go
```diff
@@ -0,0 +1,118 @@
...
+	"strconv"
+
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/iterator"
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+)
+
+// FindValuesOpts sets options for FindValues.
+type FindValuesOpts struct {
+	// SearchInArray searches by iterating through the whole array to find documents that contains path key.
```
AlekSi, line 29: Option does not search anything

internal/handlers/commonpath/commonpath.go
```diff
@@ -0,0 +1,118 @@
...
+	// Using path `v.0` and `v` is an array, it returns 0-th index element of the array.
+	// If `v` is not an array, FindArrayIndex has no impact.
+	FindArrayIndex bool
+}
+
+// FindValues goes through each key of the path iteratively on doc to find values
+// at the suffix of the path. At each key of the path, it checks:
+//   - if the document has the key;
+//   - if the array contains an element at an index where index equals to the key, and;
+//   - if the array contains one or more documents which have the key.
```
AlekSi, line 43: I don't think I understand those two lines completely, but they are more clean the option names and descriptions above. It is hard to understand what options match what line there.

internal/handlers/commonpath/commonpath.go
```diff
@@ -0,0 +1,118 @@
...
+// FindValuesOpts sets options for FindValues.
+type FindValuesOpts struct {
+	// SearchInArray searches by iterating through the whole array to find documents that contains path key.
+	// Using path `v.foo` and `v` is an array, it returns all document which has key `foo`.
+	// If `v` is not an array, SearchInArray has no impact.
+	SearchInArray bool
+	// FindArrayIndex gets an element at the specified index of an array.
+	// Using path `v.0` and `v` is an array, it returns 0-th index element of the array.
+	// If `v` is not an array, FindArrayIndex has no impact.
+	FindArrayIndex bool
```
AlekSi, line 36: What is the difference between "find" and "search"? Why names are so different?

internal/handlers/commonpath/commonpath.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// FindValues goes through each key of the path iteratively on doc to find values
+// at the suffix of the path. At each key of the path, it checks:
+//   - if the document has the key;
+//   - if the array contains an element at an index where index equals to the key, and;
+//   - if the array contains one or more documents which have the key.
+//
+// It returns a slice of values and an empty array is returned if no value was found.
```
AlekSi, line 45: I think it is extremely hard to understand what this function does due to confusing comments and inconsistent terminology. Let's read it together:

> FindValues goes through each key of the path iteratively on doc

It is not clear if we iterate over path elements (that could contain array indexes, not only document keys) or over doc fields.

> to find values at the suffix of the path.

?.. If we iterate over path elements, then what is the path suffix if we already iterate over path elements? If we iterate over doc fields, then why do we use the path's [suffix](https://github.com/FerretDB/FerretDB/blob/19e1cdbfde91cafad281d7b18e02c9dbf3a8b233/internal/types/path.go#L134) instead of just passing the single element?

> At each key of the path

Ok, at each element of the path. So we iterate over path elements after all.

> it checks if the document has the key;

But what if that's not a document or not a key?

> if the array contains an element at an index where index equals to the key

Ok, so "element at an index" is a value at the given index, I guess. Why do we compare the value with the key?..

> if the array contains one or more documents which have the key.

Ok, so we got a boolean value, true or false. What do we do with it?

> It returns a slice of values and an empty array is returned if no value was found.

I'm completely lost.

---

I understand that the whole area is complicated, but that's a reason to make it approachable. If we keep it as is, very soon, that will be magic no one would understand and touch. Personally, after spending days (and hours of wall clock time) on it, my head hurts – and I still don't get it. (If anything, I get less after today 😬)

I think we could try to simplify it by:

1. Using consistent terminology. I think it is already defined by the type package; we could also improve it if needed.
2. Splitting that function. Right now, the cyclomatic complexity is very high. If we split it into multiple functions with good names and descriptions, someone could get a better idea of what the whole thing does by reading the code.
3. Writing better tests (see below). Failing to understand description and code, someone at least could read the test.

internal/handlers/commonpath/commonpath_test.go
```diff
@@ -0,0 +1,119 @@
...
+	empty := new(types.Document)
+	doc := must.NotFail(types.NewDocument("foo", must.NotFail(types.NewDocument("bar", 1))))
+	arrayDocOne := must.NotFail(types.NewDocument("foo", must.NotFail(types.NewArray(
+		must.NotFail(types.NewDocument("bar", 1)),
+	))))
+	arrayDocTwo := must.NotFail(types.NewDocument("foo", must.NotFail(types.NewArray(
+		must.NotFail(types.NewDocument("bar", 1)),
+		must.NotFail(types.NewDocument("bar", 2)),
+	))))
+	arrayScalarThree := must.NotFail(types.NewDocument("foo", must.NotFail(types.NewArray(0, 1, 2))))
```
AlekSi, line 38: I think the test will be significantly easier to understand if there is only one document. Right now, if someone (read: me) could not understand the description or code and try to understand it all from the test, they (again, that's me 😬) have to go up and down to link tests cases (paths and expected results) that are right there with documents that are way up.

I also think there could be a better way to group test cases after that.

And there are no test cases with both options enabled.

internal/handlers/commonpath/commonpath.go
chilagrow: Let me rework on this. Reading your comment, I think I'm also getting lost which is really bad sign. Splitting to functions would improve it too.

internal/handlers/commonpath/commonpath.go
chilagrow: True, let me standadise those :hugs:

internal/handlers/commonpath/commonpath_test.go
chilagrow: Good point, let me use one document, and also keep those options set more consistently to improve this.

internal/handlers/commonpath/commonpath.go
chilagrow: I added its behaviour on option's off. See what you think.

internal/handlers/commonpath/commonpath.go
chilagrow: Thanks, updated comment.

internal/handlers/commonpath/commonpath.go
chilagrow: Updated this, feel free to suggest better wording if you can think of any as well :pray:

internal/handlers/commonpath/commonpath.go
chilagrow: I added which option is used, hopefully it's better :sweat_smile:

chilagrow:
Integration tests are failing at the moment, this PR introduces stricter validation of `NewPathFromString` and functions that use that are not behaving. I'm fixing them...

internal/handlers/commonpath/commonpath_test.go
chilagrow: I updated tests to use one document for array to make it a bit more clear.

internal/handlers/commonpath/commonpath.go
chilagrow: Splitted them to smaller functions and updated doc string to follow closer to path package. Let me know if this improves understanding. or how we can make it better :pray:

internal/handlers/commonpath/commonpath_test.go
```diff
@@ -0,0 +1,222 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/must"
+)
+
+func TestGetPathValue(t *testing.T) {
```
noisersup, line 26:
```suggestion
func TestFindValues(t *testing.T) {
```

noisersup (commented):
Changes look reasonable to me

rumyantseva (dismissed):
I like the new naming and comments! The implementation looks good to me too.

internal/handlers/commonpath/path.go
chilagrow: @rumyantseva, it went though a few iteration of renaming and re-writing. Could you resolve this comment?

noisersup (dismissed):
LGTM

internal/handlers/common/aggregations/expression.go
```diff
@@ -69,23 +70,20 @@ func (e *ExpressionError) Code() ExpressionErrorCode {
 
 // Expression is an expression constructed from field value.
```
AlekSi, line 71: > Expression is an expression

We can't define the term using the term itself.

Compare and contrast with `types.Path` description

internal/handlers/common/aggregations/expression.go
```diff
@@ -69,23 +70,20 @@ func (e *ExpressionError) Code() ExpressionErrorCode {
...
 
-// ExpressionOpts represents options used to modify behavior of Expression functions.
-type ExpressionOpts struct {
-	// TODO https://github.com/FerretDB/FerretDB/issues/2348
-
-	// IgnoreArrays disables checking arrays for provided key.
-	// So expression {"$v.foo"} won't match {"v":[{"foo":42}]}
-	IgnoreArrays bool // defaults to false
-}
+// NewExpression creates a new instance by checking expression string.
```
AlekSi, line 77: What is an expression string?

internal/handlers/common/aggregations/expression.go
```diff
@@ -121,19 +119,13 @@ func NewExpressionWithOpts(expression string, opts *ExpressionOpts) (*Expression
...
 	}, nil
 }
 
-// NewExpression creates a new instance by checking expression string.
-func NewExpression(expression string) (*Expression, error) {
-	// TODO https://github.com/FerretDB/FerretDB/issues/2348
-	return NewExpressionWithOpts(expression, new(ExpressionOpts))
-}
-
 // Evaluate gets the value at the path.
```
AlekSi, line 127: What path?

AlekSi (dismissed):

rumyantseva (approved):
In addition to the changes, I also like the documentation and examples (e.g. `$unwind`). It brings more clarity about how things work and why!

AlekSi (approved):

PR #3102, opened 2023-07-24
Reorganise aggregation by deciding how initial document is fetched

### LINKED ISSUE #2423: Move `$collStats` logic to its stage implementation
### What should be done?

`$collStats` created the whole parallel universe for itself not only in the Stage interface but also in the `aggregate` command implementation. Could we fold that special case in the general design?

At the moment, the complexity is around `$collStats` does not always need to query database, and if does it uses its specific query to fetch data. There are other stages that does not query database such as [$documents](https://github.com/FerretDB/FerretDB/issues/1419) and [$listSessions](https://github.com/FerretDB/FerretDB/issues/1426).

The proposed design separates `Stages` to `ProducerStage` and `ProcessorStage` (see https://github.com/FerretDB/FerretDB/pull/3102), each `ProducerStage` may decide how to produce document iterator.

### Where?
https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/pg/msg_aggregate.go
Also find issue link in the code base
### Definition of Done

- [x] remove `Type()` interface
- [ ] remove `$collStats` specific code from [msg_aggregate.go](https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/pg/msg_aggregate.go
), it should be in [collStats.go](https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/common/aggregations/stages/collstats.go).
- [ ] calling database is optional by design (see proposed design https://github.com/FerretDB/FerretDB/pull/3102)
- [ ] integration/compatibility tests added/updated;
- [ ] spot refactorings done;

chilagrow:
# Description

Closes #2423.

This PR demonstrates intended design for aggregation pipeline to handle special stage like `$collStats`.

`$collStats` created a bit of special case in aggregation. It's because it wants to decide if it wants to query DB or not, and if it queries DB, it uses query specific to its need.

In this PR, I'm intending to remove special handling for `$collStats` by allowing stage to decide how it's initial document iterator should be fetched. This would be useful for stages that haven't been added such as `$documents` stage which does not query DB at all.

**Do not merge, as this PR is not intended to be merged. It demonstrates the design and that's the purpose. Some tests failing, that's okay.**
## Readiness checklist


internal/handlers/common/aggregations/aggregations.go
```diff
@@ -24,6 +24,31 @@ import (
 
 // Stage is a common interface for all aggregation stages.
 type Stage interface {
+	// FetchDocuments fetches document iterator, it is only used if this is
+	// the first stage of the pipeline.
+	// The returned document iterator may be originated from querying database or
+	// from in-memory value iterator.
+	// This allows first stage of the pipeline to decide how to create initial document iterator.
+	FetchDocuments(ctx context.Context, closer *iterator.MultiCloser) (types.DocumentsIterator, error)
```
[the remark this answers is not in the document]
rumyantseva: Or maybe it's part of a "pipeline" interface. E.g. we "seed" the pipeline with some documents first, and then apply stages on it. And then it's pipeline responsibility/implementation  how to fetch these seeded documents from the DB.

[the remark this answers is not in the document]
rumyantseva: After some thinking... Actually, the current approach looks good to me too. I was confused by the phrase "it is only used if this is the first stage of the pipeline", but actually any stage could be the first stage, so it's correct that this method exists for each stage. 

On the other hand, the implementation of this method is exactly the same in most of the cases, so that's another sign to me that maybe this function shouldn't be here...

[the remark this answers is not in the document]
noisersup: This function seems reasonable if the way of fetching documents is different at the first stage for some stages, although I cannot define any of them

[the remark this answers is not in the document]
chilagrow: Valid point about the implementation of this `FetchDocuments` being _mostly_ the same. For upcoming pipelines, the ones that have different implementations are `$documents`, `$listLocalSessions`, `$listSessions` and existing `$collStats`.

It cannot be part of pipeline because it would expose specific details about each stage to the pipeline. Let me try a few other things if it's possible to make it not an interface. Yes thanks we should rename this to something better :hugs:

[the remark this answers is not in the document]
rumyantseva: @chilagrow yeah, I agree with your point. My thought was to somehow "identify" the type of pipeline based on the first stage and then fetch documents based on it, but it doesn't seem better (because then we need to come up with a "pipeline" type which will depend on the first stage, so the implementation doesn't look clear and simple). 

So, it looks to me that the current function "How documents are fetched if this stage is the first in the pipeline" is a better solution.

internal/handlers/common/aggregations/aggregations.go
```diff
@@ -24,6 +24,31 @@ import (
...
+
 	// Process applies an aggregate stage on documents from iterator.
 	Process(ctx context.Context, iter types.DocumentsIterator, closer *iterator.MultiCloser) (types.DocumentsIterator, error)
 }
+
+// AggregateQuery is a common interface for fetching from database.
+type AggregateQuery interface {
```
rumyantseva, line 39: So, this interface will grow when we have `listSessions` and other stages that need different types of documents. I think it's fine. This way, we have a unified interface to work with different backends. 

The approach looks similar to the `Backend`, `Collection`, and other interfaces that we have in the new handler.
Maybe we can have a similar naming too? And then this could be just `Aggregate`/`Aggregation`? But I think the naming it's not so critical.

internal/handlers/common/aggregations/aggregations.go
```diff
@@ -24,6 +24,31 @@ import (
...
+	// QueryDocuments fetches documents from the database.
+	QueryDocuments(ctx context.Context, closer *iterator.MultiCloser) (types.DocumentsIterator, error)
+
+	// QueryCollStats fetches collection statistics from the database.
+	QueryCollStats(ctx context.Context, closer *iterator.MultiCloser) (*CollStats, error)
+}
+
+// CollStats describes collection statistics retrieved from the database.
+type CollStats struct {
```
rumyantseva, line 48: I think, if we want a similar naming to how it's called in handler, this should be called `CollStatsResult`.

rumyantseva (dismissed):
The approach looks good to me, and I don't see a better interface/approach/implementation, so I'm ready to approve.

Maybe the only question is how similar to `internal/backends` this new interface should be. If we want more similarity, maybe some naming/structuring needs to be changed. At some point, it looks reasonable to make them similar, because the idea is quite the same - we have a common implementation of aggregation pipelines and all the backends should implement a particular interface to support it.

noisersup (approved):
This approach makes sense to me!

internal/handlers/common/aggregations/aggregations.go
```diff
@@ -24,6 +24,31 @@ import (
 
 // Stage is a common interface for all aggregation stages.
 type Stage interface {
+	// FirstStage fetches document iterator, it is only used if this is
+	// the first stage of the pipeline.
+	// The returned document iterator may be originated from querying database or
+	// from in-memory value iterator.
+	// This allows first stage of the pipeline to decide how to create initial document iterator.
+	FirstStage(ctx context.Context, closer *iterator.MultiCloser) (types.DocumentsIterator, error)
```
AlekSi, line 32: I don't think we need that method at all. First of all, it looks suspiciously like a constructor; it is almost as if the `io.Reader` interface had `NewReader` method. Second, the implication there is that `Stage` has access to the data source, but then it is not clear why `Process` can't use it.

One option could be to fit everything into the existing `Stage` interface with only the `Process` method. `$collStats` implementation, for example, could ignore incoming `iter` and use the data source instead. Then the code that arranges the pipeline (a sequence of stages in this case) could check that the order is correct.

Another option is to say that pipelines are not just a sequence of stages but a combination of zero or one Producer stage and zero or more Processor stages. That way, `$collStats` would implement only the `Producer` interface, `$count` would implement only the Processor interface, and `$facet` would implement both. Then the compiler itself could check the order.

internal/handlers/common/aggregations/aggregations.go
```diff
@@ -24,6 +24,31 @@ import (
...
+
 	// Process applies an aggregate stage on documents from iterator.
 	Process(ctx context.Context, iter types.DocumentsIterator, closer *iterator.MultiCloser) (types.DocumentsIterator, error)
 }
+
+// Aggregation is a common interface for fetching from database.
```
AlekSi, line 38: Independently of everything else, that's a bad name given this aggregation does not aggregate anything – it returns original data to be processed and aggregated, and that's pretty much the opposite.

internal/handlers/common/aggregations/aggregations.go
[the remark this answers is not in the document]
chilagrow: I tried your suggestion in another PR https://github.com/FerretDB/FerretDB/pull/3110
This approach exposes logic related to `$collStats` to aggregation package level. But it removes first stage logic from the stages. Let me know what you think :hugs:

internal/handlers/common/aggregations/aggregations.go
chilagrow: Both are definitely better options. First option being simpler. The second option seems efficient :+1:

chilagrow: I think using `ProducerStage` and `ProcessorStage` works quite well, thanks for the suggestion. Updated implementation a bit :hugs:

internal/handlers/common/aggregations/aggregations.go
```diff
@@ -22,8 +22,29 @@ import (
...
-type Stage interface {
+// ProcessorStage is a common interface for aggregation stages that process document iterator.
+type ProcessorStage interface {
 	// Process applies an aggregate stage on documents from iterator.
 	Process(ctx context.Context, iter types.DocumentsIterator, closer *iterator.MultiCloser) (types.DocumentsIterator, error)
 }
+
+// ProducerStage is a common interface aggregation stages produce documentiterator.
+type ProducerStage interface {
+	// Produce applies an aggregate stage on documents from iterator.
```
AlekSi, line 33: That's not correct

internal/handlers/common/aggregations/aggregations.go
```diff
@@ -22,8 +22,29 @@ import (
...
+	Produce(ctx context.Context, closer *iterator.MultiCloser) (types.DocumentsIterator, error)
+}
+
+// ProducerStageDataSource is a common interface for fetching from database.
+type ProducerStageDataSource interface {
+	// CollStats fetches collection statistics from the database.
+	CollStats(ctx context.Context, closer *iterator.MultiCloser) (*CollStatsResult, error)
```
AlekSi, line 40: That should be `ProducerStage` that returns an iterator with one document

internal/handlers/common/aggregations/aggregations.go
chilagrow: Indeed thanks, it should return a document iterator.

internal/handlers/common/aggregations/aggregations.go
chilagrow: So `$collStats` is a `ProducerStage` and wants to fetch data from DB and this interface allows access to DB. Maybe I misunderstood and `ProducerStage` should be implemented on pg layer? :thinking:

AlekSi: You are mixing stages and what this code called Aggregation before and calls data source now.

Data source allows one to run queries.

ProducerStage produces documents. It is not important for the interface how those documents are produced – are they fetched from the database or just made up.

There are two implementations of the produce ProducerStage. Both would use a data source. The first one would return all documents from the collection. The second one would produce collection stats. (in the future, `$match` stage could implement both `ProcessorStage` and `ProducerStage` for pushdown).

chilagrow: So hopefully I get this. We want to keep `ProducerStageDataSource` or `DataSource` to its `ProducerStage`. No need to expose it here.

And I'm missing `ProducerStage` that returns all documents :sweat_smile:

chilagrow: There is quite a bit more cleanup I can do, but overall I like it. :hugs:

internal/handlers/common/aggregations/stages/collstats.go
```diff
@@ -83,11 +115,69 @@ func newCollStats(stage *types.Document) (aggregations.Stage, error) {
 	return &cs, nil
 }
 
-// Process implements Stage interface.
+// Produce implements ProducerStage interface.
 //
-// Processing consists of modification of the input document, so it contains all the necessary fields
+// Producing consists of modification of the input document, so it contains all the necessary fields
```
AlekSi, line 120: That's not correct

internal/handlers/common/aggregations/stages/collstats.go
```diff
@@ -29,12 +31,29 @@ import (
 	"github.com/FerretDB/FerretDB/internal/util/must"
 )
 
+// CollStatsDataSource fetches collection statistics from the database.
+type CollStatsDataSource interface {
```
AlekSi, line 35: Let's call it `DataSource`. It has only `CollStats` for now, it will have `Query` or something soon and will be replaced with a backend interface later

chilagrow: Sounds good, updated this name.

internal/handlers/common/aggregations/stages/collstats.go
chilagrow: Thanks, updated.

chilagrow: Thanks, updated :pray:

internal/handlers/common/aggregations/aggregations.go
```diff
@@ -22,8 +22,14 @@ import (
...
-// Stage is a common interface for all aggregation stages.
-type Stage interface {
+// ProcessorStage is a common interface for aggregation stages that process document iterator.
+type ProcessorStage interface {
 	// Process applies an aggregate stage on documents from iterator.
 	Process(ctx context.Context, iter types.DocumentsIterator, closer *iterator.MultiCloser) (types.DocumentsIterator, error)
 }
+
+// ProducerStage is a common interface for aggregation stages that produce document iterator.
+type ProducerStage interface {
```
rumyantseva, line 32: I like this new naming for the interfaces!

internal/handlers/pg/msg_aggregate.go
```diff
@@ -243,40 +258,28 @@ func (h *Handler) MsgAggregate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
...
-		iter, err = processStagesDocuments(ctx, closer, &stagesDocumentsParams{dbPool, &qp, stagesDocuments})
 	} else {
-		// stats stages are provided - fetch stats from the DB and apply stages to them
-		// TODO move $collStatsDocuments specific logic to its stage https://github.com/FerretDB/FerretDB/issues/2423
-		statistics := stages.GetStatistics(collStatsDocuments)
-
-		iter, err = processStagesStats(ctx, closer, &stagesStatsParams{
-			dbPool, db, collection, statistics, collStatsDocuments,
-		})
+		// handle case where there are multiple producer stages
```
rumyantseva, line 273: Is it a possible situation - multiple producer stages?

rumyantseva (approved):
I like this new approach with two interfaces! We no longer duplicate any code, and the idea's easy to understand. I have some minor questions about the usage (I asked one question), but it's not blocking us, we can clarify it later.

internal/handlers/pg/msg_aggregate.go
AlekSi: No

chilagrow: For example, having two consecutive `$collStats` stage should return error.

AlekSi: That's correct, but that's not what this code does, isn't it?

I think we should return an error and exit early waaay up when we iterate over `aggregationStages`

internal/handlers/pg/msg_aggregate.go
```diff
@@ -163,10 +162,27 @@ func (h *Handler) MsgAggregate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
 	}
 
 	aggregationStages := must.NotFail(iterator.ConsumeValues(pipeline.Iterator()))
-	stagesDocuments := make([]aggregations.Stage, 0, len(aggregationStages))
-	collStatsDocuments := make([]aggregations.Stage, 0, len(aggregationStages))
+	producerStages := make([]aggregations.ProducerStage, 0, len(aggregationStages))
```
AlekSi, line 165:
```go
var producerStage aggregations.ProducerStage
```

As there could be only zero (`nil`) or one (non-nil value)

internal/handlers/pg/msg_aggregate.go
```diff
@@ -176,28 +192,27 @@ func (h *Handler) MsgAggregate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
 			)
 		}
 
-		var s aggregations.Stage
+		if _, ok := stages.ProducerStages[d.Command()]; ok {
+			var s aggregations.ProducerStage
```
AlekSi, line 196:
```go
if producerStage != nil {
  return nil, errors.New("oops, already have one")
}
```

chilagrow:
Pausing work on this PR for now

noisersup (commented):
Leaving a comment as this PR is on hold

chilagrow:
Close for now

PR #3103, opened 2023-07-24
Add wire.LoadDataDocuments function

### LINKED ISSUE #3067: A better fuzzing seed corpus collection
### What should be done?

Right now we're using `./tmp/records` that are created by the integration tests to increase the fuzzing corpus of `sjson`.

This is sub-optimal, because it seems like most of the documents are not "valid" data documents. The sjson fuzzing applies some document validation, so invalid documents are skipped.

After 35mins of integration testing, I got 3348 `.bin` entries in the `./tmp/records` folder. 3 repetitive runs of fuzzing report 0 documents being added to the corpus. I double-checked that this is due to the fact that these documents can't pass the validation.

```go
if doc.ValidateData() != nil {
	continue
}
```

We could try finding a better way to generate a corpus for sjson fuzzing purposes.

This is a separate issue from #1273, so it can be planned, analyzed, and estimated separately.

### Where?

`sjson` package + something new that provides that new corpus.

### Definition of Done

- all handlers updated;
- unit tests added/updated;
- spot refactorings done;
- something else?

quasilyte:
# Description

Unlike wire.LoadRecords, it tries to find valid data documents.

This new function is used in sjson package fuzz testing.

Closes #3067

### DIFF
--- a/internal/wire/record.go
+++ b/internal/wire/record.go
@@ -22,7 +22,9 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/FerretDB/FerretDB/internal/types"
 	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+	"github.com/FerretDB/FerretDB/internal/util/must"
 )
 
 // Record represents a single recorded wire protocol message, loaded from a .bin file.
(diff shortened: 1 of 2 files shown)


noisersup (approved):
I like the new function. although the linters need to be addressed

internal/wire/record.go
```diff
@@ -56,28 +137,68 @@ func LoadRecords(dir string, limit int) ([]Record, error) {
...
+
+	switch v := v.(type) {
+	case *types.Document:
+		if v.ValidateData() == nil {
+			dst = append(dst, v)
+			return dst, nil
 		}
 
-		res = append(res, r...)
+		if v.Has("insert") && v.Has("documents") {
```
chilagrow, line 185: This one looks like a record from insert command. Does fuzzing require or forbid having `_id` field? I'm not familiar with stored records but I suppose it may or may not have `_id` field :thinking:

internal/wire/record.go
```diff
@@ -36,8 +38,87 @@ type Record struct {
...
+	if err != nil {
+		return nil, err
+	}
+
+	var docs []*types.Document
+
+	// We can't tell whether a record have any data documents or
+	// not before parsing the bin file.
+	// Parsing all files and then using some shuffling is slow,
+	// so we're using a random window to collect the results.
```
chilagrow, line 54: Random window seems fine :+1: Given we were not getting any before, having any randomization seems positive!

chilagrow (commented):
Great work! Some questions mainly for me to be more familiar with fuzzing setup :hugs:

What's the improvement on the number of fuzzed record like?

rumyantseva (commented):
I guess this PR needs a bit further work....

PR #3114, opened 2023-07-26
Fix pushdown for `find` with `filter` and `limit`

### LINKED ISSUE #3113: Invalid results for `find` with `filter` and `limit`
### FerretDB version

#2921 / #2993 introduced a regression in v1.6.0 for the `find` command with set `filter` and `limit` parameters.

### Backend

PostgreSQL, any version

### Environment

Any

### What did you do?

Run `db.col.find(<filter>).limit(1)` or `db.col.findOne(<filter>)`. The filter should be non-empty.

The expected document or null is returned.

### What did you see instead?

The result is not determined: the expected document, some other document, or null could be returned.

AlekSi:
Closes #3113.


rumyantseva (approved):
:shipit:

PR #3137, opened 2023-07-31
Compat tests failure check proof of concept

noisersup:
# Description

That's just a draft of solution to the issue with checking for SQLite failure of compat tests. It still needs some changes (like merging it into the `expected` struct, or somehow extracting the Fail() logic between `expected` and `failChecker`, although this'll be covered as soon as we agree to specific architecture).

I'd be grateful for any opinions 

## Readiness checklist

### DIFF
--- a/integration/aggregate_documents_compat_test.go
+++ b/integration/aggregate_documents_compat_test.go
@@ -99,7 +99,6 @@ func testAggregateStagesCompatWithProviders(t *testing.T, providers shareddata.P
 			}
 
 			e := setup.NewFailCatcher(t, "ferretdb-sqlite", tc.failsForSQLite)
-			defer e.Catch()
 
 			var nonEmptyResults bool
 			for i := range targetCollections {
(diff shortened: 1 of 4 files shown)


integration/aggregate_documents_compat_test.go
```diff
@@ -39,6 +39,7 @@ type aggregateStagesCompatTestCase struct {
 	resultType     compatTestCaseResultType // defaults to nonEmptyResult
 	resultPushdown bool                     // defaults to false
 	skip           string                   // skip test for all handlers, must have issue number mentioned
+	failsForSQLite string
```
chilagrow, line 42: I expected if you leave `failsForSQLite` empty for a test that should fail, it should fail. But that doesn't seem to be the case. Could you add comment how this behaves please :hugs: 

I think if a test that passes has non-empty `failsForSQLite`, it should fail.
If a test that fails has empty `failsForSQLite`, it should fail.
Other tests should pass.

chilagrow (commented):
I can't quite get how it works :confused:

Issue #3131, opened 2023-07-31
Make all integration tests "pass" with SQLite


AlekSi:
### What should be done?

Add `setup.FailsForSQLite` with an issue URL to all tests that fail.
Create new issues as needed with `backend/sqlite` label before checking existing issues first.

If the test or test case passed unexpectedly, that might mean that the real query is never sent, and that test actually tested the driver's error handling. Change that test to test FerretDB or remove it.

### Where?

Integration tests.

Let's do one PR per integration test file, starting alphabetically.

### Definition of Done

- integration/compatibility tests added/updated;
- CI passes;
- spot refactorings done.

Issue #3148, opened 2023-08-02
Implement `aggregate` for SQLite


AlekSi:
### What should be done?

SQLite handler should support `aggregate` command.

No changes should be made to the backend code. Any pushdowns are not in the scope.

Refactoring is not in the scope.

### Where?

See references to this issue.

`$collStats` is out of scope.

### Definition of Done

- handler updated;
- integration/compatibility tests enabled and pass;
- no references to this issue are left.

PR #3150, opened 2023-08-02
Make aggregate documents compat tests pass for SQLite

noisersup:
# Description

`aggregate_documents_compat_test.go`

Closes #3131.

## Readiness checklist


rumyantseva (dismissed):
To be honest, I'm not sure if we need `failsForSQLite` param for aggregation test cases at all because aggregation is not supported, and it's enough to just fail all the aggregation tests. And also because aggregation is implemented through common, I expect it should be easy to enable it for SQLite. So, I guess as part of #3148 we will have to go through all these cases again and remove `failsForSQLite`.

However, potentially, there might be some stages that we won't enable/implement immediately, and, as this code does the job, I approve.

integration/aggregate_documents_compat_test.go
```diff
@@ -158,18 +158,17 @@ func testAggregateStagesCompatWithProviders(tt *testing.T, providers shareddata.
 				})
 			}
 
-			var t testtb.TB = tt
-			if tc.failsForSQLite != "" {
-				t = setup.FailsForSQLite(tt, tc.failsForSQLite)
+			if tc.failsForSQLite != "" && setup.IsSQLite(tt) {
```
chilagrow, line 161: Other tests checked `setup.FailsForSQLite(tt, tc.failsForSQLite)`, but this test helper skips tests marked to fail for sqlite. Why is this handled differently? :thinking: Is it related to what you said about some collection failing and others collection not failing? Or because aggregation is not supported yet for sqlite? I'm assuming all aggregation tests are currently failing though.

A consequence is that when we start implementing aggregation for sqlite, we won't know which tests start passing, because they are skipped.

chilagrow (commented):
Just one comment about how we are skipping aggregation tests

integration/aggregate_documents_compat_test.go
noisersup: This is done to handle failure correctly. The `failsForSQLite` function is used above in this function. Here, we need to return if we did the previous check, because right after this line we do a check for empty/non-empty results, which might be different if we expect test to fail.

chilagrow: I see we check `failsForSQLite` earlier. And all sqlite would have emptyResult.

chilagrow (approved):

rumyantseva (approved):
:shipit:

PR #3171, opened 2023-08-08
Do not return immutable `_id` error from `findAndModify` for upserting same `_id`

### LINKED ISSUE #2803: Error updating a document passing its same id with the $set operator

chilagrow:
# Description

Closes #2803.

## Readiness checklist

### DIFF
--- a/integration/findandmodify_compat_test.go
+++ b/integration/findandmodify_compat_test.go
@@ -588,6 +588,14 @@ func TestFindAndModifyCompatUpsert(t *testing.T) {
 				{"update", bson.D{{"_id", "int32"}, {"v", "replaced"}}},
 			},
 		},
+		"UpdateDifferentID": {
+			command: bson.D{
+				{"query", bson.D{{"_id", bson.D{{"$exists", true}}}}},
+				{"upsert", true},
+				{"update", bson.D{{"_id", "replaced"}, {"v", "replaced"}}},
+			},
+			resultType: emptyResult, // _id must be an immutable field
+		},
 		"ExistsTrue": {
 			command: bson.D{
 				{"query", bson.D{{"_id", bson.D{{"$exists", true}}}}},
--- a/integration/findandmodify_compat_test.go
+++ b/integration/findandmodify_compat_test.go
@@ -588,6 +588,7 @@ func TestFindAndModifyCompatUpsert(t *testing.T) {
 				{"update", bson.D{{"_id", "replaced"}, {"v", "replaced"}}},
 			},
 			resultType: emptyResult, // _id must be an immutable field
+			skip:       "https://github.com/FerretDB/FerretDB/issues/3017",
 		},
 		"ExistsTrue": {
 			command: bson.D{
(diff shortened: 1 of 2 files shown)


noisersup (approved):
LGTM

integration/findandmodify_compat_test.go
```diff
@@ -585,9 +585,8 @@ func TestFindAndModifyCompatUpsert(t *testing.T) {
 			command: bson.D{
 				{"query", bson.D{{"_id", bson.D{{"$exists", true}}}}},
 				{"upsert", true},
-				{"update", bson.D{{"_id", "replaced"}, {"v", "replaced"}}},
+				{"update", bson.D{{"_id", "int32"}, {"v", "replaced"}}},
```
rumyantseva, line 588: Hmm, to me this is a bit separate test case. In the previous test case ("replace") we were trying to set the value that didn't exist in any collection, just a new _id. In this test case, we try to set the value int32 which exists at least in some collections... I'd keep both of the test cases.

chilagrow: I can keep both :+1:

rumyantseva (approved):
LGTM

noisersup (approved):

PR #3188, opened 2023-08-10
Refactor aggregation operators

### LINKED ISSUE #3129: Refactor aggregation operators and accumulators
### What should be done?

Let's make our aggregation operators and accumulators take values as arguments instead of giving them documents that they need to parse.
The document parsing should be moved into the `NewOperator`, so we have less boilerplate in each operator. Example: #3122 

The second thing to do is to move operator validation from the Process function. Currently we have a lot of validation inside the Process/Accumulate, which is not great as by default we don't return errors on empty collections (because we don't call `Process` in such scenario).

https://github.com/FerretDB/FerretDB/blob/48dad105f741a9fefd8b9fcc716c59ff5320a639/internal/handlers/common/aggregations/operators/type.go#L36-L43

To fix this issue we added some `Process(nil)` calls but they are not a clean solution. We should remove those calls and move whole validation into the constructors.

### Where?

- aggregation
- aggreagation/accumulators

### Definition of Done

- spot refactorings done;
- operators/accumulators take values as arguments;
- operator/accumulator validation moved to constructors;
- validation of recursive operators should be covered either;
- unnecessary `Process`/`Accumulate` calls removed;
- something else?

noisersup:
# Description

Closes #3129.

## Readiness checklist


chilagrow:
> The second thing to do is to move operator validation from the Process function. Currently we have a lot of validation inside the Process/Accumulate, which is not great as by default we don't return errors on empty collections (because we don't call Process in such scenario).

@noisersup Do we plan to remove validation inside Process/Accumulate in separate PR too? I see you have another PR to address accumulator :hugs:

chilagrow (commented):
great seeing so many code removed :muscle:

rumyantseva (approved):
LGTM

AlekSi (approved):

PR #3194, opened 2023-08-10
Verify that client metadata not being mutated

### LINKED ISSUE #2989: Verify that client metadata document has not been mutated
### FerretDB version

v1.5.0-1-gcbd9e06a

### Backend

PostgreSQL

### Environment

- OS: Ubuntu 22.10 x86_64
- Deployment: v1.5.0-1
- Deployment details: 23.0.3

### What did you do?

See test script at #3068.

We should verify that a `isMaster` or `hello` request fails if it contains client metadata, and it is not first:

```
> db.runCommand({"hello": 1, "client": {"application": "foobar"}});
{
        "ok" : 0,
        "errmsg" : "The client metadata document may only be sent in the first hello",
        "code" : 186,
        "codeName" : "ClientMetadataCannotBeMutated"
}
>
```

See above.

### What did you see instead?

```
> db.runCommand({"hello": 1, "client": {"application": "foobar"}});
{
        "isWritablePrimary" : true,
        "maxBsonObjectSize" : 16777216,
        "maxMessageSizeBytes" : 48000000,
        "maxWriteBatchSize" : 100000,
        "localTime" : ISODate("2023-07-05T17:12:33.036Z"),
        "connectionId" : 42,
        "minWireVersion" : 0,
        "maxWireVersion" : 17,
        "readOnly" : false,
        "ok" : 1
}
> 
```

kropidlowsky:
# Description

Closes #2989.

Added support for verifying client metadata not being mutated

## Readiness checklist


rumyantseva (changes requested):
After fixing the problem with clojure, you can run the new test against MongoDB only:

```
go test -count=1 -run='TestMutatingClientMetadata' -target-url=mongodb://127.0.0.1:47017/ -target-backend=mongodb .
```

It seems to me that the test fails.

integration/basic_test.go
```diff
@@ -635,3 +635,28 @@ func TestPingCommand(t *testing.T) {
 		require.Empty(t, actualDatabases.Databases)
 	})
 }
+
+func TestMutatingClientMetadata(t *testing.T) {
```
rumyantseva, line 639: I see that the example in the original issue contains `client` field in `runCommand` call. However, in the test I don't see `client` field being set. I believe we should add this test case to ensure the problem is fixed.

internal/handlers/common/common.go
```diff
@@ -15,10 +15,31 @@
...
 const (
 	// MinWireVersion is the minimal supported wire protocol version.
 	MinWireVersion = int32(0) // needed for some apps and drivers
 
 	// MaxWireVersion is the maximal supported wire protocol version.
 	MaxWireVersion = int32(17)
 )
+
+// IsNotClientMetadata checks if the message does not contain client metadata.
+func IsNotClientMetadata(msg *wire.OpMsg) error {
```
rumyantseva, line 34: Usually, functions called `IsSomething` mean that `bool` needs to be returned. In this case, it returns an error (and it looks to me that it makes sense as then we can return the error in the callers).
 
If we want to keep this function to return an error, I'd suggest we can name it `CheckClientMedata`.

kropidlowsky:
Hi @rumyantseva thanks!
I have pushed requires changes.

With regard to running the single integration test I was trying to follow previous approach - `task test-integration-pg TEST_RUN=TestName/TestCaseName`  which used to be in CONTRIBUTING.md.

rumyantseva (commented):
Thank you very much for contributing! I tried the implementation and left my thoughts.

kropidlowsky:
Hi @rumyantseva,

Thank you very much for the review!

I have pushed the required changes.

kropidlowsky:
@noisersup PTAL

internal/handlers/common/common.go
```diff
@@ -15,10 +15,32 @@
...
 const (
 	// MinWireVersion is the minimal supported wire protocol version.
 	MinWireVersion = int32(0) // needed for some apps and drivers
 
 	// MaxWireVersion is the maximal supported wire protocol version.
 	MaxWireVersion = int32(17)
 )
+
+// CheckClientMedata checks if the message does not contain client metadata.
+func CheckClientMedata(msg *wire.OpMsg) error {
```
noisersup, line 33:
```suggestion
func CheckClientMetadata(msg *wire.OpMsg) error {
```

noisersup (commented):
LGTM! Let's just fix the typos :)

noisersup (approved):

internal/handlers/common/common.go
```diff
@@ -15,10 +15,32 @@
...
+// CheckClientMetadata checks if the message does not contain client metadata.
+func CheckClientMetadata(msg *wire.OpMsg) error {
+	document, err := msg.Document()
+	if err != nil {
+		return lazyerrors.Error(err)
+	}
+
+	if client, _ := document.Get("client"); client != nil {
+		return commonerrors.NewCommandErrorMsg(commonerrors.ErrClientMetadataCannotBeMutated,
+			"The client metadata document may only be sent in the first hello",
```
AlekSi, line 41: Okay, so the client metadata may be included in the first `hello` message but not in subsequent. But then we call that function in every `hello` message handler. We don't store the information about it being the first or not call anywhere.

I can't see how that possibly could work, tests notwithstanding.

AlekSi (changes requested):

internal/handlers/common/common.go
kropidlowsky: @AlekSi It is a bit unclear to me. Would you be able to draft/explain the scenario?
I would like to observe it on original MongoDB and then to provide next changes.

Thanks in advance!

AlekSi: Start FerretDB in the proxy-diff mode (in it the client would receive responses from MongoDB) with `task run-proxy` in one terminal and mongosh with `task mongosh` in another. The very first message `mongosh` sends is:

```json
{
  "Flags": 0,
  "FullCollectionName": "admin.$cmd",
  "NumberToReturn": -1,
  "NumberToSkip": 0,
  "Query": {
    "$k": [
      "ismaster",
      "helloOk",
      "client",
      "compression"
    ],
    "ismaster": 1,
    "helloOk": true,
    "client": {
      "$k": [
        "application",
        "driver",
        "platform",
        "os"
      ],
      "application": {
        "$k": [
          "name"
        ],
        "name": "mongosh 1.10.5"
      },
      "driver": {
        "$k": [
          "name",
          "version"
        ],
        "name": "nodejs|mongosh",
        "version": "5.7.0|1.10.5"
      },
      "platform": "Node.js v16.20.2, LE",
      "os": {
        "$k": [
          "name",
          "architecture",
          "version",
          "type"
        ],
        "name": "linux",
        "architecture": "arm64",
        "version": "5.15.49-linuxkit-pr",
        "type": "Linux"
      }
    },
    "compression": [
      "none"
    ]
  }
}
```
That's an OP_QUERY command that is handled there: https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/pg/cmd_query.go. It can also be sent as OP_MSG `hello` and `isMaster`/`ismaster` command. The `client` field is the client's metadata.

After that initial metadata is sent, all future invocations of one of those commands should not contain the metadata. (The original issue implied it can be sent if it is exactly the same, but I don't think that's true.)
```
db.runCommand({isMaster: 1}) // works

db.runCommand({isMaster: 1, client: {}})
MongoServerError: The client metadata document may only be sent in the first hello
```

So the code should handle all variants of that command, including OP_QUERY, store metadata in `ConnInfo` (or just store that it was received already), and reject future commands with metadata.

internal/clientconn/conninfo/conn_info.go
```diff
@@ -37,6 +40,35 @@ type ConnInfo struct {
 	rw       sync.RWMutex
 	username string
 	password string
+
+	clientMetadata ClientMetadata
+}
+
+type ClientMetadata struct {
+	K           []string    `json:"$k"`
```
AlekSi, line 48: Those are not really JSON values, that's just a logging representation of the `*types.Document`. On the wire, it is encoded in BSON.

I think we should be good just by keeping track of if metadata was already received:
```go
metadataReceived bool
```

internal/clientconn/conninfo/conn_info.go
```diff
@@ -37,6 +37,8 @@ type ConnInfo struct {
 	rw       sync.RWMutex
 	username string
 	password string
+
+	ClientMetadataPresence bool
```
AlekSi, line 41:
```suggestion
	clientMetadataPresence bool
```

To prevent direct access and use the same ["mutex hat"](https://dmitri.shuralyov.com/idiomatic-go#mutex-hat)

internal/clientconn/conninfo/conn_info.go
```diff
@@ -71,6 +73,14 @@ func (connInfo *ConnInfo) SetAuth(username, password string) {
 	connInfo.password = password
 }
 
+// SetClientMetadata sets the client metadata presence.
+func (connInfo *ConnInfo) SetClientMetadataPresence() {
+	connInfo.rw.Lock()
+	defer connInfo.rw.Unlock()
+
+	connInfo.ClientMetadataPresence = true
```
AlekSi, line 81: If we set it under lock, then we should read it under lock. Otherwise, there is a race condition

internal/handlers/common/metadata.go
```diff
@@ -0,0 +1,52 @@
...
+	if !doc.Has("client") {
+		return nil
+	}
+
+	clientMetadata, err := doc.Get("client")
+	if err != nil {
+		return lazyerrors.Error(err)
+	}
+
+	connInfo := conninfo.Get(ctx)
```
AlekSi, line 38: Tests explode because `connInfo` is not closed

internal/handlers/pg/cmd_query.go
```diff
@@ -33,6 +33,8 @@ import (
 
 // CmdQuery implements HandlerInterface.
 func (h *Handler) CmdQuery(ctx context.Context, query *wire.OpQuery) (*wire.OpReply, error) {
+	common.CheckClientMetadata(ctx, query.Query)
```
AlekSi, line 36: I think we should do it below, where we check for "ismaster", and don't ignore error

AlekSi (commented):
Please check tests and linters

AlekSi:
@kropidlowsky please re-request review if it is ready

internal/handlers/common/metadata.go
kropidlowsky: I have tried to add  `defer connInfo.Close()` in:
1.  where you pointed
2. only `TestCheckClientMetadata`, under creating the new context
3. disabling the unit test (`TestCheckClientMetadata`) and keeping 1 point

All scenarios still do explode tests.
I was looking for use cases of closing the connInfo. However, it is unclear to me why.

AlekSi: We should call `defer connInfo.Close()` only in handlers code, never in tests.

This place is the right one

AlekSi: #3376

kropidlowsky: Unfortunately, it explodes again

internal/clientconn/conninfo/conn_info.go
```diff
@@ -57,6 +58,22 @@ func (connInfo *ConnInfo) SetAuth(username, password string) {
 	connInfo.password = password
 }
 
+// ClientMetadataPresence returns stored client metadata presence.
+func (connInfo *ConnInfo) ClientMetadataPresence() bool {
+	connInfo.rw.RLock()
+	defer connInfo.rw.Unlock()
```
AlekSi, line 64:
```suggestion
	defer connInfo.rw.RUnlock()
```

That's why it panics

internal/handlers/common/metadata.go
AlekSi: That's a different panic now, see my other comment

AlekSi:
@kropidlowsky please re-request review if it is ready :)

kropidlowsky:
@AlekSi integration tests continue to explode. Can you please advise? I would appreciate it

AlekSi (approved):

AlekSi:
@kropidlowsky the root cause is fixed in be944e751f36005d29c0eec0396d3651345d56e5

chilagrow (approved):
Great work! Thanks for your contribution :hugs:

AlekSi:
@rumyantseva PTAL

rumyantseva (approved):
:shipit:

PR #3246, opened 2023-08-23
Cleanup SQLite tests

noisersup:
# Description

Closes #3131.

## Readiness checklist


integration/aggregate_documents_compat_test.go
```diff
@@ -247,7 +238,8 @@ func testAggregateCommandCompat(tt *testing.T, testCases map[string]aggregateCom
 				}
 			})
 
-			if tc.failsForSQLite != "" && setup.IsSQLite(tt) {
+			// TODO https://github.com/FerretDB/FerretDB/issues/3148
+			if setup.IsSQLite(tt) {
```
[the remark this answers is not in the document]
noisersup: done

integration/update_field_compat_test.go
```diff
@@ -212,17 +212,17 @@ func TestUpdateFieldCompatIncComplex(t *testing.T) {
 		"DotNotationNegativeIndex": {
 			update: bson.D{{"$inc", bson.D{{"v.-1", int32(42)}}}},
 		},
-		"DotNotationIndexOutsideArray": {
+		"DotNotIndexOutsideArray": {
 			update: bson.D{{"$inc", bson.D{{"v.100", int32(42)}}}},
 		},
-		"DotNotationArrayFieldNotExist": {
+		"DotNotArrayFieldNotExist": {
```
AlekSi, line 218: "Not array field not exists". While the test is actually about array, not "not array".

I think that abbreviation of dot notation is unfortunate. It also does not quite belong to that PR/issue.

noisersup: We need to shorten those names as they are longer than 64 characters so they always fail

AlekSi: I'm not sure why they become longer if we don't change them, but okay, let's shorten them some other non-confusing way

noisersup: They are already too long since creation. We just don't see it as they fail in the same way both on MongoDB and FerretDB. We could actually panic or fail if the name is too long to omit such situation in the future

AlekSi: Oh, that's bad…

AlekSi: @noisersup is it done?

noisersup: I replaced the names with less confusing ones, currently waiting for CI to pass, I'm also investigating why the tests were passing, although not sure if that should be covered in this PR in your opinion

AlekSi: I think we should at least understand why that was happening

AlekSi (commented):
Please re-request review once ready

integration/update_field_compat_test.go
noisersup: Found it, it's due to the fact that we append `_sqlite` at the end of the database name

noisersup: If we want to avoid name mangling I guess we are forced to shorten our db names.

AlekSi (approved):

PR #3253, opened 2023-08-27
add: freeze metod for documant type

### LINKED ISSUE #3112: Add ability to freeze `*types.Document` and `*types.Array`
### What should be done?

`*types.Document` and `*types.Array` should have a `Freeze()` method that would disallow any receiver modifications. All future calls that would modify frozen objects (such as `*types.Document.SortFieldsByKey`, `*types.Array.RemoveByPath`, etc.) should panic.

That method will be called in places where we don't expect further modifications; for example, when documents are passed to backend methods.

### Where?

https://github.com/FerretDB/FerretDB/tree/main/internal/types

### Definition of Done

- unit tests added/updated;
- spot refactorings done.

KrishnaSindhur:
# Description

Closes #3112.

## Readiness checklist

### DIFF
--- a/internal/types/document.go
+++ b/internal/types/document.go
@@ -37,6 +37,7 @@ type document interface {
 // (key/value pairs where key is a string and value is any BSON value).
 type Document struct {
 	fields []field
+	frozen bool
 }
 
 // field represents a field in the document.
@@ -77,7 +78,7 @@ func ConvertDocument(d document) (*Document, error) {
 		}
 	}
 
-	return &Document{fields}, nil
+	return &Document{fields, false}, nil
 }
 
 // MakeDocument creates an empty document with set capacity.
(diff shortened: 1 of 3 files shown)


KrishnaSindhur:
@AlekSi its just a draft pr added freeze() function for document type, just to check am I in proper path or should i change the approach I used currently?

AlekSi:
@KrishnaSindhur you are on the right track, I would do it the same way 👍

AlekSi (commented):
There are more methods where we should panic. For example, `Array.Remove`

KrishnaSindhur:
should i check the all the methods of array and document which are modifying the values?

AlekSi:
Well, I think it should be somewhat obvious from existing documentation. If it is not – sure, let's update it

KrishnaSindhur:
I can add few more test if its required for array and document method!!

internal/types/array.go
```diff
@@ -83,6 +84,10 @@ func (a *Array) GetByPath(path Path) (any, error) {
 
 // Set sets the value at the given index.
 func (a *Array) Set(index int, value any) error {
+	if a.frozen {
+		panic("this array is not mutable its been freezed")
+	}
```
AlekSi, line 89: I think we should extract that info helper method:

```go
func (a *Array) checkFrozen() {
  if a.frozen {
    panic("array is frozen")
  }
}

func (a *Array) Set(index int, value any) error {
  a.checkFrozen()
  // ...
}
```

And then do the same for Document

AlekSi (dismissed):

chilagrow (approved):
Thanks for your contribution! Great work :rocket:

PR #3264, opened 2023-08-28
Implement `explain` command for SQLite

### LINKED ISSUE #3050: Implement basic `explain` for SQLite
### What should be done?

We already have a basic implementation of `explain` command, but it does not exist in the backend. It should.

We should run `EXPLAIN QUERY PLAN` SQL command on the query (that is almost static for now due to the absence of any pushdowns) and return the SQLite-specific query plan.

`pushdown`/`sortingPushdown`/`limitPushdown` boolean values should not be returned from the backend; they are fixed on the handler layer for now.

### Where?

See references to this issue.

### Definition of Done

- SQLite backend and handler updated;
- integration/compatibility tests enabled and pass;
- no references to this issue are left.

noisersup:
# Description

Closes #3050.

I had some troubles with understanding what data should be outputted, as SQLite returns completely different data format and less information than pg

## Readiness checklist


internal/handlers/sqlite/msg_explain.go
```diff
@@ -55,20 +58,43 @@ func (h *Handler) MsgExplain(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 	cmd := params.Command
 	cmd.Set("$db", params.DB)
 
-	// TODO https://github.com/FerretDB/FerretDB/issues/3050
+	db, err := h.b.Database(params.DB)
+	if err != nil {
+		if backends.ErrorCodeIs(err, backends.ErrorCodeDatabaseNameIsInvalid) {
+			msg := fmt.Sprintf("Invalid namespace specified '%s.%s'", params.DB, params.Collection)
+			return nil, commonerrors.NewCommandErrorMsgWithArgument(commonerrors.ErrInvalidNamespace, msg, "find")
```
chilagrow, line 65:
```suggestion
			return nil, commonerrors.NewCommandErrorMsgWithArgument(commonerrors.ErrInvalidNamespace, msg, document.Command())
```

internal/handlers/sqlite/msg_explain.go
```diff
@@ -55,20 +58,43 @@ func (h *Handler) MsgExplain(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+
+		return nil, lazyerrors.Error(err)
+	}
+	defer db.Close()
+
+	coll, err := db.Collection(params.Collection)
+	if err != nil {
+		if backends.ErrorCodeIs(err, backends.ErrorCodeCollectionNameIsInvalid) {
+			msg := fmt.Sprintf("Invalid collection name: %s", params.Collection)
+			return nil, commonerrors.NewCommandErrorMsgWithArgument(commonerrors.ErrInvalidNamespace, msg, "find")
```
chilagrow, line 76: Let's replace `find` with `document.Command()`

internal/backends/sqlite/collection.go
```diff
@@ -211,8 +211,50 @@ func (c *collection) DeleteAll(ctx context.Context, params *backends.DeleteAllPa
...
+		var id int32
+		var parent int32
+		var notused int32
+		var detail string
+
+		if err := rows.Scan(&id, &parent, &notused, &detail); err != nil {
+			return nil, lazyerrors.Error(err)
+		}
+
+		queryPlan.Append(fmt.Sprintf("id=%d parent=%d notused=%d detail=%s", id, parent, notused, detail))
```
chilagrow, line 252: Does the output appear sensible?

chilagrow (commented):
Just a couple of comments :hugs:

internal/backends/sqlite/collection.go
rumyantseva: As far as I understand, explain for SQLite is a bit different than what we see for PostgreSQL. So, in addition to the plan itself, they also have those extra fields that show the relations between steps in the plan. I just added the simplest approach (we print all the fields that are returned). But maybe there is a better idea of what we can do with it. We don't have complicated plans now (as we query everything, no WHERE clauses, no indexes etc), but maybe in the future we have a better idea how to print it.

chilagrow: That's fine. I'm not familiar with SQLite execution plan so I just wanted to ask above is what we want to return :hugs:

For now, having something is good. But I'm lacking what they mean, perhaps just adding comment to link where we can find more info about id, parent etc would be good :hugs:

rumyantseva: Yeah, let's add it. I used this docs - https://www.sqlite.org/eqp.html - to understand it.

> A query plan is represented as a tree. In raw form, as returned by sqlite3_step(), each node of the tree consists of four fields: An integer node id, an integer parent id, an auxiliary integer field that is not currently used, and a description of the node.

The tricky part is that SQLite cli tools can draw quite nice plans based on this info (basically, like a tree where parents-children look quite nice), but implementing it on our side looks unnecessary for now. And also some other clients (like the one in Goland) don't show it as a tree, but just print these exact four columns... So I think we are ok for now.

rumyantseva: Added!

chilagrow: We don't have to worry about prettiness for now :hugs:
I think `id` and `detail` gives idea of what's going on which is nice!

```
test> db.collection.insertMany([{_id: 1}, {_id: 2}])

test> db.runCommand({explain: { find: "colls", query: { _id: 1 } }})
{
  queryPlanner: { Plan: [ 'id=2 parent=0 notused=0 detail=SCAN colls_367c18bc' ] },
  explainVersion: '1',
  command: { find: 'colls', query: { _id: 1 }, '$db': 'test' },
  serverInfo: {
    host: 'alma',
    version: '6.0.42',
    gitVersion: 'c6a73ee1f1f5c235b4464c2e9dadf9f033747e17',
    ferretdbVersion: 'v1.9.0-30-gc6a73ee1'
  },
  pushdown: false,
  sortingPushdown: false,
  limitPushdown: false,
  ok: 1
}
```

chilagrow (commented):
Change looks good to me! 

Why is CI failing?

rumyantseva:
@chilagrow yeah, failing CI looks strange to me. Linters and unit tests pass locally...

chilagrow:
> @chilagrow yeah, failing CI looks strange to me. Linters and unit tests pass locally...

I'm puzzled :thinking:

rumyantseva:
Ok, I think something was lost when I was moving between PRs locally 😅 

Now everything is back.

internal/backends/collection.go
```diff
@@ -64,6 +64,8 @@ type QueryParams struct {
 
 // QueryResult represents the results of Collection.Query method.
 type QueryResult struct {
+	// nothing for now - no pushdowns yet
+	// TODO https://github.com/FerretDB/FerretDB/issues/3235
```
AlekSi, line 68: What will be added there once we add pushdown?

AlekSi: But why would we need that?

rumyantseva: to show results of push down in explain?

AlekSi: `QueryResults` is a result of `Query`, not `Explain`

rumyantseva: ok, then I remove this todo for now, and we can decide later where these param should belong.

chilagrow (approved):
:rocket:

AlekSi (approved):

PR #3266, opened 2023-08-29
Support `$bit` bitwise update operator

### LINKED ISSUE #821: Support `$bit` bitwise update operator
### What should be done?

https://www.mongodb.com/docs/manual/reference/operator/update/bit/

### Where?

In the common handler code.

### Definition of Done

- all handlers updated;
- integration/compatibility tests added/updated:
- - dot notation;
- - all data types;
- - overflows and underflows;
- spot refactorings done;
- user documentation updated.

adetunjii:
# Description
Add support for $bit update operator

Closes #821.

## Readiness checklist


internal/handlers/common/update.go
```diff
@@ -746,6 +752,91 @@ func processCurrentDateFieldExpression(doc *types.Document, currentDateVal any)
...
+				if err = doc.SetByPath(path, bitOpResult); err != nil {
+					return false, lazyerrors.Error(err)
+				}
+
+				changed = true
+
+				continue
+			}
+
+			return false, err
```
chilagrow, line 833: We want to return error with appropriate message here. For example, what if the operator was not `and`, `or` or `xor`?

internal/handlers/common/update.go
```diff
@@ -746,6 +752,91 @@ func processCurrentDateFieldExpression(doc *types.Document, currentDateVal any)
...
+				command,
+			)
+		}
+
+		for _, bitOp := range nestedDoc.Keys() {
+			bitOpValue := must.NotFail(nestedDoc.Get(bitOp))
+			fieldType := commonparams.AliasFromType(bitOpValue)
+
+			// $bit operator argument must be an integer (Int32/Long)
+			if fieldType != "int" && fieldType != "long" {
```
chilagrow, line 810: We are checking int and long type here and in `performBitLogic()`? I don't think we need to check it here.

Instead, let's use the error from `performBitLogic()` and return `newUpdateError` based on that. :hugs:

chilagrow (commented):
Thanks for your contribution :hugs: Great work so far!

I've added few comments, could you add integration and compatibility tests for the implemented logic please :pray:

internal/handlers/common/update.go
adetunjii: Is `commonerrors.ErrInvalidArg` an appropriate error code for this use case. I'm not sure of the error code to use

AlekSi: In general, we should return the same error code as MongoDB. What does it produce in that case? It should be easy to check with FerretDB's proxy mode.

chilagrow: Indeed we want to return the same error code as MongoDB :hugs:. Also our compatibility tests check error codes are the same, so we should add test case for failing case and ensure that test passes.

chilagrow (approved):
Great work!

AlekSi (approved):
Thank you!

PR #3279, opened 2023-08-29
Implement `getLog` for SQLite

### LINKED ISSUE #3261: Implement `getLog` for SQLite
### What should be done?

TBD

### Where?

TBD

### Definition of Done

- all handlers updated;
- unit tests added/updated;
- spot refactorings done;
- something else?

rumyantseva:
# Description

Closes #3261.

**Note:** The new implementation of `getLog` doesn't depend on a particular handler and can be reused in a universal handler. 

## Readiness checklist


internal/handlers/sqlite/msg_getlog.go
```diff
@@ -16,12 +16,137 @@ package sqlite
...
+			"ok", float64(1),
+		))
+
+	case "startupWarnings":
+		state := h.StateProvider.Get()
+
+		info := version.Get()
+
+		startupWarnings := []string{
+			"Powered by FerretDB " + info.Version + " and SQLite.",
```
rumyantseva, line 92: Currently, we don't have handler version for SQLite, so I didn't include it.

chilagrow (dismissed):
Looks great :rocket:

rumyantseva:
[upd] I set "do not merge" label because I'm going to continue working on the PR (to have backend name and version passed to getLog)

AlekSi (commented):
Let's set backend name and version

internal/handlers/sqlite/msg_getlog.go
AlekSi: https://github.com/FerretDB/FerretDB/blob/b16fe1f85de627e6cdcee186b21e56e2f13e6974/internal/backends/sqlite/metadata/pool/pool_test.go#L169

rumyantseva: Great! Thanks!

internal/backends/sqlite/metadata/registry.go
```diff
@@ -352,6 +354,27 @@ func (r *Registry) CollectionRename(ctx context.Context, dbName, oldCollectionNa
...
+	if r.version != "" {
+		return r.version, nil
+	}
+
+	db := r.p.GetFirst(ctx)
+	if db == nil {
+		return "", nil
+	}
+
+	row := db.QueryRowContext(context.Background(), "SELECT sqlite_version()")
```
chilagrow, line 370: Let's use `ctx` passed from the function parameter :hugs:

internal/backends/sqlite/metadata/registry.go
```diff
@@ -352,6 +354,27 @@ func (r *Registry) CollectionRename(ctx context.Context, dbName, oldCollectionNa
...
+	if err := row.Scan(&r.version); err != nil {
```
chilagrow, line 371: Do we need a lock for writing to the registry? :thinking:

chilagrow (commented):
Just a small comment :hugs:

internal/backends/sqlite/metadata/registry.go
rumyantseva: Good question! Version is a minor thing compared to databases, so I don't want to lock the whole registry with the existing mutex. A new mutex that only protects this field can be set, but again I'm not sure if it brings a lot of value. In principle, I can set it to be completely accurate.

internal/backends/sqlite/metadata/registry.go
rumyantseva: fixed

internal/backends/sqlite/metadata/registry.go
chilagrow: Okay, let's leave it as it is for now. :+1:

chilagrow (dismissed):
Looks good to me!

internal/backends/sqlite/metadata/registry.go
```diff
@@ -55,6 +55,8 @@ type Registry struct {
 	p *pool.Pool
 	l *zap.Logger
 
+	version string // SQLite version
```
AlekSi, line 58: We don't need that field

internal/backends/sqlite/metadata/registry.go
```diff
@@ -352,6 +354,27 @@ func (r *Registry) CollectionRename(ctx context.Context, dbName, oldCollectionNa
 	panic("not implemented")
 }
 
+// Version returns SQLite version.
+//
+// If there are no databases in the registry, empty string is returned.
+func (r *Registry) Version(ctx context.Context) (string, error) {
```
AlekSi, line 360: > Registry provides access to SQLite databases and collections information.

SQLite library version is not a database or collection of information. It should not be there

internal/handlers/pg/pgdb/pool.go
```diff
@@ -69,6 +69,9 @@ func NewPool(ctx context.Context, uri string, logger *zap.Logger, p *state.Provi
 			return lazyerrors.Error(err)
 		}
 
+		// After moving pg to the new architecture, we should stop storing the version in the state,
```
AlekSi, line 72: No, we should keep storing it in the state because it is used in other places. See references to State's HandlerVersion field

AlekSi (changes requested):

internal/handlers/pg/pgdb/pool.go
rumyantseva: Well, it is only needed to send telemetry.

AlekSi: So? Does this PR send the SQLite version to telemetry?

rumyantseva: My comment is that "it is used in other places". There is only one place - telemetry - not "places". I'm ok to store this info twice, I just don't think it's really needed.

AlekSi: There is one place where it is stored and two places where it is used.

I agree we should not store that information twice. We also should report in the telemetry, and that code already uses state. So why this PR stores it the in second place?

internal/backends/sqlite/metadata/registry.go
AlekSi: I don't think we should use another method at all. `state.Proivder.Get().HandlerVersion` is already there

rumyantseva: Ah, ok, I can try it. I misunderstood what you said yesterday - as it sounded to me that we should have `Info` interface for the backend that should return both name and version.

AlekSi: That's indeed what I meant yesterday. But refactoring the telemetry reporter to accept the backend does not sound like a good idea, especially now. So, let's just™ store state the current backend in the state in addition to the version and use that.

AlekSi: (and if we store the handler/backend in the state, let's remove it from `NewReporterOpts`)

chilagrow (approved):
Looks good to me!

AlekSi (approved):

PR #3304, opened 2023-09-04
Implement `createIndexes` for SQLite

### LINKED ISSUE #3176: Implement `createIndexes` for SQLite
### What should be done?

tbd

### Where?

- issue number mentions

### Definition of Done

- all handlers updated;
- unit tests added/updated;
- spot refactorings done;
- something else?

rumyantseva:
# Description

Closes #3176.

## Readiness checklist


internal/backends/sqlite/metadata/registry.go
```diff
@@ -353,6 +374,94 @@ func (r *Registry) CollectionRename(ctx context.Context, dbName, oldCollectionNa
...
 }
 
+// indexesCreate creates an index in the collection.
+//
+// It does not hold the lock.
+func (r *Registry) indexesCreate(ctx context.Context, db *fsql.DB, c *Collection, indexes []IndexInfo) error {
+	err := db.InTransaction(ctx, func(tx *fsql.Tx) error {
+		for _, index := range indexes {
+			// check if index already exists
+			q := `SELECT EXISTS (SELECT 1 FROM sqlite_master WHERE type= 'index' and tbl_name = ? and name = ?)`
```
chilagrow, line 384: Can this be checked from `Settings` rather than querying `sqlite_master`? Answer might be no :stuck_out_tongue:

internal/backends/sqlite/metadata/registry.go
```diff
@@ -353,6 +374,94 @@ func (r *Registry) CollectionRename(ctx context.Context, dbName, oldCollectionNa
...
+
+			columns := make([]string, len(index.Key))
+			for i, key := range index.Key {
+				columns[i] = fmt.Sprintf("%s->'$.%s'", DefaultColumn, key.Field)
+				if key.Descending {
+					columns[i] += " DESC"
+				}
+			}
+
+			q = fmt.Sprintf(q, c.TableName+index.Name, c.TableName, strings.Join(columns, ", "))
```
chilagrow, line 413: Could we use a placeholder `?` or `$` for fields? I suppose `index.Name` can be user input.

internal/backends/sqlite/metadata/registry.go
```diff
@@ -353,6 +374,94 @@ func (r *Registry) CollectionRename(ctx context.Context, dbName, oldCollectionNa
...
+	}
+
+	c := r.collectionGet(dbName, collectionName)
+
+	err = r.indexesCreate(ctx, db, c, indexes)
+	if err != nil {
+		return lazyerrors.Error(err)
+	}
+
+	r.colls[dbName][collectionName] = c
```
chilagrow, line 460: Should we call `r.collectionGet(dbName, collectionName)` after `r.indexesCreate(ctx, db, c, indexes)`?
I assume this line updates `Settings` of Collection.

internal/backends/collection.go
```diff
@@ -237,10 +239,30 @@ type ListIndexesResult struct {
...
+	Key    IndexKey
 	Unique bool
 }
 
+// IndexKey is a list of field name + sort order pairs.
+//
+// It exists as a separate type to make it easier to compare index keys.
+type IndexKey []IndexKeyPair
+
+// Equal returns true if the given index key is equal to the current one.
```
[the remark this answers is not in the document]
rumyantseva: Nice catch! If not `slices.Equal` then ``slices.EqualFunc` should help! Let me try!

internal/backends/sqlite/metadata/registry.go
rumyantseva: In principle, we can. My idea was to use the query here as we use it in the other place in this func, but actually checking settings might be better.

internal/backends/sqlite/metadata/registry.go
rumyantseva: Yeah, it can be user input that breaks everything (there is  `DangerousKey` test that currently fails). I tried `?` but I didn't find an approach that would work.

internal/backends/sqlite/metadata/registry.go
rumyantseva: Yes, this row is to update settings. We can call `r.collectionGet`, but I'm not sure why. We already blocked everything with the mutex, so we don't need any extra check  that collection is updated.

Also please not that calling `indexCreate` modifies `c` (Settings are set). Maybe I need to document it.

internal/backends/sqlite/metadata/registry.go
chilagrow: I think this one would be important one. I wonder why :thinking:

internal/backends/sqlite/metadata/registry.go
chilagrow: Not super important if it works.

internal/backends/sqlite/metadata/registry.go
rumyantseva: There should be a better way to escape this string anyway... I'll take a look at how we do it for pg (there was sort of a hack to do it), maybe we can do smth similar here.

integration/indexes_compat_test.go
```diff
@@ -342,17 +350,19 @@ func TestDropIndexesCompat(tt *testing.T) {
...
-			dropIndexName: "*",
-			resultType:    emptyResult,
-		},
-		"NonExistent": {
-			dropIndexName: "nonexistent_1",
-			resultType:    emptyResult,
-		},
+		// We need to figure out why these test cases pass even if there is no implementation for dropIndexes.
+		// TODO https://github.com/FerretDB/FerretDB/issues/3287
+		//"AsteriskWithDropOne": {
```
rumyantseva, line 355: Ok, for this test case I figured out. It is checked by the driver itself. I will remove the test case as it rather tests how driver works, not how FerretDB works.

integration/indexes_compat_test.go
```diff
@@ -342,13 +350,6 @@ func TestDropIndexesCompat(tt *testing.T) {
 			},
 			dropIndexName: "v_-1",
 		},
-		"AsteriskWithDropOne": {
```
rumyantseva, line 345: This one is checked by the driver itself. I remove the test case as it rather tests how driver works, not how FerretDB works.

integration/explain_compat_test.go
```diff
@@ -34,7 +34,7 @@ type explainCompatTestCase struct {
 	resultType compatTestCaseResultType // defaults to nonEmptyResult
 
 	skip           string // skip test for all handlers, must have issue number mentioned
-	failsForSQLite string // optional, if set, the case is expected to fail for SQLite due to given issue}
+	failsForSQLite string // optional, if set, the case is expected to fail for SQLite due to given issue
```
AlekSi, line 37: 👍

integration/indexes_command_compat_test.go
```diff
@@ -40,7 +41,9 @@ func TestCreateIndexesCommandCompat(t *testing.T) {
 		key            any
 		unique         any
 		resultType     compatTestCaseResultType // defaults to nonEmptyResult
-		skip           string                   // optional, skip test with a specified reason
+
+		skip           string // optional, skip test with a specified reason
+		failsForSQLite string // optional, if set, the case is expected to fail for SQLite due to given issue}
```
AlekSi, line 46: But after fixing it in another place, we copy the same old comment

integration/indexes_command_test.go
```diff
@@ -221,9 +222,10 @@ func TestCreateIndexesCommandInvalidSpec(tt *testing.T) {
 		missingIndexes bool // optional, if set indexes must be nil
 		noProvider     bool // if set, no provider is added.
 
-		err        *mongo.CommandError // required, expected error from MongoDB
-		altMessage string              // optional, alternative error message for FerretDB, ignored if empty
-		skip       string              // optional, skip test with a specified reason
+		err            *mongo.CommandError // required, expected error from MongoDB
+		altMessage     string              // optional, alternative error message for FerretDB, ignored if empty
+		skip           string              // optional, skip test with a specified reason
+		failsForSQLite string              // optional, if set, the case is expected to fail for SQLite due to given issue}
```
AlekSi, line 228:
```suggestion
		failsForSQLite string              // optional, if set, the case is expected to fail for SQLite due to given issue
```

internal/backends/collection.go
```diff
@@ -259,6 +281,85 @@ func (cc *collectionContract) ListIndexes(ctx context.Context, params *ListIndex
...
+	defer observability.FuncCall(ctx)()
+
+	existingIndexes, err := cc.c.ListIndexes(ctx, new(ListIndexesParams))
+	if err != nil {
+		if ErrorCodeIs(err, ErrorCodeCollectionDoesNotExist) {
+			// If the collection doesn't exist, it's not a problem,
+			// it just means that there are no existing indexes.
+			existingIndexes = &ListIndexesResult{Indexes: []IndexInfo{}}
+		} else {
+			return nil, lazyerrors.Error(err)
```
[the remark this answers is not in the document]
AlekSi: In any case, that's too much logic in the contract anyway. Contract's CreateIndexes should validate parameters, call wrapped method, and check error code. Please cut it; all those "notes" are not important now

internal/backends/error.go
```diff
@@ -40,6 +40,11 @@ const (
 	ErrorCodeCollectionAlreadyExists
 
 	ErrorCodeInsertDuplicateID
+
+	ErrorCodeIndexNameIsEmpty
+	ErrorCodeIndexAlreadyExists
+	ErrorCodeIndexOptionsConflict
+	ErrorCodeIndexKeySpecsConflict
```
AlekSi, line 47: All those errors are never handled and therefore should not be introduced

internal/backends/sqlite/metadata/registry.go
AlekSi: It is important not to query the database after `initCollections`. Note that all other queries are inserts/updates/deletes; there are no selects.

internal/backends/sqlite/metadata/registry.go
```diff
@@ -353,6 +374,94 @@ func (r *Registry) CollectionRename(ctx context.Context, dbName, oldCollectionNa
...
+
+	r.rw.Lock()
+	defer r.rw.Unlock()
+
+	_, err := r.collectionCreate(ctx, dbName, collectionName)
+	if err != nil {
+		return lazyerrors.Error(err)
+	}
+
+	db, err := r.databaseGetOrCreate(ctx, dbName)
```
AlekSi, line 448: database creation is not possible there, so it should be DatabaseGetExisting

internal/backends/sqlite/metadata/registry.go
AlekSi: That line should be inside indexesCreate

internal/backends/validation.go
```diff
@@ -74,3 +74,38 @@ func validateCollectionName(name string) error {
...
 }
+
+// validateIndexes checks that indexes are valid for FerretDB.
+func validateIndexes(exitingIndexes, newIndexes []IndexInfo) error {
+	for _, index := range newIndexes {
+		if index.Name == "" {
+			return NewError(ErrorCodeIndexNameIsEmpty, nil)
+		}
+
+		// Validation needs to be extended (e.g. to check that index names don't contain illegal symbols, see the issue).
```
AlekSi, line 85: Please remove it completely, as it introduces error codes that are not handled yet, and we don't know if they will be handled at all

integration/indexes_command_compat_test.go
rumyantseva: ¯\_(ツ)_/¯

AlekSi: ¯\_(:_:)_/¯

internal/backends/collection.go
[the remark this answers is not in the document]
rumyantseva: Ok, then I'll create a separate issue about it (we have some tests that check these fields, and these tests will fail).

internal/backends/sqlite/metadata/registry.go
rumyantseva: @AlekSi it can't be there. If `indexesCreate` was called from `collectionCreate` to create the default index, technically, collection doesn't exist before we exit from `collectionCreate`, so we can't update it from `indexesCreate`.

AlekSi: > technically, collection doesn't exist before we exit from collectionCreate

I'm not sure why is that, but I'm sure that for any given function `foo()` I could write:

```go
func foo() {
  // ...
}

// For call foo() with lock.
func Foo() {
  l.Lock()
  defer l.Unlock()

  foo()
}
```

That's what we do for other exported/unexported function pairs, so I don't see the reason to not do the same there

rumyantseva: It's not enough. Collection creation consists of a few steps. One of the steps is index creation, another step is metadata insertion. Only after metadata insertion is complete correctly, we can consider collection is created and can be stored in `r.colls`. Otherwise, "everyone" is an owner and can modify `r.colls`, and this is not a safe approach.

rumyantseva: ok, I can copy-paste the same logic to `indexesCreate`, but I still don't think it's a good idea. two different things shouldn't both be owners. collection is already a good and suitable owner.

AlekSi: The owner is the registry. `r.colls` is modified by many of its methods already, including `databaseDrop` that is not about collections.

And to be clear – I'm not asking you to copy&paste. All I'm saying that exported/unexported method pairs should be consistent.

rumyantseva: @AlekSi then I don't understand what you mean. Could you please suggest the exact changes? (either a commit or just a code suggestion).

AlekSi: Just do it as close as possible for now

rumyantseva:
Ok, the problem with unit tests is because of `TestCreateSameStress` doesn't work. I don't see what exactly happens, they way how mutex is used looks correct to me, so I'm still investigating the problems. Suggestions are welcome.

PR #3329, opened 2023-09-08
Implement `dropIndexes` for SQLite

### LINKED ISSUE #3287: Implement `dropIndexes` for SQLite
### What should be done?

TBD

rumyantseva:
# Description

Closes #3287.

## Readiness checklist


internal/backends/collection.go
```diff
@@ -284,6 +285,28 @@ func (cc *collectionContract) CreateIndexes(ctx context.Context, params *CreateI
...
+	Msg         string
+}
+
+// DropIndexes drops indexes for the collection.
+//
+// The operation should be atomic.
+// If some indexes cannot be dropped, the operation should be rolled back,
+// and the first encountered error should be returned.
+func (cc *collectionContract) DropIndexes(ctx context.Context, params *DropIndexesParams) (*DropIndexesResult, error) {
+	return cc.c.DropIndexes(ctx, params)
```
AlekSi, line 307: `checkError`

internal/backends/error.go
```diff
@@ -65,6 +70,21 @@ func NewError(code ErrorCode, err error) *Error {
 	}
 }
 
+// NewErrorWithArgument creates a new backend error with argument to be passed to the handler.
+//
+// Code must not be 0. Err may be nil.
+func NewErrorWithArgument(code ErrorCode, err error, arg any) *Error {
```
rumyantseva, line 76: I'm not sure about this approach, please share your ideas if you know how to improve it.

noisersup (dismissed):
LGTM

internal/backends/collection.go
```diff
@@ -285,6 +286,30 @@ func (cc *collectionContract) CreateIndexes(ctx context.Context, params *CreateI
...
+
+// DropIndexes drops indexes for the collection.
+//
+// The operation should be atomic.
+// If some indexes cannot be dropped, the operation should be rolled back,
+// and the first encountered error should be returned.
+//
+// Database or collection may not exist; that's not an error.
+func (cc *collectionContract) DropIndexes(ctx context.Context, params *DropIndexesParams) (*DropIndexesResult, error) {
+	res, err := cc.c.DropIndexes(ctx, params)
```
AlekSi, line 307: PTAL at all other methods there and add what is missing

internal/backends/collection.go
```diff
@@ -285,6 +286,30 @@ func (cc *collectionContract) CreateIndexes(ctx context.Context, params *CreateI
 	return res, err
 }
 
+// DropIndexesParams represents the parameters of Collection.DropIndexes method.
+type DropIndexesParams struct {
+	Indexes []string       // Index names to drop.
+	Spec    []IndexKeyPair // Single index to drop.
+	DropAll bool           // Drop all indexes except for _id_.
```
AlekSi, line 293: `DeleteAll` does not accept a filter, `ordered`, etc. Please do the same.

internal/backends/decorators/dummy/collection.go
```diff
@@ -70,6 +70,11 @@ func (c *collection) CreateIndexes(ctx context.Context, params *backends.CreateI
 	return c.c.CreateIndexes(ctx, params)
 }
 
+// DropIndexes implements backends.Collection interface.
+func (c *collection) DropIndexes(context.Context, *backends.DropIndexesParams) (*backends.DropIndexesResult, error) {
+	return c.c.DropIndexes(context.Background(), nil)
```
AlekSi, line 75: Compare with every other method there

internal/backends/error.go
AlekSi: By removing it. We talked about backend implementation not returning error messages, yet we still do that there for some reason.

internal/backends/error.go
```diff
@@ -40,6 +40,9 @@ const (
 	ErrorCodeCollectionAlreadyExists
 
 	ErrorCodeInsertDuplicateID
+
+	ErrorCodeIndexDoesNotExist
+	ErrorCodeIndexInvalidOptions
```
AlekSi, line 45: Please remove both – we don't do validation when we create indexes

internal/backends/decorators/oplog/collection.go
```diff
@@ -185,6 +185,11 @@ func (c *collection) CreateIndexes(ctx context.Context, params *backends.CreateI
 	return c.c.CreateIndexes(ctx, params)
 }
 
+// DropIndexes implements backends.Collection interface.
+func (c *collection) DropIndexes(context.Context, *backends.DropIndexesParams) (*backends.DropIndexesResult, error) {
+	return c.c.DropIndexes(context.Background(), nil)
```
AlekSi, line 190: and there

internal/backends/postgresql/collection.go
```diff
@@ -74,6 +74,11 @@ func (c *collection) CreateIndexes(ctx context.Context, params *backends.CreateI
 	panic("not implemented")
 }
 
+// DropIndexes implements backends.Collection interface.
+func (c *collection) DropIndexes(context.Context, *backends.DropIndexesParams) (*backends.DropIndexesResult, error) {
```
AlekSi, line 78: and there

internal/backends/sqlite/collection.go
```diff
@@ -386,6 +387,107 @@ func (c *collection) CreateIndexes(ctx context.Context, params *backends.CreateI
...
+					fmt.Sprintf("index not found with name [%s]", index),
+				)
+			}
+		}
+
+		err = c.r.IndexesDropByNames(ctx, c.dbName, c.name, params.Indexes)
+
+	case len(params.Spec) > 0:
+		// default index {_id: 1} can't be dropped
+		if len(params.Spec) == 1 && params.Spec[0].Field == "_id" && !params.Spec[0].Descending {
```
chilagrow, line 425: This probably should be handler logic.

internal/backends/collection.go
chilagrow: Should it take a single parameter `Indexes []IndexInfo`? Handler already hold that info when it called `ListIndexes` before calling `DropIndexes`.

internal/backends/sqlite/metadata/registry.go
```diff
@@ -542,6 +542,75 @@ func (r *Registry) indexesDrop(ctx context.Context, dbName, collectionName strin
 	return nil
 }
 
+// IndexesDropByNames remove given connection's indexes but their names.
+func (r *Registry) IndexesDropByNames(ctx context.Context, dbName, collectionName string, indexes []string) error {
```
chilagrow, line 546: I think we can have one method that accepts `Indexes []IndexInfo` and won't need three methods `IndexesDropByNames`, `IndexesDropBySpec` and `IndexesDropAll`.

This is optional comment. It can be todo for a future to reduce complexity a bit.

chilagrow (commented):
Just some suggestions :hugs: 

great work!

internal/backends/sqlite/metadata/registry.go
AlekSi: I think it should be a list of index names because they are unique for collection, and the caller (handler) can always call ListIndexes to get them

internal/backends/sqlite/collection.go
rumyantseva: In principle, this logic could be in the handler, but isn't it a bit too dangerous? :) 

However, if we always define the list of indexes to delete then it should be safe, yeah.

internal/backends/error.go
rumyantseva: Then I need to modify tests again to have `failsForSQLite` for particular cases. Can we release it as is and cleanup after the release?

rumyantseva:
Neunundneunzig Luftballons :D

internal/backends/sqlite/collection.go
chilagrow: Yeah exactly, we define the list to delete :+1:

internal/backends/postgresql/collection.go
chilagrow, line 78:
```suggestion
func (c *collection) DropIndexes(ctx context.Context, params *backends.DropIndexesParams) (*backends.DropIndexesResult, error) {
```

internal/handlers/sqlite/msg_dropindexes.go
```diff
@@ -16,13 +16,234 @@ package sqlite
...
+		if len(spec) == 1 && spec[0].Field == "_id" && !spec[0].Descending {
+			return nil, false, commonerrors.NewCommandErrorMsgWithArgument(
+				commonerrors.ErrInvalidOptions, "cannot drop _id index", command,
+			)
+		}
+
+		for _, index := range existing {
+			for i, key := range index.Key {
+				if key.Field == spec[i].Field && key.Descending == spec[i].Descending {
+					return []string{index.Name}, false, nil
```
chilagrow, line 144: We are returning if first field matches with `spec[i].Field`. Don't we want to check if all `Field` match? :thinking: 

For example, if index contains two fields, it will match as long as first field is the same.

internal/handlers/sqlite/msg_dropindexes.go
```diff
@@ -16,13 +16,234 @@ package sqlite
...
+			}
+
+			if !found {
+				return nil, false, commonerrors.NewCommandErrorMsgWithArgument(
+					commonerrors.ErrIndexNotFound,
+					fmt.Sprintf("index not found with name [%s]", index),
+					command,
+				)
+			}
+		}
```
chilagrow, line 214: We want to return found indexes `toDrop` right? :thinking: 
```suggestion
		}
		return toDrop, false, nil
```

internal/handlers/sqlite/msg_dropindexes.go
```diff
@@ -16,13 +16,234 @@ package sqlite
...
+				return nil, false, commonerrors.NewCommandErrorMsgWithArgument(
+					commonerrors.ErrTypeMismatch,
+					fmt.Sprintf(
+						"BSON field 'dropIndexes.index' is the wrong type '%s', expected types '[string, object]'",
+						commonparams.AliasFromType(v),
+					),
+					command,
+				)
+			}
+
```
chilagrow, line 195: //  TODO cannot drop ID index `index == "_id_"`

chilagrow (commented):
Just a few comments, great work simplifying it! :hugs:

internal/backends/sqlite/metadata/registry.go
chilagrow: Yep certainly cleaner!

internal/handlers/sqlite/msg_dropindexes.go
AlekSi: That's done above, where we check for `ErrIteratorDone`

chilagrow: Yes, I missed that :sweat_smile:

AlekSi:
@rumyantseva PTAL on Chi's comments and send a follow up RP, if needed

internal/handlers/sqlite/msg_dropindexes.go
rumyantseva: Nice catch! Our tests didn't find it, so I think I should add a test where the first field matches and the second doesn't.

PR #3337, opened 2023-09-11
Implement filter pushdown for SQLite

### LINKED ISSUE #3235: Implement filter pushdown for SQLite
### What should be done?

We should support the same filter pushdowns as we currently support for PostgreSQL.
We should integrate it with `explain`.

We also should investigate how we should change our tests to make it easier for new backends: currently, they expect all backends to support the same pushdowns.

### Where?

SQLite backend and handler.

### Definition of Done

- handler updated;
- unit tests added/updated;
- spot refactorings done;
- user documentation updated.

noisersup:
# Description

Closes #3235.

## Readiness checklist


internal/backends/sqlite/collection.go
```diff
@@ -91,6 +96,142 @@ func (c *collection) Query(ctx context.Context, params *backends.QueryParams) (*
...
+			filters = append(filters, subquery)
+			args = append(args, v)
+
+		case time.Time:
+			subquery := fmt.Sprintf(`EXISTS (SELECT value FROM json_each(%v) WHERE value = ?)`, queryPath) // TODO sanitize the key
+
+			filters = append(filters, subquery)
+			val := v.UnixMilli()
+
+			args = append(args, val)
```
noisersup, line 204: @AlekSi here we need to convert time.Time to int, by using UnixMilli. I'm not quiet sure if this parsing logic should be here, and using sjson package for that seems more relevant, though the Marshal functions return strings.

AlekSi: But why `MarshalSingleValue` can't be used there?

noisersup: It returns slice of bytes, which can be parsed to string, and SQLite library will search for strings and not numbers

AlekSi:
```
test> db.test.find()
[
  {
    _id: ObjectId("65097913f4928d7f17fa78b6"),
    t: ISODate("2023-09-19T10:33:55.233Z")
  },
  {
    _id: ObjectId("65097915f4928d7f17fa78b7"),
    t: ISODate("2023-09-19T10:33:57.941Z")
  },
  {
    _id: ObjectId("65097918f4928d7f17fa78b8"),
    t: ISODate("2023-09-19T10:34:00.403Z")
  },
  {
    _id: ObjectId("65097a40e627ae53f9f9604d"),
    t: [ ISODate("2023-09-19T10:38:56.881Z") ]
  },
  {
    _id: ObjectId("65097a41e627ae53f9f9604e"),
    t: [ ISODate("2023-09-19T10:38:57.787Z") ]
  },
  {
    _id: ObjectId("65097a43e627ae53f9f9604f"),
    t: [ ISODate("2023-09-19T10:38:59.310Z") ]
  }
]
```

```
sqlite> SELECT * FROM test_afd071e5 WHERE EXISTS (SELECT value FROM json_each(_ferretdb_sjson->'$.t') WHERE value = 1695119635233);
_ferretdb_sjson
{"$s":{"p":{"_id":{"t":"objectId"},"t":{"t":"date"}},"$k":["_id","t"]},"_id":"65097913f4928d7f17fa78b6","t":1695119635233}
sqlite> SELECT * FROM test_afd071e5 WHERE EXISTS (SELECT value FROM json_each(_ferretdb_sjson->'$.t') WHERE value = 1695119939310);
_ferretdb_sjson
{"$s":{"p":{"_id":{"t":"objectId"},"t":{"t":"array","i":[{"t":"date"}]}},"$k":["_id","t"]},"_id":"65097a43e627ae53f9f9604f","t":[1695119939310]}
```

So that seems to work, and `MarshalSingleValue` would return a JSON representation of that time as a JSON slice of bytes / string without quotes, so that should work, no?

AlekSi: Then, let's do `value = %s` without a placeholder. It still sounds sus, TBH.

noisersup: That's what is produced when we provide a string value into the SQLite
```
2023-09-19T12:54:20.219+0200    DEBUG   sqlite.test     fsql/db.go:77   >>> SELECT _ferretdb_sjson FROM "values_34474c3b" WHERE EXISTS (SELECT value FROM json_each(_ferretdb_sjson->'$."v"') WHERE value = ?)       {"args": ["1695119939310"]}
```

noisersup: > Then, let's do value = %s without a placeholder. It still sounds sus, TBH.

What about sanitizing?

AlekSi: I think we should use `json`, not `value`, as we use `->` operator for indexes that returns JSON representation as text:

```
sqlite> SELECT _ferretdb_sjson->'$.t', typeof(_ferretdb_sjson->'$.t') FROM test_afd071e5;
┌────────────────────────┬────────────────────────────────┐
│ _ferretdb_sjson->'$.t' │ typeof(_ferretdb_sjson->'$.t') │
├────────────────────────┼────────────────────────────────┤
│ 1695119635233          │ text                           │
│ 1695119637941          │ text                           │
│ 1695119640403          │ text                           │
│ [1695119936881]        │ text                           │
│ [1695119937787]        │ text                           │
│ [1695119939310]        │ text                           │
└────────────────────────┴────────────────────────────────┘
sqlite> SELECT value, typeof(value) FROM test_afd071e5, json_each(_ferretdb_sjson->'$.t');
┌───────────────┬───────────────┐
│     value     │ typeof(value) │
├───────────────┼───────────────┤
│ 1695119635233 │ integer       │
│ 1695119637941 │ integer       │
│ 1695119640403 │ integer       │
│ 1695119936881 │ integer       │
│ 1695119937787 │ integer       │
│ 1695119939310 │ integer       │
└───────────────┴───────────────┘
sqlite> SELECT json, typeof(json) FROM test_afd071e5, json_each(_ferretdb_sjson->'$.t');
┌─────────────────┬──────────────┐
│      json       │ typeof(json) │
├─────────────────┼──────────────┤
│ 1695119635233   │ text         │
│ 1695119637941   │ text         │
│ 1695119640403   │ text         │
│ [1695119936881] │ text         │
│ [1695119937787] │ text         │
│ [1695119939310] │ text         │
└─────────────────┴──────────────┘
```

AlekSi: Or maybe not:

```
sqlite> SELECT json, typeof(json) FROM test_afd071e5, json_each(_ferretdb_sjson->'$.t');
┌───────────────────────────────┬──────────────┐
│             json              │ typeof(json) │
├───────────────────────────────┼──────────────┤
│ 1695119635233                 │ text         │
│ 1695119637941                 │ text         │
│ 1695119640403                 │ text         │
│ [1695119936881]               │ text         │
│ [1695119937787]               │ text         │
│ [1695119939310]               │ text         │
│ [1695121115855,1695121115855] │ text         │
│ [1695121115855,1695121115855] │ text         │
└───────────────────────────────┴──────────────┘
```

I think I will leave you with that now :) Just be careful with type-checking.

---

> What about sanitizing?

There is no need to sanitize a number. We should not do that in all cases

noisersup: Unfortunately it's not that easy to use json, as it treats arrays as whole:
```
sqlite> SELECT json, value FROM "values_34474c3b" join json_each(_ferretdb_sjson->'$."v"');
json|value
1|1
[1,2,1]|1
[1,2,1]|2
[1,2,1]|1
2|2
"foo"|foo
"65046bcc4ed83bef4404a73a"|65046bcc4ed83bef4404a73a
"65046bcc4ed83bef4404a73a"|65046bcc4ed83bef4404a73a
"65046bcc4ed83bef4404a73a"|65046bcc4ed83bef4404a73a
1695119939310|1695119939310
true|1
["65046bcc4ed83bef4404a73a"]|65046bcc4ed83bef4404a73a
```

Because of that we cannot query the items of array

integration/helpers.go
```diff
@@ -456,3 +456,34 @@ func createNestedDocument(n int, arr bool) any {
...
+// For example if both pg and SQlite backends are expected to pushdown, the `PgPushdown + SQLitePushdown` operation
+// can be used.
+type ResultPushdown uint8
+
+const (
+	NoPushdown     ResultPushdown = 0
+	PgPushdown     ResultPushdown = 1
+	SQLitePushdown ResultPushdown = 2
+
+	AllPushdown ResultPushdown = PgPushdown + SQLitePushdown
```
chilagrow, line 470: I think we want to think about how easy to adjust this when we add new backend with no pushdown. I like the introduction of current number enum thing, having constant 0, 1, 2, 4, 8 etc will allow us to do addition without collision :hugs: . By the way, naming `AllPushdown` would require modifying this name when a new backend is added or hana is using this test. I think it can be named this specifically to indicate it's pg and sqlite pushdown, something like `PgSQLitePushdown ResultPushdown = PgPushdown + SQLitePushdown`? :thinking:

integration/helpers.go
```diff
@@ -456,3 +456,34 @@ func createNestedDocument(n int, arr bool) any {
...
+)
+
+// PushdownExpected returns true if the pushdown is expected for currently running backend.
+func (res ResultPushdown) PushdownExpected(t testtb.TB) bool {
+	if setup.IsPushdownDisabled() {
```
chilagrow, line 475: Could you make sure you run entire testsuit with disable pushdown flag please :pray:

chilagrow (commented):
Great pushdown result handling :hugs: 

Current SQL filters I'm unsure the way it's done :thinking:

integration/helpers.go
AlekSi: #3381

chilagrow: Thank you :pray:

integration/helpers.go
noisersup: Good concern, I'm not quiet sure about `PgSQLitePushdown` idea though. My goal with this approach was to allow syntax like `PgPushdown+SQLitePushdown` to make it readable by just looking at the test case. What if we introduce another backend (`hana` for example), to stay consistent we would need to Introduce `HanaPgPushdown` `SQliteHanaPushdown`, etc...

I have 2 ideas though:
- we can just leave a comment above `AllPushdown` that it should be updated with each new backend
- we could assign The `0xFF` value to the AllPushdown, so it will be able to match every new backend introduced

WDYT?

internal/backends/sqlite/collection.go
```diff
@@ -91,6 +90,140 @@ func (c *collection) Query(ctx context.Context, params *backends.QueryParams) (*
 	}, nil
 }
 
+// prepareWhereClause adds WHERE clause with filters given in the document.
+// It returns the WHERE clause and the SQLite arguments.
+func prepareWhereClause(filterDoc *types.Document) (string, []any, error) {
```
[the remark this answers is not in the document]
noisersup: I will create another issue for that, or introduce it in separate PR. I believe our main priority is to make sure that pushdown tests are able to differentiate between the backends, although if I misunderstood our priorities please let me know :)

internal/backends/sqlite/collection.go
```diff
@@ -91,6 +90,140 @@ func (c *collection) Query(ctx context.Context, params *backends.QueryParams) (*
...
+		k, v, err := iter.Next()
+		if errors.Is(err, iterator.ErrIteratorDone) {
+			break
+		}
+
+		if err != nil {
+			return "", nil, lazyerrors.Error(err)
+		}
+
+		queryPath := fmt.Sprintf("%s->'$.\"%s\"'", metadata.DefaultColumn, k)
```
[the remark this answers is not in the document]
noisersup: > // TODO sanitize the key

Yep, it's not safe and I'll address it in this PR, also `%q` will make this code cleaner that's right :)

internal/backends/sqlite/collection.go
```diff
@@ -91,6 +90,140 @@ func (c *collection) Query(ctx context.Context, params *backends.QueryParams) (*
...
+
+		switch {
+		case err == nil:
+			// Handle dot notation.
+			// TODO https://github.com/FerretDB/FerretDB/issues/2069
+			if path.Len() > 1 {
+				continue
+			}
+		case errors.As(err, &pe):
+			// ignore empty key error, otherwise return error
```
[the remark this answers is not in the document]
noisersup: Yes, the same thing is done in pgdb

internal/backends/sqlite/collection.go
```diff
@@ -91,6 +90,140 @@ func (c *collection) Query(ctx context.Context, params *backends.QueryParams) (*
...
+			}
+
+			subquery := fmt.Sprintf(`EXISTS (SELECT value FROM json_each(%v) WHERE value %s)`, queryPath, sql) // TODO sanitize the key
+
+			filters = append(filters, subquery)
+			args = append(args, parseValue(v))
+
+		case float64:
+			var sql string
+			switch {
```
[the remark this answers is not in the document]
noisersup: the difference is the type of compared values, the same logic exists in pgdb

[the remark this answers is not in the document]
chilagrow: Code copy and pasted but slight differences are always hard to tell :see_no_evil: . So it's actually maxSafeLong instead of maxSafeDouble.

internal/backends/sqlite/collection.go
```diff
@@ -91,6 +90,156 @@ func (c *collection) Query(ctx context.Context, params *backends.QueryParams) (*
...
+
+			case v < -maxSafeDouble:
+				comparison = `< ?`
+				v = -maxSafeDouble
+			default:
+				// don't change the default eq query
+			}
+
+			// json_each returns top level json values, and the contents of arrays if any
+			subquery := fmt.Sprintf(`EXISTS (SELECT value FROM json_each(%v) WHERE value %s)`, queryPath, comparison)
```
chilagrow, line 208: By the way, why do we use `%v` in `json_each(%v)`?

internal/backends/sqlite/collection.go
```diff
@@ -91,6 +90,140 @@ func (c *collection) Query(ctx context.Context, params *backends.QueryParams) (*
...
+			switch {
+			case v > types.MaxSafeDouble:
+				sql = ` > ?`
+				v = types.MaxSafeDouble
+
+			case v < -types.MaxSafeDouble:
+				sql = ` < ?`
+				v = -types.MaxSafeDouble
+			default:
+				// don't change the default eq query
```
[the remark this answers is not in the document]
chilagrow: I still don't understand this. Is this because `value` is boolean? But it seems it return top level json values or array content so? :thinking:

chilagrow (commented):
Just few comments to help me understand :hugs:

internal/backends/sqlite/collection.go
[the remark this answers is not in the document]
chilagrow: Thanks for the update! Very clear why we do it even though sql is a bit more complex

integration/helpers.go
chilagrow: Great idea! I like how it's done :hugs:

internal/backends/sqlite/collection.go
```diff
@@ -91,6 +90,156 @@ func (c *collection) Query(ctx context.Context, params *backends.QueryParams) (*
 	}, nil
 }
 
+// prepareWhereClause adds WHERE clause with filters given in the document.
```
chilagrow, line 93: What do you think of moving pushdown related logic to a new file `pushdown.go`, it's gonna get big.

integration/helpers.go
```diff
@@ -456,3 +456,35 @@ func createNestedDocument(n int, arr bool) any {
 
 	return bson.D{{"v", child}}
 }
+
+// resultPushdown stores the information about expected pushdown results for a single or multiple backends.
+// For example if both pg and SQlite backends are expected to pushdown, the `pgPushdown + sqlitePushdown` operation
```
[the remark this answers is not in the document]
noisersup: What exactly should be updated here?

internal/backends/sqlite/collection.go
noisersup: fixed

internal/backends/sqlite/collection.go
noisersup: good idea

integration/pushdown.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/FerretDB/FerretDB/internal/util/testutil/testtb"
+)
+
+// resultPushdown stores the information about expected pushdown results for a single or multiple backends.
+// For example if both pg and SQlite backends are expected to pushdown, the `pgPushdown + sqlitePushdown` operation
+// can be used.
+type resultPushdown uint8
+
+const (
+	noPushdown     resultPushdown = 0 // 0000 0000
```
AlekSi, line 14: The idiomatic way is to use `iota` with a bitshift: `1 << iota`. See also all other our "enum"s where we omit type after the first constant, etc

internal/backends/sqlite/collection.go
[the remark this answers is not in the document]
chilagrow: No no, not about `json_each`, what I'm confused is about `sql` variable being empty, despite having second `%s` in the `subquery` variable. 

```
subquery := fmt.Sprintf(`EXISTS (SELECT value FROM json_each(%v) WHERE value %s)`, queryPath, sql)
```
Won't `subquery` be like `EXISTS (SELECT value FROM json_each(_json->'$.\"_id\"') WHERE value )` and I'm struggling to understand how it's a valid sql query :thinking:

integration/helpers.go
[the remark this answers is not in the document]
chilagrow: yea true, it's actually fine :hugs:

internal/backends/sqlite/collection.go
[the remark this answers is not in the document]
chilagrow: Don't you need this to be like below? 
```
comparison := ` = ?`
```

Why doesn't newly added test fail?

AlekSi:
Just not to merge it right into the coming release

internal/backends/sqlite/collection.go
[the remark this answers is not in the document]
noisersup: nice catch! It apparently worked as `EXISTS (SELECT value FROM json_each(...) WHERE value)` query returns every document where the path have any value. Fixed that.

chilagrow (approved):
Great work :+1:

integration/pushdown.go
```diff
@@ -0,0 +1,52 @@
...
+
+// resultPushdown stores the information about expected pushdown results for a single or multiple backends.
+// For example if both pg and SQlite backends are expected to pushdown, the `pgPushdown + sqlitePushdown` operation
+// can be used.
+type resultPushdown uint8
+
+const (
+	noPushdown     resultPushdown = 1 << iota // 0000 0000
+	pgPushdown                                // 0000 0001
+	sqlitePushdown                            // 0000 0010
```
AlekSi, line 30: Those comments are not actually correct. It is better to check them in the init function

integration/pushdown.go
```diff
@@ -0,0 +1,52 @@
...
+		res = noPushdown
+	}
+
+	switch {
+	case setup.IsSQLite(t):
+		return res&sqlitePushdown == sqlitePushdown
+	case setup.IsPostgres(t):
+		return res&pgPushdown == pgPushdown
+	case setup.IsMongoDB(t):
+		return false
```
AlekSi, line 48: Let's use consistent order: PostgreSQL, SQLite, other backends, MongoDB

integration/setup/test_helpers.go
```diff
@@ -35,6 +35,13 @@ func IsSQLite(tb testtb.TB) bool {
 	return *targetBackendF == "ferretdb-sqlite"
 }
 
+// IsPostgres returns true if the current test is running for pg handler.
+//
+// This function should not be used lightly.
+func IsPostgres(tb testtb.TB) bool {
```
AlekSi, line 41: Move it up to preserve the order

AlekSi (commented):
Almost there

internal/backends/sqlite/collection.go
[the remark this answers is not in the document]
chilagrow: :see_no_evil: that was unexpected. For newly added pushdown, we should have unit test confirming the built query.
For now, let's do with it and thanks for fixing it :pray:

AlekSi (commented):
As discussed on DS, let's split that PR into two: one for tests changes (that will be need for the new PG backend), one for SQLite pushdown

internal/backends/sqlite/collection.go
noisersup: Oops, I believe I have reverted this change, let me quickly fix that

chilagrow (commented):
Is this one in pause right?

PR #3345, opened 2023-09-12
Add test for inserting different data types

### LINKED ISSUE #3235: Implement filter pushdown for SQLite

noisersup:
# Description

Closes #3235.

## Readiness checklist

### DIFF
--- a/integration/insert_command_test.go
+++ b/integration/insert_command_test.go
@@ -150,7 +150,7 @@ func TestInsertIDDifferentTypes(t *testing.T) {
 	})
 
 	AssertEqualWriteError(t, mongo.WriteError{
-		Message: "E11000 duplicate key error collection: TestUpdateFieldSetIDDifferentTypes.TestUpdateFieldSetIDDifferentTypes index: _id_ dup key: { _id: 1 }",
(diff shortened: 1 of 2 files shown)


rumyantseva (approved):
LGTM.

Could you please enable auto-merge if the PR is ready for review.

integration/insert_command_test.go
```diff
@@ -132,3 +132,41 @@ func TestInsertCommandErrors(tt *testing.T) {
 		})
 	}
 }
+
+func TestInsertIDDifferentTypes(t *testing.T) {
```
chilagrow, line 136:
```suggestion
func TestInsertIDDuplicateNumberTypes(t *testing.T) {
```
Something a bit more specific to what it's testing would be better. By no means, it's not a blocker :hugs:

chilagrow (approved):

integration/insert_command_test.go
AlekSi: Let's keep as-is for now

AlekSi (approved):

PR #3356, opened 2023-09-14
Store `PostgreSQL` metadata

### LINKED ISSUE #3335: Store PostgreSQL metadata
### What should be done?

We should store collections metadata exactly the same way as an old backend does – as `*types.Document` in a single field, with the same fields, etc. However, the exposed `Registry` API should be similar to SQLite's backend.

In that issue only the `metadata` package should be changed, not the new backend package itself.

### Where?

https://github.com/FerretDB/FerretDB/tree/main/internal/backends/postgresql/metadata

### Definition of Done

- unit tests added;
- spot refactorings done.

chilagrow:
# Description

Closes #3335.

There is no integration test, because pg still uses `handlers/pg/pgdb` implementation. So unit tests added are the ones covers the logic.

Indexes will be added in https://github.com/FerretDB/FerretDB/issues/3375

## Readiness checklist


internal/backends/postgresql/metadata/metadata.go
```diff
@@ -15,21 +15,184 @@
...
+	return &Collection{
+		Name:      c.Name,
+		TableName: c.TableName,
+		Settings:  c.Settings.deepCopy(),
+	}
+}
+
+// Settings represents collection settings.
+type Settings struct {
+	Indexes []IndexInfo `json:"indexes"`
```
AlekSi, line 59: Let's remove _everything_ about indexes for now

chilagrow: Sure!

chilagrow:
Seems like `task test-short` fails on CI with `panic: *pool.Pool has not been finalized`. Any clue where I forgot to call `Close()` on pool?

```
--- FAIL: TestCreateDrop (0.00s)
    logger.go:130: 2023-09-19T10:59:01.721Z	WARN	pool/pool.go:127	Pool: connection failed	{"username": "", "error": "[opendb.go:83 pool.openDB] [opendb.go:98 pool.checkConnection] failed to connect to `host=127.0.0.1 user=username database=ferretdb`: dial error (dial tcp 127.0.0.1:5432: connect: connection refused)"}
    registry_test.go:91: 
        	Error Trace:	/home/runner/work/FerretDB/FerretDB/internal/backends/postgresql/metadata/registry_test.go:91
        	            	[registry.go:89 metadata.NewRegistry] [registry.go:218 metadata.(*Registry).getPool] [pool.go:128 pool.(*Pool).Get] [opendb.go:83 pool.openDB] [opendb.go:98 pool.checkConnection] failed to connect to `host=127.0.0.1 user=username database=ferretdb`: dial error (dial tcp 127.0.0.1:5432: connect: connection refused)
        	Test:       	TestCreateDrop
FAIL
panic: *pool.Pool has not been finalized
Object created by goroutine 53 [running]:
runtime/debug.Stack()
	/opt/hostedtoolcache/go/1.21.1/x64/src/runtime/debug/stack.go:24 +0x[67](https://github.com/FerretDB/FerretDB/actions/runs/6234621760/job/16922195744?pr=3356#step:7:68)
github.com/FerretDB/FerretDB/internal/util/debugbuild.Stack(...)
	/home/runner/work/FerretDB/FerretDB/internal/util/debugbuild/debugbuild.go:28
github.com/FerretDB/FerretDB/internal/util/resource.Track({0xf56940, 0xc00032a640}, 0x1001cd0?)
	/home/runner/work/FerretDB/FerretDB/internal/util/resource/resource.go:78 +0x1bb
github.com/FerretDB/FerretDB/internal/backends/postgresql/metadata/pool.New({0x1023b3a, 0x45}, 0xc00032e310, 0xc00032a580)
	/home/runner/work/FerretDB/FerretDB/internal/backends/postgresql/metadata/pool/pool.go:72 +0x73d
github.com/FerretDB/FerretDB/internal/backends/postgresql/metadata.NewRegistry({0x1023b3a, 0x45}, 0xc00032e310, 0x0?)
	/home/runner/work/FerretDB/FerretDB/internal/backends/postgresql/metadata/registry.go:73 +0xb1
github.com/FerretDB/FerretDB/internal/backends/postgresql/metadata.TestRenameCollection(0xc0003ca[68](https://github.com/FerretDB/FerretDB/actions/runs/6234621760/job/16922195744?pr=3356#step:7:69)0)
	/home/runner/work/FerretDB/FerretDB/internal/backends/postgresql/metadata/registry_test.go:392 +0x233
testing.tRunner(0xc0003ca6[80](https://github.com/FerretDB/FerretDB/actions/runs/6234621760/job/16922195744?pr=3356#step:7:81), 0x1046c28)
	/opt/hostedtoolcache/go/1.21.1/x64/src/testing/testing.go:1595 +0x239
created by testing.(*T).Run in goroutine 1
	/opt/hostedtoolcache/go/1.21.1/x64/src/testing/testing.go:1648 +0x[82](https://github.com/FerretDB/FerretDB/actions/runs/6234621760/job/16922195744?pr=3356#step:7:83)b
```

internal/backends/postgresql/metadata/registry.go
```diff
@@ -16,20 +16,36 @@ package metadata
 
 import (
 	"context"
+	"fmt"
```
chilagrow, line 19: github collapses this file, majority of change is in this file, please review it.

internal/backends/postgresql/metadata/registry.go
```diff
@@ -73,6 +115,108 @@ func (r *Registry) Close() {
...
+	}
+
+	return dbNames, nil
+}
+
+// initCollections loads collections metadata from the database during initialization.
+func (r *Registry) initCollections(ctx context.Context, dbName string, p *pgxpool.Pool) error {
+	defer observability.FuncCall(ctx)()
+
+	r.rw.Lock()
```
rumyantseva, line 174: Why do we need a lock here? I see that for SQLite we don't have a lock in a similar place.

internal/backends/postgresql/metadata/registry.go
```diff
@@ -65,6 +81,32 @@ func NewRegistry(u string, l *zap.Logger, sp *state.Provider) (*Registry, error)
 		colls: map[string]map[string]*Collection{},
 	}
 
+	// use defer to close registry upon error to avoid forgetting
+	defer func() {
+		if err != nil {
+			r.Close()
```
rumyantseva, line 87: Minor comment: `r.p.Close()` might be a bit more safe here.
Currently, `r.Close()` only does `r.p.Close()` but if in the future we extend it and it will do something else that belongs to `r`, we might end up here by attempting to close something that is not initialized yet. But on the other hand, if we don't call `r.Close()` we also might end up by not closing something that must be closed. so I'm not sure :)

internal/backends/postgresql/metadata/registry.go
```diff
@@ -104,3 +248,432 @@ func (r *Registry) DatabaseList(ctx context.Context) ([]string, error) {
...
+	}
+
+	return p, nil
+}
+
+// DatabaseGetOrCreate returns a connection to existing database or newly created database.
+//
+// The dbName must be a validated database name.
+//
+// If the user is not authenticated, it returns error.
```
rumyantseva, line 278: Do I understand correctly that it's because `getPool` returns an error?
Should we add the same comment to the `DatabaseList` method doc? I see all the methods that call `getPool` have this comment, but not `DatabaseList`.

rumyantseva (dismissed):
Thanks for implementing this!

The code looks reasonable to me and consistent with what we have for `sqlite` metadata, so I approve.

The only thing that looks differently - in sqlite registry file unexported method goes first, and then we have the exported method, e.g. `databaseDrop` goes before `DatabaseDrop`. In the postgresql registry it's the opposite. It doesn't affect the quality, so I don't think it's significant, just a bit inconsistent.

I leave a couple of more comments, but none of them are important.

internal/backends/postgresql/metadata/registry.go
chilagrow: I think implementation of `Close()` should handle closing something not initialised. Often `Close()` function sets variables to nil, but subsequent call to `Close()` shouldn't fail and should be idempotent. :hugs: 

```
func (r *Registry) Close() {
	r.p.Close()
	if r.something != nil {
		r.something.Close()
	}
}
```

internal/backends/postgresql/metadata/registry.go
chilagrow: Yes I'll add it.

chilagrow:
> The only thing that looks differently - in sqlite registry file unexported method goes first, and then we have the exported method, e.g. databaseDrop goes before DatabaseDrop. In the postgresql registry it's the opposite. It doesn't affect the quality, so I don't think it's significant, just a bit inconsistent.

Exported method is called from externally, then exported method calls unexported method. So so while implementing, it was easier to follow the flow of code if exported method came first then unexported method came right after. Also exported methods are short and unexported methods can be quite long, so I didn't have to go all the way up to function declaration of unexported method. Anyway, it was easier for me that's why I did it. But I agree about consistency, is having unexported method first our preferred way? :hugs:

internal/backends/postgresql/metadata/registry.go
AlekSi, line 1: Don't miss that file

internal/backends/postgresql/metadata/registry.go
AlekSi: That comment is also collapsed. But it is possible to leave a comment on the file now:

CleanShot 2023-09-20 at 12 29 42@2x

chilagrow: :scream:

That's a nice feature I didn't know!

internal/backends/postgresql/metadata/registry.go
```diff
@@ -65,6 +81,32 @@ func NewRegistry(u string, l *zap.Logger, sp *state.Provider) (*Registry, error)
...
+	defer func() {
+		if err != nil {
+			r.Close()
+		}
+	}()
+
+	connInfo := conninfo.New()
+	ctx := conninfo.Ctx(context.Background(), connInfo)
+
+	dbPool, err := r.getPool(ctx)
```
AlekSi, line 94: If PostgreSQL requires authentication and FerretDB's PostgreSQL URL does not contain a username/password, that would fail. But it does not fail now, and this is supported and documented configuration: https://docs.ferretdb.io/security/authentication/ So that's no-go.

`pool` package already stores connection pulls by full URL (that includes authentication). We should do the same there (or move a part of that logic to the pool). Each user / each URL will have its own view of the metadata.

AlekSi (changes requested):

internal/backends/postgresql/metadata/registry.go
AlekSi: > Each user / URL will have its own view of the metadata.

Actually, let's drop that last part – we don't support authorization yet, and we will not support backend-layer authorization going forward. So, let's keep the same metadata for all connected clients, as the Registry's documentation comment says. We still need to delay the initial loading of metadata until we can authenticate – and that means until the first call with conninfo.

AlekSi: No-no, let's not change the pool. Let's also store everything in the `colls` field as it is now. What we should do is to call the "load metadata if not loaded yet" method from all other methods that are called from the backend – they will pass conninfo in the context

chilagrow: I see, so we don't load `colls` at the creation of registry. Instead we load it upon first call to the registry :+1:

chilagrow: Now metadata is loaded upon first call by client :+1:

internal/backends/postgresql/metadata/registry.go
```diff
@@ -73,10 +90,153 @@ func (r *Registry) Close() {
...
 
-// getPool returns a pool of connections to PostgreSQL database
+// CheckAuthAndLoadMetadataIfEmpty returns a pool of connections to PostgreSQL database
 // for the username/password combination in the context using [conninfo].
 //
-// All methods should use that method to check authentication.
+// It loads metadata if it hasn't been loaded from the database yet.
+//
+// All methods should use this method to check authentication and load metadata.
+func (r *Registry) CheckAuthAndLoadMetadataIfEmpty(ctx context.Context) (*pgxpool.Pool, error) {
```
AlekSi, line 99: There is no need for that method to be exported or be separated from `loadMetadata`

rumyantseva (dismissed):
Everything looks good to me, I don't have further questions.

When we start using the metadata package, we can see how it works, run integration tests etc.

noisersup (approved):
Looks great!

AlekSi (approved):

PR #3358, opened 2023-09-14
Fix metadata updates for `dropIndexes`

### LINKED ISSUE #3328: Write unit tests for SQLite `createIndexes` and `dropIndexes` implementation
### What should be done?

It would be nice to add some unit tests to ensure that table names and index names in SQLite are called correctly.

### Where?

internal/handlers/registry/registry_test.go

### Definition of Done

- unit tests added/updated

### LINKED ISSUE #3366: SQLite `dropIndexes` should update metadata table
### What should be done?

Indexes are dropped in below: https://github.com/FerretDB/FerretDB/blob/22de2805f499056834c8fa97f8d89ea823188072/internal/backends/sqlite/metadata/registry.go#L540

But metadata table `_ferretdb_collections`  `settings`, still contains deleted indexes.
We should update metadata table `_ferretdb_collections`  `settings` column with updated indexes just like `indexesCreate(...)`

### Where?

https://github.com/FerretDB/FerretDB/blob/22de2805f499056834c8fa97f8d89ea823188072/internal/backends/sqlite/metadata/registry.go#L540

### Definition of Done

- unit tests added/updated;
- spot refactorings done;

rumyantseva:
# Description

Closes #3328.
Closes #3366.

## Readiness checklist


internal/backends/sqlite/metadata/registry_test.go
```diff
@@ -316,3 +316,116 @@ func TestCreateDropSameStress(t *testing.T) {
...
+	}
+
+	err = r.IndexesCreate(ctx, dbName, collectionName, toCreate)
+	require.NoError(t, err)
+
+	collection := r.CollectionGet(ctx, dbName, collectionName)
+
+	t.Run("NonUniqueIndex", func(t *testing.T) {
+		indexName := collection.TableName + "_index_non_unique"
+		q := fmt.Sprintf("SELECT sql FROM sqlite_master WHERE type = 'index' AND name = '%s'", indexName)
```
chilagrow, line 376: Very clever check :hugs:

chilagrow (approved):
Great work!

internal/backends/sqlite/metadata/registry.go
```diff
@@ -537,6 +537,11 @@ func (r *Registry) indexesDrop(ctx context.Context, dbName, collectionName strin
 		c.Settings.Indexes = slices.Delete(c.Settings.Indexes, i, i+1)
 	}
 
+	q := fmt.Sprintf("UPDATE %q SET settings = ?", metadataTableName)
+	if _, err := db.ExecContext(ctx, q, c.Settings); err != nil {
```
chilagrow, line 541: If this fails, do we want have undo `drop indexes`? I see that `indexesCreate(..)` attempts undo action :thinking: . After all undo action is also best effort though.

rumyantseva: Then we might be in an infinite loop because `indexesCreate` calls `dropIndexes` if a similar thing fails.
A better approach would be to use a transaction but we decided against transactions.

chilagrow: Yeah if we call `indexesCreate` from this, it will produce infinite loop. So solution would need to be more than just that and it would bring more logic. So for now, let's keep it as it is :hugs:

chilagrow (approved):
Thanks for quick fix as well! :hugs:

internal/backends/sqlite/metadata/registry.go
rumyantseva: Yeah... I'm a bit worried though, as with the current approach we don't have enough consistency. So, maybe having a potential infinite loop is better than having a potential data inconsistency...

@noisersup @AlekSi what do you think?

AlekSi (approved):

Issue #3409, opened 2023-09-21
Some backend tests are fluky


AlekSi:
### What should be done?

SQLite:
* https://github.com/FerretDB/FerretDB/actions/runs/6262515571/job/17004858668?pr=3407
* https://github.com/FerretDB/FerretDB/actions/runs/6301692434/job/17107283075?pr=3430
* https://github.com/FerretDB/FerretDB/actions/runs/6416041905/job/17419012857?pr=3496

PostgreSQL:
* https://github.com/FerretDB/FerretDB/actions/runs/6271012235/job/17029850391?pr=3411
* https://github.com/FerretDB/FerretDB/actions/runs/6496580977/job/17643830496?pr=3565

We should either fix the test or our code.

### Where?

Integration tests, unit tests.

### Definition of Done

Test or code fixed.

rumyantseva:
See also https://ferretdb.slack.com/archives/C02NWN1K3MJ/p1695984191528039.

PR #3419, opened 2023-09-25
Implement `InsertAll` for `PostgreSQL`

### LINKED ISSUE #3395: Implement `InsertAll`
### What should be done?

Implement `InsertAll` method.
Add `InTransaction` helper.

### Where?

### Definition of Done

- method implemented;
- spot refactorings done.

chilagrow:
# Description

Closes #3395.

## Readiness checklist


internal/backends/postgresql/metadata/pool/transaction.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/jackc/pgx/v5/pgxpool"
+
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+	"github.com/FerretDB/FerretDB/internal/util/observability"
+)
+
+// InTransaction uses pool p and wraps the given function f in a transaction.
+//
+// If f returns an error or context is canceled, the transaction is rolled back.
+func InTransaction(ctx context.Context, p *pgxpool.Pool, f func(tx *pgx.Tx) error) (err error) {
```
chilagrow, line 30: This function takes pool as parameter. Unlike sqlite, it's not using `fsql` because existing `*pgxpool.Pool` functions are quite useful.

Other option is we could create a struct similar to `fsql.DB` and wrap all existing `*pgxpool.Pool` functions like `fsql.DB` does. Then we can add our custom function like `db.InTransaction`. Do we prefer that?

AlekSi: > Do we prefer that?

Let's keep it as it is now

internal/backends/postgresql/metadata/pool/transaction.go
```diff
@@ -0,0 +1,74 @@
...
+		// including `panic(nil)` that is problematic for tests too.
+		if done {
+			return
+		}
+
+		if err == nil {
+			err = lazyerrors.Errorf("transaction was not committed")
+		}
+
+		_ = pgTx.Rollback(ctx)
```
AlekSi, line 58: What happens when `ctx` is canceled? Does it cancel the rollback?

AlekSi (approved):

noisersup (approved):

rumyantseva (approved):
LGTM

internal/backends/postgresql/metadata/pool/transaction.go
chilagrow: It should rollback, and currently it doesn't.

AlekSi: So I guess we should use `context.WithoutCancel(ctx)` instead of `ctx` there.

But then what should we use for Commit?

chilagrow: I think we should use normal context for commit. If context is canceled, it won't commit. What do you think?

AlekSi: That sounds correct

chilagrow: Just have to make sure that rollback is executed. Thanks for catching this :hugs:

internal/backends/postgresql/metadata/registry_test.go
```diff
@@ -150,7 +150,9 @@ func TestCreateDrop(t *testing.T) {
 		t.Skip("skipping in -short mode")
 	}
 
-	t.Parallel()
+	// Do not run parallel, `initDBs()` searches all schema with ferretdb metadata table and
+	// including schema created by other tests that can be deleted before `initCollections()`
+	// is called would cause pg relation does not exist error.
```
chilagrow, line 155: See https://github.com/FerretDB/FerretDB/actions/runs/6309012677/job/17128119042?pr=3419

internal/backends/postgresql/metadata/pool/transaction.go
```diff
@@ -14,8 +14,25 @@
...
+	"github.com/FerretDB/FerretDB/internal/util/observability"
+)
+
+// InTransaction uses pool p and wraps the given function f in a transaction.
+//
+// If f returns an error or context is canceled, the transaction is rolled back.
+func InTransaction(ctx context.Context, p *pgxpool.Pool, f func(tx pgx.Tx) error) error {
+	defer observability.FuncCall(ctx)()
+
+	if err := pgx.BeginFunc(ctx, p, f); err != nil {
```
chilagrow, line 32: Actually this existing function is exactly what we need for transaction, thanks @AlekSi. I kept it within `InTransaction` so I can showcase the tests. Particularly for context cancel

internal/backends/postgresql/metadata/registry_test.go
chilagrow: This will fix second case of https://github.com/FerretDB/FerretDB/issues/3409

chilagrow: will be done in https://github.com/FerretDB/FerretDB/pull/3438

internal/backends/postgresql/metadata/pool/transaction_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+				//nolint:vet // need it for testing
+				panic(nil)
+			})
+			require.ErrorContains(t, err, "no rows in result set")
+		})
+
+		var res string
+		err = p.QueryRow(ctx, `SELECT s FROM t_test WHERE s = $1`, v).Scan(&res)
+		require.ErrorContains(t, err, "no rows in result set")
```
AlekSi, line 134: I think that's `pgx.ErrNoRows`

internal/backends/postgresql/metadata/pool/transaction.go
rumyantseva: Ah, this is really great!

rumyantseva (approved):
LGTM

internal/backends/postgresql/metadata/pool/transaction_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/FerretDB/FerretDB/internal/util/state"
+	"github.com/FerretDB/FerretDB/internal/util/testutil"
+)
+
+func TestInTransaction(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping in -short mode")
+	}
+
+	// do not run parallel, there is something weird about context canceling and parallel test
```
AlekSi, line 36: That's the only test in this package. There can't be something weird due to parallel tests because there are no other parallel tests in this package, but `t.Parallel()` does not affect other packages / test binaries.

internal/backends/postgresql/metadata/pool/transaction_test.go
```diff
@@ -0,0 +1,136 @@
...
+	require.NoError(t, err)
+
+	t.Cleanup(pp.Close)
+
+	p, err := pp.Get(username, password)
+	require.NoError(t, err)
+
+	t.Cleanup(p.Close)
+
+	_, err = p.Exec(ctx, `DROP TABLE IF EXISTS t_test; CREATE TABLE t_test(s TEXT);`)
```
AlekSi, line 55: I think we should use `t.Name()` to be able to track orphan table back to test

internal/backends/postgresql/metadata/pool/transaction_test.go
```diff
@@ -0,0 +1,136 @@
...
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		_, _ = p.Exec(ctx, `DROP TABLE IF EXISTS t_test`)
```
AlekSi, line 59: It should exist, and we should check the error

internal/backends/postgresql/metadata/pool/transaction_test.go
chilagrow: Yes that's correct, replaced it :pray:

internal/backends/postgresql/metadata/pool/transaction_test.go
chilagrow: Just made sure that it creates local instance of ctx within each subtest, it was previously using canceled context because it might be cancelled by other test

internal/backends/postgresql/metadata/pool/transaction_test.go
chilagrow: Good idea, applied it.

internal/backends/postgresql/metadata/pool/transaction_test.go
chilagrow: Added error check and removed IF EXISTS :pray:

AlekSi (commented):
There is a data race: https://github.com/FerretDB/FerretDB/actions/runs/6320745819/job/17163700348?pr=3419

internal/backends/postgresql/metadata/pool/transaction_test.go
```diff
@@ -0,0 +1,151 @@
...
+	if testing.Short() {
+		t.Skip("skipping in -short mode")
+	}
+
+	t.Parallel()
+
+	ctx := conninfo.Ctx(testutil.Ctx(t), conninfo.New())
+	username, password := conninfo.Get(ctx).Auth()
+
+	sp, err := state.NewProvider("")
```
AlekSi, line 42: `err` is shared across subtests

chilagrow: Thanks :pray: updated

AlekSi (approved):

rumyantseva (approved):
:shipit:

PR #3461, opened 2023-09-28
Add TODOs for capped collections

AlekSi:
# Description

Describe future work.

## Readiness checklist

### DIFF
--- a/internal/backends/database.go
+++ b/internal/backends/database.go
@@ -17,6 +17,7 @@ package backends
 import (
 	"context"
 
+	"github.com/FerretDB/FerretDB/internal/util/must"
 	"github.com/FerretDB/FerretDB/internal/util/observability"
 )
 
@@ -83,7 +84,14 @@ type ListCollectionsResult struct {
 
 // CollectionInfo represents information about a single collection.
 type CollectionInfo struct {
-	Name string
+	Name            string
+	CappedSize      int64 // TODO https://github.com/FerretDB/FerretDB/issues/3458
+	CappedDocuments int64 // TODO https://github.com/FerretDB/FerretDB/issues/3458
+}
+
+// Capped returns true if collection is capped.
+func (ci *CollectionInfo) Capped() bool {
+	return ci.CappedSize > 0 || ci.CappedDocuments > 0
 }
 
 // ListCollections returns information about collections in the database.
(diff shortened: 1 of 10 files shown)


internal/backends/database.go
```diff
@@ -83,7 +84,14 @@ type ListCollectionsResult struct {
...
 type CollectionInfo struct {
-	Name string
+	Name            string
+	CappedSize      int64 // TODO https://github.com/FerretDB/FerretDB/issues/3458
+	CappedDocuments int64 // TODO https://github.com/FerretDB/FerretDB/issues/3458
+}
+
+// Capped returns true if collection is capped.
+func (ci *CollectionInfo) Capped() bool {
+	return ci.CappedSize > 0 || ci.CappedDocuments > 0
```
chilagrow, line 94: According to this https://mongodb.com/docs/manual/core/capped-collections/#create-a-capped-collection size is always set.

> The size field is always required, even when you specify the max number of documents. MongoDB removes older documents if a collection reaches the maximum size limit before it reaches the maximum document count.

internal/backends/database.go
```diff
@@ -109,6 +124,10 @@ type CreateCollectionParams struct {
 func (dbc *databaseContract) CreateCollection(ctx context.Context, params *CreateCollectionParams) error {
 	defer observability.FuncCall(ctx)()
 
+	must.BeTrue(params.CappedSize >= 0)
```
chilagrow, line 127: I mean this will be validated in handler layer.

```suggestion
	must.BeTrue(params.CappedSize > 0)
```

chilagrow (approved):
Looks great, just some small detail I found out while playing with capped collection :hugs:

internal/backends/database.go
AlekSi: `CappedSize == 0` means that the collection is not capped.
There is no separate boolean flag because the zero value is useful already.

chilagrow: Got it, I misunderstood :see_no_evil:

internal/backends/database.go
AlekSi: It is always set for capped collections. If both `CappedSize` and `CappedDocuments` are zero, `Capped()` returns false because that collection is not capped

chilagrow: Thanks :+1:

internal/backends/database.go
```diff
@@ -109,6 +124,10 @@ type CreateCollectionParams struct {
...
+	must.BeTrue(params.CappedSize%256 == 0)
```
chilagrow, line 128: By the way 256 increment specific to Oplog only? For normal capped collection is this requirement too strict? :thinking:

AlekSi: That restriction is documented for all collections: https://www.mongodb.com/docs/manual/core/capped-collections/#create-a-capped-collection

chilagrow: Thanks :pray: So before reaching this point, the size will be rounded :+1:

Fashander (approved):

PR #3476, opened 2023-10-02
Process collection name param using `collection` tag

### LINKED ISSUE #2786: Process collection name parameter in tags
### What should be done?

Right now when we pass a parameters structure to the ExtractParams function we expect that the command field would contain a collection name. That's not true for several commands.
For example, the dataSize command should accept a namespace in the format database.collection.

```js
db.runCommand(
    {
         dataSize: "database.collection"
    }
)
```
The explain command expects the command document to be passed as the command parameter.

```js
db.runCommand(
    {
        explain: {
        "update": {...}
        }
    }
)
```

In order to process above mentioned commands we should add a tag that would specify that the field should contain a collection name.

### Where?

internal/handlers/commonparams/extract_params.go
internal/handlers/common

### Definition of Done
- add the `collection` tag to the `tagOptions`;
- use the `collection` tag to process the collection name in `ExtractParams`;
- update the following commands to use that tag: `count`, `delete`, `distinct`, `find`, `findAndModify`, `insert`, `update`;
- replace the `collection` field with the command name in above mentioned commands;
- unit tests added/updated - check the new tag `collection`;
- spot refactorings were done;

adetunjii:
# Description

Closes #2786.

## Readiness checklist


internal/handlers/commonparams/extract_params_test.go
```diff
@@ -28,7 +28,7 @@ import (
 func TestParse(t *testing.T) {
 	type allTagsThatPass struct { //nolint:vet // it's a test struct
 		DB           string          `ferretdb:"$db"`
-		Collection   string          `ferretdb:"collection"`
+		Collection   string          `ferretdb:"find,collection"`
```
chilagrow, line 31: Could you add a test case where `Collection` is `*type.Document` please? For example, `explain` is one of those case :hugs: 

> The explain command expects the command document to be passed as the command parameter.
```
db.runCommand(
    {
        explain: {
        "update": {...}
        }
    }
)
```

chilagrow (commented):
Thank you for your contribution :hugs: I have just one comment about adding a test for new logic.

internal/handlers/commonparams/extract_params.go
```diff
@@ -43,11 +43,7 @@ import (
 //   - `wholePositiveNumber` - provided value must be of types [int, long] and greater than 0;
 //   - `numericBool` - provided value must be of types [bool, int, long, double] and would be converted to bool;
 //   - `zeroOrOneAsBool` - provided value must be of types [int, long, double] with possible values `0` or `1`.
-//
-// Collection field processed in a special way. For the commands that require collection name
-// it is extracted from the command name.
-// If the field could have different types (e.g. `*types.Document` and `*types.Array`) then
-// the field must be of type `any`.
+//   - `collection` - field value holds the name of the collection and must be of type string.
```
chilagrow, line 46:
```suggestion
//   - `collection` - Collection field value holds the name of the collection and must be of type string. An error is returned if `collection` is not set, field value that is not collection name is not supported yet.
```

chilagrow (commented):
Thanks for your quick response and updating godoc! I see that newly introduced logic has a case where `collection` tag is not set. Let's cover that test case as well :hugs:

chilagrow:
By the way, when you are ready to request review again, you can press this button. It will notify reviewers :hugs: 
image

chilagrow (approved):
Great work! Thanks again for your contribution :hugs:

chilagrow (approved):

AlekSi (approved):

internal/handlers/commonparams/extract_params.go
```diff
@@ -87,12 +84,6 @@ func ExtractParams(doc *types.Document, command string, value any, l *zap.Logger
 
 		lookup := key
 
-		// If the key is the same as the command name, then it is a collection name.
-		// Depending on the driver, the key may be camel case or lower case for a collection name.
```
AlekSi, line 91: That comment was relevant

internal/handlers/common/findandmodify.go
```diff
@@ -44,7 +44,7 @@ const (
 // FindAndModifyParams represent parameters for the findAndModify command.
 type FindAndModifyParams struct {
 	DB                string          `ferretdb:"$db"`
-	Collection        string          `ferretdb:"collection"`
+	Collection        string          `ferretdb:"findAndModify,collection"`
```
AlekSi, line 47: Some drivers send this as `findandmodify`

internal/handlers/commonparams/extract_params.go
chilagrow: Thanks for capturing this. Fix made in https://github.com/FerretDB/FerretDB/pull/3494

PR #3485, opened 2023-10-03
Add stubs for `Collection.Compact`

AlekSi:
# Description

Refs #75.

## Readiness checklist

### DIFF
--- a/internal/backends/collection.go
+++ b/internal/backends/collection.go
@@ -43,6 +43,7 @@ type Collection interface {
 	Explain(context.Context, *ExplainParams) (*ExplainResult, error)
 
 	Stats(context.Context, *CollectionStatsParams) (*CollectionStatsResult, error)
+	Compact(context.Context, *CompactParams) (*CompactResult, error)
 
 	ListIndexes(context.Context, *ListIndexesParams) (*ListIndexesResult, error)
 	CreateIndexes(context.Context, *CreateIndexesParams) (*CreateIndexesResult, error)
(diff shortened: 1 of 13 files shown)


internal/backends/collection.go
```diff
@@ -211,9 +212,13 @@ func (cc *collectionContract) Explain(ctx context.Context, params *ExplainParams
 }
 
 // CollectionStatsParams represents the parameters of Collection.Stats method.
-type CollectionStatsParams struct{}
+type CollectionStatsParams struct {
+	Refresh bool // TODO https://github.com/FerretDB/FerretDB/issues/3483
```
chilagrow, line 216: Should setting refresh call `full vacuum` or is it related to calling `analyze`?

Which handler would have Refresh = false/true then?

chilagrow (commented):
Thank you, just one comment

internal/backends/collection.go
AlekSi: This PR is a subset of #3461; we need to merge this one first because a few linked issues are easy for contributors to pick up.

`Stats` with `Refresh: true` should call `ANALYZE`. But that's a different issue, yeah. Maybe that will not be needed.

chilagrow: Cool :+1:

So for now, every handler will have `Refresh: true` set since it calls `Analyze`.

chilagrow (dismissed):
Looks good to me :+1:

internal/backends/collection.go
AlekSi: No, for now, let's change nothing yet.

AlekSi: Actually, let me remove it from that PR…

chilagrow (approved):
LGTM :rocket:

PR #3491, opened 2023-10-03
Relax restrictions when _id is not the first field in projection

### LINKED ISSUE #2826: Relax restrictions when `_id` is not the first field in projection
### FerretDB version

v1.3.0

Currently we only relax projection restrictions when `_id` is the first field in the projection spec. We should allow it be at any index, just like MongoDB.

### Backend

Any.

### Environment

- OS: Ubuntu 22.10 x86_64
- Deployment: v1.3.0
- Deployment details: 23.0.3

### What did you do?

```sh
mongosh --quiet --eval 'db.foo.insertOne({a: 1}); db.foo.find({}, {a: 1, _id: 0});'
```

```
[ { a: 1 } ]
```

### What did you see instead?

```
MongoServerError: Cannot do exclusion on field _id in inclusion projection
```

princejha95:
# Description

Closes #2826.

## Readiness checklist


integration/query_projection_test.go
```diff
@@ -252,3 +252,35 @@ func TestQueryProjectionErrors(t *testing.T) {
...
+			}
+
+			t.Parallel()
+
+			require.NotNil(t, tc.filter, "filter should be set")
+			require.NotNil(t, tc.projection, "projection should be set")
+
+			res, err := collection.Find(ctx, tc.filter, options.Find().SetProjection(tc.projection))
+			defer res.Close(ctx)
+			assert.Nil(t, err)
```
chilagrow, line 283:
```suggestion
			require.NoError(t, err)
```

integration/query_projection_test.go
```diff
@@ -252,3 +252,35 @@ func TestQueryProjectionErrors(t *testing.T) {
...
+func TestQueryProjectionSuccess(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setup.Setup(t, shareddata.Scalars, shareddata.Composites)
+	for name, tc := range map[string]struct { //nolint:vet // used for testing only
+		filter     bson.D // required
+		projection any    // required
+
+		skip       string              // optional, skip test with a specified reason
+	}{
+		"QueryProjection_v": {
```
chilagrow, line 265: We use capitalization style for test case name https://github.com/FerretDB/FerretDB/blob/main/CONTRIBUTING.md#integration-tests-naming-guidelines

integration/query_projection_test.go
```diff
@@ -252,3 +252,35 @@ func TestQueryProjectionErrors(t *testing.T) {
...
+			if tc.skip != "" {
+				t.Skip(tc.skip)
+			}
+
+			t.Parallel()
+
+			require.NotNil(t, tc.filter, "filter should be set")
+			require.NotNil(t, tc.projection, "projection should be set")
+
+			res, err := collection.Find(ctx, tc.filter, options.Find().SetProjection(tc.projection))
```
chilagrow, line 281: Let's check `res` against expected value :hugs:

For the `QueryProjection_v` test case I suggest using filter such as `bson.D{{"_id", "int32"}}` so we can assert that expected value of a single document, that'll be `bson.D{{"v", int32(42)}}` for that filter.
FYI, document with `_id` of `int32` exists `shareddata.Scalars` collection,  but not in `shareddata.Composites` so we can remove it for this test:hugs:

chilagrow (commented):
Thanks for adding test :hugs: Let's check response against expected value :hugs:

integration/query_projection_test.go
```diff
@@ -252,3 +252,35 @@ func TestQueryProjectionErrors(t *testing.T) {
...
 }
+
+func TestQueryProjectionSuccess(t *testing.T) {
+	t.Parallel()
+	ctx, collection := setup.Setup(t, shareddata.Scalars, shareddata.Composites)
+	for name, tc := range map[string]struct { //nolint:vet // used for testing only
+		filter     bson.D // required
+		projection any    // required
+
+		skip       string              // optional, skip test with a specified reason
```
chilagrow, line 263: Please don't forget to run `task all`, there is format warning. For quicker check for just formatting and linting, you could also run `task` fmt` and `task lint`. They are run by `task all` too :pray:

integration/query_projection_test.go
princejha95: > Let's check `res` against expected value 🤗
> 
> For the `QueryProjection_v` test case I suggest using filter such as `bson.D{{"_id", "int32"}}` so we can assert that expected value of a single document, that'll be `bson.D{{"v", int32(42)}}` for that filter. FYI, document with `_id` of `int32` exists `shareddata.Scalars` collection, but not in `shareddata.Composites` so we can remove it for this test🤗

I thought of checking the res against expected value but the filter i was using was `bson.D{{"v", 42}}` due to which i was getting an array of results. Hence i thought of checking only the validity of projection by checking if the error is nil.

But i have a query. Should we depend upon the value of response ? Because the test case is used to check if Query Projection is working fine. Even a slightest change in response will lead to test case failure even though the projection works fine which is not what this test case reflects.

chilagrow: Yes we should check the response. `Scalar` collection is good one to use because it is frozen, so it won't be modified, see https://github.com/FerretDB/FerretDB/blob/main/integration/shareddata/scalars.go#L39

chilagrow (approved):
Looks great! Thank you for the contribution :hugs:

AlekSi (approved):

AlekSi (approved):

rumyantseva (approved):
LGTM! Thanks for contributing.

Issue #3458, opened 2023-09-28
Support `create` for capped collections


AlekSi:
### What should be done?

It should be possible to create capped collections with a different storage layout.

We want to introduce a hidden unique property to documents for locating them in the storage – RecordID. It is not a normal field, and it is stored separately. It is `types.Timestamp` that is ordered, both time- and count-based, and can be converted to both `int64` and `uint64` for native storage.

For PostgreSQL, tables for capped collections should be created with an additional column `_ferretdb_record_id bigint PRIMARY KEY`. It will store RecordID.

For SQLite, tables for capped collections should be created with an additional column `_ferretdb_record_id INTEGER PRIMARY KEY`. It will store RecordID. That format ensures that `_ferretdb_record_id` is used for [`rowid`](https://www.sqlite.org/lang_createtable.html#rowid).

(`_ferretdb_record_id` is not used in this issue but needs to be created for future tasks. For all backends, `_ferretdb_` is a reserved prefix.)

`Database.CreateCollection` method should store capped collection parameters in metadata.
`Database.ListCollections` should include that information in response.

`create` command handler should accept `capped`, `max`, and `size` parameters. The values of those parameters should be checked and passed to the `Database.CreateCollection` method.

`collStats` command handler should return `capped`, `max`, and `maxSize` fields with the actual value.

See:
* https://www.mongodb.com/docs/manual/core/capped-collections/
* https://www.mongodb.com/docs/manual/reference/command/create/

### Where?

* https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/sqlite/msg_create.go
* https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/sqlite/msg_collstats.go
* https://github.com/FerretDB/FerretDB/blob/main/internal/backends/postgresql/database.go
* https://github.com/FerretDB/FerretDB/blob/main/internal/backends/sqlite/database.go

See references to this issue.

### Definition of Done

- PostgreSQL and SQLite backend updated;
- command handlers updated;
- integration/compatibility tests added/updated;
- spot refactorings done;
- user documentation updated.

#### Not in the scope

- `collMod`, `convertToCapped`, any conversions between capped and non-capped collections;
- RecordID for non-capped collections;
- CRUD operations for capped collections.

PR #3495, opened 2023-10-04
Add `RecordID` to `types.Document`

AlekSi:
# Description

Refs #3458.

## Readiness checklist

### DIFF
--- a/internal/types/document.go
+++ b/internal/types/document.go
@@ -35,12 +35,18 @@ type document interface {
 
 // Document represents BSON document: an ordered collection of fields
 // (key/value pairs where key is a string and value is any BSON value).
+//
+// Data documents (that are stored in the backend) have a special RecordID property
+// that is not a field and can't be accessed by most methods.
+// It use used to locate the document in the backend.
 type Document struct {
-	fields []field
-	frozen bool
+	fields   []field
+	frozen   bool
+	recordID Timestamp
 }
 
 // field represents a field in the document.
+// RecordID is not a field.
 //
 // The order of field is like that to reduce a pressure on gc a bit, and make vet/fieldalignment linter happy.
 type field struct {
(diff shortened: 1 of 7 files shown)


internal/types/document.go
```diff
@@ -35,12 +35,18 @@ type document interface {
...
+//
+// Data documents (that are stored in the backend) have a special RecordID property
+// that is not a field and can't be accessed by most methods.
+// It use used to locate the document in the backend.
 type Document struct {
-	fields []field
-	frozen bool
+	fields   []field
+	frozen   bool
+	recordID Timestamp
```
noisersup, line 45: What are the benefits of using timestamp for this field? If we want to remove the oldest document then why can't we just use incremental id for such task?

Also, what if more than one documents were inserted in a single second? Which one should be removed first?

chilagrow: The uniqueness will need to be ensured by backend, using index possibly. If new insert is not unique, another timestamp can be re-generated and set, hence recordID can be set also on frozen document I think. So neither should be remove.

Incremental ID generation will depend heavily on each backend, or if we don't want to depend on backend, we need to fetch last incremental id, then compute a new one. So transaction + more code maybe involved, so I think using timestamp is easier :hugs:

chilagrow (approved):

internal/types/document.go
AlekSi: The index will be used, yes.

`types.Timestamp` includes both the time… stamp 🥁 and a counter. Even if the second is the same, the counter value will be different. It is not the same as `time.Time`.

There are cases when we will need to know Record ID before we insert the row to the table.

chilagrow: I see, so it's more guarantee of uniqueness than I initially read.

rumyantseva (approved):
LGTM!

PR #3497, opened 2023-10-04
Extract `ReservedPrefix` constant

AlekSi:
# Description

Just a small clean-up before we use it in more places.

## Readiness checklist

### DIFF
--- a/internal/backends/sqlite/metadata/metadata.go
+++ b/internal/backends/sqlite/metadata/metadata.go
@@ -33,7 +33,7 @@ import (
 
 const (
 	// DefaultColumn is a column name for all fields.
-	DefaultColumn = backends.ReservedPrefix + "_sjson"
+	DefaultColumn = backends.ReservedPrefix + "sjson"
 
 	// IDColumn is a SQLite path expression for _id field.
 	IDColumn = DefaultColumn + "->'$._id'"
--- a/internal/backends/sqlite/metadata/metadata.go
+++ b/internal/backends/sqlite/metadata/metadata.go
@@ -22,6 +22,7 @@ import (
 
 	"golang.org/x/exp/slices"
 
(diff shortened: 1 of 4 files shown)


internal/backends/sqlite/metadata/metadata.go
```diff
@@ -32,7 +33,7 @@ import (
 
 const (
 	// DefaultColumn is a column name for all fields.
-	DefaultColumn = "_ferretdb_sjson"
+	DefaultColumn = backends.ReservedPrefix + "_sjson"
```
rumyantseva, line 36: Some tests start failing due to this test. `DefaultColumn` used to be `_ferretdb_sjson`, but now it's `_ferretdb__sjson`. Should it remain `_ferretdb_sjson` or do we use double `_` on purpose (if so, tests need to be updated)?

internal/backends/sqlite/metadata/registry.go
```diff
@@ -42,7 +42,7 @@ const (
 	reservedTablePrefix = "sqlite_"
 
 	// SQLite table name where FerretDB metadata is stored.
-	metadataTableName = "_ferretdb_collections"
+	metadataTableName = backends.ReservedPrefix + "_collections"
```
rumyantseva, line 45: Same here: now `metadataTableName` has double `_` in the name.

internal/backends/sqlite/metadata/metadata.go
rumyantseva: P.S. I guess it's not on purpose as this change breaks back compatibility. Fun fact: integration tests would pass. That's why I avoid using constants in unit tests: sometimes constant-less tests help catch such things!

chilagrow (approved):

rumyantseva (approved):
:shipit:

PR #3504, opened 2023-10-05
Implement sort pushdown for PostgreSQL

### LINKED ISSUE #3415: Implement sort pushdown
### What should be done?

We should implement collection query sort pushdown as in `pgdb`.

Relevant integration tests should pass.

### Where?

### Definition of Done

- unit tests added/updated;
- spot refactorings done.

noisersup:
# Description

Closes #3415.

## Readiness checklist


internal/handlers/sqlite/msg_explain.go
```diff
@@ -83,6 +83,10 @@ func (h *Handler) MsgExplain(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 		qp.Filter = params.Filter
 	}
 
+	if h.EnableSortPushdown {
+		qp.Sort = params.Sort
+	}
```
chilagrow, line 88: We should update the `reply` document it currently returns `"sortingPushdown", false,`

internal/backends/postgresql/query.go
```diff
@@ -0,0 +1,281 @@
...
+		if err != nil {
+			if errors.Is(err, iterator.ErrIteratorDone) {
+				break
+			}
+
+			return "", nil, lazyerrors.Error(err)
+		}
+
+		// Skip sorting if there are more than one sort parameters
+		if order != 0 {
```
chilagrow, line 186: If this happens, we should set handler response to `"sortingPushdown", false,` right?

internal/backends/postgresql/collection.go
```diff
@@ -285,9 +326,9 @@ func (c *collection) Explain(ctx context.Context, params *backends.ExplainParams
 		return nil, lazyerrors.Error(err)
 	}
 
-	return &backends.ExplainResult{
-		QueryPlanner: must.NotFail(types.NewDocument("Plan", queryPlan)),
-	}, nil
+	res.QueryPlanner = queryPlan
+
+	return res, nil
```
chilagrow, line 331: I think we need to update `backend.ExplainResult` to have SortPushdown field.

Could you ensure to run postgresql integration tests related to sorting for `explain` command, please.

internal/backends/postgresql/query.go
```diff
@@ -0,0 +1,281 @@
...
+			return "", nil, nil
+		}
+
+		order, err = common.GetSortType(k, v)
```
chilagrow, line 190: Ideally sort parameter is validated in handler level, because this returns `NewCommandErrorMsgWithArgument`. See also https://github.com/FerretDB/FerretDB/blob/main/internal/backends/backends.go#L34

internal/backends/collection.go
```diff
@@ -71,6 +71,7 @@ func CollectionContract(c Collection) Collection {
 type QueryParams struct {
 	// TODO https://github.com/FerretDB/FerretDB/issues/3235
 	Filter *types.Document
+	Sort   *types.Document
```
chilagrow, line 74: Rather than passing not validated document, we could validate `*types.Document` in handler and then pass a struct here like it's done for Index. Then we won't need to return `NewCommandErrorMsgWithArgument` from postgresql implementation. :hugs:

chilagrow (commented):
Some comment related to sort pushdown part of the code :hugs:

internal/backends/postgresql/query.go
AlekSi: #3512

internal/handlers/sqlite/msg_explain.go
noisersup: Nice catch, I've tested the pushdown, but didn't check the explain. Fixed

chilagrow (commented):
Let me know what you think about comments :hugs:

internal/backends/collection.go
```diff
@@ -197,14 +199,17 @@ func (cc *collectionContract) DeleteAll(ctx context.Context, params *DeleteAllPa
 // ExplainParams represents the parameters of Collection.Explain method.
 type ExplainParams struct {
 	// TODO https://github.com/FerretDB/FerretDB/issues/3235
-	Filter *types.Document
+	Filter    *types.Document
+	SortKey   string
+	SortOrder *types.SortType
```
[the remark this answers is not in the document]
AlekSi: Yes, `types.SortType` should not be serializable. Also, no need to have `nil` as a distinct value

chilagrow (approved):
LGTM!

internal/handlers/sqlite/msg_find.go
```diff
@@ -71,6 +71,44 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+
+		var key string
+		var order types.SortType
+		var count int
+
+		for ; ; count++ {
+			var k string
+			var v any
+
+			k, v, err = iter.Next()
```
AlekSi, line 86: I think that's a good example where iterator makes code harder to understand. Checking `params.Sort.Len() == 1` would make this part much simpler

AlekSi (commented):
Linter fails

internal/handlers/sqlite/msg_explain.go
```diff
@@ -83,6 +85,44 @@ func (h *Handler) MsgExplain(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 		qp.Filter = params.Filter
 	}
 
+	if h.EnableSortPushdown {
```
rumyantseva, line 88: I think one more place to add this thing is `msg_aggregate` (see the old `msg_aggreagte` for pg).

noisersup: For msg_aggregate we don't support yet both filter and sort pushdown. It seems to me that in old pg handler the logic for deciding when to do pushdown and not was a bit more complex, so I'd suggest to implement those pushdowns in separate issue to hopefully merge this one before the weekend :pray:

rumyantseva: @noisersup let's reuse the issue #3520 then! could you describe there what needs to be done to support it in aggregation, please?

noisersup: Sure!

internal/handlers/sqlite/msg_explain.go
```diff
@@ -83,6 +85,43 @@ func (h *Handler) MsgExplain(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+		iter := params.Sort.Iterator()
+		defer iter.Close()
+
+		var key string
+		var order types.SortType
+
+		for {
```
AlekSi, line 95: `params.Sort.Len()` is either 1 or not 1. There is no need to iterate at all

chilagrow: Do you mean not use iterator at all and get values directly like this?

```
	k := params.Sort.Keys()[0]
	v := must.NotFail(params.Sort.Get(k))
```

AlekSi: or `.Values()[0]`

internal/handlers/sqlite/msg_explain.go
```diff
@@ -83,6 +83,32 @@ func (h *Handler) MsgExplain(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 		qp.Filter = params.Filter
 	}
 
+	// Skip sorting if there are more than one sort parameters
+	if h.EnableSortPushdown && params.Sort.Len() == 1 {
+		iter := params.Sort.Iterator()
```
AlekSi, line 88: Let's not use iterator if we don't iterate

internal/handlers/sqlite/msg_explain.go
chilagrow: True, that is simpler :+1:

@noisersup let's do that :hugs:

internal/handlers/sqlite/msg_explain.go
noisersup: What do we do if we want to just get the single key:value? Should we use params.Sort.Keys()[0] and params.Sort.Values()[0] instead?

AlekSi: https://github.com/FerretDB/FerretDB/pull/3504#discussion_r1360159515

chilagrow (approved):

AlekSi (approved):

PR #3508, opened 2023-10-06
Fix `_id` restriction in aggregation `$project` stage

### LINKED ISSUE #2826: Relax restrictions when `_id` is not the first field in projection

chilagrow:
# Description

Closes #2826.

## Readiness checklist

### DIFF
--- a/integration/aggregate_documents_test.go
+++ b/integration/aggregate_documents_test.go
@@ -25,6 +25,7 @@ import (
 	"go.mongodb.org/mongo-driver/mongo"
 
 	"github.com/FerretDB/FerretDB/integration/setup"
+	"github.com/FerretDB/FerretDB/integration/shareddata"
 )
 
 func TestAggregateAddFieldsErrors(t *testing.T) {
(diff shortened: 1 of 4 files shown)


chilagrow:
:wave: @princejha95 I also imported same fix to the aggregation `$project` stage :hugs:

AlekSi (approved):

princejha95:
@chilagrow I also think that in future, we should create a constants file which will store string constants like `_id` otherwise we will be using strings everywhere and we need to make change everywhere if there's a change in string value. What are your suggestions on this ? Can it be a improvement task ?

Thanks

rumyantseva:
@princejha95 could you give an example of what change in string value might be needed in the future, please?

rumyantseva (approved):
LGTM

noisersup (approved):

chilagrow:
> > @princejha95 could you give an example of what change in string value might be needed in the future, please?
> 
> What i meant was if we create a file like `MongoConstants.go` and define a constant like `MONGO_ID = "_id"`, then we can use `MONGO_ID` everywhere instead of `"_id"` as it might get written wrongly by mistake. So it's a good practice to keep all such constants which never gets changed inside a file and use them everywhere from the file itself. Generally it is considered a good practice.
> 
> Thanks

:wave: @princejha95 thanks for your comment. For a short one like `_id`, creating a constant for it doesn't add additional context. Also it's hard to mistype without tests failing somewhere. :hugs: So I don't think it's necessary for `_id`.

PR #3523, opened 2023-10-07
Fix invalid validation for `_id` field

### LINKED ISSUE #2804: Invalid validation for `_id` field
### FerretDB version

v1.8.0

### Backend

Any.

### Environment

Does not matter.

### What did you do?

```js
db.test.insertOne({_id: 1, v: {_id: ["foo", "bar"]}})`
```

```js
```

### What did you see instead?

```
MongoServerError: The '_id' value cannot be of type array
```

---

Our validation for the `_id` field incorrectly checks nested documents. Only the top-level `_id` field should be checked. See references to this issue in tests.

slavabobik:
# Description

Closes #2804.

## Readiness checklist


slavabobik:
I'm not so sure that this is a better solution, I have a couple of thoughts:

Instead of creating every time new boolean variable `topLevel` inside `validateData(..)` function, may be better add two helper function like 
```go
func (d *Document) validateTopLevelData() error { return d.validateData(true) }
func (d *Document) validateInnerData() error { return d.validateData(false) }
```
and inside `d.validateData(...)` func we could use it like:

```go
...
case *Document:
    err := value.validateInnerData
...
```
But this approach add complexity and not clarity


May be the better solution, is add in the `Document` struct boolean field `root`, but I'm not sure how easy to populate this field.

@AlekSi what do you think?

internal/types/document_validation.go
```diff
@@ -68,6 +68,12 @@ func (e *ValidationError) Code() ValidationErrorCode {
 // It replaces negative zero -0 with valid positive zero 0.
 // If the document is not valid it returns *ValidationError.
 func (d *Document) ValidateData() error {
+	topLevel := true
+	return d.validateData(topLevel)
```
AlekSi, line 72:
```suggestion
	return d.validateData(true)
```

internal/types/document_validation.go
```diff
@@ -104,7 +110,8 @@ func (d *Document) ValidateData() error {
 
 		switch value := value.(type) {
 		case *Document:
-			err := value.ValidateData()
+			topLevel := false
+			err := value.validateData(topLevel)
```
AlekSi, line 114:
```suggestion
			err := value.validateData(false)
```

internal/types/document_validation.go
```diff
@@ -124,7 +131,8 @@ func (d *Document) ValidateData() error {
 
 				switch item := item.(type) {
 				case *Document:
-					err := item.ValidateData()
+					topLevel := false
+					err := item.validateData(topLevel)
```
AlekSi, line 135:
```suggestion
					err := item.validateData(false)
```

AlekSi (commented):
I think the current approach is simple and good, it just needs a small clean-up. There is no need to shadow and redefine that variable

slavabobik:
Agree, my first attempt was like you suggested 😄

rumyantseva:
Thanks for contributing! It would also be nice to add integration tests for this case, so we ensure that the behavior is compatible with MongoDB. 

I think `shareddata.DocumentsDocuments` might be a good candidate. For example, we can add three more docs there - for array, document and regex. Maybe something like this:

```go
var DocumentsDocuments = &Values[primitive.ObjectID]{
	name: "DocumentsDocuments",
	data: map[primitive.ObjectID]any{
		{0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01}: bson.D{{"foo", int32(42)}},
		{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff}: bson.D{{"bar", bson.D{}}},
		{0x02, 0x02, 0x02, 0x02, 0x02, 0x02, 0x02, 0x02, 0x02, 0x02, 0x02, 0x02}: bson.D{{"_id", bson.A{int32(42), int32(42)}}},
		{0x03, 0x03, 0x03, 0x03, 0x03, 0x03, 0x03, 0x03, 0x03, 0x03, 0x03, 0x03}: bson.D{{"_id", bson.D{{"foo", "bar"}}}},
		{0x04, 0x04, 0x04, 0x04, 0x04, 0x04, 0x04, 0x04, 0x04, 0x04, 0x04, 0x04}: bson.D{{"_id", primitive.Regex{Pattern: "foo", Options: "i"}}},
	},
}
```

As these documents are inserted when the integration tests are run, there will be a check that the insertion was successful. 

Alternatively, please feel free to suggest some other tests.

rumyantseva (commented):
Also, please feel free to ping me if you need any help.

AlekSi (commented):
https://github.com/FerretDB/FerretDB/pull/3523#issuecomment-1752642495

slavabobik:
Hi @rumyantseva! thank you for your examples, I slightly adjust the last example about regex field. 
Can you please help me with **Go / postgresql 1/{1,2,3} / Run (pull_request)** test? I can't get it green, and I'm not sure why? 
also when I run `task all` these tests are not green

AlekSi:
It is not you – it's us :) Just ignore it. We plan to fix that soon

internal/types/document_validation.go
```diff
@@ -115,7 +120,7 @@ func (d *Document) ValidateData() error {
 				return err
 			}
 		case *Array:
-			if key == "_id" {
+			if isTopLevel && key == "_id" {
```
rumyantseva, line 123: @slavabobik codecov still marks some lines uncovered, it's because we don't have tests where we try to insert documents with top-level `_id` set as array or regex. A good candidate for such tests could be `TestInsertCommandErrors`, we can add examples where top-level `_id` field is filled with those invalid types and we know which error we expect. Personally, I would add such tests, but in this PR I leave it up to you to do it (if you have more time and wish, please feel free to do it, otherwise we can do it separately). 

What else we don't have is a test case where top-level `_id` is set as a document (a valid case, but probably rate in real life), but it's rather a separate story :)

rumyantseva (approved):
Thank you for contributing!

The changes look good to me (I manually tested the cases with valid/invalid top-level `_id` fields as we don't have integration tests for it).

integration/shareddata/composites.go
```diff
@@ -157,6 +157,9 @@ var DocumentsDocuments = &Values[primitive.ObjectID]{
 	data: map[primitive.ObjectID]any{
 		{0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01}: bson.D{{"foo", int32(42)}},
 		{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff}: bson.D{{"bar", bson.D{}}},
+		{0x02, 0x02, 0x02, 0x02, 0x02, 0x02, 0x02, 0x02, 0x02, 0x02, 0x02, 0x02}: bson.D{{"_id", bson.A{int32(42), int32(42)}}},
+		{0x03, 0x03, 0x03, 0x03, 0x03, 0x03, 0x03, 0x03, 0x03, 0x03, 0x03, 0x03}: bson.D{{"_id", bson.D{{"foo", "bar"}}}},
+		{0x04, 0x04, 0x04, 0x04, 0x04, 0x04, 0x04, 0x04, 0x04, 0x04, 0x04, 0x04}: bson.D{{"_id", bson.D{{"_id", primitive.Regex{Pattern: "foo", Options: "i"}}}}},
```
AlekSi, line 162: I'm afraid that's not the best place to add those cases – the key in this map is an `_id` already. The result – documents with two `_id`s, one of which is unused – is quite confusing.

I think it is better to add an ad-hoc test to basic_test.go

AlekSi (approved):

internal/types/document_validation.go
slavabobik: @rumyantseva I can add more tests in separate PR :) Do I need a separate task or just create a PR with tests?

rumyantseva: Let‘s have a separate PR only. A separate task is not needed. Thank you!

PR #3527, opened 2023-10-09
Implement `Collection.Compact` for PostgreSQL

### LINKED ISSUE #3484: Implement `Collection.Compact` for PostgreSQL
### What should be done?

PostgreSQL's backend `Collection.Compact` method should call [`VACUUM ANALYZE`](https://www.postgresql.org/docs/current/sql-vacuum.html) on the given collection/table.

When `Full` is true, `VACUUM FULL ANALYZE` should be run instead.

If the database or collection does not exist, `ErrorCodeDatabaseDoesNotExist` or `ErrorCodeCollectionDoesNotExist` should be returned (that's already checked by the contact and backend test).

### Where?

* https://github.com/FerretDB/FerretDB/blob/main/internal/backends/postgresql/collection.go
* https://github.com/FerretDB/FerretDB/blob/main/internal/backends/collection_test.go

### Definition of Done

- PostgreSQL backend updated;
- `Skip` removed from the test.

alhridoy:
Implements the `Collection.Compact` method for the PostgreSQL backend (closes #3484).

In this PR, I have updated the `Compact` method in `collection.go` to call `VACUUM ANALYZE` on the given collection/table. If the `Full` parameter is true, `VACUUM FULL ANALYZE` is run instead. I have also updated the corresponding test in `collection_test.go` and ensured that it passes with the new implementation.


implement-collection-compact-postgres
AlekSi, line 1: Please remove that file

internal/backends/collection_test.go
```diff
@@ -222,9 +222,6 @@ func TestCollectionStats(t *testing.T) {
 }
 
 func TestCollectionCompact(t *testing.T) {
-	t.Skip("https://github.com/FerretDB/FerretDB/issues/3484")
-	t.Skip("https://github.com/FerretDB/FerretDB/issues/3469")
```
AlekSi, line 226: I think only one of those skips should be removed

AlekSi (commented):
Please click the CLA button in this message: https://github.com/FerretDB/FerretDB/pull/3527#issuecomment-1752497827

AlekSi:
@alhridoy Please re-request the review once it is ready. Please also click that CLA button

AlekSi:
@alhridoy unfortunately, we can't accept your PR without CLA signed

Issue #89, opened 2021-11-24
Support count function


thuan1412:
The project currently does not support the `count` function.

![image](https://user-images.githubusercontent.com/36019052/143276633-8759ead0-d36f-4df2-9b55-c02a16fc71aa.png)

OpenSauce:
I can pick this up, unless you want to give it a try @thuan1412 ?

thuan1412:
@OpenSauce yeh, I gonna try to resolve this issue this weekend. I think it's a good opportunity for me to have much understanding of the project.

dstpierre:
Hey same here, I'm trying to test mangodb for my project and I'm getting this:

`(NotImplemented) unhandled msg "aggregate"`

It's also for a "Count" command, but not receiving same error msg as above.

This is very promising project and I'd be happy to help.

dstpierre:
This is the line generating the `aggregate` error, it's a count. Maybe the Go driver of Mongo uses `aggregate` in its internal:

```go
count, err := db.Collection("accounts").CountDocuments(ctx, bson.M{"email": email})
```

PR #3559, opened 2023-10-11
Implement `compact` command

### LINKED ISSUE #3466: Implement `compact` command
### What should be done?

After #3484 and #3469, we should implement that command in the handler: https://www.mongodb.com/docs/manual/reference/command/compact/

The implementation should:
* run `Collection.Stats` method to get the current statistics;
* run `Collection.Compact` to compact it;
* run `Collection.Stats` method again to get the new statistics;
* log the difference between all metrics, and return the relevant one as `bytesFreed`.

If `force` is true, `Compact` should be called with `Full: true`. (There is a semantic difference between `force` and `full`, but that's fine.)

Errors (missing database, missing collection) should be returned as in MongoDB.

### Where?

* https://github.com/FerretDB/FerretDB/tree/main/internal/handlers/sqlite

Old `pgdb` handler should not be updated (or updated with a stub if required).

### Definition of Done

- handler updated;
- spot refactorings done;
- user documentation updated.

adetunjii:
Closes #3466.

## Readiness checklist


internal/handlers/hana/msg_compact.go
```diff
@@ -0,0 +1,27 @@
...
+package hana
+
+import (
+	"context"
+
+	"github.com/FerretDB/FerretDB/internal/util/must"
+	"github.com/FerretDB/FerretDB/internal/wire"
+)
+
+// MsgCompact implements HandlerInterface.
```
princejha95, line 24: I think the comment is wrong. Isn't it be `Handler implements Interface`. Also can you please add `var _ Interface = (*Handler)(nil)` to provide compile-time check in order to make sure Handler implements the desired interface ?

Thanks

internal/handlers/handlers.go
```diff
@@ -58,6 +58,9 @@ type Interface interface {
 	// MsgCollStats returns storage data for a collection.
 	MsgCollStats(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error)
 
+	// MsgCompact rewrites and degragments all the data and indexes in a collection.
```
princejha95, line 61: spelling of `defragments is wrong`

adetunjii: I'll correct that thanks.

internal/handlers/sqlite/msg_compact.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+// MsgCompact implements HandlerInterface.
+func (h *Handler) MsgCompact(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
+	document, err := msg.Document()
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+
+	common.Ignored(document, h.L, "comment")
```
princejha95, line 37: I will suggest not to use random strings like `comment` in between. It will be better if you can create some constant variable having value as `comment` and use that everywhere in the code. This will limit the repetition of writing same string again and again everywhere

internal/handlers/sqlite/msg_compact.go
```diff
@@ -0,0 +1,128 @@
...
+
+	command := document.Command()
+
+	dbName, err := common.GetRequiredParam[string](document, "$db")
+	if err != nil {
+		return nil, err
+	}
+
+	collection, err := common.GetRequiredParam[string](document, command)
```
princejha95, line 46: Can we try to fetch the collection after we have fetched db successfully otherwise it will be unnecessary to get collection here. It is a good practice to keep the declaration of variable closer to its usage.

internal/handlers/sqlite/msg_compact.go
```diff
@@ -0,0 +1,128 @@
...
+				command,
+			)
+		}
+
+		return nil, lazyerrors.Error(err)
+	}
+
+	var forceParam any
+
+	if forceParam, err = document.Get("force"); err != nil {
```
princejha95, line 79: Please create a constant variable to store `force`

internal/handlers/sqlite/msg_compact.go
```diff
@@ -0,0 +1,128 @@
...
+		return nil, err
+	}
+
+	force, ok := forceParam.(bool)
+	if !ok {
+		force = false
+	}
+
+	statsBefore, err := c.Stats(ctx, new(backends.CollectionStatsParams))
+	if backends.ErrorCodeIs(err, backends.ErrorCodeCollectionDoesNotExist) {
```
princejha95, line 89: please include `if err != nil` before line #89

AlekSi:
As a side note – it always better to start a new branch from the current state of the `main` branch to avoid this:

CleanShot 2023-10-11 at 21 30 10@2x

internal/handlers/commoncommands/msg_listcommands.go
```diff
@@ -63,6 +63,10 @@ var Commands = map[string]command{
 		Help:    "Returns storage data for a collection.",
 		Handler: handlers.Interface.MsgCollStats,
 	},
+	"compact": {
+		Help:    "Rewrites and defragments all data and indexes in a collection",
```
AlekSi, line 67: We should not copy descriptions as-is from MongoDB

internal/handlers/hana/msg_compact.go
AlekSi: > I think the comment is wrong. Isn't it be Handler implements Interface.

This method allows `*Handler` to implement `HandlerInterface`. This comment is also consistent with comments for all other methods.

> Also can you please add var _ Interface = (*Handler)(nil) to provide compile-time check in order to make sure Handler implements the desired interface ?

We already do that in `hana.go`.

internal/handlers/sqlite/msg_compact.go
AlekSi: I don't think we should do that. `comment` is an argument of `compact` command. The fact that many commands have the argument with the same name is not important – they are different arguments for different commands.

internal/handlers/sqlite/msg_compact.go
AlekSi: It is even better practice to do non-I/O code before I/O code. If the collection name is not present, there is no need to call `h.b.Database(dbName)` at all.

And again – this code is quite consistent with other methods. I don't think we should make one method stand out.

internal/handlers/sqlite/msg_compact.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	db, err := h.b.Database(dbName)
+	if err != nil {
+		if backends.ErrorCodeIs(err, backends.ErrorCodeDatabaseNameIsInvalid) {
+			return nil, commonerrors.NewCommandErrorMsgWithArgument(
+				commonerrors.ErrInvalidNamespace,
+				fmt.Sprintf("Invalid namespace specified '%s.%s'", dbName, collection),
```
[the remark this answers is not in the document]
AlekSi: That error message should be the same as in MongoDB. Integration tests will check that.

internal/handlers/sqlite/msg_compact.go
AlekSi: Please don't :)

internal/handlers/sqlite/msg_compact.go
AlekSi: Why do you think so?

internal/handlers/sqlite/msg_compact.go
```diff
@@ -0,0 +1,128 @@
...
+	var forceParam any
+
+	if forceParam, err = document.Get("force"); err != nil {
+		return nil, err
+	}
+
+	force, ok := forceParam.(bool)
+	if !ok {
+		force = false
+	}
```
AlekSi, line 86: That should be handled by `commonparams.GetBoolOptionalParam`

AlekSi (commented):
I think it is almost done. Great job!

AlekSi:
@adetunjii any news? Anything we could help you with? Also, could we add a basic integration test?

chilagrow:
:wave: @adetunjii I added basic integration tests to the implemented feature :hugs:

internal/handlers/handlers.go
chilagrow: This comment was updated.

chilagrow (approved):

adetunjii:
@chilagrow Thanks. I'll just pull the changes and resolve the conflicts on my end. I'd already worked on the fixes

AlekSi:
@adetunjii I think the PR is good enough already. If you want to change something, please do so today; we are going to merge that PR tomorrow.

adetunjii:
No. I’m not adding anything. Looks good to me. Thanks

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

AlekSi (approved):

AlekSi:
Thank you!

PR #3563, opened 2023-10-12
Call `ANALYZE` less often

### LINKED ISSUE #3518: Call `ANALYZE` less often
### What should be done?

PostgreSQL and SQLite backends call `ANALYZE` on the whole database when `dbStats` and `collStats` commands are called. That's slow and not necessary in real usage.

Let's call `ANALYZE` only when `Database.Stats` and `Collection.Stats` are called with `Refresh: true`.  `dbStats` and `collStats` should call those methods with `Refresh: true` for now (to do less in this issue), but we might change that in the future.

Additionally, we should call `ANALYZE` only on a given list of collections (and given PostgreSQL schema), not on the whole database.

### Where?

See references to this issue.

### Definition of Done

- new SQLite handler updated;
- new PostgreSQL and SQLite backends updated;
- spot refactorings done.

Aditya1404Sal:
# Description

Currently i've made changes to the postgreSql and sqlite backend , I need help to make changes to ANALYZE to call on a given list of collection

Closes #3518.

## Readiness checklist


internal/backends/postgresql/backend.go
```diff
@@ -128,8 +128,9 @@ func (b *backend) ListDatabases(ctx context.Context, params *backends.ListDataba
 		if err != nil {
 			return nil, lazyerrors.Error(err)
 		}
-
-		stats, err := db.Stats(ctx, new(backends.DatabaseStatsParams))
+		dbStatParams := new(backends.DatabaseStatsParams)
+		dbStatParams.Refresh = true
```
AlekSi, line 132: Why we do that?

internal/backends/postgresql/collection.go
```diff
@@ -312,42 +312,45 @@ func (c *collection) Explain(ctx context.Context, params *backends.ExplainParams
 
 // Stats implements backends.Collection interface.
 func (c *collection) Stats(ctx context.Context, params *backends.CollectionStatsParams) (*backends.CollectionStatsResult, error) {
-	p, err := c.r.DatabaseGetExisting(ctx, c.dbName)
-	if err != nil {
-		return nil, lazyerrors.Error(err)
-	}
+	if params.Refresh {
```
AlekSi, line 315: If `params.Refresh` is true, we should call [`ANALYZE`](https://www.postgresql.org/docs/current/sql-analyze.html) on that collection/table, then calculate stats as before

internal/backends/postgresql/collection.go
```diff
@@ -312,42 +312,45 @@ func (c *collection) Explain(ctx context.Context, params *backends.ExplainParams
...
-	}, nil
+		return &backends.CollectionStatsResult{
+			CountObjects:   stats.countRows,
+			CountIndexes:   stats.countIndexes,
+			SizeTotal:      stats.sizeTables + stats.sizeIndexes,
+			SizeIndexes:    stats.sizeIndexes,
+			SizeCollection: stats.sizeTables,
+		}, nil
+	}
+	return &backends.CollectionStatsResult{}, nil
```
AlekSi, line 353: `Refresh: false` does not mean "do nothing and return empty structure"

AlekSi (changes requested):

internal/backends/postgresql/backend.go
Aditya1404Sal: It was a temporary change , to call dbStat and CollStat with refresh: true.

AlekSi: I think it is better to add to this test: https://github.com/FerretDB/FerretDB/blob/2d57d3ff41bc6cf09a6d82d71424bee864b520e7/internal/backends/collection_test.go#L136

internal/backends/postgresql/collection.go
Aditya1404Sal: So the stats func should continue its execution and remain unchanged,  should i be modifying the `collectionStat` function and add a parameter, for the params.refresh to add checks locally where the `ANALYZE` is called? 
I will add a commit for these changes

internal/backends/postgresql/collection.go
Aditya1404Sal: I apologise for the mistake, I'll make changes.

internal/backends/postgresql/collection.go
AlekSi:
```go
// get p, coll as before

if params.Refresh {
  // call ANALYZE
}

stats, err := collectionsStats(...
```

internal/backends/postgresql/backend.go
Aditya1404Sal: I understand now , thank you 🙂

internal/backends/postgresql/database.go
```diff
@@ -146,7 +146,13 @@ func (db *database) Stats(ctx context.Context, params *backends.DatabaseStatsPar
 		return nil, lazyerrors.Error(err)
 	}
 
-	stats, err := collectionsStats(ctx, p, db.name, list)
+	stats, placeholders, arg, err := collectionsStats(ctx, p, db.name, list)
+	if params.Refresh {
```
AlekSi, line 150: We should call `ANALYZE schema.table1, schema.table2` before we get stats

internal/backends/postgresql/postgresql.go
```diff
@@ -80,6 +71,16 @@ func collectionsStats(ctx context.Context, p *pgxpool.Pool, dbName string, list
 	// initialization fork https://www.postgresql.org/docs/current/storage-init.html,
 	// free space map https://www.postgresql.org/docs/current/storage-fsm.html and
 	// TOAST https://www.postgresql.org/docs/current/storage-toast.html.
+
```
AlekSi, line 74: The code that calculates stats should be there, not in `AnalyzeColl`. That looks like a bad merge

internal/backends/postgresql/postgresql.go
```diff
@@ -80,6 +71,16 @@ func collectionsStats(ctx context.Context, p *pgxpool.Pool, dbName string, list
...
+	return &s, placeholders, args, nil
+}
+
+func AnalyzeColl(ctx context.Context, p *pgxpool.Pool, s stats, placeholders []string, args []any) error {
+	var err error
+	q := `ANALYZE`
+	if _, err = p.Exec(ctx, q); err != nil {
+		return lazyerrors.Error(err)
+	}
 	q = fmt.Sprintf(`
```
AlekSi, line 84: That code should be in collectionsStats

Aditya1404Sal: But shouldn't the `ANALYZE` be isolated from the `collectionStats` method, because as long as the `ANALYZE` cmd exists as `q` in the `collectionStats` , it will be called for the same number of times as before and adding the `params.Refresh : true` check would be pointless ? i'm having trouble understanding
```
if params.Refresh{
     //Calls ANALYZE
}
stats , err := collectionStats(........) // This will call ANALYZE as well
```

AlekSi: > collectionStats(........) // This will call ANALYZE as well

`collectionStats` should not call `ANALYZE`

AlekSi: And that's exactly why we work on the issue titled "Call `ANALYZE` _less often_" :)

internal/backends/postgresql/postgresql.go
```diff
@@ -41,17 +41,26 @@ type stats struct {
 // collectionsStats returns statistics about tables and indexes for the given collections.
 //
 // If the list of collections is empty, then stats filled with zero values is returned.
-func collectionsStats(ctx context.Context, p *pgxpool.Pool, dbName string, list []*metadata.Collection) (*stats, error) {
-	if len(list) == 0 {
-		return new(stats), nil
-	}
-
+func collectionsStats(ctx context.Context, p *pgxpool.Pool, dbName string, list []*metadata.Collection, val bool) (*stats, error) {
```
AlekSi, line 44: Why argument is called `val`?

internal/backends/postgresql/postgresql.go
```diff
@@ -41,17 +41,26 @@ type stats struct {
...
 	var err error
+	if len(list) == 0 && val {
```
AlekSi, line 46: The previous logic was correct, I think. `collectionStats` is always called with a list of collections. If the list is empty, then there are no collections in that database.

internal/backends/postgresql/postgresql.go
```diff
@@ -41,17 +41,26 @@ type stats struct {
...
+		if _, err = p.Exec(ctx, q); err != nil {
+			return nil, lazyerrors.Error(err)
+		}
+
+		return new(stats), nil
+	}
+	if val {
+		var placeholder []string
+		for _, c := range list {
+			placeholder = append(placeholder, c.TableName)
```
AlekSi, line 58: Placeholder is `$1`, `$2`, etc. That slices is not a slice of placeholders.

Also, we should add PostgreSQL schema names and escape identifiers properly using `pgx.Identifier`.

internal/backends/postgresql/postgresql.go
Aditya1404Sal: I wasn't able to think of a proper name for the bool arguement, so temporarily kept it as val (short for value). if you could provide a sugesstion, that would be great.

internal/backends/postgresql/postgresql.go
Aditya1404Sal: I will make necessary changes.

internal/backends/postgresql/postgresql.go
AlekSi: `refresh`?

internal/backends/postgresql/postgresql.go
Aditya1404Sal: Like so ?
```
if refresh {
		var identifiers []string
		var placeholder metadata.Placeholder
		placeholders := make([]string, len(list))
		args := []any{}
		for _, c := range list {
			tableName := pgx.Identifier{dbName, c.TableName}.Sanitize()
			identifiers = append(identifiers, tableName)
		}
		for i,identifier := range identifiers{
			placeholders[i] = placeholder.Next()
			args = append(args, identifier)
		}
		q := fmt.Sprintf(`ANALYZE %s`, strings.Join(placeholders, ", "))
		if _, err := p.Exec(ctx, q, args... ); err != nil {
		}
	}
```

Aditya1404Sal: @AlekSi ??

AlekSi: As for `pgx.Identifier` – writing and pushing a code would allow us to review on GitHub with per-line comments, etc. It is much harder to review code written in the comment itself

Aditya1404Sal: Will push a commit with these changes ASAP. Thank you for resolving my doubt !

chilagrow (dismissed):
Thanks for your contribution. I push commits for minor cleanup :hugs:

internal/handlers/sqlite/msg_listdatabases.go
```diff
@@ -73,7 +74,7 @@ func (h *Handler) MsgListDatabases(ctx context.Context, msg *wire.OpMsg) (*wire.
 			continue
 		}
 
-		stats, err := db.Stats(ctx, nil)
+		stats, err := db.Stats(ctx, &backends.DatabaseStatsParams{Refresh: true})
```
AlekSi, line 77: @chilagrow the issue is about executing `ANALYZE` less often, but after this PR we will execute as often as before :) We should not use `Refresh: true` there, and also probably in `aggregate`

internal/backends/sqlite/sqlite.go
```diff
@@ -48,13 +48,29 @@ type stats struct {
...
-	if _, err = db.ExecContext(ctx, q); err != nil {
-		return nil, lazyerrors.Error(err)
+	if refresh {
+		queries := make([]string, len(list))
+		for i, c := range list {
+			queries[i] = fmt.Sprintf("ANALYZE %q;", c.TableName)
+		}
+
+		// SQLite ANALYZE does not allow multiple tables as arguments, hence
+		// build multiple queries such as `ANALYZE 'table1'; ANALYZE 'table2';`.
```
AlekSi, line 69: @chilagrow are you sure that actually works? Is there a way to check that?

chilagrow: I added tests for error handling.
- `ANALYZE 'table1'; ANALYZE 'table2';`
- `ANALYZE 'non-existent'; ANALYZE 'table2';`
- `ANALYZE 'table1'; ANALYZE 'non-existent';`

But we both also want to know that if `ANALYZE 'table1'` and `ANALYZE 'table2'` has been executed and updated stats. Unfortunately `SQLite` seems to have updated stats without calling 

`

internal/handlers/sqlite/msg_listdatabases.go
chilagrow: That's right. Thanks for this :pray:

chilagrow: If we don't call `ANALYZE` on `aggregate`, postgresql returns negative count of document. I suppose that's okay?

```
    --- FAIL: TestAggregateCompatCollStats/CountAndStorageStats (0.09s)
        --- FAIL: TestAggregateCompatCollStats/CountAndStorageStats/TestAggregateCompatCollStats_ArrayDocuments (0.08s)
            aggregate_stats_compat_test.go:131: 
                        Error Trace:    /home/chi/go/src/github.com/chilagrow/FerretDB/integration/aggregate_stats_compat_test.go:131
                        Error:          Not equal: 
                                        expected: int32(4)
                                        actual  : int64(-1)
                        Test:           TestAggregateCompatCollStats/CountAndStorageStats/TestAggregateCompatCollStats_ArrayDocuments
```

AlekSi: > postgresql returns negative count of document

It is not PostgreSQL, it is how we use it in tests. Under normal operations, PostgreSQL refreshes stats in the background itself. In our tests, we don't give it a chance to do so. Running `ANALYZE` everywhere makes tests pass – and hurts the real usage. So, in general, we should not call `ANALYZE` ever*†‡. Exceptions include the `validate` command. So let's use it in test.

internal/backends/sqlite/sqlite.go
AlekSi: > I added tests for error handling and it seems to work

The problem with running `QUERY 1; QUERRY 2` in a single exec is that we don't know without going deep into `database/sql`'s driver code if it works correctly. Many drivers may run only the first or only the last statement. 

Let's run them one-by-one to be sure

chilagrow: Okay, let me update it to send multiple queries :+1:

AlekSi: @chilagrow edited ^

chilagrow: Okay :+1:

internal/handlers/sqlite/msg_listdatabases.go
chilagrow: So in our tests, we can use `validate` to update the stats. And we remove `refresh: true` from all other handler correct?

AlekSi: From the issue description:
> `dbStats` and `collStats` should call those methods with `Refresh: true` for now (to do less in this issue)

I think changing those commands will take more time with updating tests, so let's not do it right now

chilagrow: Yes sure, as description says. We leave `dbStats` and `collStats` with refresh true. I will fix other handler tests by calling `validate` prior. :+1:

integration/aggregate_stats_compat_test.go
```diff
@@ -81,6 +81,21 @@ func TestAggregateCompatCollStats(t *testing.T) {
...
 				compatCollection := compatCollections[i]
 				t.Run(targetCollection.Name(), func(t *testing.T) {
+					require.Equal(t, compatCollection.Name(), targetCollection.Name())
+
+					// call validate to updated statistics
+					command := bson.D{{"validate", targetCollection.Name()}}
+					targetErr := targetCollection.Database().RunCommand(ctx, command).Err()
+					compatErr := compatCollection.Database().RunCommand(ctx, command).Err()
+
+					if targetErr != nil {
```
AlekSi, line 91: Is it possible in this test?

AlekSi (approved):

integration/aggregate_stats_compat_test.go
chilagrow: For non existent collection added by `AddNonExistentCollection` flag, validate fails for both compat and target.

chilagrow (approved):

Issue #3459, opened 2023-09-28
Support capped collection clean-up


AlekSi:
### What should be done?

After #3466, #3467, and #3498, we should add support for cleaning up capped collections.

Let's start a background goroutine in the handler that would do the following every minute (configurable with a test flag) if `EnableOplog` is `true`:
* iterate over a list of databases;
* iterate over a list of capped collections in those databases;
* run `Collection.Stats` on the current collection;
* if stats are lower than capped collection parameters, continue to the next collection;
* query 10% (configurable with a test flag) RecordIDs of documents in this collection (with `Limit` and `OnlyRecordIDs` set);
* delete them;
* compact that collection.

The last step essentially involves the calling of the `compact` command with `force: false` (#3466). We should not call that command handler directly but refactor it to make the shared code callable from the handler and our background goroutine.

Databases and collections could disappear during the iteration; that should be handled.

The goroutine should not `panic` and not `recover`.
We should add basic logging and metrics for that background goroutine.

Most of the logic should already be tested. No new automated tests are necessary.

### Where?

The handler. All backend changes should already be done by that point.

### Definition of Done

- handlers updated;
- all backends tested manually;
- spot refactorings done;
- user documentation checked and updated.

slavabobik:
Hi @AlekSi! Can I grab this issue? As I can see dependent tasks are mostly done(https://github.com/FerretDB/FerretDB/issues/3498 has opened PR). Or feel free please to suggest more suitable task on your opinion

AlekSi:
@slavabobik feel free to pick up this task, just not try to do everything in a single PR

AlekSi:
@slavabobik any news? Anything we could help you with?

slavabobik:
@AlekSi Hi! sorry for the delay, I'm started work on this only today, and I'm try to keep updates more frequently

rumyantseva:
Commands for manual testing (slow but good enough for testing purposes). Create a capped collection and insert docs:

```
mongosh --port=27017 --eval="db.createCollection('coll', {capped: true, size: 1000, max: 50})" cleanup

for in in (seq 100)
   mongosh --port=27017 --eval="db.coll.insert({data: 'document $i'})" cleanup
end
```

Check metrics and logs to ensure that cleanup goroutine is running.

PR #3580, opened 2023-10-16
Implement limit pushdown for PostgreSQL

### LINKED ISSUE #3416: Implement limit pushdown
### What should be done?

We should implement collection query limit pushdown as in `pgdb`.

Relevant integration tests should pass.

### Where?

### Definition of Done

- unit tests added/updated;
- spot refactorings done.

noisersup:
# Description

Closes #3416.

## Readiness checklist


internal/handlers/common/params.go
```diff
@@ -100,6 +100,26 @@ func AssertType[T types.Type](value any) (T, error) {
...
+			`BSON field '%s' is the wrong type '%s', expected type '%s'`,
+			"limit", commonparams.AliasFromType(v), commonparams.AliasFromType(res),
+		)
+
+		return res, commonerrors.NewCommandErrorMsgWithArgument(commonerrors.ErrTypeMismatch, msg, "limit")
+	}
+
+	return res, nil
+}
+
```
noisersup, line 122: Helper needed to fix a bug (the msg_explain didn't take int values, but only longs, when msg_find and mongodb accepted all whole values).

chilagrow: Could you add a test for it please? :hugs:

internal/handlers/sqlite/msg_find.go
```diff
@@ -80,6 +80,32 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
 		qp.Filter = params.Filter
 	}
 
+	// Skip sorting if there are more than one sort parameters
```
chilagrow, line 83: Why is there no change related related to limit? 
Shouldn't it use newly added `GetLimitParam()`, also some test should fail right?

internal/handlers/sqlite/msg_explain.go
```diff
@@ -83,6 +83,34 @@ func (h *Handler) MsgExplain(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+			return nil, err
+		}
+
+		qp.Sort = &backends.SortField{
+			Key:        k,
+			Descending: order == types.Descending,
+		}
+	}
+
+	qp.Limit = params.Limit
```
chilagrow, line 112: Why is logic related to limit pushdown omitted?
https://github.com/FerretDB/FerretDB/blob/093be9878f3c6b358029648c8c57d56102460ed2/internal/handlers/pg/msg_explain.go#L70-L76

internal/handlers/sqlite/msg_find.go
noisersup: `GetLimitParams` is used from `common/GetExplainParams()`.

chilagrow: I think naming is confusing, exported `GetLimitParams` parameter indicates it should be use for limit parameters. We should change godoc and function name.

chilagrow: My question is why is there no limit pushdown related change in this file? `msg_find.go` should set query limit parameter right? :thinking: Could we add a test for it if it's easy to miss?

AlekSi (commented):
image

noisersup:
@AlekSi Not only conflicts, there are failing tests and I'm tweaking that

internal/backends/postgresql/collection.go
```diff
@@ -101,6 +101,11 @@ func (c *collection) Query(ctx context.Context, params *backends.QueryParams) (*
 		args = append(args, sortArgs...)
 	}
 
+	if params.Limit != 0 {
+		q += fmt.Sprintf(` LIMIT %s`, placeholder.Next())
```
AlekSi, line 105: Please port that basic logic to the SQLite, too. We will not use it for basic queries with filter and sort, but we will need to for #3459

AlekSi (commented):
https://github.com/FerretDB/FerretDB/pull/3580#pullrequestreview-1682555926

(that seems to be a new GitHub bug – reviews without the final review comment are not marked as reviews)

internal/handlers/common/params.go
chilagrow: This comment isn't addressed.

chilagrow: Could you add a test for int please. New logic introduction without test is not clear.

Also does this need to be a common function at all? The thing is `GetLimitParam` is not accurate description of this function as far as I understand, it cannot be used for getting limit param in find. :hugs:

chilagrow (commented):
There are failing tests

rumyantseva (dismissed):
The code looks good to me, but I agree with Chi: it would be nice to add a test case to cover the situation where you found a bug.

internal/handlers/common/params.go
noisersup: > Also does this need to be a common function at all? The thing is GetLimitParam is not accurate description of this function as far as I understand, it cannot be used for getting limit param in find. 🤗

It can be used in find, we just don't use it there.

noisersup:
CI is failing due to coveralls, the tests passed

rumyantseva (approved):
LGTM

AlekSi (approved):

PR #3602, opened 2023-10-17
Make `listIndexes` return a sorted list

### LINKED ISSUE #3589: `listIndexes` should return a sorted list
### What should be done?

Currently, `listIndexes` command returns indexes in an unspecified order. It should return them sorted by name instead.

`Collection.ListIndexes` methods should also return results sorted by name. It will allow us to use a binary search on them.

We should also update our tests to compare sorted lists.

### Where?

See references to this issue.

### Definition of Done

- handler updated;
- spot refactorings done.

codenoid:
# Description

when doing `listIndex`, the result now will be sorted in ascending order based on index name

Closes #3589.

## Readiness checklist

### DIFF
--- a/internal/backends/collection.go
+++ b/internal/backends/collection.go
@@ -242,11 +242,12 @@ type CollectionStatsParams struct {
 
 // CollectionStatsResult represents the results of Collection.Stats method.
 type CollectionStatsResult struct {
-	CountDocuments int64
-	SizeTotal      int64
-	SizeIndexes    int64
-	SizeCollection int64
-	IndexSizes     []IndexSize
+	CountDocuments  int64
+	SizeTotal       int64
+	SizeIndexes     int64
+	SizeCollection  int64
+	SizeFreeStorage int64
+	IndexSizes      []IndexSize
 }
 
 // IndexSize represents the name and the size of an index.
--- a/internal/backends/collection.go
+++ b/internal/backends/collection.go
@@ -15,7 +15,10 @@
 package backends
 
 import (
+	"cmp"
 	"context"
+	"slices"
+	"sort"
 	"time"
 
 	"github.com/FerretDB/FerretDB/internal/types"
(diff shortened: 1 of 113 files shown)


codenoid:
actually, I have no clue, should we go for explicit test or functional test

internal/backends/collection.go
```diff
@@ -315,12 +318,15 @@ func (cc *collectionContract) ListIndexes(ctx context.Context, params *ListIndex
 	res, err := cc.c.ListIndexes(ctx, params)
 	checkError(err, ErrorCodeCollectionDoesNotExist)
 
-	// TODO https://github.com/FerretDB/FerretDB/issues/3589
```
AlekSi, line 318: I think it is enough to uncomment those lines – and then ensure that existing tests pass because currently `ListIndexes` implementations do not return results sorted

internal/backends/postgresql/metadata/registry_test.go
AlekSi, line 1: I don't think we need to add tests at all

internal/backends/collection.go
```diff
@@ -315,12 +318,15 @@ func (cc *collectionContract) ListIndexes(ctx context.Context, params *ListIndex
...
-	// if res != nil && len(res.Indexes) > 0 {
-	// 	must.BeTrue(slices.IsSortedFunc(res.Indexes, func(a, b IndexInfo) int {
```
AlekSi, line 320: That code was correct – we should only verify that results are sorted, not actually sort them. That's because `ListIndexes` implementation could be called by other methods of that implementation, and we want them to be able to rely on the fact that the list is sorted.

codenoid: but the code was panicking because `.IsSortedFunc` return false

AlekSi: Yes, that's why we need to update implementations of this method:
* https://github.com/FerretDB/FerretDB/blob/d7493f75f0d8d9109dd3cc816e529425fea93cd6/internal/backends/postgresql/collection.go#L441
* https://github.com/FerretDB/FerretDB/blob/d7493f75f0d8d9109dd3cc816e529425fea93cd6/internal/backends/sqlite/collection.go#L415

codenoid: I see, I understood now, should on the pg and sqlite handler

codenoid: do you still need this after the sort process?

https://github.com/FerretDB/FerretDB/blob/d7493f75f0d8d9109dd3cc816e529425fea93cd6/internal/backends/postgresql/collection.go#L486

AlekSi: We need to uncomment those lines and fix tests

codenoid: could you check this: https://github.com/FerretDB/FerretDB/commit/c7a7f059d10eab28805db8dd27d44217b4cbcde3

I will also remove the test

AlekSi: That commit looks correct, but we need to fix integration tests after that

internal/backends/postgresql/metadata/registry_test.go
codenoid: removed

rumyantseva (approved):
LGTM

AlekSi (approved):
Thank you!

PR #3632, opened 2023-10-23
Add issue tracking to checkcomments analyzer

### LINKED ISSUE #2733: Make linter for checking that linked issues are open
### What should be done?

We have a linter that checks comments like `// TODO https://github.com/FerretDB/FerretDB/issues/<number>` in our Go code: https://github.com/FerretDB/FerretDB/tree/main/tools/checkcomments

It should also check if those issues are still open. Doing that is a bit tricky: the tool runs once for each Go package (not project or Go module) and generates so many requests (many of which are for the same issues) that we easily go over GitHub's rate limit. We should keep cache check results in a temporary file and coordinate multiple invocations using [`lockedfile`](https://github.com/rogpeppe/go-internal/tree/master/lockedfile).

Additionally (and in a separate PR), we should check those issue links in string constants like in
```go
skip: "https://github.com/FerretDB/FerretDB/issues/1522"
```

### Where?

https://github.com/FerretDB/FerretDB/tree/main/tools/checkcomments

### Definition of Done

* Linter does not check the same issue twice for one `task lint` invocation.
* String constants in Go code are checked.

raeidish:
# Description

Implemented github issue tracking for checkcomments analyzer, uses [lockedfile](https://github.com/rogpeppe/go-internal/tree/master/lockedfile) to limit number of requests.

Closes #2733 

## Readiness checklist


AlekSi:
Oh, sorry, we missed that. Will review shortly

tools/Taskfile.yml
```diff
@@ -8,6 +8,8 @@ vars:
 tasks:
   tools-test:
     cmds:
+      - ../bin/envtool{{exeExt}} shell rmdir ../tmp/checkcomments
+      - ../bin/envtool{{exeExt}} shell mkdir ../tmp/checkcomments
```
rumyantseva, line 12: Why do we need to call it twice?

tools/checkcomments/checkcomments.go
```diff
@@ -72,45 +124,75 @@ func run(pass *analysis.Pass) (any, error) {
...
+	if err != nil {
+		if errors.As(err, new(*github.RateLimitError)) && os.Getenv("GITHUB_TOKEN") == "" {
+			log.Println(
+				"Rate limit reached. Please set a GITHUB_TOKEN as described at",
+				"https://github.com/FerretDB/FerretDB/blob/main/CONTRIBUTING.md#setting-a-github_token",
+			)
+
+			return false, err
+		}
+	}
```
rumyantseva, line 179: So, if an error is not `RateLimitError` or the token is set, we ignore the error and consider the issue open? It looks ok to me (e.g. if GitHub is down, it's ok that the linter just passes), but let's maybe add a comment that we do it on purpose.

tools/checkcomments/checkcomments.go
```diff
@@ -72,45 +124,75 @@ func run(pass *analysis.Pass) (any, error) {
...
+
+	isOpen := issue.GetState() == "open"
+
+	return isOpen, nil
+}
+
+func getCacheFilePath(p string) string {
+	path := filepath.Dir(p)
+
+	readmePath := filepath.Join(path, "README.md")
```
rumyantseva, line 189: This part is not clear to me. Why do we check if README.md exists? Maybe it is worth to leave a comment in the code.

rumyantseva (commented):
Thanks for contributing! I left a couple of questions.

tools/Taskfile.yml
rumyantseva: actually, it's `rmdir` and `mkdir`, so not twice 🙈

tools/checkcomments/checkcomments.go
raeidish: Yes, I will add a comment 👍

tools/checkcomments/checkcomments.go
rumyantseva: @raeidish thanks for the clarification! Yeah, it makes sense. This PR is almost ready to be merged, so I'm just going through all my previous questions at the moment.

rumyantseva (approved):
LGTM

AlekSi (approved):

AlekSi:
Thank you!

PR #3654, opened 2023-10-25
Add graphs for Prometheus metrics

### LINKED ISSUE #3600: Add graphs for Prometheus metrics
### What should be done?

TBD

### Where?

TBD

### Definition of Done

- all handlers updated;
- unit tests added/updated;
- spot refactorings done;
- something else?

Aditya1404Sal:
# Description

I took a look at the statsviz examples , and now am aware of the various types of graph plots and have added the same for the first commit, I am now working towards finding patterns in the prometheus metrics which can be visualized through the graphs, But i need guidance on What the basic graphs should be like and how to expand the prometheus metrics.

Closes #3600
## Readiness checklist


AlekSi (changes requested):
We should use real metric values from `prometheus.DefaultGatherer`, not generate random values

Aditya1404Sal:
> We should use real metric values from `prometheus.DefaultGatherer`, not generate random values

@AlekSi  I am not able to access the real metrics ,

i have tried
```
prometheus_metrics, err := prometheus.DefaultGatherer.Gather()
```
because 
Gather calls the Collect method of the registered Collectors and then gathers the collected metrics into a lexicographically sorted slice of uniquely named MetricFamily protobufs. Gather ensures that the returned slice is valid and self-consistent so that it can be used for valid exposition.

where prometheus_metrics is : 
`var prometheus_metrics []*io_prometheus_client.MetricFamily`

and looked at prometheus as well as promhttp docs but i can't see a way to expand the real metrics

I really need some guidance.

AlekSi:
They _are_ the real metrics: https://pkg.go.dev/github.com/prometheus/client_model/go#MetricFamily

internal/util/debug/debug.go
```diff
@@ -19,35 +19,49 @@ import (
...
+	}
+
+	http.Handle("/debug/metrics", metricHandler)
 
 	opts := []statsviz.Option{
 		statsviz.Root("/debug/graphs"),
+		statsviz.TimeseriesPlot(ferretdb_postgresql_metadata_databases_barPlot()),
+		statsviz.TimeseriesPlot(ferretdb_postgresql_pool_size_barPlot()),
+		statsviz.TimeseriesPlot(promhttp_metric_handler_requests_total_stackPlot()),
+		statsviz.TimeseriesPlot(promhttp_metric_handler_requests_in_flight_scatterPlot()),
```
AlekSi, line 64: Instead of adding graphs for those 4 metrics, let's generate graphs for all metrics

internal/util/debug/debug.go
```diff
@@ -110,3 +124,190 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
...
+	return plot
+}
+
+func codeCountGen() float64 {
+	return rand.Float64()
+}
+
+func metricRetriever(prometheus_metrics []*io_prometheus_client.MetricFamily, metricName string) *io_prometheus_client.Metric {
+	for _, specificMetric := range prometheus_metrics {
+		if specificMetric.GetName() == metricName {
```
AlekSi, line 263: Let's generate graphs for all metrics

internal/util/debug/debug.go
Aditya1404Sal: Yes let's do that

AlekSi: Generate _programmatically_, not write a function for each metric. We often add and remove metrics, all of them are available in `prometheus.DefaultGatherer`, we could generate graphs on a fly

AlekSi (changes requested):

internal/util/debug/debug.go
Aditya1404Sal: My apologies. I'll make the Changes :+1:

internal/util/debug/debug.go
```diff
@@ -20,38 +20,60 @@ import (
...
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	io_prometheus_client "github.com/prometheus/client_model/go"
 	"go.uber.org/zap"
 	"golang.org/x/exp/maps"
-
-	"github.com/FerretDB/FerretDB/internal/util/must"
 )
 
+type metric struct {
```
AlekSi, line 42: https://github.com/FerretDB/FerretDB/blob/1f18df076dc2c7bc804e9358d3b933ad2f9992ec/CONTRIBUTING.md?plain=1#L263

internal/util/debug/debug.go
```diff
@@ -20,38 +20,60 @@ import (
...
+	metricName       string
+	metricType       int
+	metricLabel      string
+	metricLabelValue string
```
AlekSi, line 46: I'm not sure how we fit multiple key/value pairs into just two fields. I'm also not sure why we need that structure at all

internal/util/debug/debug.go
```diff
@@ -20,38 +20,60 @@ import (
...
 			Registry:          r,
 			EnableOpenMetrics: true,
 		}),
-	))
+	)
+	prometheus_metrics, err := prometheus.DefaultGatherer.Gather()
+	fmt.Printf("prometheus_metrics: %v\n", prometheus_metrics)
+	if err != nil {
+		log.Fatalf("error in Gathering prometheus metrics: %v", err)
+	}
```
AlekSi, line 65: We should not use `fmt.Printf` or `log.Fatalf` there

internal/util/debug/debug.go
```diff
@@ -122,3 +144,123 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
 	s.Close()
 	l.Sugar().Info("Debug server stopped.")
 }
+
+func (obj metric) getValue() float64 {
+	x := prometheusGather()
+	if obj.metricLabel == "" {
+		y := metricRetriever(x, obj.metricName)
+		switch obj.metricType {
+		case 0:
```
AlekSi, line 153: https://pkg.go.dev/github.com/prometheus/client_model@v0.5.0/go#MetricType

AlekSi (commented):
It is hard to understand why this code generates graphs this way or even what it does. There should be a comment explaining how graphs are generated from metrics.

Please also use a PR template checklist. For example, `task all` clearly wasn't run.

internal/util/debug/debug.go
Aditya1404Sal: We don't fit Multiple key/value pairs because it's not possible to fit more than one Timeseries into a single Plot programatically, hence each graph plot only represents a unique single key/value pair even if metricFamily.Name is simillar across them. We need that structure because the GetValue attribute of statsviz.Timeseries doesn't support passing arguments (in this use-case passing metricFamily.Name field). therefore creating a structure would allow data passing thru interface.

internal/util/debug/debug.go
Aditya1404Sal: would `panic(err)` be fine instead of `log.Fatalf`

AlekSi: I think we should be consistent with surrounding code that uses `must.NoError`

internal/util/debug/debug.go
Aditya1404Sal: since MetricType is of
type MetricType **int32** 

metricType field in the struct gets assigned the typecasted metricFamily.Type as `int` which helps in referencing at lines 152 and later

```
func generateGraphAll(........

finalMetricObj.metricType = int(*metricFamily.Type)
...
}
```

AlekSi:
Blocked by arl/statsviz#118.

PR #3663, opened 2023-10-30
Support limit pushdown when filter and sort can apply pushdown entirely

### LINKED ISSUE #3644: Support limit pushdown when filter and sort can apply pushdown entirely
### What should be done?

Current limit pushdown is applied only when filter and sort is not set. This could be loosen so that limit pushdown can apply to more find commands. If both filter and sort can apply pushdown entirely, then limit can also pushdown without causing errors.

There might be two options:
1.implement a function which accepts params.Filter and params.Sort as inputs,  determines if there are any parameters that can‘t apply pushdown. If params.Filter and params.Sort can both apply pushdown entirely, then limit can also apply pushdown
2.reuse current logic and make some changes to function prepareWhereClause()  and prepareOrderByClause(). Determine whether limit should be pushed down after these two funcitons

Any thoughts?

cvbhjkl:
# Description

To support limit pushdown when filter and sort can apply pushdown entirely:
1. Loose limit pushdown conditions in handler.
2. Add a boolean return value to functions prepareWhereClause and prepareOrderByClause to reuse current logic to check if some filters or sort can not be applied pushdown. Only if prepareWhereClause and prepareOrderByClause does not find any parameters that cannot be pushed down, will apply limit pushdown.
3. Update some integration tests and add comments.

Closes #3644.

## Readiness checklist

### DIFF
--- a/integration/query_test.go
+++ b/integration/query_test.go
@@ -835,7 +835,7 @@ func TestQueryCommandLimitPushDown(t *testing.T) {
 
 		len            int                 // expected length of results
 		queryPushdown  resultPushdown      // optional, defaults to noPushdown
-		limitPushdown  bool                // optional, set true for expected pushdown for limit
+		limitPushdown  resultPushdown      // optional, set true for expected pushdown for limit
 		err            *mongo.CommandError // optional, expected error from MongoDB
 		altMessage     string              // optional, alternative error message for FerretDB, ignored if empty
 		skip           string              // optional, skip test with a specified reason
@@ -844,39 +844,39 @@ func TestQueryCommandLimitPushDown(t *testing.T) {
 		"Simple": {
 			limit:         1,
 			len:           1,
-			limitPushdown: true,
+			limitPushdown: allPushdown,
 		},
 		"AlmostAll": {
 			limit:         int64(len(shareddata.Composites.Docs()) - 1),
 			len:           len(shareddata.Composites.Docs()) - 1,
-			limitPushdown: true,
+			limitPushdown: allPushdown,
 		},
 		"All": {
 			limit:         int64(len(shareddata.Composites.Docs())),
 			len:           len(shareddata.Composites.Docs()),
-			limitPushdown: true,
+			limitPushdown: allPushdown,
 		},
 		"More": {
 			limit:         int64(len(shareddata.Composites.Docs()) + 1),
 			len:           len(shareddata.Composites.Docs()),
-			limitPushdown: true,
+			limitPushdown: allPushdown,
 		},
 		"Big": {
 			limit:         1000,
 			len:           len(shareddata.Composites.Docs()),
-			limitPushdown: true,
+			limitPushdown: allPushdown,
 		},
 		"Zero": {
 			limit:         0,
 			len:           len(shareddata.Composites.Docs()),
-			limitPushdown: false,
+			limitPushdown: noPushdown,
 		},
 		"IDFilter": {
 			filter:        bson.D{{"_id", "array"}},
 			limit:         3,
 			len:           1,
 			queryPushdown: allPushdown,
-			limitPushdown: false,
+			limitPushdown: allPushdown,
 		},
 		"ValueFilter": {
 			filter:        bson.D{{"v", 42}},
@@ -884,28 +884,28 @@ func TestQueryCommandLimitPushDown(t *testing.T) {
 			limit:         3,
 			len:           3,
 			queryPushdown: pgPushdown,
-			limitPushdown: false,
+			limitPushdown: pgPushdown,
 		},
 		"DotNotationFilter": {
 			filter:        bson.D{{"v.foo", 42}},
 			limit:         3,
 			len:           3,
 			queryPushdown: noPushdown,
-			limitPushdown: false,
+			limitPushdown: noPushdown,
 		},
 		"ObjectFilter": {
 			filter:        bson.D{{"v", bson.D{{"foo", nil}}}},
 			limit:         3,
 			len:           1,
 			queryPushdown: noPushdown,
-			limitPushdown: false,
+			limitPushdown: noPushdown,
 		},
 		"Sort": {
 			sort:          bson.D{{"_id", 1}},
 			limit:         2,
 			len:           2,
 			queryPushdown: noPushdown,
-			limitPushdown: true,
+			limitPushdown: allPushdown,
 		},
 		"IDFilterSort": {
 			filter:        bson.D{{"_id", "array"}},
@@ -913,7 +913,7 @@ func TestQueryCommandLimitPushDown(t *testing.T) {
 			limit:         3,
 			len:           1,
 			queryPushdown: allPushdown,
-			limitPushdown: false,
+			limitPushdown: noPushdown,
 		},
 		"ValueFilterSort": {
 			filter:        bson.D{{"v", 42}},
@@ -921,7 +921,7 @@ func TestQueryCommandLimitPushDown(t *testing.T) {
 			limit:         3,
 			len:           3,
 			queryPushdown: pgPushdown,
-			limitPushdown: false,
+			limitPushdown: pgPushdown,
 		},
 		"DotNotationFilterSort": {
 			filter:        bson.D{{"v.foo", 42}},
@@ -929,7 +929,7 @@ func TestQueryCommandLimitPushDown(t *testing.T) {
 			limit:         3,
 			len:           3,
 			queryPushdown: noPushdown,
-			limitPushdown: false,
+			limitPushdown: noPushdown,
 		},
 		"ObjectFilterSort": {
 			filter:        bson.D{{"v", bson.D{{"foo", nil}}}},
@@ -937,13 +937,13 @@ func TestQueryCommandLimitPushDown(t *testing.T) {
 			limit:         3,
 			len:           1,
 			queryPushdown: noPushdown,
-			limitPushdown: false,
+			limitPushdown: noPushdown,
 		},
 		"Skip": {
 			optSkip:       pointer.ToInt64(1),
 			limit:         2,
 			len:           2,
-			limitPushdown: false,
+			limitPushdown: noPushdown,
 		},
 	} {
 		tc, name := tc, name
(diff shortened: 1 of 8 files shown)


internal/backends/postgresql/query.go
```diff
@@ -63,9 +63,12 @@ func prepareSelectClause(schema, table string, capped, onlyRecordIDs bool) strin
 }
 
 // prepareWhereClause adds WHERE clause with given filters to the query and returns the query and arguments.
-func prepareWhereClause(p *metadata.Placeholder, sqlFilters *types.Document) (string, []any, error) {
+func prepareWhereClause(p *metadata.Placeholder, sqlFilters *types.Document) (string, []any, bool, error) {
```
noisersup, line 66: As we start to return too many values, I'd suggest to create respectively `whereClauseResults`, and `orderByClauseResults`.

noisersup (approved):
Thank you for contribution! I've tried to find some edge cases, but everything seems to work properly. I left some minor comments. Looks good to me

internal/backends/postgresql/collection.go
```diff
@@ -321,12 +322,13 @@ func (c *collection) Explain(ctx context.Context, params *backends.ExplainParams
 
 	q += where
 
-	sort, sortArgs := prepareOrderByClause(&placeholder, params.Sort, meta.Capped())
+	sort, sortArgs, sortPushdownAll := prepareOrderByClause(&placeholder, params.Sort, meta.Capped())
```
chilagrow, line 325: In handler code, we don't set `params.Sort` if there is more than one field for sort, for example `{v: 1, _id: 1}`. So `params.Sort` is nil that case.
https://github.com/FerretDB/FerretDB/blob/0e69b3dfed7d18daafb0d7e2b1de313ed41af4e4/internal/handlers/sqlite/msg_explain.go#L91

`prepareOrderByClause()` returns boolean `true` seeing `params.Sort == nil`, which then allows `LIMIT` to be applied. This will be an issue because `LIMIT` would be applied before sorting documents in memory.

Let's be cautious and ensure we have tests to cover this case. :hugs:

internal/handlers/sqlite/msg_find.go
```diff
@@ -99,11 +99,14 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
-	//  and sort them in memory;
+	//  - `filter` is set but can not be applied pushdown entirely,
+	//  it must fetch all documents to filter them in memory;
+	//  - `sort` is set but can not be applied pushdown entirely,
+	//  or `sort` is set but `EnableSortPushdown` is not set,
+	//  it must fetch all documents and sort them in memory;
 	//  - `skip` is non-zero value, skip pushdown is not supported yet.
-	if params.Filter.Len() == 0 && (params.Sort.Len() == 0 || h.EnableSortPushdown) && params.Skip == 0 {
+	// More checks are performed when preparing LIMIT clause.
+	if (params.Sort.Len() == 0 || h.EnableSortPushdown) && params.Skip == 0 {
```
chilagrow, line 109: We cannot set limit pushdown if `h.DisableFilterPushdown` is true and filter is set. That's because filter is expected to be done purely in memory, so we need all documents fetched without any `LIMIT`. :hugs:

chilagrow (commented):
Thanks for your contribution, please see the comment :hugs:

chilagrow (changes requested):
See comments please :pray:

integration/query_test.go
```diff
@@ -995,23 +995,24 @@ func TestQueryCommandLimitPushDown(t *testing.T) {
...
 				if setup.IsPushdownDisabled() {
-					resultPushdown = noPushdown
+					queryResultPushdown = noPushdown
 					msg = "Query pushdown is disabled, but target resulted with pushdown"
 				}
 
 				doc := ConvertDocument(t, res)
 				limitPushdown, _ := doc.Get("limitPushdown")
-				assert.Equal(t, tc.limitPushdown, limitPushdown, msg)
+				assert.Equal(t, limitResultPushdown.PushdownExpected(t), limitPushdown, msg)
```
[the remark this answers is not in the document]
cvbhjkl: `PushdownExpected(t)` is replaced by a new `LimitPushdownExpected(t)`. Logic here now is: Limit pushdown is set to false if Filter pushdown is disabled but filter is not nil, or Sort pushdown is disabled but sort is not nil.

internal/backends/postgresql/collection.go
cvbhjkl: handler code is updated. And a test case with two fields for sort is added to cover this case.

internal/handlers/sqlite/msg_find.go
cvbhjkl: handler code is updated to take `h.DisableFilterPushdown` into consideration.

cvbhjkl:
Thank you for your review. Please see my recent commit and comments. And integration/compatibility tests, with one or both of flags `disable-filter-pushdown` and `enable-sort-pushdown` set, pass now.

internal/handlers/sqlite/msg_explain.go
```diff
@@ -106,11 +106,17 @@ func (h *Handler) MsgExplain(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
...
+	//  it must fetch all documents to filter them in memory;
+	//  - `sort` is set but can not be applied pushdown entirely,
+	//  or `sort` is set but `EnableSortPushdown` is not set,
+	//  it must fetch all documents and sort them in memory;
 	//  - `skip` is non-zero value, skip pushdown is not supported yet.
-	if params.Filter.Len() == 0 && (params.Sort.Len() == 0 || h.EnableSortPushdown) && params.Skip == 0 {
+	// More checks are performed when preparing LIMIT clause.
+	if (params.Filter.Len() == 0 || !h.DisableFilterPushdown) &&
+		(params.Sort.Len() == 0 || (h.EnableSortPushdown && params.Sort.Len() == 1)) &&
+		params.Skip == 0 {
```
chilagrow, line 119: This logic became a bit too complex. Although there is comment, it doesn't quite help understanding what is going on at first sight. Could we simplify it or refactor it? :thinking:

internal/backends/postgresql/query.go
```diff
@@ -63,9 +63,13 @@ func prepareSelectClause(schema, table string, capped, onlyRecordIDs bool) strin
...
 
 // prepareWhereClause adds WHERE clause with given filters to the query and returns the query and arguments.
-func prepareWhereClause(p *metadata.Placeholder, sqlFilters *types.Document) (string, []any, error) {
+// It returns false if pushdown for at least one filter couldn't be applied.
+func prepareWhereClause(p *metadata.Placeholder, sqlFilters *types.Document) (string, []any, bool, error) {
 	var filters []string
 	var args []any
+	var filterPushdownAll bool
+
+	filterPushdownAll = true
```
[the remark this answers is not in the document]
chilagrow: @noisersup do you have a suggestion of improving how to check if pushdown is applied from the entire filter?

integration/query_test.go
```diff
@@ -835,7 +835,7 @@ func TestQueryCommandLimitPushDown(t *testing.T) {
 
 		len            int                 // expected length of results
 		queryPushdown  resultPushdown      // optional, defaults to noPushdown
-		limitPushdown  bool                // optional, set true for expected pushdown for limit
+		limitPushdown  resultPushdown      // optional, set true for expected pushdown for limit
```
chilagrow, line 838: `TestQueryCommandLimitPushDown` test currently tests against `shareddata.Composites`, we would want to test with `shareddata.Scalars` provider for testing filters that are boolean, string and so on like below.

```
filter: bson.D{{"v", true}},
```

```
filter: bson.D{{"v", bson.D{{"$eq", true}}}},
```

```
filter: bson.D{{"v", bson.D{{"$ne", true}}}},
```

```
filter: bson.D{{"v", bson.D{{"$gt", true}}}},
```

internal/backends/postgresql/query.go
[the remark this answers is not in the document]
cvbhjkl: I didn't realize that some pushdowns with $eq and $ne were best-effort matches, causing some unexpected results. Adding a return value to differentiate between exact matches and best-effort matches in the `filterEqual` function should fix this, and added tests should cover the cases you described. 
In `filterEqual`, I followed your suggestion to set `exactMatch=false` and set it true when exact match satisfied. However, this may not work for `filterPushdownAll`, because determining whether limit pushdown can be applied is actually an AND logic, and limit pushdown can only be enabled when all fields in Filter, Sort and Skip can be applied pushdown.If `filterPushdownAll` is set to false at the beginning, then the number of conditions required to set it to true is actually uncertain, because each field in Filter, Sort and Skip needs to be checked. 
I agree that accidentally apply limit pushdown is a problem that needs to be avoided. As long as `filterpushdownall=false` in current code does not accidentally be removed during future development of new pushdowns , the worst scenario is that we fetch more results than necessary. 
Or do you @chilagrow @noisersup have any recommended ways to do this?

internal/handlers/sqlite/msg_explain.go
cvbhjkl: Quite agree. But this logic may be difficult to simplify, as it does indeed involve many parameters and flags. Luckily, with future work, logic here should be simplified, such as we won't need to check if skip is 0 if we support skip pushdown. Or, do you think it's a good idea to put logic here in one or two functions, like `if canApplyLimitPushdown(params,h){`?

integration/query_test.go
cvbhjkl: `shareddata.Scalars` provider and a lot more tests are added. Please see if there are still some missing cases.

cvbhjkl:
Thank you for your careful review. Fix issue with $eq and $ne and add more test cases. Please see comments.

AlekSi:
@cvbhjkl Sorry, we have to put it on hold for now – we need to deeply refactor our existing code for pushdown before we can proceed, and there will be a lot of merge conflicts.

AlekSi (commented):
https://github.com/FerretDB/FerretDB/pull/3663#issuecomment-1801182303

Issue #3601, opened 2023-10-17
`ListDatabases` and `ListCollections` should allow filtering by name


AlekSi:
### What should be done?

`Backend.ListDatabases` and `Database.ListCollections` should allow getting a database/collection by exact name. Essentially, we are combining "list" and "get" methods in one method.

Some code that currently does the filtering by name itself should be simplified.

### Where?

See references to this issue.

### Definition of Done

- all backends updated;
- handler updated;
- unit tests added/updated;
- integration/compatibility tests added/updated;
- spot refactorings done.

AlekSi:
@sachinpuranik there is another place that needs updating now

PR #3677, opened 2023-11-03
Add test for tailable cursor with non-capped collection

### LINKED ISSUE #3462: Add test for tailable cursor with non-capped collection
### What should be done?

Tailable cursors do not work with non-capped collections. Let's add an integration test for that.

The test should pass for MongoDB and the current FerretDB handler.

### Where?

https://github.com/FerretDB/FerretDB/blob/main/integration/query_test.go

### Definition of Done

- the handler updated;
- spot refactorings done.

noisersup:
# Description

Closes #3462.

## Readiness checklist


integration/query_test.go
```diff
@@ -1063,3 +1063,30 @@ func TestQueryIDDoc(t *testing.T) {
...
+func TestQueryTailableCursors(t *testing.T) {
+	t.Parallel()
+
+	ctx, collection := setup.Setup(t)
+
+	_, err := collection.InsertOne(ctx, bson.D{{"v", int32(42)}})
+	require.NoError(t, err)
+
+	expectedErr := mongo.CommandError{
+		Code:    238,
```
rumyantseva, line 1076: What is not clear to me in the initial issue is the error code that should be returned. 
Shouldn't we return the same error code as MongoDB? That tailable cursors are not supported for capped collections in principle rather than a "non-implemented" error? I think "non implemented" error is for something we might want to implement in the future.

noisersup: we don't support tailable cursors at all.

I wanted to return unimplemented only if it's ran against capped collection, the problem here is we handle unimplemented on the `FindParam` struct tags, and return them while unmarshaling.

AlekSi: > Definition of Done
> the handler updated;

To return the same error when tailable cursor is requested for non-capped collection

AlekSi (changes requested):

internal/backends/collection.go
```diff
@@ -48,6 +48,7 @@ type Collection interface {
 
 	Stats(context.Context, *CollectionStatsParams) (*CollectionStatsResult, error)
 	Compact(context.Context, *CompactParams) (*CompactResult, error)
+	Capped(context.Context) (bool, error)
```
rumyantseva, line 51: Oh, I remember this one was questionable when I was trying to offer it last time 😅 

`ListCollections` has info on wether a collection is capped. In `msg_collstats` there is an example how to use it - https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/sqlite/msg_collstats.go#L88-L94.

While I agree that the method `Capped` would be really handy to have, I also understand that we don't want to extend the interface, so it's easier to support it.

(Unless something about it has changed, cc @AlekSi)

integration/query_test.go
```diff
@@ -1203,3 +1203,21 @@ func TestQueryShowRecordIDErrors(t *testing.T) {
...
+
+	expectedErr := mongo.CommandError{
+		Code: 2,
+		Message: "error processing query: ns=TestQueryTailableCursors.TestQueryTailableCursorsTree:" +
+			"  $eq null\nSort: {}\nProj: {}\n tailable cursor requested on non capped collection",
+		Name: "BadValue",
+	}
+
+	_, err = collection.Find(ctx, bson.D{{}}, options.Find().SetCursorType(options.Tailable))
+	AssertEqualCommandError(t, expectedErr, err)
```
rumyantseva, line 1222: The test looks good to me!

internal/backends/collection.go
AlekSi: #3601

AlekSi: @noisersup please remove that method

AlekSi (changes requested):

internal/backends/collection.go
noisersup: whoops my bad, fixed

rumyantseva (approved):
LGTM

AlekSi (approved):

PR #3695, opened 2023-11-10
Add tracing to envtool tests run

### LINKED ISSUE #3665: Add tracing to `envtool tests run`
### What should be done?

We should add OpenTelemetry traces to `envtool tests run`. That would give us better visibility into the way our parallel tests started, paused, and continued. Adding it to the tests itself is not enough because we call `t.Parallel()` before anything else.

Let's use one trace per top-level test and spans for subtests. Let's not create a span when the test is paused. And let's experiment with additional events to see what gives the best visualization.

### Where?

https://github.com/FerretDB/FerretDB/tree/main/cmd/envtool

### Definition of Done

`envtool tests run` visualization is available, useful, and looks nice.

hungaikev:
# Description

Adding tracing to envtool tests run

Closes #3665

## Readiness checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,7 @@ vendor/
 website/build/
 
 **/testdata/fuzz/
+
+*.idea
+.idea
+.vscode
--- a/cmd/envtool/envtool.go
+++ b/cmd/envtool/envtool.go
@@ -162,7 +162,7 @@ func setupMongodb(ctx context.Context, logger *zap.SugaredLogger) error {
 	for ctx.Err() == nil {
 		buf.Reset()
 
-		err := runCommand("docker", args, &buf, logger)
+		err := runCommand(ctx, "docker", args, &buf, logger)
 		if err == nil {
 			break
 		}
@@ -201,7 +201,7 @@ func setup(ctx context.Context, logger *zap.SugaredLogger) error {
 }
 
 // runCommand runs command with given arguments.
-func runCommand(command string, args []string, stdout io.Writer, logger *zap.SugaredLogger) error {
+func runCommand(ctx context.Context, command string, args []string, stdout io.Writer, logger *zap.SugaredLogger) error {
 	bin, err := exec.LookPath(command)
 	if err != nil {
 		return err
@@ -213,6 +213,10 @@ func runCommand(command string, args []string, stdout io.Writer, logger *zap.Sug
 	cmd.Stderr = os.Stderr
 
 	if err := cmd.Run(); err != nil {
+		// Check if the error was due to a context cancellation.
+		if ctx.Err() != nil {
+			return fmt.Errorf("%s canceled: %s", strings.Join(cmd.Args, " "), err)
+		}
 		return fmt.Errorf("%s failed: %s", strings.Join(cmd.Args, " "), err)
 	}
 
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -14,3 +14,7 @@ vendor/
 website/build/
 
 **/testdata/fuzz/
+
+*.idea
+.idea
+.vscode
```
AlekSi, line 20: Please see the comment on top of the file

cmd/envtool/envtool.go
```diff
@@ -213,24 +213,28 @@ func runCommand(command string, args []string, stdout io.Writer, logger *zap.Sug
 	cmd.Stderr = os.Stderr
 
 	if err := cmd.Run(); err != nil {
+		// Check if the error was due to a context cancellation.
+		if ctx.Err() != nil {
+			return fmt.Errorf("%s canceled: %s", strings.Join(cmd.Args, " "), err)
```
AlekSi, line 218: But why do we need it? What is special about that error?

I think we should revert those unrelated changes that add context

AlekSi (commented):
I don't see where this PR adds spans to `envtool tests run` command implementation in `envtool.go`. This issue is about adding tracing to it, not to its tests

AlekSi (changes requested):
That's not what needs to be done in that issue.

> Let's use one trace per top-level test and spans for subtests

Not one span per `go test` invocation and one tracing event per decoded event. We need to start span when the decoded event indicates that the test started, end span when the test ends, handle paused and continued tests, etc.

go.mod
```diff
@@ -26,6 +26,8 @@ require (
 	modernc.org/sqlite v1.27.0
 )
 
+require go.opentelemetry.io/otel/trace v1.19.0
```
AlekSi, line 29: That should be in the group above

cmd/envtool/tests.go
```diff
@@ -97,6 +100,8 @@ func runGoTest(ctx context.Context, args []string, total int, times bool, logger
 
 	var done int
 	results := make(map[string]*testResult, 300)
+	contexts := make(map[string]context.Context)
+	spans := make(map[string]trace.Span)
```
AlekSi, line 104: Let's add those fields to the `testResult` struct

hungaikev: Done

AlekSi (commented):
It looks strange that we don't end the span on pass/fail/skip. How does the result look in Jaeger?

hungaikev:
> It looks strange that we don't end the span on pass/fail/skip. How does the result look in Jaeger?

I was getting a null pointer exception when the events `pass/fail/skip`  because of `event.Test` being "".  I have changed the code to factor that.

cmd/envtool/tests.go
```diff
@@ -112,13 +117,19 @@ func runGoTest(ctx context.Context, args []string, total int, times bool, logger
...
 
-		if event.Test != "" {
-			if results[event.Test] == nil {
-				results[event.Test] = &testResult{
-					outputs: make([]string, 0, 2),
-				}
+		// Skip processing if event.Test is empty
+		if event.Test == "" {
+			continue
+		}
```
AlekSi, line 123: We should not do that because there are package-level events that we want to track. This change also broke tests. Please run them locally.

AlekSi (changes requested):

hungaikev:
> @hungaikev any news? Anything we could help you with?

Hi @AlekSi  would you be willing or available for a <= 15-minute block for us to go through this together to reduce the back and forth?  I have a feeling that the part left is very small but seems to be taking 80% of the effort.

AlekSi:
I fixed all the problems. The result looks like this:

CleanShot 2023-12-06 at 18 57 19@2x

AlekSi (approved):

AlekSi:
@rumyantseva PTAL

rumyantseva (approved):
So, with the current approach, we can see when a particular event (run, continue, pause, pass) happened for each subtest. Looks good to me.

PR #3720, opened 2023-11-16
Separate safe pushdown from unsafe sort pushdown

### LINKED ISSUE #3688: Cleanup pushdown terminology
### What should be done?

Our terminology for pushdowns is a mess. Since we don't have good terms, our code is a bit messy and has suspicious places that confuse different types of pushdowns: https://github.com/FerretDB/FerretDB/blob/22421e7e4e33a492169a65a495edc0d40b9afd8a/integration/query_test.go#L1005-L1012

We should clarify terms in our glossary, documentation, tests, code, and configuration (in that order).

The proposal:
* do not use "query pushdown" or just "pushdown" at all;
* use "filter pushdown" and "sort pushdown" for both inputs (flags, parameters) and outputs (explain results);
* make it clear that "sort pushdown" input also includes and implies `limit` and `skip` pushdowns;
* use "limit pushdown" for outputs only;
* separate "unsafe" from both "filter pushdown" and "sort pushdown" – there are already safe (capped collections) and unsafe sort pushdowns, and we might introduce unsafe filter pushdowns in the future.

### Where?

- [ ] glossary
- [x] the rest of the documentation
- [x] integration tests
- [x] unit tests
- [x] code
- [x] configuration flags

### Definition of Done

- handler updated;
- all backends updated;
- unit tests updated;
- integration/compatibility tests updated;
- spot refactorings done;
- user documentation updated.

noisersup:
# Description

Closes #3688.

## Readiness checklist


internal/backends/collection.go
```diff
@@ -79,8 +79,10 @@ type SortField struct {
 
 // QueryParams represents the parameters of Collection.Query method.
 type QueryParams struct {
-	Filter        *types.Document
-	Sort          *SortField
+	Filter       *types.Document
+	Sort         *SortField
+	SortPushdown bool
```
AlekSi, line 84: There is no documentation update or contract change. For someone implementing a new backend, it is not possible to understand what those fields mean, what different combinations of them mean, how they should be used, etc.

AlekSi (changes requested):

internal/backends/collection.go
noisersup: Should we document all fields, or only those "less obvious"?

noisersup: Not sure how well I've understood. PTAL at the latest commits

Taskfile.yml
```diff
@@ -270,6 +273,7 @@ tasks:
         -hana-url=$FERRETDB_HANA_URL
         -compat-url='mongodb://username:password@127.0.0.1:47018/?tls=true&tlsCertificateKeyFile=../build/certs/client.pem&tlsCaFile=../build/certs/rootCA-cert.pem'
         -disable-filter-pushdown={{.DISABLE_FILTER_PUSHDOWN}}
+        -disable-sort-pushdown={{.DISABLE_SORT_PUSHDOWN}}
```
rumyantseva, line 276: If `-disable-sort-pushdown` is set as false and `-enable-unsafe-sort-pushdown` is set as true, is pushdown enabled?

internal/backends/collection.go
```diff
@@ -79,11 +79,16 @@ type SortField struct {
...
 type QueryParams struct {
-	Filter        *types.Document
-	Sort          *SortField
-	Limit         int64
+	Filter  *types.Document
+	Sort    *SortField
+	Limit   int64
+	Comment string
+
+	// If true, the backend might perform safe sort pushdown when possible.
```
AlekSi, line 87: That part should be in the Query method documentation.

internal/backends/collection.go
```diff
@@ -79,11 +79,16 @@ type SortField struct {
...
+	SortPushdown bool
+
+	// If true, the backend will return only recordID column for capped collection.
```
AlekSi, line 90: Move to Query method

internal/backends/collection.go
```diff
@@ -98,6 +103,9 @@ type QueryResult struct {
 // The passed context should be used for canceling the initial query.
 // It also can be used to close the returned iterator and free underlying resources,
 // but doing so is not necessary - the handler will do that anyway.
+//
+// If SortPushdown is set to false, any sort, or limit pushdown won't be performed,
+// even when Sort, or Limit fields are set.
```
AlekSi, line 108: If that's true, then we don't need that field at all – the handler should just not set `Sort` and `Limit` fields.

AlekSi (changes requested):
Let's focus on QueryParams fields and Query method documentation. Right now, it is not possible to understand how they should be interpreted by the new backend implementation.

internal/backends/collection.go
```diff
@@ -105,17 +105,29 @@ type QueryResult struct {
 // It also can be used to close the returned iterator and free underlying resources,
 // but doing so is not necessary - the handler will do that anyway.
 //
-// If SortPushdown is false, any sort, or limit pushdown won't be performed.
-// If SortPushdown is false, Sort and Limit fields cannot be set.
-// If SortPushdown is true, but Sort is equal to nil, sort pushdown can still occure
+// - If SortPushdown is false, any sort, or limit pushdown won't be performed.
+// - If SortPushdown is false, Sort and Limit fields cannot be set.
```
AlekSi: That's the same point.

internal/backends/collection.go
```diff
@@ -105,17 +105,29 @@ type QueryResult struct {
...
+// - If SortPushdown is true, but Sort is equal to nil, sort pushdown can still occure
 // for capped collection if recordID is not set.
```
AlekSi: What does it mean?

internal/backends/collection.go
```diff
@@ -105,17 +105,29 @@ type QueryResult struct {
...
+//
+// - If Filter is provided, the pushdown will be performed if supported.
+// Results may contain more documents than necessary, those should be filtered out at the handler level.
+// - If Sort is provided, the unsafe sort pushdown will be performed;
```
AlekSi, line 107: What makes it unsafe?

internal/backends/collection.go
```diff
@@ -105,17 +105,29 @@ type QueryResult struct {
...
+// - If Limit is provided standalone, the safe sort pushdown will be performed;
```
AlekSi, line 108: What is "standalone"? Why we do sort pushdown for limit?

internal/backends/collection.go
```diff
@@ -105,17 +105,29 @@ type QueryResult struct {
...
+// - If Filter and Sort are provided, the unsafe sort pushdown for both will be performed if supported;
```
AlekSi, line 109: And if not? How will the handler know if sorting was performed and if there is no need to fetch all documents and sort them?

internal/backends/collection.go
```diff
@@ -105,17 +105,29 @@ type QueryResult struct {
...
+// - Filter and Limit cannot be provided simultaneously;
```
AlekSi, line 110: Why not? `db.collection.find({v: 42}).limit(1)` is pretty common

internal/backends/collection.go
```diff
@@ -105,17 +105,29 @@ type QueryResult struct {
...
+// - If Sort and Limit are provided, the unsafe sort pushdown for both will be performed; (?)
```
AlekSi: Well, so you are not sure? :)

noisersup: I think I posted the link to specific commit from PR which is not the latest now, sorry.

I left this at the beginning of investigating, and forgot to remove it afterwards, but fixed that in the later commit

internal/backends/collection.go
noisersup: But currently our handler doesn't support it

internal/backends/collection.go
noisersup: I think me might want to introduce such information in `QueryResult`

internal/backends/collection.go
noisersup: should we call it safe limit pushdown?

internal/backends/collection.go
noisersup: Currently the only case we pushdown safely is default sorting of capped collection. We don't perform any safe pushdown if Sort is provided.

internal/backends/collection.go
noisersup: It means that sort pushdown can be present even if Sort field is not set in one specific case. 
Also, I've fixed the typo.

internal/backends/collection.go
noisersup: fixed

internal/backends/collection.go
AlekSi: So, why did you re-request the review?

AlekSi:
Temporary on hold

AlekSi (commented):
On hold

Issue #3413, opened 2023-09-22
Add MySQL backend


AlekSi:
### What should be done?

There is a [significant community interest](https://github.com/FerretDB/FerretDB/discussions/50) in the MySQL backend for FerretDB. While FerretDB Inc. does not have plans to author one, we could help maintain it if someone in the community wrote it.

FerretDB's architecture was recently changed to make adding new backends easier. See https://blog.ferretdb.io/ferretdb-v1-10-production-ready-sqlite/

To add a new backend, three interfaces should be implemented: https://pkg.go.dev/github.com/FerretDB/FerretDB/internal/backends. Please take a look at existing backends for examples.

Additionally, it should be hooked into our build system, integration tests, and CI configuration.

### Where?

https://github.com/FerretDB/FerretDB/tree/main/internal/backends

### Definition of Done

- backend created;
- unit tests added/updated;
- integration/compatibility tests added/updated;
- spot refactorings done;
- user documentation updated or an issue to create documentation created;
- something else?

adetunjii:
@AlekSi Please I would like to work on this ticket.

AlekSi:
@adetunjii any news? Anything we could help you with?

adetunjii:
@AlekSi I'm on it. I should be able to raise the first PR on it soon

AlekSi:
@adetunjii please start with the backend code. We are going to change a lot of code around handlers: #3662

AlekSi:
@adetunjii poke :)

adetunjii:
Hi @AlekSi. Sorry I've been quite busy lately. I've just pushed a draft PR. It's still very scattered, I'm trying to fit the pieces together.

AlekSi:
@adetunjii, for the next step, I think we should integrate MySQL backend stub into our integration tests: add `test-integration-mysql` Taskfile.yml task, flags to setup.go, etc

AlekSi:
@adetunjii Is anything left in this issue?

AlekSi:
@adetunjii As a next step, we probably should add MySQL backend there: https://github.com/FerretDB/FerretDB/blob/main/internal/backends/helpers_test.go#L40

AlekSi:
We don't have immediate plans to support the MySQL backend as we re-focus on PostgreSQL using the DocumentDB extension: https://blog.ferretdb.io/ferretdb-releases-v2-faster-more-compatible-mongodb-alternative/

PR #3782, opened 2023-11-29
Fix `checkcomments` linter

noisersup:
# Description

I've noticed that we don't check if the issue exists, though didn't know if that's important for us, so I've used it in this way.
If we prefer to add such check then let's add it to #2733.

## Readiness checklist

### DIFF
--- a/tools/checkcomments/testdata/testdata.go
+++ b/tools/checkcomments/testdata/testdata.go
@@ -16,7 +16,7 @@
 package testdata
 
 func testCorrect() {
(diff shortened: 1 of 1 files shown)


henvic (approved):

rumyantseva (approved):
Nice catch!

If the issue doesn't exist, we should report that it's invalid. That's definitely smth for my PR!

AlekSi:
The correct temporary fix is just ot reopen #3413 because it is not done :)

PR #3793, opened 2023-11-30
Use single flag to disable all pushdowns

### LINKED ISSUE #3760: Use a single flag to disable all pushdowns
### What should be done?

We should rename "disable filter pushdown" flags to "disable pushdown" that should disable all possible pushdowns: filtering, sorting, etc.

### Where?

All code and docs.

### Definition of Done

- handler updated;
- all backends updated;
- unit tests added/updated;
- spot refactorings done;
- user documentation updated.

noisersup:
# Description

Closes #3760.

## Readiness checklist

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -8,7 +8,7 @@ env:
 vars:
   SHARD_INDEX: 1
   SHARD_TOTAL: 1
-  DISABLE_FILTER_PUSHDOWN: false
+  DISABLE_PUSHDOWN: false
   TEST_RUN: ""
   TEST_TIMEOUT: 35m
   BENCH_TIME: 5s
@@ -216,7 +216,7 @@ tasks:
         -target-tls
         -postgresql-url='postgres://username@127.0.0.1:5432/ferretdb?search_path='
         -compat-url='mongodb://username:password@127.0.0.1:47018/?tls=true&tlsCertificateKeyFile=../build/certs/client.pem&tlsCaFile=../build/certs/rootCA-cert.pem'
-        -disable-filter-pushdown={{.DISABLE_FILTER_PUSHDOWN}}
+        -disable-pushdown={{.DISABLE_PUSHDOWN}}
 
   test-integration-sqlite:
     desc: "Run integration tests for `sqlite` backend"
@@ -241,7 +241,7 @@ tasks:
         -sqlite-url=file:../tmp/sqlite-tests/
         -target-tls
         -compat-url='mongodb://username:password@127.0.0.1:47018/?tls=true&tlsCertificateKeyFile=../build/certs/client.pem&tlsCaFile=../build/certs/rootCA-cert.pem'
-        -disable-filter-pushdown={{.DISABLE_FILTER_PUSHDOWN}}
+        -disable-pushdown={{.DISABLE_PUSHDOWN}}
 
   test-integration-mysql:
     desc: "Run integration tests for `mysql` handler"
@@ -266,7 +266,7 @@ tasks:
         -target-tls
         -mysql-url='mysql://username:password@127.0.0.1:3306/ferretdb'
         -compat-url='mongodb://username:password@127.0.0.1:47018/?tls=true&tlsCertificateKeyFile=../build/certs/client.pem&tlsCaFile=../build/certs/rootCA-cert.pem'
-        -disable-filter-pushdown={{.DISABLE_FILTER_PUSHDOWN}}
+        -disable-pushdown={{.DISABLE_PUSHDOWN}}
 
   test-integration-hana:
     desc: "Run integration tests for `hana` handler"
@@ -291,7 +291,7 @@ tasks:
         -target-tls
         -hana-url=$FERRETDB_HANA_URL
         -compat-url='mongodb://username:password@127.0.0.1:47018/?tls=true&tlsCertificateKeyFile=../build/certs/client.pem&tlsCaFile=../build/certs/rootCA-cert.pem'
-        -disable-filter-pushdown={{.DISABLE_FILTER_PUSHDOWN}}
+        -disable-pushdown={{.DISABLE_PUSHDOWN}}
 
   test-integration-mongodb:
     desc: "Run integration tests for MongoDB"
--- a/cmd/ferretdb/main.go
+++ b/cmd/ferretdb/main.go
@@ -92,8 +92,8 @@ var cli struct {
 	Test struct {
 		RecordsDir string `default:"" help:"Testing: directory for record files."`
 
-		DisableFilterPushdown bool `default:"false" help:"Experimental: disable filter pushdown."`
-		EnableOplog           bool `default:"false" help:"Experimental: enable capped collections, tailable cursors and OpLog." hidden:""`
+		DisablePushdown bool `default:"false" help:"Experimental: disable all pushdown."`
+		EnableOplog     bool `default:"false" help:"Experimental: enable capped collections, tailable cursors and OpLog." hidden:""`
 
 		//nolint:lll // for readability
 		Telemetry struct {
@@ -395,8 +395,8 @@ func run() {
 		MySQLURL: mySQLFlags.MySQLURL,
 
 		TestOpts: registry.TestOpts{
-			DisableFilterPushdown: cli.Test.DisableFilterPushdown,
-			EnableOplog:           cli.Test.EnableOplog,
+			DisablePushdown: cli.Test.DisablePushdown,
+			EnableOplog:     cli.Test.EnableOplog,
 		},
 	})
 	if err != nil {
--- a/integration/Taskfile.yml
+++ b/integration/Taskfile.yml
@@ -75,7 +75,7 @@ tasks:
         -bench-docs={{.BENCH_DOCS}}
         -target-backend=ferretdb-postgresql
         -postgresql-url=postgres://username@127.0.0.1:5432/ferretdb
-        -disable-filter-pushdown
+        -disable-pushdown
         | tee new-postgresql.txt
       - ../bin/benchstat{{exeExt}} old-postgresql.txt new-postgresql.txt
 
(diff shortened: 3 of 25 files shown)


cmd/ferretdb/main.go
```diff
@@ -92,8 +92,8 @@ var cli struct {
 	Test struct {
 		RecordsDir string `default:"" help:"Testing: directory for record files."`
 
-		DisableFilterPushdown bool `default:"false" help:"Experimental: disable filter pushdown."`
-		EnableOplog           bool `default:"false" help:"Experimental: enable capped collections, tailable cursors and OpLog." hidden:""`
+		DisablePushdown bool `default:"false" help:"Experimental: disable all pushdown."`
```
AlekSi, line 95: I think it should be "all pushdowns" or (probably better) just "pushdown", without "all"

integration/setup/setup.go
```diff
@@ -60,7 +60,7 @@ var (
 	debugSetupF = flag.Bool("debug-setup", false, "enable debug logs for tests setup")
 	logLevelF   = zap.LevelFlag("log-level", zap.DebugLevel, "log level for tests")
 
-	disableFilterPushdownF = flag.Bool("disable-filter-pushdown", false, "disable filter pushdown")
+	disablePushdownF = flag.Bool("disable-pushdown", false, "disable all pushdown")
```
AlekSi, line 63: The same

AlekSi (dismissed):

Taskfile.yml
```diff
@@ -216,7 +216,7 @@ tasks:
         -target-tls
         -postgresql-url='postgres://username@127.0.0.1:5432/ferretdb?search_path='
         -compat-url='mongodb://username:password@127.0.0.1:47018/?tls=true&tlsCertificateKeyFile=../build/certs/client.pem&tlsCaFile=../build/certs/rootCA-cert.pem'
-        -disable-filter-pushdown={{.DISABLE_FILTER_PUSHDOWN}}
+        -disable-all-pushdown={{.DISABLE_ALL_PUSHDOWN}}
```
AlekSi, line 219: > We should rename "disable filter pushdown" flags to "disable pushdown"

Not to "disable all pushdown"

cmd/ferretdb/main.go
AlekSi: That comment was about the help text, not field name

Taskfile.yml
noisersup: Yeah sorry, noticed that right after requesting review

integration/Taskfile.yml
```diff
@@ -133,8 +133,8 @@ tasks:
         -log-level=error
         -bench-docs={{.BENCH_DOCS}}
         -target-backend=ferretdb-mysql
-        -mysql-url=mysql://username:password@127.0.0.1:3306/ferretdb
-        -disable-filter-pushdown
+        -mysql-url=mysql://username:password@127.0.0.1:3306/ferretdb
```
AlekSi, line 136:
```suggestion
        -mysql-url=mysql://username:password@127.0.0.1:3306/ferretdb
```

internal/backends/collection.go
```diff
@@ -76,8 +76,11 @@ func CollectionContract(c Collection) Collection {
 
 // QueryParams represents the parameters of Collection.Query method.
 type QueryParams struct {
-	Filter        *types.Document
-	Sort          *types.Document
+	Filter *types.Document
+	Sort   *types.Document
+
+	DisablePushdown bool
```
AlekSi, line 82: We use it only for checking other parameters, but that should be in the caller. Let's remove it from there

internal/backends/collection.go
```diff
@@ -134,9 +145,10 @@ func (cc *collectionContract) Query(ctx context.Context, params *QueryParams) (*
 
 // ExplainParams represents the parameters of Collection.Explain method.
 type ExplainParams struct {
-	Filter *types.Document
-	Sort   *types.Document
-	Limit  int64
+	Filter          *types.Document
+	Sort            *types.Document
+	DisablePushdown bool
```
AlekSi, line 150: Let's remove it

internal/handler/msg_aggregate.go
```diff
@@ -259,14 +259,17 @@ func (h *Handler) MsgAggregate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
...
+		var sort *types.Document
+
+		if !h.DisablePushdown {
+			filter, sort = aggregations.GetPushdownQuery(aggregationStages)
+		}
 
 		// only documents stages or no stages - fetch documents from the DB and apply stages to them
 		qp := new(backends.QueryParams)
 
-		if !h.DisableFilterPushdown {
```
AlekSi, line 267: That check should be there

internal/backends/collection.go
noisersup: That's not true. It is used by backend to disable capped collection sort pushdown here: https://github.com/FerretDB/FerretDB/pull/3793/files#diff-ed1b75cf02347cc3668ac3770a3e8b1197504bd78835652b97516604eb9f62b8R99

AlekSi: Well, yeah, you added it in this PR. Let's remove that field and its usage

internal/backends/collection.go
```diff
@@ -96,6 +99,9 @@ type QueryResult struct {
 // It also can be used to close the returned iterator and free underlying resources,
 // but doing so is not necessary - the handler will do that anyway.
 //
+// If DisablePushdown is true, no pushdown will be performed.
+// In such case Filter, Limit, and Sort cannot be set.
```
AlekSi, line 103: Essentially, you are having a function `func foo(a *A, b bool)` that says something like "if b is `true`, `foo` behaves as if `a` was `nil`". But then the caller could just pass `nil` to `a`. Parameter `b` is not needed.

So please remove `DisablePushdown` from the backend and arrange the handler to call backend methods with correct parameters.

AlekSi (changes requested):

noisersup:
@mergify yeah yeah I see...

internal/backends/collection.go
```diff
@@ -96,7 +97,14 @@ type QueryResult struct {
 // It also can be used to close the returned iterator and free underlying resources,
 // but doing so is not necessary - the handler will do that anyway.
 //
-// Passed sort document should be already validated. If sort document is invalid, function panics.
+// Filter may be ignored, or safely applied partially or entirely.
+// Extra documents will be filtered out by the handler.
+//
+// Sort should have one of the following forms: nil, {}, {"$natural": int64(1)} or {"$natural": int64(-1)}.
+// Other field names are not supported.
+// If non-empty, it should be applied.
```
AlekSi, line 105: @noisersup PTAL

internal/backends/postgresql/query.go
```diff
@@ -230,21 +230,26 @@ func prepareWhereClause(p *metadata.Placeholder, sqlFilters *types.Document) (st
...
+func prepareOrderByClause(p *metadata.Placeholder, sort *types.Document) (string, []any) {
 	if sort.Len() == 0 {
-		if capped {
-			return fmt.Sprintf(" ORDER BY %s", metadata.RecordIDColumn), nil
-		}
-
 		return "", nil
 	}
 
+	// the following code could be simplified now
```
AlekSi, line 239: Because it is checked by the contract now

internal/backends/postgresql/query.go
```diff
@@ -255,6 +260,7 @@ func prepareOrderByClause(p *metadata.Placeholder, sort *types.Document, capped
 		order = " DESC"
 	}
 
+	// FIXME remove it, it is not safe
```
AlekSi, line 263: Only safe pushdowns now

internal/backends/postgresql/query.go
```diff
@@ -230,21 +230,26 @@ func prepareWhereClause(p *metadata.Placeholder, sqlFilters *types.Document) (st
...
+
 	k := sort.Keys()[0]
 	v := sort.Values()[0].(int64)
 
+	if k == "$natural" {
+		if v == 1 {
+			return fmt.Sprintf(" ORDER BY %s", metadata.RecordIDColumn), nil
+		}
+
+		// FIXME support -1 for $natural
```
AlekSi, line 249: That's an important use-case

internal/backends/sqlite/query.go
```diff
@@ -45,14 +45,27 @@ func prepareSelectClause(table, comment string, capped, onlyRecordIDs bool) stri
...
+// The provided sort document should be already validated.
+// Provided document should only contain a single value.
+func prepareOrderByClause(sort *types.Document) string {
+	if sort.Len() == 0 {
+		return ""
 	}
 
-	// TODO https://github.com/FerretDB/FerretDB/issues/3181
-	return ""
+	// the following code could be simplified now
```
AlekSi, line 57: See PostgreSQL

internal/backends/sqlite/query.go
```diff
@@ -45,14 +45,27 @@ func prepareSelectClause(table, comment string, capped, onlyRecordIDs bool) stri
...
+
+	v, err := sort.Get("$natural")
+	if err != nil {
+		return ""
+	}
+
+	sortOrder := v.(int64)
+	if sortOrder != 1 {
+		// FIXME support -1 for $natural
```
AlekSi, line 66: See PostgreSQL

internal/backends/collection.go
noisersup: Okay, I was a bit worried that we remove and implement the same things all over again, though the need of simplification of code on the backend level is understandable, gonna remove support for any other sort

AlekSi: Yeah, before, we had support for unsafe sorting under the experimental flag, but that flag is gone, but the sorting remained. That's practically a bug on that layer

henvic (approved):

PR #3826, opened 2023-12-07
Extract `find` helper functions

AlekSi:
# Description

Closes #<issue_number>.

## Readiness checklist

### DIFF
--- a/internal/backends/backends.go
+++ b/internal/backends/backends.go
@@ -36,6 +36,7 @@
 //     Contracts enforce error codes; they are not documented in the code comments
 //     but are visible in the contract's code (to avoid duplication).
 //     Methods should return different error codes only if the difference is important for the handler.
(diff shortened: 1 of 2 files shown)


internal/backends/backends.go
```diff
@@ -36,6 +36,7 @@
 //     Contracts enforce error codes; they are not documented in the code comments
 //     but are visible in the contract's code (to avoid duplication).
 //     Methods should return different error codes only if the difference is important for the handler.
+//  6. Passed parameters should not be modified because the handler may reuse them.
```
noisersup, line 39: Understandable, couldn't we pass a struct rather than pointer to struct though?

AlekSi: Yes, except that if that struct contains pointers, slices, or maps, we get a false sense of safety. Go doesn't have immutability; let's not pretend that value semantics is it

PR #3831, opened 2023-12-08
Support capped collection cleanup

### LINKED ISSUE #3459: Support capped collection clean-up

rumyantseva:
# Description

Closes #3459.

## Readiness checklist


internal/handler/handler.go
```diff
@@ -65,31 +85,209 @@ func New(opts *NewOpts) (*Handler, error) {
...
+	h.cleanupedCappedCollections.Collect(ch)
+}
+
+// cleanupCappedCollections drops the given percent of documents from all capped collections.
+func (h *Handler) cleanupCappedCollections(ctx context.Context) error {
+	h.L.Debug("CleanupCappedCollections: started", zap.Uint8("percentage", h.cappedCleanupPercentage))
+	defer h.L.Debug("CleanupCappedCollections: finished")
+
+	// fixme: conninfo must be set, otherwise backend panics
+	connInfo := conninfo.New()
```
AlekSi, line 164:
```suggestion
	connInfo := conninfo.New()
	connInfo.BypassAuth = true
```

see #3840

rumyantseva: Thanks!

integration/commands_administration_test.go
```diff
@@ -1720,6 +1720,68 @@ func TestCommandsAdministrationCompactForce(t *testing.T) {
...
+	t.Parallel()
+
+	s := setup.SetupWithOpts(t, &setup.SetupOpts{
+		DatabaseName: "admin",
+		Providers:    []shareddata.Provider{},
+	})
+
+	for name, tc := range map[string]struct {
+		force          bool
+		skipForMongoDB string // optional, skip test for mongoDB with a specific reason
```
AlekSi, line 1733: Remove it

integration/commands_administration_test.go
```diff
@@ -1720,6 +1720,68 @@ func TestCommandsAdministrationCompactForce(t *testing.T) {
...
+			collName := testutil.CollectionName(t) + name
+
+			opts := options.CreateCollection().SetCapped(true).SetSizeInBytes(1024).SetMaxDocuments(10)
+			err := s.Collection.Database().CreateCollection(s.Ctx, collName, opts)
+			require.NoError(t, err)
+
+			collection := s.Collection.Database().Collection(collName)
+
+			// maxDocuments is set as 10, we insert 11 documents to have one document to compact
+			arr, _ := GenerateDocuments(0, 11)
```
AlekSi, line 1759:
```suggestion
			arr, _ := GenerateDocuments(0, max+1)
```

internal/handler/handler.go
```diff
@@ -65,31 +85,230 @@ func New(opts *NewOpts) (*Handler, error) {
 		b = oplog.NewBackend(b, opts.L.Named("oplog"))
 	}
 
+	if opts.CappedCleanupPercentage > 100 {
```
AlekSi, line 88: I think we should return an error there (and also check for `> 99`, not 100)

internal/handler/handler.go
```diff
@@ -65,31 +85,230 @@ func New(opts *NewOpts) (*Handler, error) {
...
+		opts.CappedCleanupPercentage = 100
+	}
+
+	if opts.CappedCleanupPercentage == 0 {
+		opts.CappedCleanupPercentage = 10
+	}
+
+	if opts.CappedCleanupInterval == 0 {
+		opts.CappedCleanupInterval = 1 * time.Minute
+	}
```
AlekSi, line 98: While we check things, we should check that it is not negative

internal/handler/handler.go
```diff
@@ -65,31 +85,230 @@ func New(opts *NewOpts) (*Handler, error) {
 		b = oplog.NewBackend(b, opts.L.Named("oplog"))
 	}
 
+	if opts.CappedCleanupPercentage > 100 {
+		opts.CappedCleanupPercentage = 100
+	}
+
+	if opts.CappedCleanupPercentage == 0 {
+		opts.CappedCleanupPercentage = 10
```
AlekSi, line 93: I think that defaults belong to the CLI, not to the logic there

internal/handler/handler.go
```diff
@@ -65,31 +85,230 @@ func New(opts *NewOpts) (*Handler, error) {
...
+	}
+
 	return h, nil
 }
 
 // Close gracefully shutdowns handler.
 // It should be called after listener closes all client connections and stops listening.
 func (h *Handler) Close() {
 	h.cursors.Close()
+	h.cappedCleanupStop <- struct{}{}
```
AlekSi, line 156: We should also wait for that goroutine to stop

internal/handler/handler.go
```diff
@@ -65,31 +88,243 @@ func New(opts *NewOpts) (*Handler, error) {
...
+		defer h.wg.Done()
+
+		h.runCappedCleanup()
+	}()
+
 	return h, nil
 }
 
+// runCapppedCleanup calls capped collections cleanup function according to the given interval.
+func (h *Handler) runCappedCleanup() {
```
henvic, line 138:
```suggestion
func (h *Handler) runCappedCleanup(ctx context.Context) {
```

internal/handler/handler.go
```diff
@@ -65,31 +88,243 @@ func New(opts *NewOpts) (*Handler, error) {
...
+		return
+	}
+
+	ticker := time.NewTicker(h.CappedCleanupInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			if err := h.cleanupAllCappedCollections(context.Background()); err != nil {
```
henvic, line 155:
```suggestion
			if err := h.cleanupAllCappedCollections(ctx); err != nil {
```

Otherwise, context.TODO() is, perhaps, more appropriate.

internal/handler/handler.go
```diff
@@ -65,31 +88,243 @@ func New(opts *NewOpts) (*Handler, error) {
...
 
 	h.initCommands()
 
+	h.wg.Add(1)
+
+	go func() {
+		defer h.wg.Done()
+
+		h.runCappedCleanup()
+	}()
```
henvic, line 132:
```suggestion
```

henvic: I'd consider exposing a `StartCappedCleanup` or `RunCappedCleanup` function which you'd use as following as it might come as a surprise that a New function is starting a goroutine. It also might make it easier to test some internal details, if we ever want to.

```go
go r.RunCappedCleanup(ctx)
```

AlekSi: Yeah, we discussed just that, but decided to keep it as is for now

AlekSi (approved):

PR #3840, opened 2023-12-12
Allow bypassing authentication

### LINKED ISSUE #3459: Support capped collection clean-up

AlekSi:
# Description

Closes #3459.

## Readiness checklist

### DIFF
--- a/internal/backends/postgresql/metadata/pool/pool.go
+++ b/internal/backends/postgresql/metadata/pool/pool.go
@@ -134,6 +134,22 @@ func (p *Pool) Get(username, password string) (*pgxpool.Pool, error) {
 	return res, nil
 }
 
+// GetAny returns a random open pool of connections to PostgreSQL, or nil if none are available.
+func (p *Pool) GetAny() *pgxpool.Pool {
+	p.rw.RLock()
+	defer p.rw.RUnlock()
+
+	for _, pool := range p.pools {
+		p.l.Debug("Pool.GetAny: returning existing pool")
+
+		return pool
+	}
+
+	p.l.Debug("Pool.GetAny: no existing pools")
+
+	return nil
+}
+
 // Describe implements prometheus.Collector.
 func (p *Pool) Describe(ch chan<- *prometheus.Desc) {
 	prometheus.DescribeByCollect(p, ch)
--- a/internal/backends/postgresql/metadata/registry.go
+++ b/internal/backends/postgresql/metadata/registry.go
@@ -106,7 +106,8 @@ func (r *Registry) Close() {
 }
 
 // getPool returns a pool of connections to PostgreSQL database
-// for the username/password combination in the context using [conninfo].
+// for the username/password combination in the context using [conninfo]
+// (or any pool if authentication is bypassed).
 //
 // It loads metadata if it hasn't been loaded from the database yet.
 //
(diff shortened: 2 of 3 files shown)


rumyantseva (approved):
LGTM.

Side question: so, we always store every new username-password group in the connection pool. But we don't cleanup the pool, right? E.g. if 100 users connect with different usernames and passwords, we will store them all unless FerretDB is stopped?

internal/backends/postgresql/metadata/pool/pool.go
```diff
@@ -134,6 +134,22 @@ func (p *Pool) Get(username, password string) (*pgxpool.Pool, error) {
 	return res, nil
 }
 
+// GetAny returns a random open pool of connections to PostgreSQL, or nil if none are available.
+func (p *Pool) GetAny() *pgxpool.Pool {
+	p.rw.RLock()
+	defer p.rw.RUnlock()
+
+	for _, pool := range p.pools {
```
henvic, line 142: I think this clever approach of getting a first map element might backfire as the same element might be returned multiple times (as in, with a bad distribution). What do you think?

AlekSi: Well, more reasons to finish authentication faster 😆 

Seriously, though, the runtime actually makes effort to randomize iterations over maps, so it is as bad as map implementation (which is good enough)

AlekSi:
> we always store every new username-password group in the connection pool.

Pools.

> But we don't cleanup the pool, right?

Right, we don't clean-up pools.

> E.g. if 100 users connect with different usernames and passwords, we will store them all unless FerretDB is stopped?

If there are 100 _valid_ username/password combinations, then yes,  in that case we store 100 strings and 100 pointer. Each pointer point to a pool of connections which automatically scales down to zero: https://github.com/FerretDB/FerretDB/blob/6737d3260e4059182e8fe59af64a3549dcb2ecf9/internal/backends/postgresql/metadata/pool/uri.go#L28-L31

henvic (approved):

xaitx:
Why would this cause authentication bypass

AlekSi:
Authentication is bypassed when `ConnInfo`'s (object representing client connection information) field `BypassAuth` is set to true. It is not set for normal connections, but needed for background workers such as capped collection clean-up (#3459).

Issue #3784, opened 2023-11-30
Support `SCRAM-SHA-256` authentication mechanism


AlekSi:
### What should be done?

We should implement `SCRAM-SHA-256` authentication mechanism. Credentials should be stored and validated on login. Enforcing authentication is in #1877.

The research prototype is in #3904, but its code should not be used. And we probably need to figure out integration testing first: both auth mechanisms (`PLAIN` and `SCRAM-SHA-256`) and new authentication enabled ~and disabled~.

### Where?

Handler.

### Definition of Done

- integration/compatibility tests added/updated;
- handler updated;
- spot refactorings done.

b1ron:
Credentials should be stored when a user is created in https://github.com/FerretDB/FerretDB/blob/c8f8c6851e4a07ae5cc9297ae2a5019bcea6c7f1/internal/handler/msg_createuser.go#L34

Neustradamus:
It is an important feature!

Linked to:
- https://github.com/scram-sasl/info/issues/1

Neustradamus:
Linked to:
- https://github.com/scram-sasl/info/issues/1
- https://github.com/FerretDB/FerretDB/issues/2012
- https://github.com/FerretDB/FerretDB/issues/3784
- https://github.com/FerretDB/FerretDB/pull/3904
- https://github.com/FerretDB/FerretDB/issues/3938
- https://github.com/FerretDB/FerretDB/pull/3970
- https://github.com/FerretDB/FerretDB/pull/3989

AlekSi:
@henvic There are TODO comments that refer to this issue

henvic:
@AlekSi, closing it with #4086.

Issue #3778, opened 2023-11-29
Support `saslSupportedMechs`


AlekSi:
### What should be done?

We should add new authentication mechanisms `SCRAM-SHA-1` and `SCRAM-SHA-256` where they are advertised in addition to `PLAIN`:
* `getParameter`'s `authenticationMechanisms`;
* somewhere else?

### Where?

TBD

### Definition of Done

- handler updated;
- integration/compatibility tests added/updated;
- spot refactorings done.

Issue #3777, opened 2023-11-29
Support `speculativeAuthenticate`


AlekSi:
### What should be done?

TBD

### Where?

TBD

### Definition of Done

TBD

- handler updated;
- all backends updated;
- unit tests added/updated;
- integration/compatibility tests added/updated;
- spot refactorings done;
- user documentation updated or an issue to create documentation created;
- something else?

AlekSi:
There are open TODOs

chilagrow:
Let me take a look for those todos

PR #3848, opened 2023-12-13
Implement `createUser` command

### LINKED ISSUE #1491: Implement `createUser` command
### What should be done?

We should implement the basic [`createUser` command](https://www.mongodb.com/docs/manual/reference/command/createUser/).

We should store the following information in the `system.users` collection in the `admin` database:

```
{
  _id: "<database name>.<username>",
  userID: types.Binary{Subtype: types.BinaryUUID, B: <UUID>},
  user: "<username>",
  credentials: {}
}
```

We should accept the following `createUser` arguments:
* `$db` – ignore for now;
* `createUser`;
* `pwd` – ignore for now;
* `customData` – ignore non-default value for now;
* `roles` – ignore for now;
* `digestPassword` – should be `true` or absent/null;
* `writeConcern` – ignore;
* `authenticationRestrictions` – ignore;
* `mechanisms` – ignore for now;
* `comment` – ignore for now.

We should test that data is inserted using the regular `find` command.

### Where?

### Definition of Done

- handler updated;
- spot refactorings done.

henvic:
# Description

Implemente createUser command.

Closes #1491

## Readiness checklist


integration/setup/setup.go
```diff
@@ -60,6 +60,7 @@ var (
 	logLevelF   = zap.LevelFlag("log-level", zap.DebugLevel, "log level for tests")
 
 	disablePushdownF = flag.Bool("disable-pushdown", false, "disable pushdown")
+	enableNewAuthF   = flag.Bool("enable-new-auth", false, "enable new auth")
```
AlekSi, line 63: Let's remove that flag and always enable it in tests

AlekSi (commented):
CI fails

integration/setup/setup.go
henvic: Fixed.

integration/users/create_user_test.go
```diff
@@ -0,0 +1,185 @@
...
+				{"createUser", "missing_roles"},
+				{"pwd", "password"},
+			},
+			err: &mongo.CommandError{
+				Code:    40414,
+				Name:    "Location40414",
+				Message: "BSON field 'createUser.roles' is missing but a required field",
+			},
+		},
+	}
```
rumyantseva, line 119: Are there any other validation-related cases? E.g. user name contains invalid symbols or smth like this. Or is it out of scope for now?

internal/handler/msg_createuser.go
```diff
@@ -29,6 +37,116 @@ func (h *Handler) MsgCreateUser(ctx context.Context, msg *wire.OpMsg) (*wire.OpM
...
+	}
+
+	if dbName != "$external" && !document.Has("pwd") {
+		return nil, handlererrors.NewCommandErrorMsg(
+			handlererrors.ErrBadValue,
+			"Must provide a 'pwd' field for all user documents, except those with '$external' as the user's source db",
+		)
+	}
+
+	// https://www.mongodb.com/docs/manual/reference/command/createUser/
```
rumyantseva, line 52: Hm, do we need this link here? We don't have links in the implementation of other commands.

internal/handler/msg_createuser.go
```diff
@@ -29,6 +37,116 @@ func (h *Handler) MsgCreateUser(ctx context.Context, msg *wire.OpMsg) (*wire.OpM
...
 	}
 
-	// TODO https://github.com/FerretDB/FerretDB/issues/1491
-	return nil, common.Unimplemented(document, document.Command())
+	dbName, err := common.GetRequiredParam[string](document, "$db")
+	if err != nil {
+		return nil, err
+	}
+
+	if dbName != "$external" && !document.Has("pwd") {
```
rumyantseva, line 45: We don't handle `$external` in the current issue, right? Do we need a separate issue to support it?

internal/handler/msg_createuser.go
```diff
@@ -29,6 +37,116 @@ func (h *Handler) MsgCreateUser(ctx context.Context, msg *wire.OpMsg) (*wire.OpM
...
+				handlererrors.ErrMissingField,
+				"BSON field 'createUser.roles' is missing but a required field",
+			)
+		}
+
+		return nil, lazyerrors.Error(err)
+	}
+
+	if err = common.UnimplementedNonDefault(document, "roles", func(v any) bool {
+		if v == nil || v == types.Null {
```
rumyantseva, line 85: I think this line might conflict with the previous check where we ask `roles` to be required (so `v` can't be `nil` here).

internal/handler/msg_createuser.go
```diff
@@ -29,6 +37,116 @@ func (h *Handler) MsgCreateUser(ctx context.Context, msg *wire.OpMsg) (*wire.OpM
...
+	common.Ignored(document, h.L, "pwd", "writeConcern", "authenticationRestrictions", "mechanisms", "comment")
+
+	id := uuid.New()
+	saved := must.NotFail(types.NewDocument(
+		"_id", dbName+"."+username,
+		"credentials", types.MakeDocument(0),
+		"user", username,
+		"db", dbName,
+		"roles", types.MakeArray(0),
+		"userId", types.Binary{Subtype: types.BinaryUUID, B: id[:]},
```
rumyantseva, line 116: I think `id[:]` could be `id.MarshalBinary()`, we used it somewhere else like this.

integration/users/create_user_test.go
```diff
@@ -0,0 +1,185 @@
...
+	"go.mongodb.org/mongo-driver/mongo"
+
+	"github.com/FerretDB/FerretDB/integration"
+	"github.com/FerretDB/FerretDB/integration/setup"
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/must"
+	"github.com/FerretDB/FerretDB/internal/util/testutil"
+)
+
+func TestCreateUser(t *testing.T) {
```
rumyantseva, line 34: Do we also need a test where we try to create a user while being connected to a non-`admin` DB?

internal/handler/msg_createuser.go
```diff
@@ -29,6 +37,116 @@ func (h *Handler) MsgCreateUser(ctx context.Context, msg *wire.OpMsg) (*wire.OpM
...
+	))
+
+	// Users are saved in the "admin" database.
+	adminDB, err := h.b.Database("admin")
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+
+	users, err := adminDB.Collection("system.users")
```
rumyantseva, line 125: I think `users` can be `nil` (and no error) if `system.users` doesn't exist.

internal/handler/msg_createuser.go
```diff
@@ -29,6 +37,116 @@ func (h *Handler) MsgCreateUser(ctx context.Context, msg *wire.OpMsg) (*wire.OpM
...
+		"_id", dbName+"."+username,
+		"credentials", types.MakeDocument(0),
+		"user", username,
+		"db", dbName,
+		"roles", types.MakeArray(0),
+		"userId", types.Binary{Subtype: types.BinaryUUID, B: id[:]},
+	))
+
+	// Users are saved in the "admin" database.
+	adminDB, err := h.b.Database("admin")
```
rumyantseva, line 120: I think `adminDB` can be `nil` (and no error) if `admin` database doesn't exist.

rumyantseva (commented):
Overall, the implementation looks according to the task to me. I asked a few questions.

integration/users/create_user_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+			payload := integration.ConvertDocument(t, tc.payload)
+			assertUserExists(ctx, t, users, db.Name(), payload)
+		})
+	}
+}
+
+// assertUserExists checks it the user was created in the admin.system.users collection.
+// All users are created in the "admin" database.
+func assertUserExists(ctx context.Context, t testing.TB, users *mongo.Collection, dbName string, payload *types.Document) {
```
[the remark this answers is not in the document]
henvic: No good reason besides I thought it'd be used in more than a point in the test but I didn't end up using it this way. Getting rid of it now.

integration/users/create_user_test.go
rumyantseva: (maybe out of scope for this task, but then we might need a todo in implementation)

integration/users/create_user_test.go
henvic: I haven't found what rules are used for validation and testing it seems to be pretty liberal, but one thing I missed was a test case for checking emptiness. Doing it now.

internal/handler/msg_createuser.go
henvic: We don't handle it at this current issue. I think this will be handled by #3777, #3778, and #3784 if I'm not mistaken.

rumyantseva: Let's add a `TODO` then, so we don't miss it ;)

integration/users/create_user_test.go
henvic: We use the non-admin DB `TestCreateUser` for the connection.
Internally, we're using the `admin` database to hold the users data, but this is transparent for the user.

internal/handler/msg_createuser.go
henvic: I took a look at the code for the Collection method and did a small test calling `.Collection("xuxu")` and it worked fine, rather than returning nil. It looks like it does some magic behind the scenes.

internal/handler/msg_createuser.go
henvic: Similar to Collection. I was also able to insert a value on a collection that didn't exist on a database that didn't exist without getting nulls in either steps.

internal/handler/msg_createuser.go
rumyantseva: ah, you are right! actually, no magic, a particular implementation for SQLite/PG just always returns a collection.

AlekSi: > a particular implementation

It is a part of the contract: https://github.com/FerretDB/FerretDB/blob/4de769a5d6fb2f73fa6ab4468c7eda9d0786d619/internal/backends/database.go#L65

internal/handler/msg_createuser.go
henvic: Added.

rumyantseva (approved):
LGTM

noisersup (approved):
Makes sense! LGTM

integration/users/create_user_test.go
```diff
@@ -0,0 +1,185 @@
...
+			payload: bson.D{
+				{"createUser", "success_user"},
+				{"roles", bson.A{}},
+				{"pwd", "password"},
+			},
+			expected: bson.D{
+				{
+					"ok", float64(1),
+				},
+			},
```
AlekSi, line 103:
```suggestion
			expected: bson.D{
				{"ok", float64(1)},
			},
```

or

```suggestion
			expected: bson.D{{
				"ok", float64(1),
			}},
```

Both suggestions are common in our code base, the current style of placing brackets is not

integration/users/create_user_test.go
```diff
@@ -0,0 +1,185 @@
...
+		},
+	}
+
+	err := db.RunCommand(ctx, bson.D{
+		{"createUser", "should_already_exist"},
+		{"roles", bson.A{}},
+		{"pwd", "password"},
+	}).Err()
+	assert.NoError(t, err)
```
AlekSi, line 136: require

integration/users/create_user_test.go
```diff
@@ -0,0 +1,185 @@
...
+			err: &mongo.CommandError{
+				Code:    40414,
+				Name:    "Location40414",
+				Message: "BSON field 'createUser.roles' is missing but a required field",
+			},
+		},
+	}
+
+	err := db.RunCommand(ctx, bson.D{
+		{"createUser", "should_already_exist"},
```
AlekSi, line 132: I think that's the place where a comment would help. Otherwise, it like: "ok, so we create a user that should already exist… but why is that does not return an error? and where this pre-existing user is created?.."

internal/handler/msg_createuser.go
```diff
@@ -29,6 +37,120 @@ func (h *Handler) MsgCreateUser(ctx context.Context, msg *wire.OpMsg) (*wire.OpM
...
+			return true
+		}
+
+		dp, ok := v.(bool)
+		return ok && dp
+	}); err != nil {
+		return nil, err
+	}
+
+	// NOTE: In MongoDB, the comment field isn't saved in the database, but used for log and profiling.
```
AlekSi, line 110: That's true for all `comment` parameters; they are very common in MongoDB. That comment does not add anything, in my opinion

AlekSi (approved):
LGTM overall. Please ack my comments, resolve them and fix/implement in the next PR for another issue

integration/users/create_user_test.go
```diff
@@ -0,0 +1,185 @@
...
+	users := client.Database("admin").Collection("system.users")
+
+	// TODO https://github.com/FerretDB/FerretDB/issues/1492
+	if setup.IsMongoDB(t) {
+		assert.NoError(t, collection.Database().RunCommand(ctx, bson.D{
+			{"dropAllUsersFromDatabase", 1},
+		}).Err())
+	} else {
+		// Erase any previously saved user in the database.
+		_, err := users.DeleteMany(ctx, bson.D{{"db", db.Name()}})
```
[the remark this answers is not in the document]
henvic: I don't see how this can be a problem from the point-of-view of this test given that db.Name() is the name of the the test (same as t.Name()), and a `t.Parallel()` only runs a test with other tests.

From the docs:

```
// Parallel signals that this test is to be run in parallel with (and only with)
// other parallel tests.
```

website/versioned_docs/version-v1.15/reference/supported-commands.md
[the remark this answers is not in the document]
henvic: Oops! Sorry about that and thanks for letting me know!

PR #3850, opened 2023-12-13
Add MySQL backend registry

adetunjii:
Closes #3413

## Readiness checklist


internal/util/testutil/mysql.go
```diff
@@ -63,7 +64,7 @@ func TestMySQLURI(tb testtb.TB, ctx context.Context, baseURI string) string {
 		User:   u.User.Username(),
 		Passwd: password,
 		Addr:   u.Host,
-		DBName: u.Path,
+		DBName: strings.TrimPrefix(u.Path, "/"),
```
AlekSi, line 67: I think we should do the same thing as we do for PostgreSQL and SQLite now – create test-specific MySQL databases. That solves a lot of problems with parallel tests.

https://github.com/FerretDB/FerretDB/blob/c4b84a39094ab71ab5fb3ef33d9c74bda7de05d2/internal/util/testutil/postgresql.go#L53-L55

https://github.com/FerretDB/FerretDB/blob/c4b84a39094ab71ab5fb3ef33d9c74bda7de05d2/internal/util/testutil/postgresql.go#L71-L92

adetunjii: I am doing the same thing as we do in PostgreSQL. What I'm doing here is to first create a connection string from the config struct to connect to MySQL so that I can create a test database then return a URI with the newly created database as the path.

AlekSi (approved):

henvic (dismissed):

AlekSi (commented):
Test fails

AlekSi (commented):
Test fails

AlekSi (changes requested):
Test fails

henvic (approved):

AlekSi (approved):

PR #3854, opened 2023-12-14
Simplify tailable cursor tests

AlekSi:
# Description

Use a simpler helper, check where `t` vs `tt` is used.

## Readiness checklist


internal/util/debug/debug.go
```diff
@@ -108,7 +109,7 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
 			l.Sugar().Infof("%s%s - %s", root, path, handlers[path])
 		}
 
-		if err := s.Serve(lis); err != http.ErrServerClosed {
+		if err := s.Serve(lis); !errors.Is(err, http.ErrServerClosed) {
```
henvic, line 112: By the Go compat guarantee promise, this should always be http.ErrServerClosed for this case.

henvic (approved):

internal/util/debug/debug.go
AlekSi: That's true, but the linter is not intelligent enough to understand that, and using `//nolint` with an explanation would take more effort 😆

noisersup (approved):

PR #3890, opened 2023-12-20
Add utilities for working with passwords

### LINKED ISSUE #3785: Store passwords for `PLAIN` authentication mechanism
### What should be done?

For the `PLAIN` authentication method, we should store the password securely. We should do it in a way similar to MongoDB:

```
[
  {
    _id: 'test.username',
    userId: UUID('74d81c52-0b4c-4904-8930-7bd1b72f4e04'),
    user: 'username',
    db: 'test',
    credentials: {
      PLAIN: {
        algo: 'argon2id',
        t: int32(3),
        p: int32(4),
        m: int32(65536),
        hash: types.Binary{…},
        salt: types.Binary{…},
      }
    },
    roles: []
  }
]
```

`credentials.PLAIN` document is provided by the `internal/util/password.PlainHash` function.

We should update `createUser` method to store password.
We should update `usersInfo` method to return credentials when called with `showCredentials: true`.
Actually using/validating password for authentication is not in the scope of this issue.

### Where?

`createUser` and `usersInfo` command handlers.

### Definition of Done

- handler updated;
- spot refactorings done.

AlekSi:
# Description

Closes #3785.

## Readiness checklist


internal/util/password/plain.go
```diff
@@ -0,0 +1,162 @@
...
+	p, _ := v.(int32)
+
+	v, _ = doc.Get("m")
+	m, _ := v.(int32)
+
+	v, _ = doc.Get("hash")
+	hash, _ := v.(types.Binary)
+
+	v, _ = doc.Get("salt")
+	salt, _ := v.(types.Binary)
```
henvic, line 126: Is it okay if it panics due to incompatible type? Asking because we might want to be more strict here.

AlekSi: All those values can't be zero/empty, so the normal error will be returned below.

henvic: oh, now I've noticed that _ was used.

rumyantseva (approved):
Looks clear! 👍

PR #3895, opened 2023-12-21
Implement `/debug/archive` handler

### LINKED ISSUE #3584: Implement `/debug/archive`
### What should be done?

We should implement a new debug handler at `/debug/archive` that would:
* self-scrape Prometheus metrics via `/debug/metrics`;
* take a binary memory pprof profile via `/debug/pprof/heap`;
* pack everything into a .zip archive and send it to the client.

We will add parameters to that handler and add more files to that archive in the future.

The file should be sent with a good content-disposition header and streamed to the client; for example, .zip header and Prometheus metrics could be sent before the memory profile is taken.

### Where?

https://github.com/FerretDB/FerretDB/blob/main/internal/util/debug/debug.go

### Definition of Done

- handler implemented;
- basic unit tests added;
- spot refactorings done.

sachinpuranik:
Closes #3584.

## Readiness checklist


internal/util/debug/debug.go
```diff
@@ -37,11 +41,23 @@ import (
 	"github.com/FerretDB/FerretDB/internal/util/must"
 )
 
+var (
+	graphicsPath = "/debug/graphs"
```
AlekSi, line 45:
```suggestion
	graphsPath = "/debug/graphs"
```

internal/util/debug/debug.go
```diff
@@ -37,11 +41,23 @@ import (
...
+	metricsPath  = "/debug/metrics"
+	archivePath  = "/debug/archive"
+	pprofPath    = "/debug/pprof"
+	varsPath     = "/debug/vars"
+)
+
+func generateFileName(prefix, filename string) string {
```
AlekSi, line 52: I don't think it is worth extracting that function

internal/util/debug/debug.go
```diff
@@ -123,3 +142,80 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
 	s.Close()
 	l.Sugar().Info("Debug server stopped.")
 }
+
+func getArchiveHandler() http.HandlerFunc {
```
AlekSi, line 146: https://github.com/FerretDB/FerretDB/blob/b9eafc581920ab30b1546cc46141486ee9945a1a/CONTRIBUTING.md?plain=1#L263

internal/util/debug/debug.go
```diff
@@ -123,3 +142,80 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
...
+	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
```
AlekSi, line 147: I think the function itself could have that signature instead of returning a function with that signature

internal/util/debug/debug.go
```diff
@@ -123,3 +142,80 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
...
+		var (
+			scheme          string
+			u               url.URL
+			responses       = make(map[string][]byte, 2)
+			debugFilePrefix = "FerretDB"
+			debugFileName   = "debug.zip"
+			err             error
+		)
```
AlekSi, line 155: I think we should define variable when we need them, not in advance

internal/util/debug/debug.go
```diff
@@ -123,3 +142,80 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
...
+
+		u.Path = metricsPath
+		u.Host = req.Host
+
+		if req.URL.Scheme == "" {
```
AlekSi, line 160: It can't be anything other than "http"

internal/util/debug/debug.go
```diff
@@ -123,3 +142,80 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
...
+
+	req.Header.Set("Accept", "application/json")
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
```
AlekSi, line 216: There should not be a need to read all data into a slice and then write this slice. We have a reader there; we use `io.Copy` above that could accept that reader.

internal/util/debug/debug.go
```diff
@@ -123,3 +142,80 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
...
+
+		for fileName, response := range responses {
+			fileWriter, err := zipWriter.Create(fileName)
+			if err != nil {
+				http.Error(rw, fmt.Sprintf("fail creating file %s , (%s)", fileName, err.Error()), http.StatusInternalServerError)
+				return
+			}
+			_, err = io.Copy(fileWriter, bytes.NewReader(response))
+			if err != nil {
+				http.Error(rw, fmt.Sprintf("failed - adding %s to zip , (%s)", fileName, err.Error()), http.StatusInternalServerError)
```
AlekSi, line 195: I think we should log error like that, but don't return them

internal/util/debug/debug.go
```diff
@@ -123,3 +142,80 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
...
+		metricsFileName := generateFileName(debugFilePrefix, filepath.Base(metricsPath)+".txt")
+		responses[metricsFileName], err = performRequest(u)
+		if err != nil {
+			http.Error(rw, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		u.Path = pprofPath
+		pprofFileName := generateFileName(debugFilePrefix, filepath.Base(pprofPath)+".html")
+		responses[pprofFileName], err = performRequest(u)
```
AlekSi, line 175: We should get a memory profile, not HTML page with a list of profiles

internal/util/debug/debug_test.go
```diff
@@ -14,8 +14,83 @@
...
+	stateProvider, err := state.NewProvider(filename)
+	require.NoError(t, err)
+
+	metricsRegisterer := prometheus.DefaultRegisterer
+	metricsProvider := stateProvider.MetricsCollector(true)
+	metricsRegisterer.MustRegister(metricsProvider)
+
+	l := zap.S()
+
+	go RunHandler(ctx, host, metricsRegisterer, l.Named("debug").Desugar())
```
AlekSi, line 54: That goroutine is never stopped

AlekSi (changes requested):

internal/util/debug/debug_test.go
sachinpuranik: It's a test and thread shall stop automatically as soon as test ends. Do we need explicit stop?

sachinpuranik: WDYT?

internal/util/debug/debug.go
sachinpuranik: ok. Optimised in new impl.

internal/util/debug/debug.go
sachinpuranik: ok.

internal/util/debug/debug.go
sachinpuranik: done.

internal/util/debug/debug_test.go
henvic: I'd gracefully stop by setting a cancellable context and calling `cancel()` from a `t.Cleanup()` just to avoid any log messages that might confuse me.

internal/util/debug/debug.go
sachinpuranik: done

internal/util/debug/debug.go
sachinpuranik: done. added comments.

internal/util/debug/debug.go
sachinpuranik: addressed.

internal/util/debug/debug.go
sachinpuranik: During testing I found that scheme was coming as "", hence this was required.
Any other suggestion?

internal/util/debug/debug.go
sachinpuranik: we can do that. But since this is specifically for debugging , isn't it good to return an error , so for any issues , error is clearly known. it shall not be lost in the logs. WDYT?

sachinpuranik:
Hi Team , PTAL.
@AlekSi @henvic

sachinpuranik:
Question - on the pprof page we have two links... Do we want both the files to come in the archive?
http://127.0.0.1:8088/debug/pprof/goroutine?debug=1
http://127.0.0.1:8088/debug/pprof/goroutine?debug=2

internal/util/debug/debug.go
AlekSi: I'm not sure why we decided to parse HTML, but we definitely should not do that. The URL for the memory profile is fixed: `debug/pprof/heap` (without query parameters).

Also, please don't resolve conversations: https://github.com/FerretDB/FerretDB/blob/c8f8c6851e4a07ae5cc9297ae2a5019bcea6c7f1/CONTRIBUTING.md?plain=1#L358-L362

It might help to re-read the whole file.

AlekSi (changes requested):
https://github.com/FerretDB/FerretDB/pull/3895#discussion_r1441993707

internal/util/debug/debug.go
sachinpuranik: Hi @AlekSi  ,HTML parsing is required since we add dynamic resource tracking in pprof, I am assuming we need them as well in zip. if not , we should improve the task by specifying what's exactly required.  Additionally Debug=1 , exports the data in the legacy text format. I tried exporting few files without this option and they were binary , some non readable info.

sachinpuranik:
request your comments on other places too.

internal/util/debug/debug.go
sachinpuranik: @AlekSi , WDYT?

internal/util/debug/debug.go
sachinpuranik: @AlekSi , WDYT?

internal/util/debug/debug.go
sachinpuranik: Do we only one URL - "debug/pprof/heap" to be grabbed and nothing else.

AlekSi: > we add dynamic resource tracking in pprof

Why do we need that?

> Do we only one URL - "debug/pprof/heap" to be grabbed and nothing else.

Yes.

> Additionally Debug=1 , exports the data in the legacy text format. I tried exporting few files without this option and they were binary , some non readable info.

That format is readable with [`go tool pprof`](https://go.dev/doc/diagnostics#profiling) and contains more information than the legacy text format. We need that.

sachinpuranik: Sure. Updated the code to gather only two files in zip. matrics and heap. test updated accordingly.

AlekSi (commented):
No amount of merging with the `main` branch will make CI pass :)

sachinpuranik:
Hi @AlekSi, I was more of focusing on approach and core implementation.
Updated the go.mod and fixed the issue.
Though "Go / sqlite 3/3 / Run (pull_request) " Still fails due to some cursor test fail. It's unrelated to change ?
How do we tackle it.

internal/util/debug/archive_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+
+	"github.com/FerretDB/FerretDB/internal/util/state"
+)
+
+func findInStringsArray(findStr string, list []string) bool {
```
AlekSi, line 37: https://pkg.go.dev/slices@go1.21.6#Contains

internal/util/debug/archive.go
```diff
@@ -0,0 +1,125 @@
...
+	var (
+		debugFilePrefix = "FerretDB"
+		debugFileName   = "debug.zip"
+	)
+
+	zipName := fmt.Sprintf("%s-%s", debugFilePrefix, debugFileName)
+	rw.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=%s", zipName))
+
+	zipWriter := zip.NewWriter(rw)
+	defer zipWriter.Close() //nolint:errcheck // we are only reading it
```
AlekSi, line 71: I don't think we are only reading _the writer_. It is not safe to ignore that error.

internal/util/debug/archive.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	"archive/zip"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"path/filepath"
+)
+
+type archiveUrlList []url.URL
```
AlekSi, line 27: https://github.com/FerretDB/FerretDB/blob/8d37e8cdc915593107c4eba3f02c50bba70a0e45/CONTRIBUTING.md?plain=1#L263

That's everywhere. Please re-read out contributing documentation.

internal/util/debug/archive.go
```diff
@@ -0,0 +1,125 @@
...
+
+// archiveHandler - is a handler for creating the archive containing all the debug and metrics info.
```
AlekSi, line 29:
```suggestion
// archiveHandler is a handler for creating the archive containing all the debug and metrics info.
```

To follow all other declaration comments.

But that comment is also incorrect – that archive does not contain all debug and metrics info.

internal/util/debug/archive.go
```diff
@@ -0,0 +1,125 @@
...
+func archiveHandler(rw http.ResponseWriter, req *http.Request) {
+	var (
+		scheme  string
+		urlList archiveUrlList
+	)
+
+	urlList = make(archiveUrlList, 0)
```
AlekSi, line 36: https://dave.cheney.net/practical-go/presentations/gophercon-singapore-2019.html#_use_a_consistent_declaration_style

(the page might be useful)

internal/util/debug/archive.go
```diff
@@ -0,0 +1,125 @@
...
+	u := new(url.URL)
+	u.Path = metricsPath
+	u.Host = req.Host
+
+	if req.URL.Scheme == "" {
+		scheme = "http"
+	}
+
+	u.Scheme = scheme
+	urlList = append(urlList, *u)
```
AlekSi, line 47: I'm sorry, but that's a very complicated way. Just use `u := &url.URL{...}`

internal/util/debug/archive.go
```diff
@@ -0,0 +1,125 @@
...
+
+	rw.Header().Set("Content-Type", "application/zip")
+
+	var (
+		debugFilePrefix = "FerretDB"
+		debugFileName   = "debug.zip"
+	)
+
+	zipName := fmt.Sprintf("%s-%s", debugFilePrefix, debugFileName)
```
AlekSi, line 67:
```suggestion
	zipName := "FerretDB-debug.zip"
```

In general, that's the biggest problem in that PR – it is way too complicated for simple things like that one. Simplify. Then simplify some more :)

internal/util/debug/debug.go
```diff
@@ -37,11 +37,20 @@ import (
 	"github.com/FerretDB/FerretDB/internal/util/must"
 )
 
+var (
+	graphsPath  = "/debug/graphs"
```
AlekSi, line 41: Why do we have those variables and why they are variables?

AlekSi (changes requested):

internal/util/debug/archive.go
sachinpuranik: simplified.

internal/util/debug/archive.go
sachinpuranik: simplified the usage.

internal/util/debug/archive.go
sachinpuranik: ok. Made it uniform.

internal/util/debug/archive.go
sachinpuranik: Updated the sentence. This was added when i was adding all the files to zip.

internal/util/debug/archive.go
sachinpuranik: done.

internal/util/debug/archive_test.go
sachinpuranik: done.

internal/util/debug/archive.go
sachinpuranik: Finding : 
Linter shows a message as below for usage of &url . That was the reason I made it like that earlier.

./internal/util/debug/archive.go:35:7: zero value ptr alloc: use new(T) for *T allocation
./internal/util/debug/archive.go:46:6: zero value ptr alloc: use new(T) for *T allocation

internal/util/debug/debug.go
sachinpuranik: Changed to constants. And used these constants everywhere in the code. Some if the URL are used in test as well. hence converted all the URL to constants and used uniformly across.

internal/util/debug/archive.go
sachinpuranik: Done.
modified the code , if error comes now we will return the internal error.
Also lint bypass was not needed post modification. so removed.

sachinpuranik:
HI @AlekSi , do we have a way to trigger failed CI - steps again?

internal/util/debug/archive.go
```diff
@@ -0,0 +1,91 @@
...
+			return
+		}
+	}
+
+	rw.Header().Set("Content-Type", "application/zip")
+	rw.Header().Set("Content-Disposition", "attachment; filename=FerretDB-debug.zip")
+}
+
+// addFileToArchive function responsible for adding content to zip file.
+func addFileToArchive(fileName string, fileReader io.ReadCloser, zipWriter *zip.Writer) error {
```
chilagrow, line 77: Do we need this separate function @noisersup? :thinking:

internal/util/debug/archive_test.go
```diff
@@ -0,0 +1,96 @@
...
+	require.NoError(t, err)
+
+	zipReader, err := zip.NewReader(bytes.NewReader(body), int64(len(body)))
+	require.NoError(t, err)
+
+	require.Equal(t, len(fileList), len(zipReader.File), "needs to be same as length of fileList")
+
+	for _, file := range zipReader.File {
+		t.Logf("\nverifying file : %s", file.Name)
+		require.Equal(t, true, slices.Contains(fileList, file.FileHeader.Name), "file should be present in archive")
```
chilagrow, line 94: Shall we check file is not empty?

chilagrow (commented):
Just minor comments @noisersup

internal/util/debug/debug.go
AlekSi: @noisersup PTAL ^ We don't need `debug=1`

internal/util/debug/archive.go
```diff
@@ -0,0 +1,91 @@
...
+	} {
+		values := url.Values{}
+		values.Add("debug", "1")
+
+		u := url.URL{
+			Scheme:   "http",
+			Host:     req.Host,
+			Path:     path,
+			RawQuery: values.Encode(),
+		}
```
AlekSi, line 49: I still don't understand why we make it so hard.

```
uri := fmt.Sprintf("http://%s%s", addr, path)
```

`addr` is should come from the listener (`req.Host` is just not correct when, for example, a reverse proxy is present).

And we don't need `debug=1` as mentioned elsewhere.

The same in the test.

internal/util/debug/archive.go
```diff
@@ -0,0 +1,91 @@
...
+		if err != nil {
+			http.Error(rw, lazyerrors.Error(err).Error(), http.StatusInternalServerError)
+			return
+		}
+
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			http.Error(rw, lazyerrors.Error(err).Error(), http.StatusInternalServerError)
+			return
+		}
```
AlekSi, line 61: All that is just `http.Get`.

The same in the test

internal/util/debug/archive.go
AlekSi: I guess yeah, for `defer` to work

internal/util/debug/archive.go
```diff
@@ -0,0 +1,91 @@
...
+
+	rw.Header().Set("Content-Type", "application/zip")
+	rw.Header().Set("Content-Disposition", "attachment; filename=FerretDB-debug.zip")
+}
+
+// addFileToArchive function responsible for adding content to zip file.
+func addFileToArchive(fileName string, fileReader io.ReadCloser, zipWriter *zip.Writer) error {
+	defer fileReader.Close() //nolint:errcheck // we are only reading it
+
+	fileWriter, err := zipWriter.Create(fileName)
```
AlekSi, line 80:
```suggestion
	f, err := zipWriter.Create(fileName)
```

https://go.dev/wiki/CodeReviewComments#variable-names

internal/util/debug/archive_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+func TestArchiveHandler(t *testing.T) {
+	t.Parallel()
+
+	// during test some of the files from the below list shall be absent
+	fileList := []string{
+		"metrics", "heap",
+	}
+
+	host := "127.0.0.1:5454"
```
AlekSi, line 44: That will fail if the port is occupied by something else.

Listening on `0.0.0.0:0` never fails.

internal/util/debug/archive_test.go
```diff
@@ -0,0 +1,96 @@
...
+	metricsProvider := stateProvider.MetricsCollector(true)
+	metricsRegisterer.MustRegister(metricsProvider)
+
+	l := zap.S()
+
+	ctx, cancelCtx := context.WithCancel(context.Background())
+	go RunHandler(ctx, host, metricsRegisterer, l.Named("debug").Desugar())
+
+	// Wait for the server to start
+	time.Sleep(time.Second)
```
AlekSi, line 60: Not good enough

internal/util/debug/archive_test.go
```diff
@@ -0,0 +1,96 @@
...
+	require.NoError(t, err)
+
+	metricsRegisterer := prometheus.DefaultRegisterer
+	metricsProvider := stateProvider.MetricsCollector(true)
+	metricsRegisterer.MustRegister(metricsProvider)
+
+	l := zap.S()
+
+	ctx, cancelCtx := context.WithCancel(context.Background())
+	go RunHandler(ctx, host, metricsRegisterer, l.Named("debug").Desugar())
```
AlekSi, line 57: We don't want to leak that goroutine

internal/util/debug/archive_test.go
```diff
@@ -0,0 +1,96 @@
...
+	body, err := io.ReadAll(resp.Body)
+	require.NoError(t, err)
+
+	zipReader, err := zip.NewReader(bytes.NewReader(body), int64(len(body)))
+	require.NoError(t, err)
+
+	require.Equal(t, len(fileList), len(zipReader.File), "needs to be same as length of fileList")
+
+	for _, file := range zipReader.File {
+		t.Logf("\nverifying file : %s", file.Name)
```
AlekSi, line 93: `\n`

chilagrow (dismissed):
One minor comment :hugs:

internal/util/debug/debug.go
```diff
@@ -88,6 +91,29 @@ func Listen(opts *ListenOpts) (*Handler, error) {
...
+	archiveDurations := prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "archive_response_seconds",
+			Help:      "Archive response time seconds.",
+			Buckets:   []float64{0.1, 0.5, 1, 5},
+		},
+		[]string{"code"},
+	)
```
AlekSi, line 103: I don't see how that would be useful given that amount of information in the archive will change over time

internal/util/debug/debug.go
```diff
@@ -97,25 +123,98 @@ func Listen(opts *ListenOpts) (*Handler, error) {
...
+				fmt.Sprintf("attachment; filename=FerretDB-debug-%d.zip", time.Now().UnixMilli()),
+			)
+
+			ctx := req.Context()
+			zipWriter := zip.NewWriter(rw)
+
+			defer func() {
+				if err := zipWriter.Close(); err != nil {
+					opts.L.Error("Archive handler failed", zap.Error(err))
+					rw.WriteHeader(http.StatusInternalServerError)
```
AlekSi, line 142: It is not possible to send a header with status after some content was already sent

AlekSi (commented):
That code should use slog, not zap

internal/util/debug/debug.go
```diff
@@ -97,25 +123,98 @@ func Listen(opts *ListenOpts) (*Handler, error) {
...
+				rw.WriteHeader(http.StatusInternalServerError)
+
+				return
+			}
+
+			// we use *http.Request instead of http.Get function to provide the ctx
+			scrapeReq := must.NotFail(http.NewRequestWithContext(
+				ctx,
+				http.MethodGet,
+				fmt.Sprintf("http://%s%s", req.Host, "/debug/metrics"),
```
chilagrow, line 160: Why not use https://pkg.go.dev/net/url#JoinPath?

internal/util/debug/debug.go
```diff
@@ -97,25 +123,98 @@ func Listen(opts *ListenOpts) (*Handler, error) {
...
+			if err != nil {
+				opts.L.Error("Archive handler failed - metrics failed", zap.Error(err))
+				rw.WriteHeader(http.StatusInternalServerError)
+
+				return
+			}
+
+			_, err = io.Copy(metricsFile, resp.Body)
+
+			_ = resp.Body.Close()
```
chilagrow, line 174: Error from Close() should be logged right?

internal/util/debug/debug.go
```diff
@@ -97,25 +123,98 @@ func Listen(opts *ListenOpts) (*Handler, error) {
...
+			resp, err = http.DefaultClient.Do(scrapeReq)
+			if err != nil {
+				opts.L.Error("Archive handler failed - pprof failed", zap.Error(err))
+				rw.WriteHeader(http.StatusInternalServerError)
+
+				return
+			}
+
+			_, err = io.Copy(heapFile, resp.Body)
+			_ = resp.Body.Close()
```
chilagrow, line 202: Same here, error from Close should be logged

internal/util/debug/debug.go
noisersup: Not really, it returns error to conform to `Closer` interface. The `net/http` calls `Response.Body.Close` without handling error in multiple places: [1](https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/net/http/response.go;l=338), [2](https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/net/http/client.go;l=708).

In `net/http` docs there's an example with `resp.Body.Close()` not being checked: https://pkg.go.dev/net/http#pkg-overview.

Also PTAL at discussion here: securego/gosec#925.

Overall I didn't check for error, because in previous approach I've used deferred close. I'd suggest to keep it like that to not introduce another 3-5 lines after each Close to handle error that doesn't occur. It'll be also less confusing for the future if someone might want to use deferred call here.

internal/util/debug/debug.go
noisersup: https://github.com/FerretDB/FerretDB/pull/3895#discussion_r1684277509

internal/util/debug/debug.go
noisersup: We don't handle user provided paths here, path traversal is not a problem here.
We can use `JoinPath` here nevertheless, but the code'll be just less simple:
```
			must.NotFail(url.JoinPath("http://"+req.Host, "/debug/metrics")),
```

Although I think what could be done is to not use `req.Host` at all. It doesn't make sense to use and potentially might be harmful, thank you for noticing the line!

internal/util/debug/debug.go
noisersup: My bad, fixed

chilagrow:
> That code should use slog, not zap

@noisersup this comment was missed.

internal/util/debug/debug.go
chilagrow: Ok. What is harmful though?

internal/util/debug/debug_test.go
```diff
@@ -58,21 +62,73 @@ func TestProbes(t *testing.T) {
...
+			require.NoError(t, err)
+
+			content := make([]byte, 1)
+
+			var n int
+			n, err = f.Read(content)
+			require.NoError(t, err)
+
+			assert.Equal(t, 1, n, "file should contain any data, but was empty")
+			require.NoError(t, f.Close())
```
chilagrow, line 129: Can you put `f.Close` closer to  `file.Open()` using `t.Cleanup`. Not all the code path reaches here.

internal/util/debug/debug.go
noisersup: The `Host` field is not validated by the server, and it can be manipulated by the client. That might lead to things such as forcing the server to query user-specified server, or something unexpected and more harmful.

internal/util/debug/debug_test.go
noisersup: done

chilagrow (approved):

internal/util/debug/debug.go
```diff
@@ -100,25 +116,84 @@ func Listen(opts *ListenOpts) (*Handler, error) {
...
+			fmt.Sprintf("attachment; filename=FerretDB-debug-%d.zip", time.Now().UnixMilli()),
+		)
+
+		ctx := req.Context()
+		zipWriter := zip.NewWriter(rw)
+
+		defer func() {
+			if err := zipWriter.Close(); err != nil {
+				l.ErrorContext(ctx, "Archive handler failed", logging.Error(err))
+				return
```
rumyantseva, line 133: I'm not sure if this `return` is needed, as we are already inside `defer` :)

internal/util/debug/debug.go
```diff
@@ -100,25 +116,84 @@ func Listen(opts *ListenOpts) (*Handler, error) {
...
+			}
+		}()
+
+		metricsFile, err := zipWriter.Create("metrics")
+		if err != nil {
+			l.ErrorContext(ctx, "Archive handler failed", logging.Error(err))
+			return
```
rumyantseva, line 140: What does the response to the user look like in this case and all the other cases when an error is returned? An empty zip file?

As this is a debug handler, I don't think it's super important to return something particular, but on the other hand, it's nice if the response is clear (like `500` or something). However, I'm not sure if it is possible to return something like 500 page, as we have already set some content-type headers. So, this is not critical, just some thinking.

rumyantseva (approved):
Looks good to me, asked a couple of questions for my understanding.

internal/util/debug/debug.go
noisersup: The request will just end with StatusOK. I think we could use multipart here to handle error properly, but I'd suggest doing that in separate PR. @FerretDB/core WDYT?

noisersup: and about multipart? :)

AlekSi: I don't know how one can send a file using the multipart encoding, but I know for sure that nothing would handle that response. Definitely not `curl` or browser that would be used to download that debug archive.

internal/util/debug/debug.go
```diff
@@ -100,25 +116,84 @@ func Listen(opts *ListenOpts) (*Handler, error) {
...
+		}
+
+		debugAddr := ctx.Value(http.LocalAddrContextKey).(net.Addr)
```
AlekSi, line 143: Oooh, that's clever

AlekSi (approved):

internal/util/debug/debug.go
AlekSi: 🏅

PR #3897, opened 2023-12-21
Implement `usersInfo` command

### LINKED ISSUE #1497: Implement `usersInfo` command
Blocked by #1491 and #1493.

### What should be done?

With `createUser` and `dropUser` implemented, we should implement the basic [`usersInfo ` command](https://www.mongodb.com/docs/manual/reference/command/usersInfo/).

We should accept the following `usersInfo` arguments:
* `usersInfo`;
* `showCredentials` – ignore for now;
* `showCustomData` – ignore;
* `showPrivileges` – ignore;
* `showAuthenticationRestrictions` – ignore;
* `filter`;
* `comment`.

Tests should use `createUser` and `dropUser` commands, not `insert` and `delete`.

### Where?

### Definition of Done

- handler updated;
- spot refactorings done;
- basic hidden user documentation page created.

henvic:
# Description

Implement `usersInfo` command.

Closes #1497

## Readiness checklist


AlekSi (commented):
Please re-request review when ready

integration/users/usersinfo_test.go
```diff
@@ -0,0 +1,426 @@
...
+func TestUsersinfo(t *testing.T) {
+	t.Parallel()
+
+	ctx, collection := setup.Setup(t)
+	client := collection.Database().Client()
+
+	// Map of users per database suffix
+	dbToUsers := map[string][]bson.D{
+		"": {
+			{{"createUser", "one"}, {"roles", bson.A{}}, {"pwd", "pwd1"}},
```
AlekSi, line 41: I think we could extract common parts like `{"roles", bson.A{}}` into the test code to make test cases simpler

integration/users/usersinfo_test.go
```diff
@@ -0,0 +1,426 @@
...
+		},
+		"_few": {
+			{{"createUser", "i"}, {"roles", bson.A{}}, {"pwd", "password1"}},
+			{{"createUser", "j"}, {"roles", bson.A{}}, {"pwd", "password2"}},
+		},
+		"_another": {
+			{{"createUser", "singleuser"}, {"roles", bson.A{}}, {"pwd", "123456"}},
+		},
+	}
+	// allInsertedUsers is a list of all users inserted in the database that you want to verify
```
AlekSi, line 57: Well, I think we should say "create" instead of "insert" because we don't use `insert` anymore there

integration/users/usersinfo_test.go
```diff
@@ -0,0 +1,426 @@
...
+
+			foundUsers := map[string]struct{}{}
+			for i := 0; i < actualUsers.Len(); i++ {
+				au, err := actualUsers.Get(i)
+				require.NoError(t, err)
+				actualUser := au.(*types.Document)
+
+				if tc.hasUser == nil && tc.expected == nil {
+					t.Errorf("set at least one expectation for %q", name)
+				}
```
AlekSi, line 389: `assert.False(t, tc.hasUser == nil && tc.expected == nil, …)`

internal/handler/msg_usersinfo.go
```diff
@@ -29,6 +38,240 @@ func (h *Handler) MsgUsersInfo(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
...
+
+		users = append(users, u)
+	case string:
+		var u usersInfoPair
+		if err = u.extract(user, dbName); err != nil {
+			return nil, lazyerrors.Error(err)
+		}
+
+		users = append(users, u)
+	case int, int32: // {usersInfo: 1 } // int for MongoDB and int32 for FerretDB
```
AlekSi, line 102: It can't be `int`. You probably want `int64`.

AlekSi:
(also, CI fails)

internal/handler/msg_usersinfo.go
henvic: Oops. It's always int32. Removing `int`.

internal/handler/msg_usersinfo.go
```diff
@@ -29,6 +38,238 @@ func (h *Handler) MsgUsersInfo(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
...
+	}
+
+	if singleDB {
+		filter.Set("db", dbName)
+		return filter, nil
+	}
+
+	ids := types.MakeArray(len(pairs))
+	for i, p := range pairs {
+		if err := ids.Set(i, p.db+"."+p.username); err != nil {
```
rumyantseva, line 267: ids.Append

internal/handler/msg_usersinfo.go
```diff
@@ -29,6 +38,241 @@ func (h *Handler) MsgUsersInfo(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
...
+	for _, p := range pairs {
+		ps = append(ps, p.db+"."+p.username)
+	}
+
+	ids, err := types.NewArray(ps...)
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+
+	filter.Set("_id", bson.D{{Key: "$eq", Value: ids}})
```
AlekSi, line 275: That should be `*types.Document`, not `bson.D`

internal/handler/msg_usersinfo.go
```diff
@@ -29,6 +38,241 @@ func (h *Handler) MsgUsersInfo(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
...
+	for _, p := range pairs {
+		ps = append(ps, p.db+"."+p.username)
+	}
+
+	ids, err := types.NewArray(ps...)
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+
+	filter.Set("_id", must.NotFail(types.NewDocument(bson.D{{Key: "$eq", Value: ids}})))
```
rumyantseva, line 275: Hmm, this one won't work.

rumyantseva, line 275: I guess you need smth like 

```suggestion
	filter.Set("_id", must.NotFail(types.NewDocument("$eq", ids)))
```

noisersup (approved):
LGTM thanks for the bug fix!

integration/users/usersinfo_test.go
```diff
@@ -0,0 +1,434 @@
...
+				"TestUsersinfo.two":                {},
+				"TestUsersinfo_example.a":          {},
+				"TestUsersinfo_example.b":          {},
+				"TestUsersinfo_example.c":          {},
+				"TestUsersinfo_few.i":              {},
+				"TestUsersinfo_few.j":              {},
+				"TestUsersinfo_another.singleuser": {},
+			},
+		},
+	}
```
rumyantseva, line 369: Maybe one more case (or a few cases) with `payload` being set as something like

```
"usersInfo", bson.D{{"foo", "bar"}}
```

and some variations of it - the cases when `usersInfo` is set and is a document (or an array of documents) but `user` field is not set in this document, or `db` is not set or some combinations of it.

internal/handler/msg_usersinfo.go
```diff
@@ -29,6 +35,248 @@ func (h *Handler) MsgUsersInfo(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
...
+			return handlererrors.NewCommandErrorMsg(
+				handlererrors.ErrBadValue,
+				"UserName must contain a field named: db",
+			)
+		}
+
+		db, err := vt.Get("db")
+		if err != nil {
+			return lazyerrors.Error(err)
+		}
```
rumyantseva, line 229: I think this can be simplified a bit. `vt.Get` will return error in a single case - if `db` field doesn't exist. So, no need to have an extra check.

```suggestion
		db, err := vt.Get("db")
		if err != nil {
		     return handlererrors.NewCommandErrorMsg(
				handlererrors.ErrBadValue,
				"UserName must contain a field named: db",
			)
		}
```

rumyantseva (commented):
Great job overall. I left a few small questions/suggestions.
Also, let's check where the current issue number is mentioned across the whole database. As now `usersInfo` command is supported, we need to update user docs - https://docs.ferretdb.io/reference/supported-commands/.

internal/handler/msg_usersinfo.go
```diff
@@ -29,6 +35,248 @@ func (h *Handler) MsgUsersInfo(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
...
+
+	usersInfo, err := document.Get(document.Command())
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+
+	if usersInfo == nil {
+		msg := fmt.Sprintf("required parameter %q is missing", document.Command())
+		return nil, handlererrors.NewCommandErrorMsgWithArgument(handlererrors.ErrBadValue, msg, document.Command())
+	}
```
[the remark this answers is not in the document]
henvic: You are right. It's not reachable if `usersInfo` is missing. Removing the check.

internal/handler/msg_usersinfo.go
```diff
@@ -29,6 +35,248 @@ func (h *Handler) MsgUsersInfo(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
...
+// When singleDB is true, it returns a filter for a single database (case when usersInfo: 1 is invoked).
+// Otherwise, it filters by any pair of user and database.
+func usersInfoFilter(allDBs, singleDB bool, dbName string, pairs []usersInfoPair) (*types.Document, error) {
+	filter := must.NotFail(types.NewDocument())
+
+	if allDBs {
+		return filter, nil
+	}
+
+	if singleDB {
```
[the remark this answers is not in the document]
henvic: Unm. I think that's almost it, but not quite exactly. For a counter-example, see subtest FromOthersMultipleDatabases.

SingleDB is a flag only set when `{usersInfo: 1 }`.

integration/users/usersinfo_test.go
henvic: Added a few extra test cases.

A question: should we fail if there is an unknown field in the payload? I've noticed MongoDB fails if this happens on the usersInfo command. There isn't an API on *types.Document to check for this error, though 🤔

rumyantseva: Well, if MongoDB returns an error, it sounds reasonable for us to return an error too :)

website/docs/reference/supported-commands.md
```diff
@@ -204,12 +204,12 @@ Related [issue](https://github.com/FerretDB/FerretDB/issues/78).
...
-| `usersInfo`                |                                  | ❌     | [Issue](https://github.com/FerretDB/FerretDB/issues/1497) |
+| `usersInfo`                |                                  | ✅     |                                                           |
 |                            | `showCredentials`                | ⚠️     |                                                           |
 |                            | `showCustomData`                 | ⚠️     |                                                           |
 |                            | `showPrivileges`                 | ⚠️     |                                                           |
 |                            | `showAuthenticationRestrictions` | ⚠️     |                                                           |
-|                            | `filter`                         | ⚠️     |                                                           |
+|                            | `filter`                         | ✅     |                                                           |
```
rumyantseva, line 212: I think this one is about a field called `filter` which we don't support in this PR.

rumyantseva (approved):
LGTM, only one tiny detail about docs

rumyantseva (approved):
LGTM

noisersup (approved):

henvic:
Needs review from @AlekSi to trigger auto-merge due to /website/ update.

rumyantseva (approved):

integration/users/usersinfo_test.go
```diff
@@ -0,0 +1,496 @@
...
+}
+
+func TestUsersinfo(t *testing.T) {
+	t.Parallel()
+
+	ctx, collection := setup.Setup(t)
+	client := collection.Database().Client()
+
+	// Map of users per database suffix
```
AlekSi, line 47:
```suggestion
	// Map of users per database suffix.
```
(full sentence)

or

```suggestion
	// map of users per database suffix
```
(sentence fragment; see CONTRIBUTING.md)

or (the best option)

```suggestion
```
(since this commend doesn't add anything)

integration/users/usersinfo_test.go
```diff
@@ -0,0 +1,496 @@
...
+			dbSuffix: "_another",
+			payloads: []bson.D{
+				createUser("singleuser", "123456"),
+			},
+		},
+	}
+
+	dbPrefix := testutil.DatabaseName(t)
+
+	// Create users into the database to test userInfo.
```
AlekSi, line 84: That's pretty obvious in the test called `TestUsersinfo`

integration/users/usersinfo_test.go
```diff
@@ -0,0 +1,496 @@
...
+	for _, inserted := range dbToUsers {
+		dbName := testutil.DatabaseName(t) + inserted.dbSuffix
+		db := client.Database(dbName)
+
+		// Clear any residual database users before recreating them.
```
AlekSi, line 89: I don't understand how any residual database users could remain. How that's possible? Why do we need to call `dropAllUsersFromDatabase` manually?

integration/users/usersinfo_test.go
```diff
@@ -0,0 +1,496 @@
...
+				require.Equal(t, expectedUsers.Len(), actualUsers.Len(), "users length")
+			}
+
+			foundUsers := map[string]struct{}{}
+			for i := 0; i < actualUsers.Len(); i++ {
+				au, err := actualUsers.Get(i)
+				require.NoError(t, err)
+				actualUser := au.(*types.Document)
+
+				assert.False(t, tc.hasUser == nil && tc.expected == nil, "set at least one expectation for %q", name)
```
AlekSi, line 471:
```suggestion
				require.False(t, tc.hasUser == nil && tc.expected == nil, "set at least one expectation for %q", name)
```

There is no point in continuing with a broken test case.

But that also does not test the case when both _are_ set. So I guess it should be `require.True((tc.hasUser == nil) != (tc.expected == nil))` or something

integration/users/usersinfo_test.go
henvic: You're correct. This shouldn't happen. I had wrongly assumed that a test failure that wasn't cleaned via `t.Cleanup()` might leave data on the database early on when creating the tests, but this doesn't apply here as we store all user information in the admin database. Removing the code.

henvic:
@AlekSi, I've just addressed the comments!

AlekSi (approved):

noisersup (approved):

PR #3900, opened 2023-12-21
Implement `awaitData` tailable cursors

### LINKED ISSUE #2283: Support tailable cursors
### What should be done?

https://www.mongodb.com/docs/manual/core/tailable-cursors/

Tailable cursors do not close when it exhausted results, the cursor is kept open waiting for more documents to be added. It's available for find cursor. Not available for aggregate cursor.

Support both `tailable` and `awaitData`.

```[tasklist]
### Tasks
- [ ] Support `awaitData`
- [x] Check test coverage, add tests if needed
- [x] Clean-up implementation
```

### Where?

TBD

### Definition of Done

TBD

- all handlers updated;
- spot refactorings done;

noisersup:
# Description

Closes #2283.

## Readiness checklist


internal/handler/msg_getmore.go
```diff
@@ -266,3 +273,81 @@ func (h *Handler) makeNextBatch(c *cursor.Cursor, batchSize int64) (*types.Array
...
+		}
+
+		switch {
+		case resBatch.Len() == 0:
+			h.L.Debug(
+				"awaitData: Waiting for new documents", zap.Int64("cursor_id", c.ID), zap.Stringer("type", c.Type),
+				zap.Duration("time_left", sleepDur-time.Since(startTime)),
+			)
+
+			resBatch, err = h.makeNextBatch(c, batchSize)
```
rumyantseva, line 331: Don't we need to `return` if `resBatch` is not empty?

internal/handler/msg_getmore.go
```diff
@@ -266,3 +273,81 @@ func (h *Handler) makeNextBatch(c *cursor.Cursor, batchSize int64) (*types.Array
...
+			err = nil
+			return
+		}
+
+		err = lazyerrors.Error(err)
+		resBatch = nil
+		return
+	}()
+
+	for {
```
rumyantseva, line 305: Don't we need any `time.Sleep` in this `for` loop? It works but the amount of queries it makes is a bit crazy...

internal/handler/msg_getmore.go
```diff
@@ -266,3 +273,81 @@ func (h *Handler) makeNextBatch(c *cursor.Cursor, batchSize int64) (*types.Array
...
+		c.Close()
```
rumyantseva, line 306: How does the whole code work overall?

Let's say, there is a collection X with some data. I query this data with `awaitData = true`.
I got some data, but because `awaitData` is set as `true`, I just continue waiting for the data. As soon as there is new data, I should receive it in addition to what I already received, only the new data, nothing else. Then I received that new data and I continue waiting. When there is some new data I should receive that new data again. What I don't understand in this code is how you ensure that next time I call `getMore` with `awaitData`, I only receive the new data? I expected some `recordId` or something like this being involved. Or do I misunderstand something?

rumyantseva: (Also I see you call `c.Close()` but don't I need this cursor to actually get more data?..)

internal/handler/msg_getmore.go
noisersup: We can implement it yeah

internal/handler/msg_getmore.go
rumyantseva: Yeah, I think my understanding is not quite right.

noisersup: > (Also I see you call c.Close() but don't I need this cursor to actually get more data?..)

`c.Close()` if tailable, doesn't close whole cursor, but only underlying iterator.
We change the iterator on every for loop iteration, as we do new backend queries.

The overusage of `Close` is a result of bugs from previous week, as in some cases underlying iterator wasn't closed properly which resulted in panic. I'm currently cleaning up everything, so some `Close` calls will be removed, however its overusage shouldn't result in any issues so I think it's ready to run your tests

henvic (approved):

internal/handler/msg_getmore.go
```diff
@@ -266,3 +273,68 @@ func (h *Handler) makeNextBatch(c *cursor.Cursor, batchSize int64) (*types.Array
 
 	return nextBatch, nil
 }
+
+// awaitData stops the goroutine, and waits for a new data for the cursor.
```
rumyantseva, line 277: To me "stops the goroutine" phrase is a bit unclear. I don't see it stopping anything, it rather retries unless `maxTimeMS` exceeds. But having in mind that this is a temporary solution, it's not that important.

rumyantseva (approved):
LGTM, I also tried `awaitData` with meteor (in thishttps://github.com/FerretDB/FerretDB/pull/3913 draft), and it worked!

internal/handler/msg_getmore.go
```diff
@@ -226,7 +228,12 @@ func (h *Handler) MsgGetMore(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg,
 		}
 
 	case cursor.TailableAwait:
-		panic("not supported yet")
+		if nextBatch.Len() == 0 {
```
AlekSi, line 231: It looks a bit sus that it is not `if next batch.Len() < int(batchSize)`. Should it be? Or we could simplify the code above to use `if nextBatch.Len() == 0`?

Please look into it in the next PR. Maybe there is a way to test this.

Ack and resolve.

internal/handler/msg_getmore.go
```diff
@@ -266,3 +273,68 @@ func (h *Handler) makeNextBatch(c *cursor.Cursor, batchSize int64) (*types.Array
...
 	return nextBatch, nil
 }
+
+// awaitData stops the goroutine, and waits for a new data for the cursor.
+// If there's a new document, or the maxTimeMS have passed it returns the nextBatch.
+func (h *Handler) awaitData(
+	ctx context.Context,
+	c *cursor.Cursor,
+	maxTimeMS, batchSize int64,
+) (resBatch *types.Array, err error) {
```
AlekSi, line 283:
```suggestion
//
//nolint:lll // for readability
func (h *Handler) awaitData(ctx context.Context, c *cursor.Cursor, maxTimeMS, batchSize int64) (resBatch *types.Array, err error) {
```

or use a `struct` for parameters.

Please ack, resolve, and tweak in the next PR

internal/handler/msg_getmore.go
```diff
@@ -266,3 +273,68 @@ func (h *Handler) makeNextBatch(c *cursor.Cursor, batchSize int64) (*types.Array
...
+			return
+		}
+
+		resBatch, err = h.makeNextBatch(c, batchSize)
+		if err != nil {
+			return
+		}
+
+		if maxTimeMS > 10 {
+			time.Sleep(10 * time.Millisecond)
```
AlekSi, line 337: That sleep should be cancelable (`ctxutil.Sleep`).

For the next PR

AlekSi (approved):
Please ack comments and resolve conversations

website/versioned_docs/version-v1.16/reference/supported-commands.md
AlekSi, line 1: That was the wrong file (version) to edit.

PR #3919, opened 2024-01-01
Fix `maxTimeMS` for `getMore` command

### LINKED ISSUE #2984: Support `maxTimeMS` in `getMore` command parameter
### What should be done?

After https://github.com/FerretDB/FerretDB/issues/2283

`maxTimeMS` for `getMore` is supported only for tailable cursor.

### Where?

https://github.com/FerretDB/FerretDB/blob/main/internal/handlers/common/getmore.go

### Definition of Done

- all handlers updated;
- spot refactorings done;

noisersup:
# Description

Closes #2984.

## Readiness checklist


internal/handler/msg_find.go
```diff
@@ -102,10 +104,19 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
 
 	cancel := func() {}
+	var findDone atomic.Bool
+
 	if params.MaxTimeMS != 0 {
-		// It is not clear if maxTimeMS affects only find, or both find and getMore (as the current code does).
-		// TODO https://github.com/FerretDB/FerretDB/issues/2984
-		ctx, cancel = context.WithTimeout(ctx, time.Duration(params.MaxTimeMS)*time.Millisecond)
+		ctx, cancel = context.WithCancel(ctx)
+		go func() {
```
rumyantseva, line 111: Hmm, this gorutine will run for the whole duration of `MaxTimeMS`, even if `findDone.Load()` was true a while ago. 

Wouldn't something like this work?

```go
for {
		select {
		case <-ticker.C:
			canel()
		case <-signalChan:
			return
		}
	}
```

and `signalChan` is closed the same way as `findDone` is set as true.

noisersup: I thought about something like that, though at this point it felt that introducing another go routine instead of ticker might be more suitable (as we only need one signal and not entire ticker). Although I remember that I've implemented something similar in some previous PR, and there were some objections in terms of complexity because of goroutines.

But yeah, in this case it seems to be more reasonable so will change that quickly and let's wait for the comments:)

internal/handler/msg_aggregate.go
```diff
@@ -248,10 +248,21 @@ func (h *Handler) MsgAggregate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
...
-		// TODO https://github.com/FerretDB/FerretDB/issues/2983
-		ctx, cancel = context.WithTimeout(ctx, time.Duration(maxTimeMS)*time.Millisecond)
+		ctx, cancel = context.WithCancel(ctx)
+
+		go func() {
+			select {
+			case <-timeout.C:
+				cancel()
+			case <-findDone:
+				return
```
henvic, line 263: Maybe use a `case <-ctx.Done()`?

internal/handler/msg_aggregate.go
```diff
@@ -248,10 +248,21 @@ func (h *Handler) MsgAggregate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
 	}
 
 	cancel := func() {}
```
henvic, line 250: Maybe you can remove this line and/or make it `var cancel func()` for the sake of brevity.

internal/handler/msg_aggregate.go
```diff
@@ -248,10 +248,21 @@ func (h *Handler) MsgAggregate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
...
+
+	timeout := time.NewTicker(time.Duration(maxTimeMS) * time.Millisecond)
```
henvic, line 252: It's important to call `ticker.Stop()` to clear up resources when finished with a ticker.

internal/handler/msg_find.go
```diff
@@ -98,10 +98,22 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+		ctx, cancel = context.WithCancel(ctx)
+
+		timeout := time.NewTicker(time.Duration(params.MaxTimeMS) * time.Millisecond)
+
+		go func() {
+			select {
+			case <-timeout.C:
+				cancel()
+			case <-findDone:
+				return
```
henvic, line 114: Similar remarks to msg_aggregate.go.

internal/handler/msg_aggregate.go
noisersup: The PR is not ready yet for another round of reviews but thanks!

internal/handler/msg_aggregate.go
noisersup: In which case? Those lines are managing the context cancellation, not waiting for it

AlekSi (commented):
Please re-request review when ready

internal/handler/msg_aggregate.go
noisersup: The func's default value is nil, which cannot be called

internal/handler/msg_aggregate.go
henvic: Oh, I see. I was mostly worried about the case when the function returns due to an error and `findDone <- struct{}{}` doesn't happen.

internal/handler/msg_find.go
```diff
@@ -98,10 +98,23 @@ func (h *Handler) MsgFind(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, er
...
+		defer timeout.Stop()
+
+		go func() {
+			select {
+			case <-timeout.C:
+				cancel()
+			case <-findDone:
+				return
+			}
+		}()
```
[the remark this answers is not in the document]
AlekSi: (not to mention that _ticker_ looks very strange there; if anything, it should be a timer)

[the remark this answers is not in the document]
noisersup: It won't as underlying ticker has buffered channel. Also I've updated ticker to timer

AlekSi (approved):

henvic (approved):

PR #3925, opened 2024-01-02
Add the necessary for replica set fields to `ismaster` response

### LINKED ISSUE #3556: Insert `i`/`u`/`d` documents into OpLog
### What should be done?

Insert i/u/d documents in the same format as MongoDB.
Do enough to make it barely compatible with Meteor apps.
Add issues and TODOs.

### Where?

https://github.com/FerretDB/FerretDB/tree/main/internal/backends/decorators/oplog

### Definition of Done

- spot refactorings done.

### LINKED ISSUE #3861: Add integration tests for OpLog i/u/d
### What should be done?

We should add integration tests for OpLog for simple operations: `insert` `update` and `delete`

Tests should `insert`, `update`, and `delete` some documents in some collection, then check that documents `local.oplog.rs` are what is expected. This should not be a compat test but a regular integration test.

Tests should not use tailable cursors.

They should pass for MongoDB and are expected to fail for FerretDB (for now).

### Where?

`integration/oplog` (new directory).

### Definition of Done

- integration tests added;
- they pass for MongoDB;
- they expectedly fail for FerretDB.

rumyantseva:
# Description

Closes #3556.

In this PR: the minimal necessary changes related to replica set params necessary to connect meteor to FerretDB.
Oplog collection needs to be created manually (as usual in FerretDB).

## Readiness checklist


internal/handler/common/ismaster.go
```diff
@@ -51,5 +52,16 @@ func IsMasterDocuments() []*types.Document {
 		"maxWireVersion", MaxWireVersion,
 		"readOnly", false,
 		"ok", float64(1),
-	))}
+	))
+
+	if name != "" {
+		if strings.HasPrefix(host, ":") {
+			host = "localhost" + host
```
rumyantseva, line 59: This is not the right approach, but I'm not sure what would be better. Maybe a cli flag to set the host?..

If the app that wants to connect to FerretDB is not running on the same host as FerretDB, this field should contain the address to connect. Another way is probably to parse `rs.initiate` command and set hosts according to the params of this command.

integration/setup/listener.go
```diff
@@ -172,6 +172,8 @@ func setupListener(tb testtb.TB, ctx context.Context, logger *zap.Logger) string
 		Logger:        logger,
 		ConnMetrics:   listenerMetrics.ConnMetrics,
 		StateProvider: sp,
+		// Host:          "127.0.0.1:0",
+		// RSName:        "rs",
```
henvic, line 176: I think that was unintentionally left commented, right?

internal/handler/common/ismaster.go
henvic: Reading the docs here it seems like `rs.initiate` would be better.
Any reason why you're considering a CLI flag besides being simpler?

rumyantseva: No, cli flag being simpler is the only reason. I don't see a quick way to implement `rs.initiate` command handling.

henvic (dismissed):

noisersup (approved):
LGTM

internal/handler/common/ismaster.go
AlekSi: Well, another reason is that this works only for TCP listen addresses and does not work for TLS-only setups…

AlekSi (approved):

AlekSi:
@henvic @noisersup PTAL

henvic (approved):

noisersup (approved):
LGTM

PR #3928, opened 2024-01-03
Store passwords for PLAIN authentication mechanism

### LINKED ISSUE #3785: Store passwords for `PLAIN` authentication mechanism

henvic:
# Description

Closes #3785.

## Readiness checklist


henvic:
Marking as ready for review because the functionality is ready, but I am checking something on `integration/setup` that might need to be tweaked, so please don't enable auto-merge just yet.

AlekSi:
> please don't enable auto-merge just yet.

We use https://github.com/FerretDB/FerretDB/labels/do%20not%20merge for this that you already added, so please enable auto-merge as usual

noisersup (approved):

internal/handler/msg_createuser.go
```diff
@@ -107,12 +109,64 @@ func (h *Handler) MsgCreateUser(ctx context.Context, msg *wire.OpMsg) (*wire.OpM
...
+				break
+			}
+
+			if err != nil {
+				return nil, lazyerrors.Error(err)
+			}
+
+			if v == "PLAIN" {
+				plainAuth = true
+			} else {
```
AlekSi, line 141: https://go.dev/wiki/CodeReviewComments#indent-error-flow

integration/users/create_user_test.go
```diff
@@ -125,6 +152,10 @@ func TestCreateUser(t *testing.T) {
 	for name, tc := range testCases {
 		name, tc := name, tc
 		t.Run(name, func(t *testing.T) {
+			if tc.hasCred && setup.IsMongoDB(t) {
+				t.Skip("PLAIN credentials are only supported via LAP (PLAIN) by MongoDB Enterprise")
```
AlekSi, line 156:
```suggestion
				t.Skip("PLAIN credentials are only supported via LDAP (PLAIN) by MongoDB Enterprise")
```

integration/users/create_user_test.go
```diff
@@ -40,6 +40,7 @@ func TestCreateUser(t *testing.T) {
 
 	testCases := map[string]struct { //nolint:vet // for readability
 		payload    bson.D
+		hasCred    bool
```
AlekSi, line 43: What does that field mean? I would assume "has credentials", but then there are cases with password and `hasCred` being `false`

integration/users/create_user_test.go
AlekSi: In general, we should prefer expected failures to just skips

integration/users/create_user_test.go
```diff
@@ -170,3 +205,17 @@ func TestCreateUser(t *testing.T) {
 		})
 	}
 }
+
+// assertPlainCredentials checks if the credential is a valid PLAIN credential.
+func assertPlainCredentials(t testing.TB, key string, cred *types.Document) {
+	assert.Truef(t, cred.Has(key), "missing credential %q", key)
```
AlekSi, line 211:
```suggestion
	t.Helper()

	assert.Truef(t, cred.Has(key), "missing credential %q", key)
```

integration/users/usersinfo_test.go
```diff
@@ -76,11 +77,27 @@ func TestUsersinfo(t *testing.T) {
...
+				},
+			},
+			skipForMongoDB: true,
+		},
 	}
 
 	dbPrefix := testutil.DatabaseName(t)
 
 	for _, inserted := range dbToUsers {
+		if inserted.skipForMongoDB && setup.IsMongoDB(t) {
```
AlekSi, line 97: That should be a subtest with expected failure. Otherwise, it just too easy to silently skip it

integration/users/usersinfo_test.go
```diff
@@ -91,12 +108,14 @@ func TestUsersinfo(t *testing.T) {
...
 
 	testCases := map[string]struct { //nolint:vet // for readability
-		dbSuffix   string
-		payload    bson.D
-		err        *mongo.CommandError
-		altMessage string
-		expected   bson.D
-		hasUser    map[string]struct{}
+		dbSuffix        string
+		showCredentials bool // showCredentials should also be set on the payload
```
AlekSi, line 112: They why we have a separate field instead of just using `payload.Get("showCredentials")`?

integration/users/create_user_test.go
henvic: Should I replace this with something like the following?

```go
t.Run(name, func(tt *testing.T) {
	tt.Parallel()

	var t testtb.TB = tt
	if tc.hasCred && setup.IsMongoDB(t) {
		t = testfail.Expected(t, "PLAIN credentials are only supported via LDAP (PLAIN) by MongoDB Enterprise")
	}

	// ...
}
```

I'm asking because I haven't seen this pattern anywhere and it seems we don't have an analogue to `setup.FailsForMongoDB` to when we expect a test run with MongoDB to fail.

integration/users/create_user_test.go
henvic: It meant "has credential" indeed. I was able to improve this, removing the variable.

integration/users/create_user_test.go
henvic: Flagged as a helper function.

integration/users/usersinfo_test.go
henvic: The reason why I haven't added it as a subtest is that this is for setting up the environment for multiple tests, rather than for testing something. I want it to run before any subtests (that are executed as a second step).

integration/users/usersinfo_test.go
henvic: I did so trying to simplify the tests, but I've refactored the test to get this from the payload directly.
The downside is that there is some extra complexity in retrieving the value of showCredentials, but it seems like a good trade-off.

AlekSi:
https://github.com/FerretDB/github-actions/pull/235

integration/users/create_user_test.go
AlekSi: > and it seems we don't have an analogue to `setup.FailsForMongoDB`

So let's make one in the next PR

AlekSi (approved):

integration/users/create_user_test.go
henvic: Okay! I thought it didn't exist on purpose. Should've have asked! 🤣

PR #3941, opened 2024-01-06
Allow Native Passwords for testcase

adetunjii:
Closes #3413 .

## Readiness checklist


adetunjii:
@AlekSi I'm still not sure this update fixes the issue. I suspect that the root cause of this issue is that when trying to authenticate with a non-existing user, MySQL tries to use the `mysql_native_password` auth plugin instead of the default `caching_sha2_password`. I'm unable to recreate this issue locally, so if it still fails after this, we'd have to skip the test like you previously did until a more concrete solution is found.

AlekSi (approved):

PR #3973, opened 2024-01-15
Implement `updateUser` command

### LINKED ISSUE #1496: Implement `updateUser` command
### What should be done?

With passwords being stored, we should implement the basic [`updateUser ` command](https://www.mongodb.com/docs/manual/reference/command/updateUser/).

Also, in this issue, we should update _all_ integration tests to use `createUser`, `dropUser`, `dropAllUsersFromDatabase`, `usersInfo` commands instead of using `find`/`insert`/`delete`/etc on the `admin.system.users` collection.

### Where?

### Definition of Done

- handler updated;
- spot refactorings done.

henvic:
# Description

Initial implementation of the updateUser command.

Closes #1496.

## Readiness checklist

### DIFF
--- a/integration/users/update_user_test.go
+++ b/integration/users/update_user_test.go
@@ -76,6 +76,22 @@ func TestUpdateUser(t *testing.T) {
 				Message: "User not_found@TestUpdateUser not found",
 			},
 		},
+		"EmptyUsername": {
+			createPayload: bson.D{
+				{"createUser", "not_empty_username"},
+				{"roles", bson.A{}},
+				{"pwd", "password"},
+			},
+			updatePayload: bson.D{
+				{"updateUser", ""},
+				{"pwd", "anewpassword"},
+			},
+			err: &mongo.CommandError{
+				Code:    11,
+				Name:    "UserNotFound",
+				Message: "User @TestUpdateUser not found",
+			},
+		},
 		"EmptyPassword": {
 			createPayload: bson.D{
 				{"createUser", "a_user_bad_password"},
--- a/integration/users/update_user_test.go
+++ b/integration/users/update_user_test.go
@@ -1,0 +1,246 @@
+// Copyright 2021 FerretDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package users
+
+import (
+	"testing"
+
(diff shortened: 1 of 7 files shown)


internal/handler/msg_updateuser.go
```diff
@@ -29,6 +38,201 @@ func (h *Handler) MsgUpdateUser(ctx context.Context, msg *wire.OpMsg) (*wire.OpM
 		return nil, lazyerrors.Error(err)
 	}
 
-	// TODO https://github.com/FerretDB/FerretDB/issues/1496
-	return nil, common.Unimplemented(document, document.Command())
+	dbName, err := common.GetRequiredParam[string](document, "$db")
+	if err != nil {
+		return nil, lazyerrors.Error(err)
```
rumyantseva, line 43: Minor comment for here and below: I think errors returned after `GetRequiredParam` don't have to be `lazyerrors`, because `GetRequiredParam` returns a user-friendly error that we just need to show as is.

integration/users/update_user_test.go
```diff
@@ -0,0 +1,263 @@
...
+			updatePayload: bson.D{
+				{"updateUser", "missing_fields"},
+			},
+			err: &mongo.CommandError{
+				Code:    2,
+				Name:    "BadValue",
+				Message: "Must specify at least one field to update in updateUser",
+			},
+		},
+		"UserNotFound": {
```
rumyantseva, line 63: Do we need a separate test where username is set as an empty string when update is called? Or is this one enough for this case too?

internal/handler/msg_updateuser.go
```diff
@@ -29,6 +38,201 @@ func (h *Handler) MsgUpdateUser(ctx context.Context, msg *wire.OpMsg) (*wire.OpM
...
+
+		var matches bool
+		matches, err = common.FilterDocument(v, filter)
+
+		if err != nil {
+			return nil, lazyerrors.Error(err)
+		}
+
+		if matches {
+			saved = v
```
rumyantseva, line 199: do also we need to close the iterator or exit from the loop or something?

rumyantseva (commented):
Looks great to me overall, I especially like how test cases cover various scenarios! I added a few minor questions/comments (feel free to disagree).

integration/users/update_user_test.go
henvic: Good idea!

internal/handler/msg_updateuser.go
henvic: Iterator is deferred closed on preceding line `qr.Iter.Close()`.

This loop should always contain a maximum 1 item, but I'm adding a `break` just to make it more explicit.

rumyantseva: Hmm, could you help me understand why this loop should always contain a maximum 1 item? Doesn't it contain all the users?

henvic: Oops! You're correct. Please disregard my comment.

It should contain at maximum one item, but we're not filtering at the database layer at this moment.

rumyantseva (approved):
LGTM

noisersup (approved):
LGTM

PR #3970, opened 2024-01-12
Use PBKDF2 for storing `PLAIN` passwords

AlekSi:
# Description

It is required for SCRAM anyway.

Closes #3938.

## Readiness checklist

- [x] I added/updated unit tests (and they pass).
- [ ] I added/updated integration/compatibility tests (and they pass).
- [x] I added/updated comments and checked rendering.
- [x] I made spot refactorings.
- [ ] I updated user documentation.
- [x] I ran `task all`, and it passed.
- [x] I ensured that PR title is good enough for the changelog.
- [x] (for maintainers only) I set Reviewers ([`@FerretDB/core`](https://github.com/orgs/FerretDB/teams/core)), Milestone (`Next`), Labels, Project and project's Sprint fields.
- [x] I marked all done items in this checklist.

Issue #3938, opened 2024-01-05
Use PBKDF2 for storing password for `PLAIN` authentication method


AlekSi:
### What should be done?

PBKDF2 is required for SCRAM anyway, so it does not make much sense to use Argon2 there.

### Where?

TBD

### Definition of Done

TBD

- handler updated;
- all backends updated;
- unit tests added/updated;
- integration/compatibility tests added/updated;
- spot refactorings done;
- user documentation updated or an issue to create documentation created;
- something else?

Neustradamus:
Linked to:
- https://github.com/scram-sasl/info/issues/1
- https://github.com/FerretDB/FerretDB/issues/2012
- https://github.com/FerretDB/FerretDB/issues/3784
- https://github.com/FerretDB/FerretDB/pull/3904
- https://github.com/FerretDB/FerretDB/issues/3938
- https://github.com/FerretDB/FerretDB/pull/3970
- https://github.com/FerretDB/FerretDB/pull/3989

Issue #2012, opened 2023-02-17
Support `SCRAM-SHA-1` authentication mechanism


b1ron:
### What should be done?

* https://www.mongodb.com/docs/manual/core/security-scram/

Neustradamus:
@AlekSi: SCRAM-SHA-1 is needed too, do not forget!

Neustradamus:
Linked to:
- https://github.com/scram-sasl/info/issues/1
- https://github.com/FerretDB/FerretDB/issues/2012
- https://github.com/FerretDB/FerretDB/issues/3784
- https://github.com/FerretDB/FerretDB/pull/3904
- https://github.com/FerretDB/FerretDB/issues/3938
- https://github.com/FerretDB/FerretDB/pull/3970
- https://github.com/FerretDB/FerretDB/pull/3989

PR #3989, opened 2024-01-18
Add SCRAM-SHA-256 authentication support

### LINKED ISSUE #3784: Support `SCRAM-SHA-256` authentication mechanism

henvic:
# Description

We want to support SCRAM-SHA-256 as an authentication method.

Closes #3784.

## Readiness checklist


integration/users/connection_test.go
```diff
@@ -0,0 +1,148 @@
...
+			username:       "scramsha256updated",
+			password:       "pass123",
+			updatePassword: "anotherpassword",
+			mechanism:      "SCRAM-SHA-256",
+		},
+		"BadPassword": {
+			username:      "baduser",
+			password:      "something",
+			mechanism:     "SCRAM-SHA-256",
+			wrongPassword: true,
```
rumyantseva, line 70: Do we test a wrong password for an existing user too?

Neustradamus:
@henvic: Nice, good job!

Linked to:
- https://github.com/scram-sasl/info/issues/1
- https://github.com/FerretDB/FerretDB/issues/2012
- https://github.com/FerretDB/FerretDB/issues/3784
- https://github.com/FerretDB/FerretDB/pull/3904
- https://github.com/FerretDB/FerretDB/issues/3938
- https://github.com/FerretDB/FerretDB/pull/3970
- https://github.com/FerretDB/FerretDB/pull/3989

internal/util/password/scramsha256.go
```diff
@@ -0,0 +1,54 @@
...
+		return nil, err
+	}
+
+	salt := make([]byte, 45)
+	if _, err := rand.Read(salt); err != nil {
+		return nil, fmt.Errorf("failed to generate salt: %w", err)
+	}
+
+	kf := scram.KeyFactors{
+		Salt:  base64.StdEncoding.EncodeToString(salt),
```
b1ron, line 41: That should not be [encoded](https://pkg.go.dev/github.com/xdg-go/scram@v1.1.2#KeyFactors).

```suggestion
		Salt:  string(salt),
```

That will fix the invalid client [proof](https://github.com/xdg-go/scram/blob/v1.1.2/server_conv.go#L143).

internal/util/password/scramsha256.go
```diff
@@ -0,0 +1,54 @@
...
+		Iters: 15000,
+	}
+
+	cred := client.GetStoredCredentials(kf)
+
+	doc := must.NotFail(types.NewDocument())
+	doc.Set("storedKey", base64.StdEncoding.EncodeToString(cred.StoredKey))
+	doc.Set("iterationCount", int32(kf.Iters))
+	doc.Set("salt", kf.Salt)
```
b1ron, line 50:
```suggestion
	doc.Set("salt", base64.StdEncoding.EncodeToString(salt))
```

internal/util/password/scramsha256.go
```diff
@@ -0,0 +1,54 @@
...
+	"encoding/base64"
+	"fmt"
+
+	"github.com/xdg-go/scram"
+
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/must"
+)
+
+// NewSCRAMSHA256 authentication.
```
b1ron, line 28:
```suggestion
// NewSCRAMSHA256 computes the SCRAM-SHA-256 credentials and returns the document that should be stored.
```

internal/util/password/scramsha256.go
AlekSi: Without actually looking into it, may I suggest checking if `string` BSON type is not confused with `types.Binary`?

integration/users/connection_test.go
henvic: Yes. This tested user actually exists.

noisersup (dismissed):
Looks safe to me :)

internal/handler/msg_saslstart.go
```diff
@@ -129,3 +147,117 @@ func saslStartPlain(doc *types.Document) (string, string, error) {
...
+
+	dbName, err := common.GetRequiredParam[string](doc, "$db")
+	if err != nil {
+		return "", err
+	}
+
+	scramServer, err := scram.SHA256.NewServer(func(username string) (scram.StoredCredentials, error) {
+		return h.credentialLookup(ctx, username, dbName)
+	})
+	must.NoError(err)
```
chilagrow, line 251: Should we handle error gracefully rather than panic?
```suggestion
	if err != nil{
		return "", err
	}
```

chilagrow (commented):
Great work a few minor questions and comments :hugs:

internal/util/password/scramsha256.go
AlekSi, line 1: Please extract changes in this package into a separate PR and make them closer to what we already have there, with a similar API

internal/util/password/scramsha256.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+
+	"github.com/xdg-go/scram"
+
+	"github.com/FerretDB/FerretDB/internal/types"
+	"github.com/FerretDB/FerretDB/internal/util/must"
+)
+
+// NewSCRAMSHA256 computes SCRAM-SHA-256 credentials and returns the document that should be stored.
+func NewSCRAMSHA256(username, password string) (*types.Document, error) {
```
AlekSi, line 29:
```suggestion
func SCRAMSHA256Hash(username, password string) (*types.Document, error) {
```

internal/util/password/scramsha256.go
```diff
@@ -0,0 +1,54 @@
...
+	client, err := scram.SHA256.NewClient(username, password, "")
+	if err != nil {
+		return nil, err
+	}
+
+	salt := make([]byte, 45)
```
AlekSi, line 35: That constant should be in the `fixedSCRAMSHA256Params` package-level variable

internal/util/password/scramsha256.go
```diff
@@ -0,0 +1,54 @@
...
+	if _, err := rand.Read(salt); err != nil {
+		return nil, fmt.Errorf("failed to generate salt: %w", err)
+	}
+
+	kf := scram.KeyFactors{
+		Salt:  string(salt),
+		Iters: 15000,
```
AlekSi, line 42: The same

internal/util/password/scramsha256.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	cred := client.GetStoredCredentials(kf)
+
+	doc := must.NotFail(types.NewDocument())
+	doc.Set("storedKey", base64.StdEncoding.EncodeToString(cred.StoredKey))
+	doc.Set("iterationCount", int32(kf.Iters))
+	doc.Set("salt", base64.StdEncoding.EncodeToString(salt))
+	doc.Set("serverKey", base64.StdEncoding.EncodeToString(cred.ServerKey))
```
AlekSi, line 51: Are you sure that MongoDB stores those values as base64-encoded strings and not in types.Binary?

internal/util/password/scramsha256_test.go
```diff
@@ -0,0 +1,52 @@
...
+	for name, tc := range tests {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			got, err := NewSCRAMSHA256(tc.username, tc.password)
+			require.NoError(t, err)
+			assert.Equal(t, must.NotFail(got.Get("iterationCount")), int32(15000))
+			assert.Len(t, must.NotFail(got.Get("salt")).(string), 60)
+			assert.Len(t, must.NotFail(got.Get("serverKey")).(string), 44)
+			assert.Len(t, must.NotFail(got.Get("storedKey")).(string), 44)
```
AlekSi, line 49: The test should compare values, not their length

AlekSi (changes requested):

internal/handler/msg_saslstart.go
```diff
@@ -61,25 +65,39 @@ func (h *Handler) MsgSASLStart(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
...
+	case "SCRAM-SHA-256":
+		response, err := h.saslStartSCRAMSHA256(ctx, document)
+		if err != nil {
+			return nil, err
+		}
+
+		must.NoError(reply.SetSections(wire.MakeOpMsgSection(
+			must.NotFail(types.NewDocument(
+				"ok", float64(1),
+				"conversationId", int32(1),
```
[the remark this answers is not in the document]
b1ron: The `conversationId` doesn't seem to change for the duration of a conversation. See here for a [definition](https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#L431).

internal/handler/msg_saslstart.go
```diff
@@ -61,25 +65,39 @@ func (h *Handler) MsgSASLStart(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
...
+				"done", false,
```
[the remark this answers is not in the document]
b1ron: This indicates that the conversation is not yet complete and that the client will send a proof in the subsequent `saslContinue` [command](https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#L651-L654).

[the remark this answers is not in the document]
henvic: Also, we don't have an explicit test case for this as it'd be hard to write such a test given that the `saslStart` command is triggered by the MongoDB client.

internal/handler/msg_saslstart.go
```diff
@@ -129,3 +147,117 @@ func saslStartPlain(doc *types.Document) (string, string, error) {
...
+	}
+
+	return scram.StoredCredentials{}, errors.New("user not found")
+}
+
+func (h *Handler) saslStartSCRAMSHA256(ctx context.Context, doc *types.Document) (string, error) {
+	var payload []byte
+
+	binaryPayload, err := common.GetRequiredParam[types.Binary](doc, "payload")
+	if err == nil {
```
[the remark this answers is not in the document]
henvic: This wouldn't work. Fixing.

internal/util/password/scramsha256_test.go
henvic: The iterationCount is the only one I can compare value directly.
NewSCRAMSHA256 actually generates the other values:

* salt by calling crypto/rand.Read within NewSCRAMSHA256 itself
* serverKey and storedKey by calling `client.GetStoredCredentials`

AlekSi: Yet existing code in that package _somehow_ solves that problem

internal/handler/msg_saslstart.go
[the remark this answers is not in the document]
chilagrow: It seems strange to have a field that never changes. Perhaps, conversationId is used by other authentication mechanism and SCRAM-SHA-256 does not use it, I'm not sure how it works

internal/handler/msg_saslstart.go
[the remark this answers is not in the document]
chilagrow: Thanks :pray:

internal/util/password/scramsha256.go
b1ron: It stores credentials as base64 encoded strings except for the iteration count. But there's a new problem see [thread](https://ferretdb-inc.slack.com/archives/C034LBM1J5V/p1706706734612669?thread_ts=1706704672.835109&cid=C034LBM1J5V) for details.

```
byron@byron-ThinkCentre-M910q:~/workdir/FerretDB$ mongo --quiet --port 47017
rs0:PRIMARY> use admin
switched to db admin
rs0:PRIMARY> typeof db.system.users.findOne().credentials['SCRAM-SHA-256'].storedKey
string
rs0:PRIMARY> db.system.users.findOne().credentials['SCRAM-SHA-256']
{
	"iterationCount" : 15000,
	"salt" : "2alXmLg2Yiz6XVWum5aYfW0OjqGrJfHukMSd1A==",
	"storedKey" : "y+bjwKtKYc66+5Tb+PZK5pngmsz/hsItkH0p7TPDg3g=",
	"serverKey" : "RRqDHT16fc82Kqykgbn4poCKfDkqOaDpPGx3QN9c0A0="
}
rs0:PRIMARY> 
```

internal/handler/msg_saslstart.go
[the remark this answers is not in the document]
b1ron: It seems to indicate some state that's remembered during the conversation, but it doesn't make a lot of sense to me either.

> conversationId: the conversation identifier. This will need to be remembered and used for the duration of the conversation.

internal/handler/msg_createuser.go
```diff
@@ -218,3 +183,49 @@ func (h *Handler) MsgCreateUser(ctx context.Context, msg *wire.OpMsg) (*wire.OpM
 
 	return &reply, nil
 }
+
+// makeCredentials creates a document with credentials for the chosen mechanisms.
+func makeCredentials(mechanisms *types.Array, username, pwd string) (*types.Document, error) {
+	if mechanisms == nil {
```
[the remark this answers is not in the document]
henvic: It's never nil. Removed this check.

internal/handler/msg_saslcontinue.go
```diff
@@ -0,0 +1,67 @@
...
+// MsgSASLContinue implements `saslContinue` command.
+func (h *Handler) MsgSASLContinue(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
+	doc, err := msg.Document()
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+
+	var payload []byte
+
+	// most drivers follow spec and send payload as a binary
```
chilagrow, line 38: This comment suggests there was a driver that sends payload in a different format (string?), do we want to keep allowing that?

internal/handler/handler.go
```diff
@@ -192,7 +192,6 @@ func (h *Handler) cleanupAllCappedCollections(ctx context.Context) error {
 	}()
 
 	connInfo := conninfo.New()
-	connInfo.BypassBackendAuth = true
```
chilagrow, line 195: Does cleaning capped collection work without this?

internal/handler/msg_saslstart.go
```diff
@@ -129,3 +149,122 @@ func saslStartPlain(doc *types.Document) (string, string, error) {
 
 	return string(authcid), string(passwd), nil
 }
+
+func (h *Handler) credentialLookup(ctx context.Context, username, dbName string) (scram.StoredCredentials, error) {
```
chilagrow, line 153: Let's have comment for this function :hugs:

integration/users/connection_test.go
```diff
@@ -165,15 +200,21 @@ func TestAuthentication(t *testing.T) {
...
-			require.NotEmpty(t, id)
-
-			var result bson.D
-			err = connCollection.FindOne(ctx, bson.D{{"_id", id}}).Decode(&result)
-			require.NoError(t, err, "cannot find document")
-			assert.Equal(t, bson.D{{"_id", id}, {"ping", "pong"}}, result)
+
+			require.NotNil(t, connCollection, "cannot get collection")
+
+			// TODO https://github.com/FerretDB/FerretDB/issues/3121
```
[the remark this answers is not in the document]
henvic: @chilagrow, yes. The problem is with the insertion. I discussed this with @b1ron once I tested this code and gotten an `unknown field "apiVersion"` error.

internal/handler/msg_saslstart.go
henvic: Added.

internal/handler/handler.go
henvic: That was mistakenly removed and not readded. I added it back.

internal/handler/msg_saslcontinue.go
henvic: Discussed this with @b1ron. He checked that no driver seems to be doing this for SCRAM and we decided to remove this comment for SCRAM. The alternative would be to [actually] add code to handle that:

```go
stringPayload, err := common.GetRequiredParam[string](doc, "payload")
if err == nil {
	if payload, err = base64.StdEncoding.DecodeString(stringPayload); err != nil {
		return "", "", handlererrors.NewCommandErrorMsgWithArgument(
			handlererrors.ErrBadValue,
			fmt.Sprintf("Invalid payload: %v", err),
			"payload",
		)
	}
}
```

internal/handler/msg_saslstart.go
```diff
@@ -61,25 +65,41 @@ func (h *Handler) MsgSASLStart(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
...
+		}
+
+		conninfo.Get(ctx).BypassBackendAuth()
+
+		must.NoError(reply.SetSections(wire.MakeOpMsgSection(
+			must.NotFail(types.NewDocument(
+				"ok", float64(1),
+				"conversationId", int32(1),
+				"done", false,
+				"payload", response,
```
b1ron, line 93: I checkout the PR and was unable to authenticate because the payloads in the responses need to be binary. The tweak below allows a client to successfully authenticate with the exception of the legacy shell due to an invalid salt length and the dotnet driver because it uses the `skipEmptyExchange` option to avoid an extra round trip.

```go
	b := types.Binary{}
	b.B = []byte(resp)
	b.Subtype = 0

	var reply wire.OpMsg
	must.NoError(reply.SetSections(wire.MakeOpMsgSection(
		must.NotFail(types.NewDocument(
			"conversationId", int32(1),
			"done", true,
			"payload", b,
			"ok", float64(1),
		)),
	)))
```

With the exception of the legacy shell due to an invalid salt length and the dotnet driv

internal/handler/msg_saslcontinue.go
```diff
@@ -0,0 +1,66 @@
...
+
+	resp, err := conv.Step(string(payload))
+	if err != nil {
+		return nil, handlererrors.NewCommandErrorMsg(
+			handlererrors.ErrAuthenticationFailed,
+			"Authentication failed.",
+		)
+	}
+
+	var reply wire.OpMsg
```
b1ron, line 55: Here too.

```go
	b := types.Binary{}
	b.B = []byte(resp)
	b.Subtype = 0

	var reply wire.OpMsg
	must.NoError(reply.SetSections(wire.MakeOpMsgSection(
		must.NotFail(types.NewDocument(
			"conversationId", int32(1),
			"done", true,
			"payload", b,
			"ok", float64(1),
		)),
	)))
```

internal/handler/msg_saslcontinue.go
```diff
@@ -0,0 +1,66 @@
...
+	must.NoError(reply.SetSections(wire.MakeOpMsgSection(
+		must.NotFail(types.NewDocument(
+			"conversationId", int32(1),
+			"done", true,
+			"payload", resp,
```
b1ron, line 60: Here too, it needs to be of a `types.Binary`.

internal/handler/msg_saslstart.go
b1ron: Just ignore what `skipEmptyExchange` means for now as it requires further investigation.

internal/handler/msg_saslcontinue.go
chilagrow: Let's not add code. It's good news if all driver use binary :+1:

integration/users/connection_test.go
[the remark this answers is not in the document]
chilagrow: Ouch, `apiVersion` error seems to be a blocker for the new authentication code.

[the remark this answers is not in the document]
chilagrow: Insert will work if we don't set `.SetServerAPIOptions(...)` on client. Is there any reason we wanted to specify API version explicitly?

internal/handler/msg_saslstart.go
chilagrow: Does this mean existing integration tests are not sufficient? :thinking: I mean manual testing is great, but I thought such scenario is covered by integration tests

integration/users/connection_test.go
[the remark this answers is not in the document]
henvic: @chilagrow, oh! Good point. Thanks! I've tried removing the server API version field and it worked flawlessly. I'm going to commit & push the change now.

internal/handler/msg_saslstart.go
chilagrow: Yeah that's a nice to have.

internal/handler/msg_saslstart.go
```diff
@@ -129,3 +153,124 @@ func saslStartPlain(doc *types.Document) (string, string, error) {
...
+				)
+			}
+
+			credentials := must.NotFail(v.Get("credentials")).(*types.Document)
+
+			if !credentials.Has("SCRAM-SHA-256") {
+				return scram.StoredCredentials{}, handlererrors.NewCommandErrorMsgWithArgument(
+					handlererrors.ErrMechanismUnavailable,
+					"User has no SCRAM-SHA-256 based authentication credentials registered",
+					"SCRAM-SHA-256",
```
chilagrow, line 218: Can we test this scenario?

chilagrow (dismissed):
Overall looks great. See comment about 2 untested errors, which would be nice to test if possible.

internal/util/password/scramsha256.go
```diff
@@ -52,7 +52,7 @@ type scramSHA256Params struct {
 // fixedScramSHA256Params represent fixed password parameters for SCRAM-SHA-256 authentication.
 var fixedScramSHA256Params = &scramSHA256Params{
 	iterationCount: 15_000,
-	saltLen:        45,
+	saltLen:        30,
```
henvic, line 55: Adjusting to the same length used by MongoDB.

internal/handler/msg_saslstart.go
```diff
@@ -129,3 +147,117 @@ func saslStartPlain(doc *types.Document) (string, string, error) {
...
+		}
+
+		matches, err := common.FilterDocument(v, filter)
+		if err != nil {
+			return scram.StoredCredentials{}, lazyerrors.Error(err)
+		}
+
+		if matches {
+			if !v.Has("credentials") {
+				return scram.StoredCredentials{}, handlererrors.NewCommandErrorMsgWithArgument(
```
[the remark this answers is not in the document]
henvic: It's not reachable given that credentials should always exist. I'm removing the check.

chilagrow (approved):

internal/handler/msg_usersinfo.go
```diff
@@ -169,6 +169,18 @@ func (h *Handler) MsgUsersInfo(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
 			return nil, lazyerrors.Error(err)
 		}
 
+		if v.Has("credentials") {
+			credentials := must.NotFail(v.Get("credentials")).(*types.Document)
+			if credentialsKeys := credentials.Keys(); len(credentialsKeys) > 0 {
+				mechanisms := must.NotFail(types.NewArray())
+				for _, k := range credentialsKeys {
```
rumyantseva, line 176: If `credentialsKeys` is a slice of strings, I guess we can do something like `types.NewArray(credentialsKeys...)` to save a few lines of code :)

internal/handler/msg_saslstart.go
```diff
@@ -129,3 +153,116 @@ func saslStartPlain(doc *types.Document) (string, string, error) {
 
 	return string(authcid), string(passwd), nil
 }
+
+// scramCredentialLookup looks up an user's credentials in the database.
+func (h *Handler) scramCredentialLookup(ctx context.Context, username, dbName string) (scram.StoredCredentials, error) {
```
rumyantseva, line 158: Why do we return `scram.StoredCredentials` and not `*scram.StoredCredentials` in this case? 
(Personally, I'm ok with both option but the option with pointer looks a bit more typical for the rest of the code, and also semantically it's easier to interpret (nil pointer is returner in case of an error)

internal/handler/msg_saslstart.go
```diff
@@ -129,3 +153,116 @@ func saslStartPlain(doc *types.Document) (string, string, error) {
...
+				StoredKey: storedKey,
+				ServerKey: serverKey,
+			}, nil
+		}
+	}
+
+	return scram.StoredCredentials{}, errors.New("user not found")
+}
+
+// saslStartSCRAMSHA256 extracts the initial challenge and respond the client.
```
rumyantseva, line 234: I don't quite understand the godoc line, so I suggest this (please check if it's correct):
```suggestion
// saslStartSCRAMSHA256 extracts the initial challenge, checks credentials, and returns the response that needs to be sent to the client.
//
// If credentials are not found, it returns XYZ.
```

internal/handler/msg_saslstart.go
```diff
@@ -129,3 +153,116 @@ func saslStartPlain(doc *types.Document) (string, string, error) {
...
+					Salt:  string(salt),
+					Iters: int(must.NotFail(cred.Get("iterationCount")).(int32)),
+				},
+				StoredKey: storedKey,
+				ServerKey: serverKey,
+			}, nil
+		}
+	}
+
+	return scram.StoredCredentials{}, errors.New("user not found")
```
rumyantseva, line 231: This error message doesn't seem to be checked anywhere (in tests or in some parts of the code).

The other places in this function either return `lazyerror` or `handlererrors.NewCommandErrorMsg`. Should this one be a `lazyerror` or a `handlererrors.NewCommandErrorMsg` error too?

rumyantseva (dismissed):
The code looks good. I added a few suggestions (mostly about incosistency), but they are not super crucial.

internal/handler/msg_usersinfo.go
henvic: I'm afraid we can't do that because types.NewArray receives a slice of []any and credentialsKeys is a slice of []string.

internal/handler/msg_saslstart.go
henvic: I didn't do that because I was trying to avoid this weird `return *cred, nil` line here, so as it wasn't necessary I left it as this, but I see it's happening just on a single place, so I'm changing it now.

```go
scramServer, err := scram.SHA256.NewServer(func(username string) (scram.StoredCredentials, error) {
		cred, err := h.scramCredentialLookup(ctx, username, dbName, "SCRAM-SHA-256")
		if err != nil {
			return scram.StoredCredentials{}, err
		}
		return *cred, nil
	})
	if err != nil {
		return "", err
	}
```

internal/handler/msg_saslstart.go
henvic: Probably

```
// saslStartSCRAMSHA256 extracts the initial challenge and attempts to move the
// authentication conversation forward.
```

(similar to how `*scram.ServerConversation.Step` explains) is more precise as it checks if the challenge can be resolved, rather than the credential itself.

internal/handler/msg_saslstart.go
b1ron: Sorry nit: the payload variable is called `response` here, but `resp` in msg_saslcontinue.go.

internal/handler/msg_saslstart.go
AlekSi: FWIW, it is our documented best practice: https://github.com/FerretDB/FerretDB/blob/e1eaaad891265ac947fbf55fd5f736ba301c9bea/CONTRIBUTING.md?plain=1#L260-L261

internal/clientconn/conninfo/conn_info.go
```diff
@@ -35,12 +37,14 @@ type ConnInfo struct {
...
 
+	sc *scram.ServerConversation // protected by rw
+
 	// If true, backend implementations should not perform authentication
 	// by adding username and password to the connection string.
 	// It is set to true for background connections (such us capped collections cleanup)
 	// and by the new authentication mechanism.
 	// See where it is used for more details.
-	BypassBackendAuth bool
+	bypassBackendAuth bool
```
AlekSi, line 47:
```suggestion
	bypassBackendAuth bool // protected by rw
```

Normally, there would be a [mutex hat](https://dmitri.shuralyov.com/idiomatic-go/entries/2), but to make this struct smaller the order of fields is weird. So we should have a comment

internal/clientconn/conninfo/conn_info.go
```diff
@@ -91,6 +112,22 @@ func (connInfo *ConnInfo) SetMetadataRecv() {
...
+	connInfo.bypassBackendAuth = true
+}
+
+// GetBypassBackendAuth returns whether the connection requires backend authentication.
+func (connInfo *ConnInfo) GetBypassBackendAuth() bool {
+	connInfo.rw.RLock()
+	defer connInfo.rw.RUnlock()
+
+	return connInfo.bypassBackendAuth
+}
```
AlekSi, line 129: Note how other methods are called `X` and `SetX`, not `GetX` and `X`.

internal/handler/msg_saslstart.go
henvic: Fixed.

internal/clientconn/conninfo/conn_info.go
henvic: Changed!

chilagrow (approved):

AlekSi (approved):

PR #4031, opened 2024-02-01
Add utility for hashing SCRAM-SHA-256 password

### LINKED ISSUE #3784: Support `SCRAM-SHA-256` authentication mechanism

henvic:
# Description

Closes #3784.

## Readiness checklist


internal/util/password/scramsha256_test.go
```diff
@@ -0,0 +1,230 @@
...
+			"salt", "c2EAbHQ=",
+			"serverKey", "qpmMln7z50yoT66R1lT55pUvBxu8BqZU4X8dRp38NWo=",
+		)),
+	},
+}
+
+func TestSCRAMSHA256(t *testing.T) {
+	for i, tc := range scramSHA256TestCases {
+		i, tc := i, tc
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
```
chilagrow, line 164: Just a suggestion :hugs: To identify failing test case easily, we use often `map[string]struct` for test cases and name each test, that way we can identify failing test with name instead of an arbitrary number. Also allow us to see what each test case is testing from the name
```suggestion
		name, tc := name, tc
		t.Run(name, func(t *testing.T) {
```

internal/util/password/scramsha256_test.go
```diff
@@ -0,0 +1,230 @@
...
+			t.Parallel()
+
+			doc, err := scramSHA256HashParams(tc.username, tc.password, tc.salt, &tc.params)
+
+			if tc.err != "" {
+				assert.Contains(t, err.Error(), tc.err)
```
chilagrow, line 170:
```suggestion
				assert.ErrorContains(t, err, tc.err)
```

internal/util/password/scramsha256_test.go
```diff
@@ -0,0 +1,230 @@
...
+				return scram.StoredCredentials{
+					KeyFactors: scram.KeyFactors{
+						Salt:  string(tc.salt),
+						Iters: tc.params.iterationCount,
+					},
+					StoredKey: []byte(must.NotFail(doc.Get("storedKey")).(string)),
+					ServerKey: []byte(must.NotFail(doc.Get("serverKey")).(string)),
+				}, nil
+			})
+			must.NoError(err)
```
chilagrow, line 187:
```suggestion
			require.NoError(t, err)
```

internal/util/password/scramsha256_test.go
```diff
@@ -0,0 +1,230 @@
...
+	params   scramSHA256Params
+	username string
+	password string
+	salt     []byte
+
+	want *types.Document
+	err  string
+}
+
+var scramSHA256TestCases = []scramSHA256TestCase{
```
chilagrow, line 43: The exported `SCRAMSHA256Hash()` uses fixed length. Why do we test many combination len and count of `scramSHA256Params`?

internal/util/password/scramsha256.go
```diff
@@ -0,0 +1,87 @@
...
+var fixedScramSHA256Params = &scramSHA256Params{
+	iterationCount: 15_000,
+	saltLen:        45,
+}
+
+// scramSHA256HashParams hashes the password with the given salt and parameters,
+// and returns the document that should be stored.
+func scramSHA256HashParams(username, password string, salt []byte, params *scramSHA256Params) (*types.Document, error) {
+	if len(salt) != int(params.saltLen) {
+		return nil, lazyerrors.Errorf("unexpected salt length: %d", len(salt))
```
chilagrow, line 43: Salt is created from `scramSHA256Params.saltLen`, so how could this happen?

internal/util/password/scramsha256_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+			resp, err := client.NewConversation().Step("")
+			require.NoError(t, err)
+
+			resp, err = conv.Step(resp)
+			require.NoError(t, err)
+			assert.NotEmpty(t, resp)
+
+			_, err = conv.Step("wrong")
+			assert.NotNil(t, err)
```
chilagrow, line 203: Or `assert.EqualError()` for comparing specific error. 
```suggestion
			assert.Error(t, err)
```

internal/util/password/scramsha256_test.go
```diff
@@ -0,0 +1,230 @@
...
+						Salt:  string(tc.salt),
+						Iters: tc.params.iterationCount,
+					},
+					StoredKey: []byte(must.NotFail(doc.Get("storedKey")).(string)),
+					ServerKey: []byte(must.NotFail(doc.Get("serverKey")).(string)),
+				}, nil
+			})
+			must.NoError(err)
+
+			// Check if the generated authentication is valid by simulating a conversation.
```
chilagrow, line 189: Thanks for this check, great example :hugs:

chilagrow (commented):
Just a few minor comments :hugs:

internal/util/password/scramsha256.go
henvic: @AlekSi , what do you think about removing this assertion and the one for PLAIN as well? Or should we keep it as a safeguard as it's something critical?

internal/util/password/scramsha256_test.go
AlekSi: The idea is to simplify migration from MongoDB by supporting their parameters. So we support _importing_ `system.users` collection from MongoDB with different parameters, but use fixed parameters for _creating_ new users. Part of it is not there yet.

chilagrow (approved):
Great work!

internal/util/password/scramsha256.go
chilagrow: Just saying for me either is fine. Let the stronger voice speak :hugs:

noisersup (approved):

internal/util/password/plain_test.go
```diff
@@ -37,8 +37,6 @@ type plainTestCase struct {
 }
 
 // https://github.com/brycx/Test-Vector-Generation/blob/master/PBKDF2/pbkdf2-hmac-sha2-test-vectors.md
-//
-// Test Case 4 is skipped because it takes too long.
```
AlekSi, line 41: https://github.com/brycx/Test-Vector-Generation/blob/master/PBKDF2/pbkdf2-hmac-sha2-test-vectors.md#test-case-4

internal/util/password/scramsha256_test.go
```diff
@@ -0,0 +1,233 @@
...
+	params   scramSHA256Params
+	username string
+	password string
+	salt     []byte
+
+	want *types.Document
+	err  string
+}
+
+var scramSHA256TestCases = map[string]scramSHA256TestCase{
```
AlekSi, line 42: What is the source of those test cases?

henvic: I've tried to find SCRAM-SHA-256 test vectors but haven't found any source (e.g., NIST Cryptographic Algorithm Validation Program, PostgreSQL, and other open-source code) with precise input/output values I could use to recalculate and match against our implementation.

So I added test cases with values generated by MongoDB and explained how to generate such values in a comment.

internal/util/password/scramsha256_test.go
```diff
@@ -0,0 +1,275 @@
...
+	// Test vector generated with db.runCommand({createUser: "user", pwd: "pencil", roles: []})
+	"FromMongoDB": {
+		params: scramSHA256Params{
+			iterationCount: 15000,
+			saltLen:        28,
+		},
+		password: "pencil",
+		salt:     decodeBase64("vXan6ZbWmm5i+f+mKY598rnIfoAGGp+G9NP0qQ=="),
```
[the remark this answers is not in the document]
henvic: as in the following?

```go
salt: func() []byte {
	b, _ := base64.StdEncoding.DecodeString("vXan6ZbWmm5i+f+mKY598rnIfoAGGp+G9NP0qQ==")
		return b
}(),
```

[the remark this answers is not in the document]
AlekSi: As in

```go
salt string
// …
salt: "vXan6ZbWmm5i+f+mKY598rnIfoAGGp+G9NP0qQ=="
// …
t.Run(// …
// …
salt := must.NotFail(base64.StdEncoding.DecodeString(tc.salt))
```

internal/util/password/scramsha256.go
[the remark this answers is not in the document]
henvic: Note for clarification: in the case of SCRAM, the verification might fail after each step, and not the first one, so that's why it has a Step function receiving a challenge that it's checked at the first time when `MsgSASLStart` is called, and subsequently, at each `MsgSASLContinue` call.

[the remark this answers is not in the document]
AlekSi: Ok, I think I understand where the confusion is. Let's finish that PR as it is now, without the verification function. Then work with @b1ron to figure out how to add support for MongoDB's flavor of SCRAM-SHA-1 using https://github.com/xdg-go/scram

internal/util/password/scramsha256.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// scramSHA256HashParams hashes the password with the given salt and parameters,
+// and returns the document that should be stored.
+func scramSHA256HashParams(password string, salt []byte, params *scramSHA256Params) (*types.Document, error) {
+	if len(salt) != int(params.saltLen) {
+		return nil, lazyerrors.Errorf("unexpected salt length: %d", len(salt))
+	}
+
+	// username is omitted because it is not used in the hash computation
```
chilagrow, line 46: This confuses me, client is created only using password? Then its stored credential has no information about its username :thinking:

henvic: @chilagrow,

Yes, that client wouldn't work to establish a connection. I removed the username from the public API as it's not needed and we discard this client immediately after password generation.

The client is only used there as this library doesn't provide a function for hashing the password directly, without creating one.

Instead of using

```go
// username is omitted because it is not used in the hash computation
client, err := scram.SHA256.NewClient("", password, "")
if err != nil {
}

kf := scram.KeyFactors{
	Salt:  string(salt),
	Iters: params.iterationCount,
}

cred := client.GetStoredCredentials(kf)
```

We could also use something like this, avoiding using this library altogether:

```go
func computeHMAC( key, data []byte) []byte {
	mac := hmac.New(sha256.New, key)
	mac.Write(data)
	return mac.Sum(nil)
}

func computeHash(b []byte) []byte {
	h := sha256.New()
	h.Write(b)
	return h.Sum(nil)
}

var prepPassword string

if prepPassword, err = stringprep.SASLprep.Prepare(password); err != nil {
	return nil, fmt.Errorf("Error SASLprepping password '%s': %v", password, err)
}

saltedPassword := pbkdf2.Key([]byte(prepPassword), []byte(kf.Salt), kf.Iters, 32, aha256.New)
clientKey := computeHMAC(saltedPassword, []byte("Client Key"))
storedKey := computeHash(clientKey)
serverKey := computeHMAC(saltedPassword, []byte("Server Key"))
```

Another option would be to try to open a pull request upstream creating a function for that, but that might take time or not get accepted.

What do you prefer? Either is fine, and I'd usually prefer the second choice, despite being more complex.

henvic: Also notice that this function only generates the password credentials, and the callee that uses it is where the data is actually recorded.

chilagrow: I see, we are using that library to just hash the password. Now your comment makes sense. I agree with you, second choice is preferable, even complex. Because the usage of library misguides creating client to establish connection which we are not doing. :see_no_evil:

henvic: Going to do it now then.

henvic: @chilagrow, you can take a look now. I've done the change :)

chilagrow (approved):
LGTM :hugs:

AlekSi (approved):

PR #4046, opened 2024-02-07
Minor cleanups

AlekSi:
# Description

Bump dependencies and port small changes from other PRs.

## Readiness checklist


internal/clientconn/conn.go
```diff
@@ -581,26 +581,31 @@ func (c *conn) handleOpMsg(ctx context.Context, msg *wire.OpMsg, command string)
...
+			ok = v == 1
 		}
+
+		if !ok {
+			level = zap.WarnLevel
+		}
+	}
+
+	if closeConn {
+		level = zap.ErrorLevel
```
chilagrow, line 608: Is it intended to change log level to errorLevel for `closeConn == true` even if `ok == 1`?

chilagrow (approved):
Just one question, thank you for updating dependencies :hugs:

internal/clientconn/conn.go
AlekSi: I don't think that case exists: `closeConn` represents connection closing as a result of handling some command, and that's always something unrecoverable

chilagrow: I see, so it would never be ok == 1 that case. Thanks for the explanation :hugs:

Issue #3121, opened 2023-07-26
We should ignore Stable API / `apiVersion` / strict parameters


b1ron:
### FerretDB version

v1.7.0

### Backend

PostgreSQL

### Environment

- OS:
- Deployment:
- Deployment details:

### What did you do?

See repro in #3120.

This is not exactly a bug because it's a specific MongoDB parameter which will prevent a client from using commands that are deemed "unstable". The stable API encompasses a subset of commands that will **not** break your application when you upgrade. However, if a client has these parameters enabled it will send them in a command document as top-level fields. We _should_ ignore them.

### What did you expect to see?

I think it's the same issue as #3010, extracting parameters breaks stuff.

### What did you see instead?

```
dance_ferretdb  |         "errmsg": "insert: unknown field \"apiVersion\"",
dance_ferretdb  |         "code": 9,
dance_ferretdb  |         "codeName": "FailedToParse"
```

chilagrow:
Do we have a list of commands that sends Stable API parameters? I found this page https://www.mongodb.com/docs/manual/reference/stable-api-changelog/#std-label-stable-api-changelog

AlekSi:
on @Evengard

PR #4066, opened 2024-02-12
Fixes inability to authenticate through the older cmdQuery mechanism

### LINKED ISSUE #4065: C# MongoDB driver: "no PostgreSQL user name specified in startup packet"
### FerretDB version

1.19.0

### Backend

PostgreSQL 15.2

### Environment

- OS: Debian 11.8
- Deployment: k8s deployment
- Deployment details: Both FerretDB and PostgreSQL are deployed in an k8s instance

### What did you do?

I either can't authenticate at all (if not using StableAPI), or can't Insert documents (if using the StableAPI).

Consider this sample which will fail no matter if I set StableAPI or not:

```csharp
using MongoDB.Driver;

var mongoClientSettings = MongoClientSettings.FromConnectionString("mongodb://username:password@myferretdb:25017/?authMechanism=PLAIN");
mongoClientSettings.ServerApi = new ServerApi(ServerApiVersion.V1);
var mongoClient = new MongoClient(mongoClientSettings);
var mongoDb = mongoClient.GetDatabase("dm3dev");
var collection = mongoDb.GetCollection<Test>("test");
collection.InsertOne(new Test("World"));

public record Test(string Hello);
```

If I run it as it is, I get ` 'Command insert failed: insert: unknown field "apiVersion".'`, which seems to be related to https://github.com/FerretDB/FerretDB/issues/3121

If I run it without the line `mongoClientSettings.ServerApi = new ServerApi(ServerApiVersion.V1);` which sets the StableAPI, I then get:

 `
'Command insert failed: [msg_insert.go:169 handler.(*Handler).MsgInsert] [collection.go:123 postgresql.(*collection).InsertAll] [registry.go:479 metadata.(*Registry).CollectionCreate] [registry.go:132 metadata.(*Registry).getPool] [pool.go:128 pool.(*Pool).Get] [opendb.go:95 pool.openDB] [opendb.go:110 pool.checkConnection] failed to connect to `host=postgres-db-rw.postgres.svc.cluster.local user= database=dm3_ferret_dev`: server error (FATAL: no PostgreSQL user name specified in startup packet (SQLSTATE 28000)).'
`

Which looks like the username from the client isn't passed down to the Postgres server.

I expect it to work at least with some settings...

Evengard:
# Description

Apparently, the older opQuery authentication mechanism was just a stub which wasn't doing anything. Because of that, the DotNet MongoDB client wasn't able to authenticate, as it was using this older mechanism when not using StableAPI. I mimicked the mechanism from opMsg, and now DotNet works.

Closes #4065.

## Readiness checklist

- [X] I ensured that PR title is good enough for the changelog.

### DIFF
--- a/internal/handler/cmd_query.go
+++ b/internal/handler/cmd_query.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/FerretDB/FerretDB/internal/clientconn/conninfo"
 	"github.com/FerretDB/FerretDB/internal/handler/common"
 	"github.com/FerretDB/FerretDB/internal/handler/handlererrors"
 	"github.com/FerretDB/FerretDB/internal/types"
@@ -28,12 +29,13 @@ import (
 
 // CmdQuery implements deprecated OP_QUERY message handling.
 func (h *Handler) CmdQuery(ctx context.Context, query *wire.OpQuery) (*wire.OpReply, error) {
-	cmd := query.Query().Command()
+	body := query.Query()
+	cmd := body.Command()
 	collection := query.FullCollectionName
 
 	// both are valid and are allowed to be run against any database as we don't support authorization yet
 	if (cmd == "ismaster" || cmd == "isMaster") && strings.HasSuffix(collection, ".$cmd") {
-		return common.IsMaster(ctx, query.Query(), h.TCPHost, h.ReplSetName)
+		return common.IsMaster(ctx, body, h.TCPHost, h.ReplSetName)
 	}
 
 	// TODO https://github.com/FerretDB/FerretDB/issues/3008
@@ -41,6 +43,16 @@ func (h *Handler) CmdQuery(ctx context.Context, query *wire.OpQuery) (*wire.OpRe
 	// database name typically is either "$external" or "admin"
 
 	if cmd == "saslStart" && strings.HasSuffix(collection, ".$cmd") {
+		mechanism, err := common.GetRequiredParam[string](body, "mechanism")
+		if err == nil && mechanism == "PLAIN" {
+			username, password, err := saslStartPlain(body)
+			if err != nil {
+				return nil, err
+			}
+
+			conninfo.Get(ctx).SetAuth(username, password)
+		}
+
 		var emptyPayload types.Binary
 		reply := wire.OpReply{
 			NumberReturned: 1,


internal/handler/cmd_query.go
```diff
@@ -28,19 +29,30 @@ import (
...
+		return common.IsMaster(ctx, body, h.TCPHost, h.ReplSetName)
 	}
 
 	// TODO https://github.com/FerretDB/FerretDB/issues/3008
 
 	// database name typically is either "$external" or "admin"
 
 	if cmd == "saslStart" && strings.HasSuffix(collection, ".$cmd") {
+		mechanism, err := common.GetRequiredParam[string](body, "mechanism")
+		if err == nil && mechanism == "PLAIN" {
```
chilagrow, line 47: Let's return the error for the case where `err != nil`. :hugs: 

Let's also handle the case where `mechanism != "PLAIN"`. I think we can handle it similarly to how it's done in `msg_saslstart.go`

chilagrow (commented):
Thanks for your contribution :hugs:

internal/handler/cmd_query.go
Evengard: The idea was to provide extreme backwards compatibility by just falling back to the "stub" behavior if something went wrong, as I have no ideas if that would break any other clients. But I don't mind make explicit errors here...

chilagrow: Thanks for your quick response :hugs:. Let us check, @b1ron do you know which other driver uses `cmd_query.go`? My guess is, it doesn't work for those drivers? If that's the case having an appropriate error is sensible here for a client attempting to authenticate with other mechanisms.

b1ron: It depends on the specific driver and version. Looking at this [compatibility matrix](https://www.mongodb.com/docs/drivers/csharp/current/compatibility/#compatibility-table-legend) drivers that do not support MongoDB 6.0 will use legacy opcodes. I suspect @Evengard is using mongocsharpdriver version 2.15 or older.

> OP_QUERY
Deprecated in MongoDB 5.0. Removed in MongoDB 5.1.

Fixing #3121 might be easier.

AlekSi: > Removed in MongoDB 5.1.

*Narrator: It wasn't actually removed.*

But yeah, to move forward, we have to know the driver version.

Evengard: изображение

I specifically made sure to upgrade to the latest revision. This is the behaviour with latest stable.

Evengard: https://github.com/mongodb/mongo-csharp-driver/blob/20391c999873faf71e96e5109e2fda00a84a6e9a/src/MongoDB.Driver.Core/Core/WireProtocol/CommandWireProtocol.cs#L181

There's that method which seems to fallback to the OP_QUERY if it deemed the server version is too low. Maybe you have some pointers about how to force it use OP_MSG instead?

Evengard: I've actually found that the commit that fixes that behaviour in the C# client is there: https://github.com/mongodb/mongo-csharp-driver/commit/7be56f72ff484ffdeb9f021a8d582b4b6d62eeb8 except it is really recent, it didn't make yet into an official MongoDB.Driver release. I tried to compile the main branch of the MongoDB.Driver and tried to connect with it without this PR, and it actually worked. Not with the latest official 2.23.1 though.

Evengard: Looks like it is nearing the release: https://github.com/mongodb/mongo-csharp-driver/pull/1261

b1ron: I was able to force an OP_MSG using version 2.19.1, strange.

Evengard: The thing is I can force an OP_MSG authentication with StableAPI, it's the inserts which bugs then.

I can do a PR in attempts to ignore theese apiVersions parameters probably instead. But looking that the actual issue seemed abandoned wasn't sure it is the way to go.

Evengard: https://github.com/FerretDB/FerretDB/pull/4067 Went ahead with the StableAPI parameter ignoring... It works with 2.23.1 too.

b1ron: Thanks @Evengard, and apologies for my confusion. I was able to reproduce the error in both 2.19.1 and 2.23.1.

```
Unhandled exception. MongoDB.Driver.MongoCommandException: Command ping failed: [msg_ping.go:52 handler.(*Handler).MsgPing] [backend.go:67 postgresql.(*backend).Status] [registry.go:259 metadata.(*Registry).DatabaseList] [registry.go:132 metadata.(*Registry).getPool] [pool.go:128 pool.(*Pool).Get] [opendb.go:95 pool.openDB] [opendb.go:110 pool.checkConnection] failed to connect to `host=postgres user= database=ferretdb`: server error (FATAL: no PostgreSQL user name specified in startup packet (SQLSTATE 28000)).
```
#4067 fixes the issue I also tested it. The core team will review it.

AlekSi: We should do https://github.com/FerretDB/dance/issues/769 first

AlekSi (commented):
We should do https://github.com/FerretDB/dance/issues/769 first

AlekSi:
That issue was fixed in #4549, #4550, and #4552. Please let us know if it works for you.

PR #4075, opened 2024-02-14
Enforce new authentication

### LINKED ISSUE #1877: Enforce new authentication
### What should be done?

- `hello`
- `ping`
- `isMaster`
- `whatsmyuri`
- `buildInfo`
- `authenticate`
- `connectionStatus`
- `saslStart` 
- `saslContinue`
- `logout`

### Where?

Handler, backends.

### Definition of Done

- handler updated;
- all backends updated;
- spot refactorings done;

chilagrow:
# Description

Closes #1877.

## Readiness checklist


internal/handler/authenticate.go
```diff
@@ -0,0 +1,145 @@
...
+			break
+		}
+	}
+
+	if !hasUser {
+		// There is no user in the database, let the backend check the authentication.
+		// We do not want unauthenticated users accessing the database, while allowing
+		// users with valid backend credentials to access the database
+		// until local exception is implemented.
+		conninfo.Get(ctx).UnsetBypassBackendAuth()
```
chilagrow, line 102: Alternatively, we can just let any `PLAIN` user authenticated if there is no user, but I consider that could be dangerous in scenario if all users were dropped. FYI, for `SCRAM` authentication it fails if there is no user.

henvic: May we add a TODO here for the local exception implementation ticket? It'd be helpful to keep track of it :)

internal/handler/msg_aggregate.go
```diff
@@ -42,6 +42,10 @@ import (
 
 // MsgAggregate implements `aggregate` command.
 func (h *Handler) MsgAggregate(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
+	if err := h.authenticate(ctx); err != nil {
```
henvic, line 45: One alternative to adding this code wherever authentication is desired is to make authentication opt-out rather than opt-in. You might want to consider this to avoid needing to add this everywhere as this is a common security fragility (eventually someone might forget to add this on a new endpoint).

So, for endpoints that require no authentication, we could require them to be wrapped by a function on commands.go. For example, say MsgPing doesn't require authentication. Rather than

```go
"ping": {
	Handler: h.MsgPing,
	Help:    "Returns a pong response.",
},
```

We could have to write it as the following for it to work for unauthenticated users:

```go
"ping": {
	Handler: UnsafeHandler(h.MsgPing),
	Help:    "Returns a pong response.",
},
```

or

```go
"ping": {
	Handler: h.MsgPing,
	Unsafe: true,
	Help:    "Returns a pong response.",
},

// With a Unsafe boolean, it'd be easy to check programmatically what commands
// doesn't require authentication.
// Maybe useful to write a test to automatically check it & maintain a list of
// allowed commands for security purposes.
```

And then UnsafeHandler would mark that the handler as safe to call without authentication.

By the way, am I correct to assume most handlers need authentication, btw? :)

henvic (commented):
Great work! I've just a few questions and suggestions which you might want to consider.

integration/users/connection_test.go
```diff
@@ -38,6 +38,8 @@ func TestAuthentication(t *testing.T) {
 	collection := s.Collection
 	db := collection.Database()
 
+	createTestRunnerUser(t, ctx, s.Collection.Database())
+
```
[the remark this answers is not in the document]
henvic: Thanks for the explanation!

internal/backends/postgresql/metadata/registry.go
```diff
@@ -122,7 +122,11 @@ func (r *Registry) getPool(ctx context.Context) (*pgxpool.Pool, error) {
 
 	if connInfo.BypassBackendAuth() {
 		if p = r.p.GetAny(); p == nil {
-			return nil, lazyerrors.New("no connection pool")
+			var err error
+			// use credential from the base URI by passing empty values
+			if p, err = r.p.Get("", ""); err != nil {
+				return nil, lazyerrors.Error(err)
+			}
```
henvic, line 129: 🥇

internal/clientconn/conninfo/conn_info.go
```diff
@@ -120,6 +120,14 @@ func (connInfo *ConnInfo) SetBypassBackendAuth() {
...
 }
 
+// UnsetBypassBackendAuth marks the connection as requiring backend authentication.
+func (connInfo *ConnInfo) UnsetBypassBackendAuth() {
+	connInfo.rw.Lock()
+	defer connInfo.rw.Unlock()
+
+	connInfo.bypassBackendAuth = false
+}
+
```
henvic, line 130: Let's make sure we remove this once it's no longer needed.

internal/handler/msg_createuser.go
```diff
@@ -217,7 +221,7 @@ func makeCredentials(mechanisms *types.Array, username, pwd string) (*types.Docu
 
 		switch v {
 		case "PLAIN":
-			credentials.Set("PLAIN", must.NotFail(password.PlainHash(username)))
+			credentials.Set("PLAIN", must.NotFail(password.PlainHash(pwd)))
```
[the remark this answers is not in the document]
henvic: We might want to add a check on the existing tests then 👀

henvic (approved):
LGTM

internal/handler/msg_createuser.go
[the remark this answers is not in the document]
chilagrow: Yes test was added, how I discovered it. I believe it was this test `TestAuthenticationEnableNewAuthPLAIN/Success`.

henvic (dismissed):

internal/handler/authenticate.go
```diff
@@ -0,0 +1,147 @@
...
+
+		if errors.Is(err, iterator.ErrIteratorDone) {
+			break
+		}
+
+		if err != nil {
+			return lazyerrors.Error(err)
+		}
+
+		hasUser = true
```
rumyantseva, line 84: So, `hasUser` means that there is at least some user in the database, not necessarily the current user? And if there is at least one user in the database, we consider that new authentication is wanted, so we don't authenticate against the backend. Sounds reasonable!

rumyantseva (approved):
Looks good!

internal/handler/authenticate.go
chilagrow: Correct. It's rather unpleasant work around.

internal/handler/msg_createuser.go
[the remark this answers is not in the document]
AlekSi: That (bad) issue could be prevent with a custom type

```go
type Password struct {
  p string
}

func WrapPassword(p string) Password { return Password{p:p} }

func (p Password) Password() string { return p.p }
```

The functions could accept passwords as `Password`, not as `string`. In that case, the original code would have to say `password.PlainHash(password.WrapPassword(username)))` that is noticeably incorrect.

Tests are good, and Go type system is weak, but it is not _so_ weak so we can't rely on it.

---

Another thing – please don't abbreviate "password" as pwd (print working directory).

integration/users/connection_test.go
```diff
@@ -38,6 +38,8 @@ func TestAuthentication(t *testing.T) {
 	collection := s.Collection
 	db := collection.Database()
 
+	createTestRunnerUser(t, ctx, s.Collection.Database())
```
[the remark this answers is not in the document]
chilagrow: Yes sure, I think so. Let me try :hugs:

internal/handler/authenticate.go
chilagrow: I see your point @AlekSi, I'm not particularly pleased with current logic, feel free to suggest if you have a though on a better choice?

Meanwhile I'll add `h.L.DPanic` :+1:

internal/handler/msg_createuser.go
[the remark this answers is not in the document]
AlekSi: I will try a custom type in a separate PR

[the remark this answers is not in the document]
chilagrow: True, that makes a lot more noticeable. Also at the time of implementation, there wasn't a way to test it with integration.

Anyway, thanks :pray:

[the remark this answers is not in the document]
AlekSi: https://github.com/FerretDB/FerretDB/actions/runs/7986094127/job/21805793593?pr=4107

> Error: internal/handler/msg_createuser.go:220:61: cannot use username (variable of type string) as password.Password value in argument to password.PlainHash

Looks promising

internal/handler/authenticate.go
chilagrow: With `h.L.DPanic`, the first user creation in integration test setup won't work. I suppose I can create them earlier

henvic (approved):

integration/setup/setup.go
```diff
@@ -322,3 +324,27 @@ func insertBenchmarkProvider(tb testtb.TB, ctx context.Context, collection *mong
 
 	return
 }
+
+// setupUser creates a user in admin database with PLAIN mechanism.
```
rumyantseva, line 328: Why do we create it with `PLAIN`? Is there any special logic related to `PLAIN`, or could it potentially be created with another mechanism too?

integration/users/connection_test.go
```diff
@@ -238,3 +238,167 @@ func TestAuthentication(t *testing.T) {
...
+		t.Run(name, func(t *testing.T) {
+			if tc.mechanism == "PLAIN" {
+				setup.SkipForMongoDB(t, "PLAIN mechanism is not supported by MongoDB")
+			}
+
+			t.Parallel()
+
+			credential := options.Credential{
+				AuthMechanism: tc.mechanism,
+				AuthSource:    db.Name(),
```
rumyantseva, line 297: This is also a param I don't understand :) Shouldn't it be `admin` instead of the current DB name?
I see parsing this param is not implemented on the FerretDB side yet, and I don't see an issue for it... Maybe we need to clarify how we expect it to work.

internal/handler/authenticate.go
chilagrow: Let's do h.L.Error instead. Because I want to test this case in integration test, and panic is not ideal for that.

integration/setup/setup.go
rumyantseva: Cool! Thank you :) It will help me in my task about sessions!

rumyantseva (approved):
LGTM!

integration/setup/setup.go
```diff
@@ -163,6 +163,8 @@ func SetupWithOpts(tb testtb.TB, opts *SetupOpts) *SetupResult {
 
 	collection := setupCollection(tb, setupCtx, client, opts)
 
+	setupUser(tb, setupCtx, client)
```
rumyantseva, line 166: btw, does it make sense to call `dropUser` before we set it up? so, if the user already exists (e.g. from the previous test run), `createUser` doesn't return an error.

AlekSi: No. Please explain why :)

rumyantseva: Ah, nevermind, In my case, I do it for MongoDB, not for FerretDB (and it's MongoDB that returns an error, not FerretDB).  And in this setup it's not created for MongoDB anyway.

rumyantseva: There should be a cleanup in `tb.Cleanup` being called - `dropAllUsersFromDatabase` is executed. What I don't understand is why it doesn't work for MongoDB. But probably I created it with some different params.

internal/handler/authenticate.go
AlekSi: Updated #4100

internal/handler/msg_saslstart.go
```diff
@@ -78,6 +82,13 @@ func (h *Handler) MsgSASLStart(ctx context.Context, msg *wire.OpMsg) (*wire.OpMs
 		)))
 
 	case "SCRAM-SHA-1", "SCRAM-SHA-256":
+		if !h.EnableNewAuth {
```
chilagrow, line 85: @henvic see this change, highlighting SCRAM is only enabled with new authentication

AlekSi (approved):

PR #4101, opened 2024-02-20
Fix `envtool run test` `-run` and `-skip` flags

### LINKED ISSUE #3969: Process `--run` and `--skip` in the `envtool` itself
### What should be done?

Currently, when `envtool tests run` is invoked with `--run` and/or `--skip` flags, we just pass those values to `go test`'s `-run` and `-skip`, skip tests listing and sharding. That makes the output less useful: there is no known number of total tests, the progress indicator is worse, etc.

Let's get a list of tests as usual, then apply `--run` and `--skip`'s regular expressions ourselves, then shard the resulting list.

### Where?

https://github.com/FerretDB/FerretDB/blob/1e736b9d4a52fee51e6121c94fd048e3a136a1fb/cmd/envtool/tests.go#L356-L398

### Definition of Done

- `envtool` code updated;
- `envtool`'s unit tests added/updated;
- spot refactorings done.

henvic:
# Description

The logic for filtering the tests should be applied before sharding and use the same logic as the Go standard testing library.

Closes #3969.

## Readiness checklist


cmd/envtool/tests_test.go
```diff
@@ -427,34 +233,34 @@ func TestShardTestFuncs(t *testing.T) {
 	t.Run("InvalidIndex", func(t *testing.T) {
 		t.Parallel()
 
-		_, err := shardTestFuncs(0, 3, testFuncs)
+		_, _, err := shardTestFuncs(0, 3, testFuncs)
```
chilagrow, line 236: Is it possible to test skip shard []string?

henvic: Yes. Let me think how would be a good way to do so 🤔

.golangci.yml
```diff
@@ -3,6 +3,8 @@
 
 run:
   timeout: 3m
+  skip-dirs:
+    - cmd/envtool/internal/testmatch # due to files match.go and match_test.go from the Go standard library
```
AlekSi, line 7: Why do we need that change?

henvic: I've removed it (it was leftover from my first submitted change), but here is the context why it was there in the first place:

This change was needed because files files match.go and match_test.go (5ff7190927a25e30f619279b5e5078b98c7cbb9a) were imported from the standard testing library and the linter is incompatible with it, and maintaining it 'as is' would be much easier than adapting the code to satisfy the linter configuration.

cmd/envtool/tests_test.go
chilagrow: If you want to test string produced for skip, you could do use `.testdata/`, see `TestListTestFuncs`.

rumyantseva (approved):
I tried the fixes locally, I think it works!

noisersup (approved):
linters need to be addressed

chilagrow (approved):
Great work!

cmd/envtool/tests.go
```diff
@@ -363,21 +363,30 @@ func testsRun(ctx context.Context, index, total uint, run, skip string, args []s
...
 	}
 
-	if len(all) == 0 {
-		return fmt.Errorf("no tests to run")
-	}
-
-	shard, err := shardTestFuncs(index, total, all)
+	// Then, shard all the tests but only run the ones that match the regex and that should
+	// be run on the specific shard.
+	shard, skipShard, err := shardTestFuncs(index, total, tests)
```
AlekSi, line 373: Note that `shard`'s content is not actually used

cmd/envtool/tests.go
```diff
@@ -363,21 +363,30 @@ func testsRun(ctx context.Context, index, total uint, run, skip string, args []s
...
+
+	if len(skipShard) > 0 {
+		if skip != "" {
+			skip += "|"
+		}
+		skip += "^(" + strings.Join(skipShard, "|") + ")$"
+	}
+
+	if skip != "" {
+		args = append(args, "-skip="+skip)
```
AlekSi, line 388: I don't understand what we are doing there. #3969's title is "Process `--run` and `--skip` in the `envtool` itself". Somehow, we both process those flags ourselves, then ignore the result for the `--run` processing, but pass `-skip`.

What I think we should do is what the issue says. In more detail:
* get `--run` and `--skip` values;
* get a list of all tests;
* filter that list by applying `--run` and `--skip` values;
* run those tests by invoking `go test -run` – without `-skip` because we already process `--skip`'s regular expression.

PR #4111, opened 2024-02-22
Support speculative authenticate

### LINKED ISSUE #3777: Support `speculativeAuthenticate`

chilagrow:
<!--
Please remove comments like this one, but keep and use the rest of the template.
See CONTRIBUTING.md for details.
-->

# Description

<!--
Write a short description to explain changes that are not mentioned in the initial issue.
What were the reasons for those changes?
Which decisions did you make and why?
What else should reviewers know about your changes?
-->

<!-- Replace <issue_number> with an actual issue number. -->

Closes #3777.

`speculativeAuthenticate` is a field on `isMaster` command. If `speculativeAuthenticate` succeeds, it skips `saslStart`. If `speculativeAuthenticate` fails, it doesn't throw error, but instead let `saslStart` handle the error.

`isMaster` request may contain speculative authenticate field, it does in go driver (maybe other drivers too).
```
    "speculativeAuthenticate": {                                                                                                                                                                  [749/1855]
      "$k": [                                                                                                                                                                                               
        "saslStart",                                                                                                                                                                                        
        "mechanism",                                                                                                                                                                                        
        "payload",                                                                                                                                                                                          
        "autoAuthorize",                                                                                                                                                                                    
        "options",                                                                                                                                                                                          
        "db"                                                                                                                                                                                                
      ],                                                                                                                                                                                                    
      "saslStart": 1,                              
      "mechanism": "SCRAM-SHA-1",                  
      "payload": {                                 
        "$b": "biwsbj11c2VyMixyPVlnU2NNaFZvcng5QzFPRDVWWW1iMjFuUjhWeUJDVVM0",
        "s": 0                                     
      },                                           
      "autoAuthorize": 1,                          
      "options": {                                 
        "$k": [                                    
          "skipEmptyExchange"                      
        ],                                         
        "skipEmptyExchange": true                  
      },                                           
      "db": "test"                                 
    }                          
```

A successful `speculativeAuthenticate` sets a field on `isMaster` response.
```
"speculativeAuthenticate": {                                                                                                                                                                            
      "$k": [                                                                                                                                                                                               
        "conversationId",                                                                                                                                                                                   
        "done",                                                                                                                                                                                             
        "payload"                                                                                                                                                                                           
      ],                                           
      "conversationId": 1,                         
      "done": false,                               
      "payload": {                                 
        "$b": "cj1ZZ1NjTWhWb3J4OUMxT0Q1VlltYjIxblI4VnlCQ1VTNDZRZ0U5S1JJSlRMT3hsR3dNcitLNFNNRlJVR3ZSQ3NxLHM9Q3R3VlYzM3NidExYdEhuZVdlQ1FiS1hRLGk9MTAwMDA=",
        "s": 0                                     
      }                                            
    }           
```


## Readiness checklist

<!-- Please check all items in this checklist that were done. -->


internal/handler/cmd_query.go
```diff
@@ -19,18 +19,44 @@ import (
...
 
 // CmdQuery implements deprecated OP_QUERY message handling.
 func (h *Handler) CmdQuery(ctx context.Context, query *wire.OpQuery) (*wire.OpReply, error) {
-	cmd := query.Query().Command()
+	q := query.Query()
+	cmd := q.Command()
 	collection := query.FullCollectionName
 
+	v, _ := q.Get("speculativeAuthenticate")
+	if v != nil && (cmd == "ismaster" || cmd == "isMaster" || cmd == "hello") {
```
chilagrow, line 38: @b1ron what driver or situation `hello` is used for speculative authenticate? Is using response of `IsMaster` correct for `hello` command as well?

internal/handler/cmd_query.go
```diff
@@ -58,3 +84,65 @@ func (h *Handler) CmdQuery(ctx context.Context, query *wire.OpQuery) (*wire.OpRe
...
+		return nil, lazyerrors.Error(err)
+	}
+
+	mechanism, err := common.GetRequiredParam[string](document, "mechanism")
+	if err != nil {
+		return nil, lazyerrors.Error(err)
+	}
+
+	switch mechanism {
+	case "PLAIN":
```
chilagrow, line 102: More tests will be covered once #1877 is merged.

henvic (approved):
I don't see anything problematic, but should we hold to merge this once #1877 is merged (and then include tests)?

chilagrow:
> I don't see anything problematic, but should we hold to merge this once #1877 is merged (and then include tests)?

I marked it as do not merge for now. I don't think there is need for additional test. But if there is uncovered path, yes we can add tests.

internal/handler/cmd_query.go
b1ron: > If `speculativeAuthenticate` succeeds, it skips `saslStart`.

Where did you get this information from?

Based on my research and testing `speculativeAuthenticate` is used to reduce connection overhead time by embedding the body of the `saslStart` command into the command `hello` under the field `speculativeAuthenticate`. No steps are skipped AFAIK. After further testing it can also be sent in an OP_MSG, see below packet.

> what driver or situation `hello` is used for speculative authenticate

I am unsure of every situation this can occur. A packet from the legacy shell which is using an OP_MSG: 

Note I am connecting to MongoDB here.

```
Frame 16: 532 bytes on wire (4256 bits), 532 bytes captured (4256 bits) on interface lo, id 0
Ethernet II, Src: 00:00:00_00:00:00 (00:00:00:00:00:00), Dst: 00:00:00_00:00:00 (00:00:00:00:00:00)
Internet Protocol Version 4, Src: 127.0.0.1, Dst: 127.0.0.1
Transmission Control Protocol, Src Port: 35596, Dst Port: 27017, Seq: 1, Ack: 1, Len: 466
Mongo Wire Protocol
    Message Length: 466
    Request ID: 0x00000000 (0)
    Response To: 0x00000000 (0)
    OpCode: Extensible Message Format (2013)
    Message Flags: 0x00000001, ChecksumPresent
    Section
        Kind: Body (0)
        BodyDocument
            Document length: 441
            Elements
                Element: hello
                Element: speculativeAuthenticate
                    Type: Document (0x03)
                    Document
                        Document length: 122
                        Elements
                            Element: saslStart
                            Element: mechanism
                            Element: payload
                            Element: db
                Element: saslSupportedMechs
                Element: client
                Element: $db
[Malformed Packet: MONGO]
```

> Is using response of IsMaster correct for hello command as well?

The `hello` and `isMaster` responses are nearly identical, so I think either will work  given that we continue the SCRAM conversation.

internal/handler/cmd_query.go
chilagrow: >Where did you get this information from?

By seeing `task run-proxy` and how this PR behaves on `task run` with successful authentication. I tried it by `task mongosh` then running `db.auth({user:'user',pwd:'pwd',mechanism: 'SCRAM-SHA-256'})`.

When speculative authentication is not implemented, client makes `isMaster`, `saslStart` then `saslContinue`. Once speculative authenticate is implemented and the speculative authenticate succeeds with `isMaster` then `saslContinue` is called.

Okay skip might not be the right term but what I'm emphasizing is that the client does not call `saslStart` command once speculative authenticate has succeeded. So one round trip of saslStart command is skipped. Am I understanding this wrong?

> Note I am connecting to MongoDB here, it uses an OP_QUERY when connecting to FerretDB so I don't think we need to handle the OP_MSG case.

Good, let's just handle it in op_query of isMaster and possibly hello.

> The hello and isMaster responses are nearly identical, so I think either will work given that we continue the SCRAM conversation.

Yes they are similar but the difference is `"ismaster", true` field is set on `isMaster` response. Does this have an impact :thinking:

Also I'm not sure how to test speculative authenticate works with `hello`, but it seems that you have `hello` op query with speculative authenticate field. Could you test my PR it works with `hello` the way you did that above?

Below is the request for speculative authenticate captured by FerretDB. See how it's `isMaster` op query.
```
{                                                                                                                                                                                                           
  "Flags": 0,                                                                                                                                                                                               
  "FullCollectionName": "admin.$cmd",                                                                                                                                                                       
  "NumberToReturn": -1,                                                                                                                                                                                     
  "NumberToSkip": 0,                                                                                                                                                                                        
  "Query": {                                                                                                                                                                                                
    "$k": [                                                                                                                                                                                                 
      "ismaster",                                                                                                                                                                                           
      "helloOk",                                                                                                                                                                                            
      "client",                                                                                                                                                                                             
      "compression",                                                                                                                                                                                        
      "speculativeAuthenticate"                                                                                                                                                                             
    ],                                                                          
    "ismaster": 1,                      
    "helloOk": true,                                                            
    "client": {                                                                 
      "$k": [                                                                   
        "application",                                                          
        "driver",                                                               
        "platform",                                                             
        "os"                                                                    
      ],                                                                        
      "application": {                                                          
        "$k": [                                                                 
          "name"                                                                
        ],                                                                      
        "name": "mongosh 2.1.4"                                                 
      },                                                                        
      "driver": {                                                               
        "$k": [                                                                 
          "name",                                                               
          "version"                                                             
        ],                                                                      
        "name": "nodejs|mongosh",                                               
        "version": "6.3.0|2.1.4"                                                
      },                                                                        
      "platform": "Node.js v20.11.0, LE",                                       
      "os": {                                                                   
        "$k": [                                                                 
          "name",                                                               
          "architecture",                                                       
          "version",                                                            
          "type"                                   
        ],                                         
        "name": "linux",                           
        "architecture": "x64",                     
        "version": "5.15.49-linuxkit-pr",                                                             
        "type": "Linux"                            
      }                                            
    },                                             
    "compression": [                               
      "none"                                       
    ],                                             
    "speculativeAuthenticate": {                   
      "$k": [                                      
        "saslStart",                               
        "mechanism",                               
        "payload",                                 
        "autoAuthorize",                           
        "options",                                 
        "db"                                       
      ],                                           
      "saslStart": 1,                              
      "mechanism": "SCRAM-SHA-256",                
      "payload": {                                 
        "$b": "biwsbj11c2VyMSxyPUNCUnJLelJQRy9qOEY2UUxlcUZwVDFDdGxuaGtzc0xI",                         
        "s": 0                                     
      },                                           
      "autoAuthorize": 1,                          
      "options": {                                 
        "$k": [                                    
          "skipEmptyExchange"                      
        ],                                         
        "skipEmptyExchange": true                  
      },                                           
      "db": "test"                                 
    }                                              
  }                                                
}                                                  
```

b1ron: > Okay skip might not be the right term but what I'm emphasizing is that the client does not call `saslStart` command once speculative authenticate has succeeded.

Ah I see you meant, and your understanding is correct. It just avoids a round-trip by embedding the `saslStart` command in the initial handshake.

The above request is something referred as helloOk Protocol Negotation and is used to check that the server supports the `hello` command because it's a newish command. If the server responds back with `helloOk: true` as well then both parties can start sending `hello` instead of `isMaster` on the connection.

IMO if there's a field with `helloOk` or if the command is a `hello` then we should probably respond with a proper `hello` response, but I also used the `isMaster` [response](https://github.com/FerretDB/FerretDB/pull/3904/files#diff-87d3df5f25dccdd7b09cc9e6bf04964eff5a2882f82f6c3fcdef423ed27d8f41R46) in my research prototype and it worked.

chilagrow: > IMO if there's a field with helloOk or if the command is a hello then we should probably respond with a proper hello response, but I also used the isMaster [response](https://github.com/FerretDB/FerretDB/pull/3904/files#diff-87d3df5f25dccdd7b09cc9e6bf04964eff5a2882f82f6c3fcdef423ed27d8f41R46) in my research prototype and it worked.

Yes I noticed your example has `isMaster` response for both `isMaster` and `hello`. If that's expected, all is good. I just wanted to ensure, because I didn't test it myself :hugs:

chilagrow:
There is now one code for `saslStart`, so it's clear that `saslStart` isn't skiped, it's called either by isMaster op query or saslstart command

b1ron:
Checking out this PR and testing it now.

internal/handler/cmd_query.go
b1ron: @chilagrow I checked out the PR and tested it with multiple drivers and it works as intended.

The only issue is with the legacy shell using an OP_MSG, but I think we can handle that in a separate PR.

internal/handler/cmd_query.go
```diff
@@ -22,15 +22,45 @@ import (
...
 
 // CmdQuery implements deprecated OP_QUERY message handling.
 func (h *Handler) CmdQuery(ctx context.Context, query *wire.OpQuery) (*wire.OpReply, error) {
-	cmd := query.Query().Command()
+	q := query.Query()
+	cmd := q.Command()
 	collection := query.FullCollectionName
 
+	v, _ := q.Get("speculativeAuthenticate")
+	if v != nil && (cmd == "ismaster" || cmd == "isMaster" || cmd == "hello") {
```
b1ron, line 37:
```suggestion
	if v != nil && (cmd == "ismaster" || cmd == "isMaster") {
```

I believe @AlekSi is correct, the `hello` command is never sent in an OP_QUERY so we can probably remove it from the check.

henvic (approved):

noisersup (dismissed):

internal/handler/cmd_query.go
chilagrow: Thanks a lot @b1ron :hugs: . Sure we can handle in separate PR _if_ we want to support that for legacy shell. By the way, I created an issue for hello ok https://github.com/FerretDB/FerretDB/issues/4132

internal/handler/cmd_query.go
chilagrow: Removed `hello`

henvic (approved):

internal/handler/cmd_query.go
b1ron: It still works with the removal of `hello`.

chilagrow: Thank a lot :pray:

Please resolve this comment if it's already addressed :hugs:

internal/handler/cmd_query.go
```diff
@@ -22,15 +22,45 @@ import (
...
 )
 
 // CmdQuery implements deprecated OP_QUERY message handling.
 func (h *Handler) CmdQuery(ctx context.Context, query *wire.OpQuery) (*wire.OpReply, error) {
-	cmd := query.Query().Command()
+	q := query.Query()
+	cmd := q.Command()
 	collection := query.FullCollectionName
 
+	v, _ := q.Get("speculativeAuthenticate")
```
rumyantseva, line 36: I see codecov doesn't complain about this line being uncovered, so it means that `speculativeAuthenticate` is used somewhere. Do we see where it's used? Is it used by the driver under the hood?

rumyantseva (approved):

internal/handler/cmd_query.go
chilagrow: Yes exactly, see https://github.com/FerretDB/FerretDB/pull/4111#discussion_r1505376061

noisersup (approved):

PR #4118, opened 2024-02-23
Improve capped collection cleanup

### LINKED ISSUE #4138: Capped collection cleanup runs when it is not necessary
### What should be done?

1. When a capped collection is configured only with size, without max option, cleanup function deletes documents from the collection even without size threshold getting reached.
2. With default CappedCleanupPercentage set to 10, if we insert less than 10 documents in a capped collection, all of the documents gets deleted during cleanup cycle.

### Where?

https://github.com/FerretDB/FerretDB/blob/main/internal/handler/handler.go

### Definition of Done

- handler updated;
- unit tests added/updated;
- spot refactorings done;
- something else?

wazir-ahmed:
# Description
This PR fixes two minor issues with `cleanupCappedCollection()`
1. When a capped collection is configured only with `size`, without `max` option, cleanup function deletes documents from the collection even without size threshold getting reached.
2. With default `CappedCleanupPercentage` set to 10, if we insert less than 10 documents in a capped collection, all of the documents gets deleted during cleanup cycle.
3.  We are dropping 10% of the documents if document count exceeds the capped configuration. Imagine, the max doc configured is 1M and user inserts the 1M+1th document, we would drop 100k documents, instead of dropping 1 document. The new logic addressed this issue and only drops the surplus documents when document count exceeds.

Closes #4138.

## Readiness checklist


internal/handler/handler.go
```diff
@@ -258,13 +258,19 @@ func (h *Handler) cleanupCappedCollection(ctx context.Context, db backends.Datab
...
+	}
+
+	cleanupCount := int64(float64(statsBefore.CountDocuments) * float64(h.CappedCleanupPercentage) / 100)
+	if cleanupCount == 0 {
 		return 0, 0, nil
 	}
 
 	res, err := coll.Query(ctx, &backends.QueryParams{
 		Sort:          must.NotFail(types.NewDocument("$natural", int64(1))),
-		Limit:         int64(float64(statsBefore.CountDocuments) * float64(h.CappedCleanupPercentage) / 100),
```
wazir-ahmed, line 267: With default `CappedCleanupPercentage` set to 10, whenever `statsBefore.CountDocuments` is less than 10, this equation results in `0`. When limit is set to 0, all the documents in the collection is selected and gets deleted.

internal/handler/handler.go
```diff
@@ -258,13 +258,19 @@ func (h *Handler) cleanupCappedCollection(ctx context.Context, db backends.Datab
 
 	h.L.Debug("cleanupCappedCollection: stats before", zap.Any("stats", statsBefore))
 
-	if statsBefore.SizeCollection < cInfo.CappedSize && statsBefore.CountDocuments < cInfo.CappedDocuments {
```
wazir-ahmed, line 261: When  `cInfo.CappedDocuments == 0`, `statsBefore.CountDocuments < cInfo.CappedDocuments` always results in false, triggering unnecessary cleanup.

chilagrow (commented):
Thanks for these! Is it possible to added tests for those cases you've fixed?

wazir-ahmed:
@chilagrow I'm not sure what would be the right way to verify this.

- The problem is, the default cleanup interval is 1min (configurable through CLI). Because of this, we cannot add an integration test right away, the test case has to wait for 1min before verifying the behavior.
- To override this, we can introduce new fields in `setup.SetupOpts` to configure the cleanup percentage and interval. We can reduce the interval to 1sec (or 5 second) and test the cleanup behavior.

Let me know your thoughts.

chilagrow:
> @chilagrow I'm not sure what would be the right way to verify this.
> 
> * The problem is, the default cleanup interval is 1min (configurable through CLI). Because of this, we cannot add an integration test right away, the test case has to wait for 1min before verifying the behavior.
> * To override this, we can introduce new fields in `setup.SetupOpts` to configure the cleanup percentage and interval. We can reduce the interval to 1sec (or 5 second) and test the cleanup behavior.
> 
> Let me know your thoughts.

I see your point. I think you've identified bugs that we were not able to see which is awesome. I believe we couldn't identify it due to not being able to test the logic. How about extracting the logic of cleanup count to a function something such as `getCleanupCount(cappedCleanupPercentage, cInfo, statsBefore) int64` and assign returned value to the limit if that's not zero? That function can be tested with varying values of cInfo and statsBefore :hugs:

internal/handler/handler.go
rumyantseva: Nice catch!

rumyantseva (commented):
Thanks for contributing, great job! I like Chi's proposal to extract some logic and add a unit test, so we can be sure that no changes will break this code in the future.

wazir-ahmed:
@rumyantseva @chilagrow Hey, I didn't add a new unit test as suggested. Instead, I have used existing integration test `TestCommandsAdministrationCompactCapped` and added additional test case. 

Also, I have refactored `cleanupCappedCollection` to make the feature more close to MongoDB w.r.t max document parameter. Previously, we were dropping 10% of the documents if document count exceeds. Imagine, the max doc configured is `1M` and user inserted `1M+1`th document, we would drop 100k documents, instead of dropping 1 document. With the new logic, when document count exceeds, only the surplus documents will be deleted.

internal/handler/handler.go
```diff
@@ -246,69 +246,61 @@ func (h *Handler) cleanupAllCappedCollections(ctx context.Context) error {
...
-		h.L.Debug("cleanupCappedCollection: no documents to delete")
-		return 0, 0, nil
+		statsBefore = statsAfter
 	}
 
-	deleteRes, err := coll.DeleteAll(ctx, &backends.DeleteAllParams{RecordIDs: recordIDs})
-	if err != nil {
-		return 0, 0, lazyerrors.Error(err)
+	if count := getSizeCleanupCount(cInfo, statsBefore, h.CappedCleanupPercentage); count > 0 {
+		err = deleteFirstNDocuments(ctx, coll, count)
```
chilagrow, line 284: At the moment, `DeleteAll` document could be called twice, if there are too many document which deletes some, then if the collection sizes are too big it calls delete again. Calling delete requires postgresql delete operation so it would be better if we only call it once to minimize overhead :hugs:

I think it should be simple to do, call `getDocCleanupCount()` and `getSizeCleanupCount()`, then use which ever is bigger count.

integration/setup/setup.go
```diff
@@ -94,6 +95,18 @@ type SetupOpts struct {
...
+	// Options to override default backend configuration.
+	BackendOptions *BackendOpts
+}
+
+// BackendOpts represents backend configuration used for test setup.
+type BackendOpts struct {
+	// Capped collections cleanup interval.
+	CappedCleanupInterval time.Duration
+
+	// Percentage of documents to cleanup for capped collections.
```
chilagrow, line 108: Let's document default value 20 if not set.

integration/setup/setup.go
```diff
@@ -142,7 +155,7 @@ func SetupWithOpts(tb testtb.TB, opts *SetupOpts) *SetupResult {
 
 	uri := *targetURLF
 	if uri == "" {
-		uri = setupListener(tb, setupCtx, logger)
+		uri = setupListener(tb, setupCtx, logger, opts.BackendOptions)
```
chilagrow, line 158: I think we should set default value of opts.BackendOptions in this file, because this is where `BackendOptions` is defined :hugs:

chilagrow (commented):
Thanks for detailing the change! Great approach of testing this :hugs: I have a few comments, see what you think

internal/handler/handler.go
wazir-ahmed: This is done so that we will be more precise w.r.t how many documents we are dropping. If both doc count and collection size exceeds the threshold, first we delete docs based on count, check the stats again and if collection size is within the limit, then we don't take further action.

> call getDocCleanupCount() and getSizeCleanupCount(), then use which ever is bigger count.

I think this will lead to the old behaviour, where there are chances of deleting docs unnecessarily (as pointed out in my `1M` example above).

chilagrow: Ah I see your point. So your point is it has to delete by document count to then get the statistics to avoid deleting too many. Could you document that in the code please :pray: 

FYI I think we need to update `bytesFreed` after your second delete.

integration/setup/setup.go
wazir-ahmed: Added `NewBackendOpts()` to handle default values.

internal/handler/handler.go
wazir-ahmed: 1. Documented the code.
2. We don't collect the stats immediately after second deletion. `bytesFreed` is updated after compact and after collecting the stats.

internal/handler/handler.go
```diff
@@ -317,5 +309,64 @@ func (h *Handler) cleanupCappedCollection(ctx context.Context, db backends.Datab
 		bytesFreed = 0
 	}
 
-	return deleteRes.Deleted, bytesFreed, nil
+	return docsDeleted, bytesFreed, nil
+}
+
+func getDocCleanupCount(cInfo *backends.CollectionInfo, cStats *backends.CollectionStatsResult) int64 {
```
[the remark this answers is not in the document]
wazir-ahmed: 1. Added function documentation.
2. Regarding making it inline, I wrote these simple functions to de-clutter the main function. IMO, the code is more readable in the current form.

integration/setup/setup.go
wazir-ahmed: Done.

chilagrow (approved):
Thanks for your contribution! :hugs:

internal/handler/handler.go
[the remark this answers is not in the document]
chilagrow: Thank you. Let's keep it as it is then!

rumyantseva (approved):
LGTM!

PR #4129, opened 2024-02-28
Fix `saslContinue` crashing due to not found authentication conversation

### LINKED ISSUE #4120: `saslContinue` crashes in proxy mode
### FerretDB version

0aad83ae686a9ef2c973c6591f98758c778c7b27

### Backend

Any

### Environment

Any

### What did you do?

1. `task env-up`
2. `task run-proxy-secured`
3. `docker compose exec mongodb mongosh 'mongodb://username:password@host.docker.internal:27018/?tls=true&tlsCertificateKeyFile=/etc/certs/client.pem&tlsCaFile=/etc/certs/rootCA-cert.pem'`

FerretDB not crashing.

### What did you see instead?

```
2024-02-25T09:57:05.181+0400	DPANIC	// 127.0.0.1:49990 -> 127.0.0.1:27018 	clientconn/conn.go:166	runtime error: invalid memory address or nil pointer dereference
(err = <nil>)
github.com/FerretDB/FerretDB/internal/clientconn.(*conn).run.func3
	/Users/aleksi/Code/FerretDB/FerretDB/internal/clientconn/conn.go:166
runtime.gopanic
	/opt/homebrew/Cellar/go/1.22.0/libexec/src/runtime/panic.go:770
runtime.panicmem
	/opt/homebrew/Cellar/go/1.22.0/libexec/src/runtime/panic.go:261
runtime.sigpanic
	/opt/homebrew/Cellar/go/1.22.0/libexec/src/runtime/signal_unix.go:881
github.com/xdg-go/scram.(*ServerConversation).Step
	/Users/aleksi/Soft/GOPATH/pkg/mod/github.com/xdg-go/scram@v1.1.2/server_conv.go:47
github.com/FerretDB/FerretDB/internal/handler.(*Handler).MsgSASLContinue
	/Users/aleksi/Code/FerretDB/FerretDB/internal/handler/msg_saslcontinue.go:47
github.com/FerretDB/FerretDB/internal/clientconn.(*conn).handleOpMsg
	/Users/aleksi/Code/FerretDB/FerretDB/internal/clientconn/conn.go:571
github.com/FerretDB/FerretDB/internal/clientconn.(*conn).route
	/Users/aleksi/Code/FerretDB/FerretDB/internal/clientconn/conn.go:428
github.com/FerretDB/FerretDB/internal/clientconn.(*conn).run
	/Users/aleksi/Code/FerretDB/FerretDB/internal/clientconn/conn.go:312
github.com/FerretDB/FerretDB/internal/clientconn.acceptLoop.func1
	/Users/aleksi/Code/FerretDB/FerretDB/internal/clientconn/listener.go:305
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
(err = <nil>)
```

henvic:
# Description

Closes #4120.

## Readiness checklist


henvic:
Marking this as `do not merge` for now as I'm working on an integration test that I want to push together with the application code changes.

internal/handler/msg_saslcontinue.go
```diff
@@ -44,11 +44,18 @@ func (h *Handler) MsgSASLContinue(ctx context.Context, msg *wire.OpMsg) (*wire.O
 
 	conv := conninfo.Get(ctx).Conv()
 
+	if conv == nil {
```
chilagrow, line 47: Is it possible to add a test for calling saslcontinue directly? I'm not sure if that's possible

chilagrow:
> Marking this as not ready for now as I'm working on an integration test that I want to push together with the application code changes.

@henvic is this still the case?

internal/handler/msg_saslcontinue.go
henvic: It has to happen on an specific order or it'll fail for different reasons. I've tried a few strategies and failed so far to identify a way to do so that doesn't involve mimicking the wire protocol entirely (which looks complicated by what I've seen using a Wireshark sample of MongoDB data traffic) or finding a way to hijack the connection. Any ideas? /cc @b1ron

chilagrow: I think capturing this scenario is hard in test is hard, so I'm happy as it is. But let's see what @b1ron says.

chilagrow (dismissed):

integration/users/connection_test.go
```diff
@@ -412,3 +414,49 @@ func TestAuthenticationEnableNewAuthPLAIN(t *testing.T) {
...
+		{"conversationId", int32(1)},
+		{"payload", []byte("Yz1iaXdzLHI9UnVBWkJSQVpnL1JlZ1FSamF2d2ZqNmhVRDBSUXRJL2VaUTQvemdBUU9NZk83dDdDUGRsR3hnQW9JQkpEODYrMixwPU52VTFQWW1GcldUT1N2ZVBFb05SdDVuMkplNHc2aVBMS2phL2phSC9XZmc9")},
+	}).Err()
+
+	integration.AssertEqualCommandError(t, mongo.CommandError{
+		Name:    "AuthenticationFailed",
+		Code:    18,
+		Message: "Authentication failed.",
+	}, err)
+}
```
henvic, line 462: So, this is what calling the saslContinue method explicitly like that would return, and even though the test is passing on FerretDB with PostgreSQL/SQLite, the failure comes from a different code path, so this is a false positive. We need to call saslContinue on the exact same moment as it happens when using mongosh, and this isn't the case here: that's why I said we can't get away with just calling the method.
/cc @AlekSi

internal/handler/msg_saslcontinue.go
```diff
@@ -44,11 +44,18 @@ func (h *Handler) MsgSASLContinue(ctx context.Context, msg *wire.OpMsg) (*wire.O
...
+		return nil, handlererrors.NewCommandErrorMsg(
+			handlererrors.ErrAuthenticationFailed,
+			"Authentication failed.",
+		)
```
chilagrow, line 51: On CI, `test-integration-mongodb` fails differently, it returns `ProtocolError`.
```
  17.82s          	            	Diff:
  17.82s          	            	--- Expected
  17.82s          	            	+++ Actual
  17.82s          	            	@@ -1,6 +1,6 @@
  17.82s          	            	 (mongo.CommandError) {
  17.82s          	            	- Code: (int32) 18,
  17.82s          	            	- Message: (string) (len=22) "Authentication failed.",
  17.82s          	            	+ Code: (int32) 17,
  17.82s          	            	+ Message: (string) (len=27) "No SASL session state found",
  17.82s          	            	  Labels: ([]string) <nil>,
  17.82s          	            	- Name: (string) (len=20) "AuthenticationFailed",
  17.82s          	            	+ Name: (string) (len=13) "ProtocolError",
  17.82s          	            	  Wrapped: (error) <nil>,
  17.82s          	Test:       	TestAuthenticationSASLContinueFailure
```

integration/users/connection_test.go
chilagrow: So the error comes from different code path because, `saslStart` passed successfully so conv is not nil. When `saslContinue` is called, it fails because payload is wrong.

I tried another way using invalid user credentials so that `saslStart` fails (conv is nil), then call `saslContinue` but the driver does not allow calling `saslContinue` before successful `saslStart`. So it fails with `saslStart` error and doesn't test what we want.

internal/handler/msg_saslcontinue.go
henvic: @chilagrow , yes. I'm going to take a look at why it is so, but this different failure is likely for a different ticket. I'm gonna check this once later today.

chilagrow: Anything left, I will pick it up tomorrow :hugs:

b1ron: Chiming in here a bit because I see some possible things that can be addressed.

That protocol error is expected there, because in this [test](https://github.com/FerretDB/FerretDB/pull/4129/files/b2a023501384274ba3d010a37634a55d901227d9#diff-980d60c3913f30481316d120bf2a14061fb6349f514e7b99404513ca7df10a19R418) we have already established a [authenticated connection](https://github.com/FerretDB/FerretDB/pull/4129/files/b2a023501384274ba3d010a37634a55d901227d9#diff-980d60c3913f30481316d120bf2a14061fb6349f514e7b99404513ca7df10a19R443). We have completed a full SCRAM conversation in the driver. At step 7 it is improper and MongoDB will a return `ProtocolError` because no session state is ever created for this particular conversation.

Below are the sequence of operations from the test's output:

```
1. createUser OK
2. isMaster with helloOk OK so speculativeAuthenticate is done
3. saslStart
4. saslContinue payload Yz1iaXdz...
5. done true we are authenticated
6. ping OK
7. saslContinue payload WXoxaWFY... FAIL
```

henvic:
We'll be skipping automated testing of this fix as it's tricky to do with the tools we currently have for integration tests.

If dance also had tests with mongosh, we could use it, but this is not the case, and adding mongosh as a dependency for the integration tests here deviates a bit from what we want.

chilagrow (approved):
Looks good to me

rumyantseva (approved):
LGTM!

internal/handler/msg_saslcontinue.go
b1ron: Just curious about the solve. So this one line fixed the issue because `conv` is always nil when in proxy mode. But we only return responses from the proxy server thus we never return the error allowing us to complete the conversation.

Is my understanding correct?

cc. @henvic @chilagrow

chilagrow: See the issue description and reproducible step in https://github.com/FerretDB/FerretDB/issues/4120.
This line avoids panic, that simple. It's because conversation was never stared in FerretDB. In that issue, what has happened is below.

Proxy: saslStart req-> saslStart ok -> saslContinue
FerretDB: saslStart req > saslStart not ok -> saslContinue panic

FerretDB failed because `username`/`password` with `SCRAM-SHA-265` does not exist

PR #4133, opened 2024-02-29
Create user in `envtool` for PostgreSQL

### LINKED ISSUE #1877: Enforce new authentication

chilagrow:
# Description

To test it, I added `--test-enable-new-auth` flag in task file `run:` and `run-secured:`.
It does not create user for SQLite, that will be handled in a separate PR.

Closes #1877.

## Readiness checklist


cmd/envtool/envtool.go
```diff
@@ -219,6 +222,118 @@ func setupMongodb(ctx context.Context, logger *zap.SugaredLogger) error {
 	return ctx.Err()
 }
 
+// setupUser creates a user in admin database with supported mechanisms.
+// The user uses username/password credential which is the same as the PostgreSQL
+// credentials.
+//
+// Without this, once the first user is created, the authentication fails
+// as username/password does not exist in admin.system.users collection.
+func setupUser(ctx context.Context, logger *zap.SugaredLogger, postgreSQLPort uint16) error {
```
chilagrow, line 231: This does not replace [setupUser()](https://github.com/FerretDB/FerretDB/blob/main/integration/setup/setup.go#L334) in integration test.
Because this function creates user in ferretdb database which postgreSQL schema, `postgres://username@127.0.0.1:5432/ferretdb?pool_max_conns=50`.

Each integration test create its own schema like so, `postgres://username@127.0.0.1:5432/testauthentication?pool_max_conns=50`.

Taskfile.yml
```diff
@@ -419,7 +419,7 @@ tasks:
         --proxy-addr=127.0.0.1:47017
         --mode=diff-normal
         --handler=pg
-        --postgresql-url='postgres://127.0.0.1:5433/ferretdb?search_path='
+        --postgresql-url='postgres://username:password@127.0.0.1:5433/ferretdb?search_path='
```
chilagrow, line 422: For run-secured to work, what other better thing can I do?

cmd/envtool/envtool.go
henvic: Thanks for the clarification! I was worried this could jeopardize running tests directly with `go test` for a moment 😄

integration/commands_authentication_test.go
```diff
@@ -27,12 +33,55 @@ import (
 func TestCommandsAuthenticationLogout(t *testing.T) {
```
chilagrow, line 33: Logout does not fail on authentication check, if user is not authenticated, it just returns success.

cmd/envtool/envtool.go
chilagrow: It's still there for SQLite. It only has handler level authentication, and once the first user is created without this code, auth would fail.

It will be handled in a separate PR.

integration/setup/setup.go
```diff
@@ -334,7 +334,8 @@ func insertBenchmarkProvider(tb testtb.TB, ctx context.Context, collection *mong
 func setupUser(tb testtb.TB, ctx context.Context, client *mongo.Client) {
 	tb.Helper()
 
-	if IsMongoDB(tb) {
+	if IsPostgreSQL(tb) || IsMongoDB(tb) {
+		// the user is created in env tool for PostgreSQL
```
chilagrow, line 338: For SQLite we need something similar to have a database with default user. Will be handle in a separate PR.

henvic: What about for MySQL?

henvic (approved):

integration/setup/setup.go
chilagrow: Yes that's true. MySQL hasn't been handled. let me see the status of MySQL backend

cmd/envtool/user.json
```diff
@@ -0,0 +1,145 @@
...
+      "_id",
+      "credentials",
+      "user",
+      "db",
+      "roles",
+      "userId"
+    ]
+  },
+  "_id": "admin.username",
+  "credentials": {
```
henvic, line 121: Have we decided to use a default username/password?
Does MongoDB do the same? I'm asking because this makes me uncomfortable as it's one of the fragilities most attacked everywhere regarding security.

chilagrow: This user needs to be the same as the user used in mongodb uri for our test and dev setup to work, for example https://github.com/FerretDB/FerretDB/blob/main/Taskfile.yml#L219]

This env-setup is for our development environment purpose so I'm not concerned with security, we have the same credentials visible in many other places :thinking:

chilagrow: Or perhaps you mean we should not show hashed value?

henvic: Got it! Would it make sense to have this named something else like testuser.json? Because from the path alone and no other context I can't know that it's only used for dev / tests. Physically not, though!

henvic (dismissed):

cmd/envtool/user.json
henvic: No. I was wondering about the default value. It doesn't matter if it's hashed or not for me.

henvic (dismissed):

rumyantseva (approved):
LGTM

Taskfile.yml
AlekSi: That target was supposed to be for the "old" authentication that does not need credentials in the PostgreSQL URI. Why are they needed?

cmd/envtool/envtool.go
```diff
@@ -136,14 +144,46 @@ func setupAnyPostgres(ctx context.Context, logger *zap.SugaredLogger, uri string
...
+	if err != nil {
+		return err
+	}
+
+	err = setupUserInPostgres(ctx, logger, "postgres://username:password@127.0.0.1:5433/ferretdb")
+	if err != nil {
+		return err
+	}
+
+	return setupUserInPostgres(ctx, logger, "postgres://username:password@127.0.0.1:5433/template1")
```
AlekSi, line 186: I don't understand why those calls are not inside `setupAnyPostgres` and what we are doing with credentials there

cmd/envtool/envtool.go
```diff
@@ -219,6 +259,92 @@ func setupMongodb(ctx context.Context, logger *zap.SugaredLogger) error {
...
+	defer dbPool.Close()
+
+	var pgErr *pgconn.PgError
+
+	q := `CREATE SCHEMA admin`
+	if _, err = dbPool.Exec(ctx, q); err != nil && (!errors.As(err, &pgErr) || pgErr.Code != pgerrcode.DuplicateSchema) {
+		return err
+	}
+
+	if err == nil {
```
AlekSi, line 304: I think we could simply that by:

```go
if errors.As(err, &pgErr) && pgErr.Code == pgerrcode.DuplicateSchema {
  err = nil
}
if err != nil {
  return err
}
```

cmd/envtool/envtool.go
```diff
@@ -219,6 +259,92 @@ func setupMongodb(ctx context.Context, logger *zap.SugaredLogger) error {
...
+		}
+	}
+
+	q = `CREATE TABLE admin.system_users_aff2f7ce (_jsonb jsonb)`
+	if _, err = dbPool.Exec(ctx, q); err != nil && (!errors.As(err, &pgErr) || pgErr.Code != pgerrcode.DuplicateTable) {
+		return err
+	}
+
+	if err == nil {
+		q = `CREATE UNIQUE INDEX "system_users_aff2f7ce__id__67399184_idx" ON admin.system_users_aff2f7ce (((_jsonb->'_id')))`
```
AlekSi, line 327: Yikes!..
I wonder if we should step away from that and just implement the localhost exception that we are already working on.

integration/setup/setup.go
AlekSi: MySQL and HANA backends are not yet supported, no need to do anything for them

cmd/envtool/envtool.go
chilagrow: Yes unfortunately this is something needed to be exposed.
It was copy and paste from log output, logic is too difficult to copy accurately :see_no_evil:

You know what, let me leave this PR for now and progress with local host exception.

Taskfile.yml
chilagrow: It's because postgres started on port 5433 is created from `postgres_secured` docker https://github.com/FerretDB/FerretDB/blob/main/docker-compose.yml#L28.
Unlike `postgres` docker which has `POSTGRES_HOST_AUTH_METHOD=trust`, `postgres_secured` doesn't have `trust` so username/password is required to connect to it.

internal/handler/commands.go
```diff
@@ -203,8 +203,9 @@ func (h *Handler) initCommands() {
 			Help:    "Returns a summary of indexes of the specified collection.",
 		},
 		"logout": {
-			Handler: h.MsgLogout,
-			Help:    "Logs out from the current session.",
+			Handler:   h.MsgLogout,
+			anonymous: true,
```
AlekSi, line 207: Let's not forget to port fixes from that PR

PR #4140, opened 2024-03-05
Basic sessions functionality

### LINKED ISSUE #1554: Implement `startSession` command
**Blocked by #4166 (see the comment https://github.com/FerretDB/FerretDB/issues/4166#issuecomment-1990994494).**

## What should be done?

* https://www.mongodb.com/docs/manual/reference/command/startSession/#mongodb-dbcommand-dbcmd.startSession

### Identifiers
When referenced as `lsid` (e.g. in command requests), a session is represented by UUID.

When the session is returned in the list of sessions (and also used in some commands), it's represented as an object like this:
```json5
_id: {
      id: UUID('b5309a1f-c9c3-4608-9d4e-e058281e4737'),
      uid: Binary.createFromBase64('hobAqu900H1RX9+3UKSg6zEaoyg2IU0ukeytvJ2Vgbo=', 0)
    }
``` 

Here `id` is a UUID (same as `lsid` above) and  `uid` base64 string is a sha256 of a user@authenticationDatabase string (in this example, `username@admin`).

### Authentication
If authentication is not enforced (even if a user was used to connect to MongoDB), uid will always be a sha256 of an empty string. Its base64 is `47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=`. 

If authentication is enforced, each user will have their own uid (username + authenticationDatabase).

### The logic & decisions

- Right now, we only store sessions in memory (not in a collection).
- For each session, we need to know user, authentication database, uuid and the date when the session was used last time.
- Each time when a session is used, the used time should be updated. 
   - It can also be updated through `refreshSessions` (to be implemented in #1553).
- To ensure that sessions are not stored forever, we need to cleanup sessions older than the given timeout (default value is 30 min).
- It is possible to mark session as expired before the timeout, see `endSessions` (to be implemented in #1549).
- Sessions can be killed immediately (to be implemented in #1552, #1550, #1551).

rumyantseva:
# Description

Closes #1554.

## Readiness checklist

### DIFF
--- a/integration/sessions/startsession_test.go
+++ b/integration/sessions/startsession_test.go
@@ -1,0 +1,168 @@
+// Copyright 2021 FerretDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sessions
+
+import (
+	"context"
+	"testing"
+
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/mongo-driver/bson"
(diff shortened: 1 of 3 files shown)


integration/sessions/startsession_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	err = db.RunCommand(ctx, bson.D{
+		{"createUser", username2},
+		{"roles", bson.A{}},
+		{"pwd", "password2"},
+		{"mechanisms", bson.A{"SCRAM-SHA-256"}},
+	}).Err()
+	require.NoError(t, err)
+
+	credential := options.Credential{ // as created in setup.setupUser
```
chilagrow, line 71: Just noting setup.setupUser does not create for Mongo

PR #4160, opened 2024-03-08
Use authentication enabled docker for integration test

### LINKED ISSUE #4100: Implement localhost exception
### What should be done?

When new authentication is enabled, `admin.system.users` collection does not exist, and the client is [connected from the localhost](https://pkg.go.dev/net/netip#Addr.IsLoopback), bypass credentials checking instead of what #4075 did (falling back to old backend authentication).

It is similar to https://www.mongodb.com/docs/manual/core/localhost-exception/

For testing, we probably should add a fixed user to MongoDB and FerretDB in the `envtool` and use MongoDB URI with those credentials in tests for the initial connection.

We probably should also clean up the way we create and drop users in tests, including test setup. We also should remove or simplify `if IsMongo` / `if !IsMongo` branches.

### Where?

See issue number in the codebase add by https://github.com/FerretDB/FerretDB/pull/4075.

### Definition of Done

- handler updated;
- all backends updated;
- unit tests added/updated;
- spot refactorings done;
- user documentation updated;
- something else?

chilagrow:
# Description

For `task test-integration-mongodb`, the image `mongodb_secured` is used which has authentication and authorization enable. It already has an user created in [docker](https://github.com/FerretDB/FerretDB/blob/main/docker-compose.yml#L102).
Calling `setup.SetupWithOpts(t, &setup.SetupOpts{SetupUser: true})` creates an user with `root` authorization permission at the beginning of the test, use that authenticated user for the test and drops the user at the end. Since all test share the users in `admin.system.users` database, the unique username is used for each test.

For `task test-integration-postgresql` and `task test-integration-sqlite` the listener is created within setup, and each test creates its own independent postgresql database or sqlite directory, so no other test accesses it.
Calling `setup.SetupWithOpts(t, &setup.SetupOpts{SetupUser: true})` creates an user using local exception at the beginning of the test, and entire postgresql database is dropped at the end of the test.
- postgresql example `postgres://username@127.0.0.1:5432/testcommandsauthenticationlogoutauthenticateduser`
- sqlite example `file:../tmp/sqlite-tests/testcommandsauthenticationlogoutauthenticateduser/`

Closes #4100.

## Readiness checklist


integration/setup/setup.go
```diff
@@ -161,9 +166,12 @@ func SetupWithOpts(tb testtb.TB, opts *SetupOpts) *SetupResult {
 	// register cleanup function after setupListener registers its own to preserve full logs
 	tb.Cleanup(cancel)
 
-	collection := setupCollection(tb, setupCtx, client, opts)
+	if opts.SetupUser {
+		// user is created before the collection so that user can run collection cleanup
```
chilagrow, line 170: @henvic creating user before collection and using authenticated client works :+1:

integration/query_test.go
```diff
@@ -39,104 +39,117 @@ func TestQueryBadFindType(t *testing.T) {
...
-		value any
-		err   *mongo.CommandError
+		value      any
+		err        *mongo.CommandError
+		altMessage string
 	}{
 		"Document": {
 			value: bson.D{},
 			err: &mongo.CommandError{
-				Code:    2,
```
chilagrow, line 48: MongoDB sends different error code for authenticated user. Well, FerretDB will always send the same error regardless or authenticated or not.

rumyantseva (approved):
Cool! This will also help with sessions testing  too :)

noisersup (approved):

AlekSi (approved):

AlekSi:
I will check the security workflow failure

PR #4163, opened 2024-03-11
Fix PLAIN mechanism authentication incorrectly working

### LINKED ISSUE #1877: Enforce new authentication

chilagrow:
# Description

I found a bug related to enforcing new authentication, see added test.

Closes #1877.

## Readiness checklist

### DIFF
--- a/internal/handler/msg_logout.go
+++ b/internal/handler/msg_logout.go
@@ -25,8 +25,7 @@ import (
 
 // MsgLogout implements `logout` command.
 func (h *Handler) MsgLogout(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
-	conninfo.Get(ctx).SetAuth("", "")
-	conninfo.Get(ctx).SetMechanism("")
+	conninfo.Get(ctx).SetAuth("", "", "")
 
 	var reply wire.OpMsg
 	must.NoError(reply.SetSections(wire.MakeOpMsgSection(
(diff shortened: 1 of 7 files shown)


internal/handler/msg_logout.go
```diff
@@ -26,6 +26,7 @@ import (
 // MsgLogout implements `logout` command.
 func (h *Handler) MsgLogout(ctx context.Context, msg *wire.OpMsg) (*wire.OpMsg, error) {
 	conninfo.Get(ctx).SetAuth("", "")
+	conninfo.Get(ctx).SetMechanism("")
```
henvic, line 29: This looks fine to me, although I'd almost always rather have the auth and mechanism reset at once on a single mutex with something like `conninfo.Get(ctx).LogoutSCRAM()` or `ResetSCRAM()`.

henvic (approved):

internal/handler/msg_logout.go
chilagrow: How about if I make mechanism part of Auth so I can do below?
`conninfo.Get(ctx).SetAuth("", "","")`

henvic: Sounds even better to me.

henvic (approved):

AlekSi (approved):

PR #4193, opened 2024-03-15
Make our own low-level driver for testing

### LINKED ISSUE #4146: Make our own low-level driver for testing
### What should be done?

We should make our own low-level driver (essentially, a set of small helpers for writing binary data to TCP connection) for testing of sessions, transactions, etc.

Let's test it with one `insert` command with many documents and `find` + `getMore` for getting them back with a cursor.

### Where?

https://github.com/FerretDB/FerretDB/tree/main/internal/driver

### Definition of Done

- code written;
- unit tests added.

noisersup:
# Description

Closes #4146.

## Readiness checklist


internal/driver/driver_test.go
```diff
@@ -33,6 +36,81 @@ func TestDriver(t *testing.T) {
...
+
+		msgBin, err := body.MarshalBinary()
+		require.NoError(t, err)
+
+		// TODO verify header
+		header := wire.MsgHeader{
+			MessageLength: int32(len(msgBin) + wire.MsgHeaderLen),
+			RequestID:     13,
+			OpCode:        wire.OpCodeMsg,
+		}
```
noisersup, line 70: - Should we verify headers and bodies of every step here, or just check if documents are present in getMore calls?
- If we should, are we going to implement similar checks in future tests later? In that case I'll create helper function for that

internal/driver/driver_test.go
```diff
@@ -33,6 +36,81 @@ func TestDriver(t *testing.T) {
...
+
+		_, resBody, err := c.Request(ctx, &header, body)
```
noisersup, line 112: What's the expected way of getting specific value from `wire.MsgBody` should I convert it to `*types.Document`?

noisersup (commented):
@AlekSi some questions to you, thanks!

internal/driver/driver.go
```diff
@@ -166,6 +166,14 @@ func (c *Conn) WriteRaw(b []byte) error {
...
-	panic("not implemented")
+	if err := c.Write(header, body); err != nil {
+		return nil, nil, lazyerrors.Error(err)
+	}
+
+	header, body, err := c.Read()
+	if err != nil {
+		return nil, nil, lazyerrors.Error(err)
+	}
+
```
AlekSi, line 177: At the very least, we should check that the message we read is actually a response (by checking RequestID / ResponseTo). Maybe do something else too

internal/driver/driver_test.go
AlekSi: I think we should do the full check for Insert, Find, and GetMore. But then I'm not sure what other commands or tests we would need.

Overall, that should not be an integration test of FerretDB or MongoDB

internal/driver/driver.go
```diff
@@ -166,6 +166,14 @@ func (c *Conn) WriteRaw(b []byte) error {
 
 // Request sends the given request to the connection and returns the response.
 func (c *Conn) Request(ctx context.Context, header *wire.MsgHeader, body wire.MsgBody) (*wire.MsgHeader, wire.MsgBody, error) {
```
AlekSi, line 168: Maybe the atomic for the next RequestID should be maintained as a part of the driver, not be the caller's responsibility.

To some degree, the driver should protect the caller from making wire protocol mistakes (but not from sending valid non-sensical commands)

internal/driver/driver_test.go
AlekSi: We should avoid conversions where possible.

In that case, I think the caller should use `resBody.(*wire.OpMsg).RawDocument().Decode()` or something, at least for now. We might add more helpers to OpMsg and bson.Document later

chilagrow (commented):
Looking good overall, see the linter warnings

internal/driver/driver.go
```diff
@@ -37,6 +37,8 @@ type Conn struct {
 	r *bufio.Reader
 	w *bufio.Writer
 	l *slog.Logger
+
+	lastRequestID int32
```
AlekSi, line 41: I think it should be a package-level `atomic.Int32` to easily differentiate between the requests of different clients.

internal/driver/driver.go
```diff
@@ -165,7 +167,71 @@ func (c *Conn) WriteRaw(b []byte) error {
...
+		header.OpCode = wire.OpCodeMsg
+	}
+
+	if header.RequestID == 0 {
+		header.RequestID = c.nextRequestID()
+	}
+	c.lastRequestID = header.RequestID
+
+	if header.ResponseTo != 0 {
+		return nil, nil, lazyerrors.Errorf("response_to is not allowed")
```
AlekSi, line 192:
```suggestion
		return nil, nil, lazyerrors.Errorf("setting response_to is not allowed")
```

Missing verb

internal/driver/driver.go
```diff
@@ -165,7 +167,71 @@ func (c *Conn) WriteRaw(b []byte) error {
...
+	}
+
+	resHeader, resBody, err := c.Read()
+	if err != nil {
+		return nil, nil, lazyerrors.Error(err)
+	}
+
+	if resHeader.ResponseTo != header.RequestID {
+		c.l.Error(
+			fmt.Sprintf("response_to not equal to request_id"),
```
AlekSi, line 212:
```suggestion
			fmt.Sprintf("response_to is not equal to request_id"),
```

internal/driver/driver.go
```diff
@@ -165,7 +167,71 @@ func (c *Conn) WriteRaw(b []byte) error {
...
+			slog.Int("request_id", int(header.RequestID)),
+			slog.Int("response_id", int(resHeader.RequestID)),
+			slog.Int("response_to", int(resHeader.ResponseTo)),
+		)
+		return nil, nil, lazyerrors.Errorf(
+			"response_to not equal to request_id (response_to=%d; expected=%d)",
```
AlekSi, line 218:
```suggestion
			"response_to is not equal to request_id (response_to=%d; expected=%d)",
```

internal/driver/driver_test.go
```diff
@@ -33,6 +38,147 @@ func TestDriver(t *testing.T) {
 	require.NoError(t, err)
 	t.Cleanup(func() { require.NoError(t, c.Close()) })
 
-	// TODO https://github.com/FerretDB/FerretDB/issues/4146
-	_ = c
+	dbName := "TestDriver"
```
AlekSi, line 41: `testutil.DatabaseName`

internal/driver/driver_test.go
```diff
@@ -33,6 +38,147 @@ func TestDriver(t *testing.T) {
...
+		doc := must.NotFail(bson.NewDocument(
+			"dropDatabase", int32(1),
+			"lsid", must.NotFail(bson.NewDocument("id", lsid)),
+			"$db", dbName,
+		))
+
+		body, err := wire.NewOpMsg(must.NotFail(doc.Encode()))
+		require.NoError(t, err)
+
+		_, _, err = c.Request(ctx, new(wire.MsgHeader), body)
```
AlekSi, line 63: Let's check the response

internal/driver/driver.go
```diff
@@ -165,7 +167,71 @@ func (c *Conn) WriteRaw(b []byte) error {
 }
 
 // Request sends the given request to the connection and returns the response.
+// If header MessageLength or RequestID is not specified, it assings the proper values.
+// For header.OpCode the wire.OpCodeMsg is used as default.
 func (c *Conn) Request(ctx context.Context, header *wire.MsgHeader, body wire.MsgBody) (*wire.MsgHeader, wire.MsgBody, error) {
```
AlekSi, line 172: I think we should also document what kind of errors are returned. For example, response with `{"ok": 0}` will not be returned as an error

internal/driver/driver_test.go
```diff
@@ -33,6 +38,147 @@ func TestDriver(t *testing.T) {
...
+		require.NoError(t, err)
+	})
+
+	insertDocs := must.NotFail(bson.NewArray(
```
AlekSi, line 67: I think we can remove that variable and use inline value below

internal/driver/driver_test.go
```diff
@@ -33,6 +38,147 @@ func TestDriver(t *testing.T) {
...
+			"documents", insertDocs,
+			"ordered", true,
+			"lsid", must.NotFail(bson.NewDocument("id", lsid)),
+			"$db", dbName,
+		))
+
+		body, err := wire.NewOpMsg(must.NotFail(insertCmd.Encode()))
+		require.NoError(t, err)
+
+		resHeader, _, err := c.Request(ctx, new(wire.MsgHeader), body)
```
AlekSi, line 85: We should check response

internal/driver/driver_test.go
```diff
@@ -33,6 +38,147 @@ func TestDriver(t *testing.T) {
...
+			resMsg, err := must.NotFail(resBody.(*wire.OpMsg).RawDocument()).Decode()
+			require.NoError(t, err)
+
+			cursor, err := resMsg.Get("cursor").(bson.RawDocument).Decode()
+			require.NoError(t, err)
+
+			nextBatch := cursor.Get("nextBatch").(bson.RawArray)
+			newCursorID := cursor.Get("id").(int64)
+
+			expectedDocs := must.NotFail(bson.NewArray(
```
AlekSi, line 153: Let's define `expectedDocs` as a slice of expected values per iteration above as data, not as code. Otherwise, there is a dependency between `insertDocs` (constant data) and `expectedDocs` (data generated with code)

internal/driver/driver.go
noisersup: I didn't use atomic, as the Conn is not safe for concurrent use, but when I move it to package level, yes sure

internal/driver/driver.go
noisersup: this one is done I believe

internal/driver/driver.go
```diff
@@ -165,7 +167,71 @@ func (c *Conn) WriteRaw(b []byte) error {
...
+	}
+
+	if header.OpCode == 0 {
+		header.OpCode = wire.OpCodeMsg
+	}
+
+	if header.RequestID == 0 {
+		header.RequestID = c.nextRequestID()
+	}
+	c.lastRequestID = header.RequestID
```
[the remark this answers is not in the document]
noisersup: I placed it here, as users may want to specify their own requestID, and then I think it'd be nice to start incrementing from this variable, right?

internal/driver/driver_test.go
noisersup: I didn't check it here, as it's mostly here for cleanup, the results may vary

internal/driver/driver_test.go
noisersup: What do you mean, should I put the `[]byte` as expected?

AlekSi: Basically, `expectedDocs` should not depend on `i`. Instead

```go
expectedDocs := []bson.Array{...}

...

for i := 0; i < 2; i++ {
  testutil.AssertEqual(t, must.NotFail(expectedDocs[i].Convert()), must.NotFail(nextBatch.Convert()))
}
```

internal/driver/driver_test.go
AlekSi: At the very least, we should check `{"ok": 1}`

AlekSi (approved):

PR #4211, opened 2024-03-26
Use non-privileged `scratch` for production Docker images

### LINKED ISSUE #3992: Use non-privileged `scratch` for production Docker images
### What should be done?

Currently, our production Docker image uses `scratch` as a base and runs the FerretDB as a root. It turns out it is possible to run it as non-root user even with the `scratch` image: https://medium.com/@lizrice/non-privileged-containers-based-on-the-scratch-image-a80105d6d341. Let's do that.

### Where?

https://github.com/FerretDB/FerretDB/blob/main/build/ferretdb/production.Dockerfile

### Definition of Done

The production Docker image runs FerretDB as non-root user, but still does not have `/bin/sh`, `/bin/bash`, etc.

rumyantseva:
Fun fact: surprisingly, I found the answer in the only book about Docker I own 😄 

# Description

In this change, in the Dockerfile `state` dir is created explicitly, and in the final stage, its ownership is assigned to the necessary user through `--chown` flag. This explicitly sets ownership, and even if a docker container would be run with an anonymous volume (e.g. through `docker run` without setting `-v` or `--mount`), the `ferretdb` user will have the necessary permission.

In case, volume is specified when starting the container, the specified volume will be used. 
E.g. something like `docker run -v ./state:/state my-ferretdb-image` would work without having those `mkdir` and `COPY`.

Closes #3992.

## Readiness checklist

### DIFF
--- a/build/ferretdb/production.Dockerfile
+++ b/build/ferretdb/production.Dockerfile
@@ -86,9 +86,9 @@ go build -v -o=bin/ferretdb ./cmd/ferretdb
 
 go version -m bin/ferretdb
 bin/ferretdb --version
-EOF
 
-RUN mkdir /state
+mkdir /state
+EOF
 
 # stage for binary only
 
@@ -101,13 +101,12 @@ COPY --from=production-build /src/bin/ferretdb /ferretdb
 
 FROM scratch AS production
 
-COPY --from=production-build /src/bin/ferretdb /ferretdb
-
 COPY build/ferretdb/passwd /etc/passwd
 COPY build/ferretdb/group  /etc/group
 USER ferretdb:ferretdb
 
-COPY --chown=ferretdb:ferretdb --from=production-build /state /state
+COPY --from=production-build /src/bin/ferretdb /ferretdb
+COPY --from=production-build  --chown=ferretdb:ferretdb /state /state
 
 ENTRYPOINT [ "/ferretdb" ]
 
--- a/build/ferretdb/production.Dockerfile
+++ b/build/ferretdb/production.Dockerfile
@@ -106,7 +106,7 @@ COPY build/ferretdb/group  /etc/group
 USER ferretdb:ferretdb
 
 COPY --from=production-build /src/bin/ferretdb /ferretdb
-COPY --from=production-build  --chown=ferretdb:ferretdb /state /state
+COPY --from=production-build --chown=ferretdb:ferretdb /state /state
 
 ENTRYPOINT [ "/ferretdb" ]
 
--- a/build/ferretdb/production.Dockerfile
+++ b/build/ferretdb/production.Dockerfile
@@ -88,11 +88,7 @@ go version -m bin/ferretdb
 bin/ferretdb --version
 EOF
 
-RUN groupadd -g 1000 ferretdb
-RUN useradd -u 1000 -g 1000 ferretdb
-
 RUN mkdir /state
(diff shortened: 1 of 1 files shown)


build/ferretdb/production.Dockerfile
```diff
@@ -88,6 +88,7 @@ go version -m bin/ferretdb
 bin/ferretdb --version
 EOF
 
+RUN mkdir /state
 
 # stage for binary only
```
rumyantseva, line 93: @AlekSi why is this stage needed at all? I see it's not used anywhere later.

AlekSi: See `TARGET: production-binary` in the Taskfile

build/ferretdb/production.Dockerfile
```diff
@@ -102,10 +103,11 @@ FROM scratch AS production
...
 
-# TODO https://github.com/FerretDB/FerretDB/issues/3992
-# COPY build/ferretdb/passwd /etc/passwd
-# COPY build/ferretdb/group  /etc/group
-# USER ferretdb:ferretdb
+COPY build/ferretdb/passwd /etc/passwd
+COPY build/ferretdb/group  /etc/group
+USER ferretdb:ferretdb
+
+COPY --chown=ferretdb:ferretdb --from=production-build /state /state
```
AlekSi, line 110:
```
COPPY --from=production-build /src/bin/ferretdb /ferretdb
COPY --chown=ferretdb:ferretdb --from=production-build /state /state

COPY build/ferretdb/passwd /etc/passwd
COPY build/ferretdb/group  /etc/group
USER ferretdb:ferretdb
```

build/ferretdb/production.Dockerfile
```diff
@@ -88,6 +88,7 @@ go version -m bin/ferretdb
 bin/ferretdb --version
 EOF
 
+RUN mkdir /state
```
AlekSi, line 91: Please move it to the block above to avoid creating another layer

build/ferretdb/production.Dockerfile
AlekSi: (edited)

build/ferretdb/production.Dockerfile
rumyantseva: fixed

build/ferretdb/production.Dockerfile
rumyantseva: fixed

build/ferretdb/production.Dockerfile
```diff
@@ -100,12 +101,12 @@ COPY --from=production-build /src/bin/ferretdb /ferretdb
...
+COPY build/ferretdb/passwd /etc/passwd
+COPY build/ferretdb/group  /etc/group
+USER ferretdb:ferretdb
 
-# TODO https://github.com/FerretDB/FerretDB/issues/3992
-# COPY build/ferretdb/passwd /etc/passwd
-# COPY build/ferretdb/group  /etc/group
-# USER ferretdb:ferretdb
+COPY --from=production-build /src/bin/ferretdb /ferretdb
+COPY --from=production-build  --chown=ferretdb:ferretdb /state /state
```
AlekSi, line 109:
```suggestion
COPY --from=production-build --chown=ferretdb:ferretdb /state /state
```

henvic (approved):

PR #4289, opened 2024-05-21
Fix `env-data`

chilagrow:
# Description

`task env-data` does not populate database like it should, because `env-data` was not run on `ferretdb` postgresql database. It was using temporary test database that exists duration of the test.

## Readiness checklist


integration/shareddata_test.go
```diff
@@ -25,23 +25,41 @@ package integration
...
 		t.Run(name, func(t *testing.T) {
-			setup.SetupWithOpts(t, &setup.SetupOpts{
+			s := setup.SetupWithOpts(t, &setup.SetupOpts{
 				DatabaseName:   "test",
-				CollectionName: p.Name(),
+				CollectionName: name,
 				Providers:      []shareddata.Provider{p},
 			})
+
+			// envData must persist, ensure the documents from the previous provider are found
```
rumyantseva, line 50: I'm not sure I understand the idea. 
In which case this might fail? I mean, in which case the documents won't be found here?

chilagrow: This test would fail if you run this on `main`, because a temporary postgresql database is created for each test and dropped upon t.Cleanup, so previous collection no longer exists.

What is needed for `env-data` is not creating temporary database, but using `ferretdb` postgresql database and populating it with this test. Does this make sense?

chilagrow: Updated the comment :hugs:

rumyantseva (approved):

rumyantseva (approved):
👍

noisersup (approved):

PR #4310, opened 2024-05-29
Allow database and user setup with CLI flags

### LINKED ISSUE #4243: Implement `ferretdb --setup` flags
### What should be done?

```sh
ferretdb --test-enable-new-auth --setup-timeout=1m --setup-username=user --setup-password=pass ...
```

When `--setup-username` flag is passed, FerretDB should create the specified user with the given password (that may be empty). It should not be an error if a user with that name already exists. It should do so without starting the handler but by using backends directly. It should try to establish a working backend connection up `--setup-timeout`.

`test-enable-new-auth`'s value must be `true`.

Let's split the implementation into multiple PRs to make reviews easier and faster:
- [x] Implement dummy flags that only check that `--test-enable-new-auth` is `true` when `--setup-username` is non-empty.
- [x] Extract user creation logic into a separate function in the `internal/backends` package. It should accept a backend, username, and password. Use that function in the `createUser` command handler.
- [x] Use that function in the CLI implementation. Implement retries by using `Backend.Status` method and `ctxutil.SleepWithJitter` function.

Testing should be done manually.

### Where?

- https://github.com/FerretDB/FerretDB/tree/main/cmd/ferretdb
- https://github.com/FerretDB/FerretDB/tree/main/internal/backends

### Definition of Done

- handler updated;
- spot refactorings done.

AlekSi:
# Description

Closes #4243.

## Readiness checklist


internal/handler/handler.go
```diff
@@ -139,6 +151,69 @@ func New(opts *NewOpts) (*Handler, error) {
...
+		ctxutil.SleepWithJitter(ctx, time.Second, retry)
+	}
+
+	res, err := h.b.ListDatabases(ctx, &backends.ListDatabasesParams{Name: h.SetupDatabase})
+	if err != nil {
+		return err
+	}
+
+	if len(res.Databases) > 0 {
+		l.Debug("Database already exists")
```
chilagrow, line 190: If the database exists but the user doesn't exist, we don't create user here?

AlekSi: The idea is to do nothing if some initialization was already performed before

noisersup (approved):

chilagrow (approved):

PR #4364, opened 2024-06-14
Implement Docker healthcheck

### LINKED ISSUE #4246: Implement Docker's HEALTHCHECK
### What should be done?

https://docs.docker.com/reference/dockerfile/#healthcheck

Create `ferretdb ping` subcommand. Use it for `HEALTHCHECK`.

### Where?

TBD

### Definition of Done

TBD

- handler updated;
- all backends updated;
- unit tests added/updated;
- spot refactorings done;
- something else?

noisersup:
# Description

Closes #4246.

## Readiness checklist


AlekSi (commented):
There are merge conflicts now

cmd/ferretdb/main.go
```diff
@@ -217,12 +221,97 @@ func main() {
...
+		l.Info("Ping successful.")
+	}
+
+	if cli.Listen.Unix != "" {
+		ctx, cancel := context.WithTimeout(context.Background(), cli.Setup.Timeout)
+		defer cancel()
+
+		l.Infof("--listen-unix flag is set. Pinging %s...", cli.Listen.Unix)
+
+		u := "mongodb://" + strings.ReplaceAll(cli.Listen.Unix, "/", "%2F")
```
chilagrow, line 292: Let's use PathEscape https://pkg.go.dev/net/url#PathEscape

cmd/ferretdb/main.go
```diff
@@ -217,12 +221,97 @@ func main() {
...
 		run()
 	case "ping":
-		// TODO https://github.com/FerretDB/FerretDB/issues/4246
+		ping()
 	default:
 		panic("unknown sub-command")
 	}
 }
 
+// ping creates connection to FerretDB instance specified by the flags, and runs `ping` command against it.
```
chilagrow, line 230: For new authentication without  `--setup-database`, is it expected to skip ping? I think that should be documented.

cmd/ferretdb/main.go
```diff
@@ -217,12 +221,97 @@ func main() {
...
+				l.Fatal(err)
+			}
+		}()
+
+		if err = client.Ping(ctx, nil); err != nil {
+			l.Fatal(err)
+		}
+
+		l.Info("Ping successful.")
+	}
```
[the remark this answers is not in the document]
noisersup: In which case not successful ping result in no proper response? We handle all of the errors with `l.Fatal` which does exactly that.

If setup database is not set, we inform that ping check is skipped. The only scenario which is not covered yet, is the one when no listen address/unix socket is set. I'll cover that.

cmd/ferretdb/main.go
```diff
@@ -217,12 +221,97 @@ func main() {
...
 }
 
+// ping creates connection to FerretDB instance specified by the flags, and runs `ping` command against it.
+func ping() {
+	logger := setupLogger(cli.Log.Format, "")
+	checkFlags(logger)
+
+	l := logger.Sugar()
+
+	if cli.Setup.Database == "" {
```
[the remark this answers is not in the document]
noisersup: That's right. We don't have currently any way of pinging with the old type of authentication, as it requires credentials to PostgreSQL.

cmd/ferretdb/main.go
```diff
@@ -217,12 +221,97 @@ func main() {
...
+		l.Info("Setup database not specified - skipping ping.")
```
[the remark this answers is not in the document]
noisersup: Yes, that's right

cmd/ferretdb/main.go
```diff
@@ -272,7 +361,20 @@ func setupMetrics(stateProvider *state.Provider) prometheus.Registerer {
 }
 
 // setupLogger setups zap logger.
-func setupLogger(stateProvider *state.Provider, format string) *zap.Logger {
+// It should be preffered for specific single commands/operations over setupLoggerWithStartupMsg.
+func setupLogger(format string, uuid string) *zap.Logger {
```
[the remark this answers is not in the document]
noisersup: The old `setupLogger` - currently `setupLoggerWithStartupMsg` is used only once as well. The same goes for most of the functions in the main file.

I believe it was done like that, because we initialize a lot of stuff in `run` function, which can affect readability.

I'm not quiet sure if we want to output all of the startup messages such as `"Starting FerretDB"` inside of administration/diagnostic subcommand.
In the same way `docker ps` doesn't return all of the build information.

The second thing is that `ferretdb ping` will be called on each container initialization, so all of the build information will be already logged by `ferretdb` listener itself. If we output another message like that again, the logs will be just confusing.

If we don't want to introduce another function, I propose moving the startup message to the `run` function itself.

cmd/ferretdb/main.go
noisersup: done

cmd/ferretdb/main.go
noisersup: done

cmd/ferretdb/main.go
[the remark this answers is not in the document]
AlekSi: And we might have them in the PostgreSQL URL. But that's not worth doing it now.

cmd/ferretdb/main.go
```diff
@@ -217,12 +221,108 @@ func main() {
...
+
+	if cli.Listen.Addr != "" {
+		pingExecuted = true
+
+		host, port, err := net.SplitHostPort(cli.Listen.Addr)
+		if err != nil {
+			l.Fatal(err)
+		}
+
+		l.Infof("--listen-addr flag is set. Pinging %s...", cli.Listen.Addr)
```
AlekSi, line 253: Let's log one info+ message per successful/unsuccessful ping (debug logs are fine). Otherwise, Docker's status and logs will be swarmed by ping logs

cmd/ferretdb/main.go
```diff
@@ -217,12 +221,108 @@ func main() {
...
+				l.Fatal(err)
+			}
+		}()
+
+		if err = client.Ping(ctx, nil); err != nil {
+			l.Fatal(err)
+		}
+
+		l.Info("Ping successful.")
+	}
```
AlekSi, line 289: If we return then and in another branch, we could drop `pingExecuted`

cmd/ferretdb/main.go
```diff
@@ -217,12 +221,108 @@ func main() {
...
+			if err = client.Disconnect(ctx); err != nil {
+				l.Fatal(err)
+			}
+		}()
+
+		if err = client.Ping(ctx, nil); err != nil {
+			l.Fatal(err)
+		}
+
+		l.Info("Ping successful.")
```
AlekSi, line 318: That's the same code as in another branch, we should refactor it: create URI first, then use it

cmd/ferretdb/main.go
```diff
@@ -272,41 +372,16 @@ func setupMetrics(stateProvider *state.Provider) prometheus.Registerer {
...
-	}
-	logUUID := stateProvider.Get().UUID
-
-	// Similarly to Prometheus, unless requested, don't add UUID to all messages, but log it once at startup.
-	if !cli.Log.UUID {
-		startupFields = append(startupFields, zap.String("uuid", logUUID))
-		logUUID = ""
-	}
-
+// It should be preffered for specific single commands/operations over setupLoggerWithStartupMsg.
```
AlekSi, line 375: preferred

cmd/ferretdb/main.go
```diff
@@ -217,12 +221,108 @@ func main() {
...
-		// TODO https://github.com/FerretDB/FerretDB/issues/4246
+		ping()
 	default:
 		panic("unknown sub-command")
 	}
 }
 
+// ping creates connection to FerretDB instance specified by the flags, and runs `ping` command against it.
+// The check is only executed if --setup-database flag is set.
+func ping() {
```
AlekSi, line 232: Let's move it to `ping.go`

cmd/ferretdb/main.go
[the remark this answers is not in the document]
chilagrow: I like the new change :+1:, it's clear when ping is not executed now.

cmd/ferretdb/main.go
chilagrow: `setupLoggerWithStartupMsg` doesn't exist anymore

cmd/ferretdb/main.go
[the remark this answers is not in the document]
chilagrow: Thanks for the explanation. Looks great!

chilagrow (commented):
See comments from @AlekSi.

cmd/ferretdb/ping.go
```diff
@@ -0,0 +1,100 @@
...
+
+	for _, u := range urls {
+		l.Debugf("Pinging %s...", u)
+
+		ctx, cancel := context.WithTimeout(context.Background(), cli.Setup.Timeout)
+		defer cancel()
+
+		client, err := mongo.Connect(ctx, options.Client().ApplyURI(u))
+		if err != nil {
+			l.Fatal(err)
```
rumyantseva, line 85: There is a tricky thing about logs of Fatal level. Fatal causes `os.Exit` which means that `defer`s won't be executed. In this particular case, let's say we have two items in the `urls` slice.  Let's say the first item was connected successfully. Disconnect of this item is deferred. Let's say the second item's connection failed here. This defer from the first item will never be called.

So, we either shouldn't defer disconnect (e.g. we can disconnect right after ping) or we shouldn't use Fatal.

cmd/ferretdb/ping.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		defer func() {
+			if err = client.Disconnect(ctx); err != nil {
+				l.Fatal(err)
```
rumyantseva, line 90: Same here about defers. If more than one things were deferred, the first defer's execution will call os.Exit, and the second one won't be executed.

rumyantseva (commented):
Left some thoughts.

cmd/ferretdb/ping.go
```diff
@@ -0,0 +1,98 @@
...
+		ctx, cancel := context.WithTimeout(context.Background(), cli.Setup.Timeout)
+		defer cancel()
+
+		client, err := mongo.Connect(ctx, options.Client().ApplyURI(u))
+		if err != nil {
+			l.Fatal(err)
+		}
+
+		if err = client.Ping(ctx, nil); err != nil {
+			l.Fatal(err)
```
AlekSi, line 89: Unfortunately, process termination might not close all TCP connections properly in some configurations. We have to call `client.Disconnect` before checking ping result

cmd/ferretdb/ping.go
```diff
@@ -0,0 +1,98 @@
...
+
+		if host == "" {
+			host = "127.0.0.1"
+
+			l.Debugf("Host not specified, defaulting to %s.", host)
+		}
+
+		u := &url.URL{
+			Scheme: "mongodb",
+			Host:   fmt.Sprintf("%s:%s", host, port),
```
AlekSi, line 58: `net.JoinHostPort`

cmd/ferretdb/ping.go
```diff
@@ -0,0 +1,98 @@
...
+	l := logger.Sugar()
+
+	if cli.Setup.Database == "" {
+		l.Info("Setup database not specified - skipping ping.")
+		return
+	}
+
+	var urls []string
+
+	if cli.Listen.Addr != "" {
```
AlekSi, line 42: What about `cli.Listen.TLS`?

cmd/ferretdb/ping.go
```diff
@@ -0,0 +1,98 @@
...
+
+		if err = client.Ping(ctx, nil); err != nil {
+			l.Fatal(err)
+		}
+
+		if err = client.Disconnect(ctx); err != nil {
+			l.Fatal(err)
+		}
+
+		l.Infof("Ping to %s successful.", u)
```
AlekSi, line 96: We should not log secrets (such as passwords) on level higher than debug

chilagrow: I think there is one too many log I think it was commented before, at the moment see for listen addr 
`l.Debugf("--listen-addr flag is set. Ping to %s will be performed.", cli.Listen.Addr)`
`l.Debugf("Pinging %s...", u)`
`l.Infof("Ping to %s successful.", u)`

Can we have one for starting such as `Pinging blah blah...` and one for success such as `Ping success blah blah`?

cmd/ferretdb/ping.go
chilagrow, line 89: Let's add a helpful message, `logger.Fatal("Ping failed", zap.Error(err))`. All other places too.

AlekSi:
@chilagrow https://github.com/FerretDB/FerretDB/pull/4364#discussion_r1647816460
I will send a PR codifying that

cmd/ferretdb/ping.go
AlekSi: @chilagrow https://github.com/FerretDB/FerretDB/pull/4364#discussion_r1647816460
I will send a PR codifying that

cmd/ferretdb/ping.go
noisersup: How do we want to implement it? We don't store CA key anywhere to sign temporary client certificate.

cmd/ferretdb/ping.go
noisersup: @AlekSi Oops, fixed

@chilagrow 2 of the lines you mentioned are only displayed in the debug mode. Normally (on the Info level) we only log the ping success, or when no ping is executed - the skip message.

cmd/ferretdb/ping.go
AlekSi: Why do we need a temporary client certificate?.. Starting FerretDB with TLS requires passing all needed TLS files that we can use.
If your question is about "Extended Key Usage", I think we can safely ignore it, at least for now

cmd/ferretdb/ping.go
chilagrow: Shouldn't we avoid displaying password even in debug?

cmd/ferretdb/ping.go
chilagrow: I don't think I got my idea across, see this PR https://github.com/noisersup/FerretDB/pull/5 of what I mean.

chilagrow (commented):
I think I didn't get across what I meant about comments related to logging. Codified the suggestion https://github.com/noisersup/FerretDB/pull/5

cmd/ferretdb/ping.go
AlekSi: No, everything is a fair game on debug level. First, all information is needed for debugging. Second, we log full requests and responses on debug level that are much more secret than anything else, as they contain actual user data.

chilagrow (approved):

cmd/ferretdb/ping.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	if cli.Listen.Unix != "" {
+		l.Debugf("--listen-unix flag is set. Ping to %s will be performed.", cli.Listen.Unix)
+
+		urls = append(urls, "mongodb://"+url.PathEscape(cli.Listen.Unix))
+	}
+
+	if len(urls) == 0 {
+		l.Info("Neither --listen-addr nor --listen-unix flags were specified - skipping ping.")
```
AlekSi, line 108: What about TLS?

cmd/ferretdb/ping.go
```diff
@@ -0,0 +1,141 @@
...
+		return
+	}
+
+	for _, u := range urls {
+		l.Debugf("Pinging %s...", u)
+
+		ctx, cancel := context.WithTimeout(context.Background(), cli.Setup.Timeout)
```
AlekSi, line 115: I think we should start from `ctxutil.SigTerm`, not `context.Background()`

chilagrow (approved):

AlekSi:
Milestone wasn't set

AlekSi (approved):

PR #4373, opened 2024-06-19
Remove `PLAIN` mechanism from new authentication

### LINKED ISSUE #4365: Remove `PLAIN` support for new authentication
### What should be done?

When FerretDB is started with old authentication, only `PLAIN` should be supported (as it is now).
When FerretDB is started with new authentication, only `SCRAM` authentication methods should be supported.

We should refactor our code to split `PLAIN` and `SCRAM` code paths into separate functions/methods as much as possible. That should simplify old authentication and `PLAIN` removal in 2.0.

We don't test new authentication in integration tests, so this issue should be tested manually.

### Where?

Handler and other places.

### Definition of Done

- handler updated;
- all backends updated;
- unit tests updated;
- integration/compatibility tests updated;
- refactoring done.

chilagrow:
# Description

Closes #4365.

## Readiness checklist


integration/users/connection_test.go
```diff
@@ -331,160 +307,57 @@ func TestAuthenticationOnAuthenticatedConnection(t *testing.T) {
...
 		"Success": {
-			username:  "plain-user",
-			password:  "correct",
+			username:  "username", // when using the PLAIN mechanism we must use user "username"
+			password:  "password",
 			mechanism: "PLAIN",
 		},
 		"BadPassword": {
-			username:  "plain-user",
+			// succeeds due to `POSTGRES_HOST_AUTH_METHOD=trust` for PostgreSQL, and no backend authentication for SQLite
```
chilagrow, line 332: I'm not sure if we want to keep such test here, as this can be misleading

integration/setup/listener.go
```diff
@@ -183,11 +189,19 @@ func setupListener(tb testtb.TB, ctx context.Context, logger *zap.Logger, opts *
...
 			CappedCleanupPercentage: opts.CappedCleanupPercentage,
 			CappedCleanupInterval:   opts.CappedCleanupInterval,
-			EnableNewAuth:           true,
+			EnableNewAuth:           opts.EnableNewAuth,
 			BatchSize:               *batchSizeF,
 			MaxBsonObjectSizeBytes:  opts.MaxBsonObjectSizeBytes,
 		},
 	}
+
+	if opts.EnableNewAuth {
```
chilagrow, line 198: This was needed to test user management integration tests. Essentially, most tests run with new authentication like it is on `main`. Exception is `TestAuthenticationPLAIN` test that runs with old authentication.

integration/setup/listener.go
```diff
@@ -81,6 +83,10 @@ func listenerMongoDBURI(tb testtb.TB, hostPort, unixSocketPath string, tlsAndAut
 		user = url.UserPassword("username", "password")
 	}
 
+	if enableNewAuth {
+		q.Set("authMechanism", "SCRAM-SHA-256")
```
rumyantseva, line 87: So, in MongoDB we always set `SCRAM-SHA-256`, but in FerretDB tests we sometimes use `SHA-1` too, right? Does it cause any problems with tests?

internal/handler/msg_saslstart.go
```diff
@@ -68,18 +68,37 @@ func (h *Handler) saslStart(ctx context.Context, dbName string, document *types.
...
+			)), nil
+		default:
+			msg := fmt.Sprintf("Unsupported authentication mechanism %q.\n", mechanism) +
+				"See https://docs.ferretdb.io/security/authentication/ for more details."
+			return nil, handlererrors.NewCommandErrorMsgWithArgument(handlererrors.ErrAuthenticationFailed, msg, "mechanism")
+		}
+	}
+
 	switch mechanism {
 	case "PLAIN":
```
rumyantseva, line 95: I don't understand this part. In some other parts of the code, `PLAIN` is removed completely, so if `PLAIN` is called, an "Unknown auth mechanism" error will be returned. But here we still keep it. Does it make sense, if `PLAIN` is not supported in the other pars?

AlekSi: In old authentication, we don't have user records, so we don't have anything to update. Still, we allow users to authenticate. https://docs.ferretdb.io/security/authentication/

integration/setup/setup.go
```diff
@@ -123,6 +120,7 @@ func NewBackendOpts() *BackendOpts {
 	return &BackendOpts{
 		CappedCleanupInterval:   time.Duration(0),
 		CappedCleanupPercentage: uint8(20),
+		EnableNewAuth:           true,
```
AlekSi, line 123: I think we should drop the whole function and make zero value (`nil`) useful. If most tests use new auth, we could change this field to `DisableNewAuth` and flip the value

chilagrow: That would simplify things :+1:

integration/setup/listener.go
chilagrow: FerretDB uses `SHA-256` for new authentication, `PLAIN` old authentication. Both FerretDB and MongoDB tests `SHA-1` in specific integration tests by creating user and creating client with `SHA-1`. I don't see a problem :thinking:

internal/handler/handler.go
```diff
@@ -96,6 +96,10 @@ type NewOpts struct {
 func New(opts *NewOpts) (*Handler, error) {
 	b := oplog.NewBackend(opts.Backend, opts.L.Named("oplog"))
 
+	if opts.CappedCleanupPercentage == 0 {
+		opts.CappedCleanupPercentage = 10
```
chilagrow, line 100: same as default set on main.go https://github.com/FerretDB/FerretDB/blob/main/cmd/ferretdb/main.go#L112

integration/setup/listener.go
AlekSi: > So, in MongoDB we always set `SCRAM-SHA-256`, but in FerretDB tests we sometimes use `SHA-1` too, right?

This code configures in-process FerretDB with either `PLAIN` or `SCRAM-SHA-256`

AlekSi (dismissed):

AlekSi (approved):

PR #4380, opened 2024-06-21
Setup OpenTelemetry exporter for FerretDB

### LINKED ISSUE #4345: Setup Otel exporter for FerretDB
### What should be done?

We have some code that setups Otel exporter for integration tests and in envtool. We should refactor it and use it for setting Otel exporter in the FerretDB process itself. It should be configurable with `--test-XXX` flags.

We should not "fix" traces in this issue.

### Where?

TBD

### Definition of Done

TBD

- handler updated;
- all backends updated;
- unit tests added/updated;
- spot refactorings done;
- something else?

rumyantseva:
# Description

Closes #4345.

## Readiness checklist


cmd/ferretdb/main.go
```diff
@@ -125,6 +126,14 @@ var cli struct {
 			Package        string        `default:""                             help:"Telemetry: custom package type."`
 		} `embed:"" prefix:"telemetry-"`
 	} `embed:"" prefix:"test-"`
+
+	OTEL struct {
```
rumyantseva, line 130: Flags are named after `OTEL_*` env variables for better consistency: https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/.

cmd/ferretdb/main.go
```diff
@@ -125,6 +126,14 @@ var cli struct {
...
+		OTLPEndpoint     string        `default:"127.0.0.1:4318" help:"OTLP exporter endpoint."`
```
AlekSi, line 131: A non-empty value should enable the exporter. Remove all other options.

cmd/ferretdb/main.go
AlekSi: > It should be configurable with `--test-XXX` flags.

They should be under the Test group

cmd/ferretdb/main.go
rumyantseva: fixed

cmd/ferretdb/main.go
```diff
@@ -124,6 +125,8 @@ var cli struct {
 			ReportTimeout  time.Duration `default:"5s"                           help:"Telemetry: report timeout."`
 			Package        string        `default:""                             help:"Telemetry: custom package type."`
 		} `embed:"" prefix:"telemetry-"`
+
+		OTLPEndpoint string `default:"" help:"OTLP exporter endpoint, if unset, OpenTelemetry is disabled."`
```
AlekSi, line 129: Please move it before telemetry to other experimental flags (and "Experimental: " should be in the help text)

cmd/ferretdb/main.go
```diff
@@ -381,6 +402,19 @@ func run() {
 
 	logger := setupLogger(stateProvider, cli.Log.Format)
 
+	otelShutdown := setupTracer()
```
AlekSi, line 405: We should pass and use that logger instead of log.Fatal

cmd/ferretdb/main.go
```diff
@@ -381,6 +402,19 @@ func run() {
...
+
+	if otelShutdown != nil {
+		defer func() {
```
AlekSi, line 408: It should be moved below to use context from `ctxutil.SigTerm`

integration/setup/startup.go
```diff
@@ -118,7 +118,10 @@ func Startup() {
 		zap.S().Infof("Compat system: none, compatibility tests will be skipped.")
 	}
 
-	shutdownOtel = observability.SetupOtel("integration")
+	shutdownOtel = must.NotFail(observability.SetupOtel(observability.Config{
+		Service:  "integration-tests",
```
AlekSi, line 122: Why "integration-tests", but "envtool tests"?

cmd/ferretdb/main.go
```diff
@@ -309,6 +312,24 @@ func setupLogger(stateProvider *state.Provider, format string) *zap.Logger {
 	return l
 }
 
+// setupTracer sets up OpenTelemetry exporter.
+func setupTracer() observability.ShutdownFunc {
```
AlekSi, line 316: So is it tracer or exporter?

internal/util/observability/observability.go
```diff
@@ -26,27 +26,44 @@ import (
...
 		otlptracehttp.WithInsecure(),
-	))
+	)
+	if err != nil {
+		return nil, err
+	}
 
+	// Tracer is configured with the particular params on purpose.
+	// We don't want to let them being set through OTEL_* environment variables,
+	// but we set them explicitly.
```
AlekSi, line 56: The same comment applies to options above, so it should be above

cmd/ferretdb/main.go
rumyantseva: Isn't it a different context? When `SigTerm` is executed, we need to give more time to let parts of FerretDB stop gracefully. This is why we use a new context with timeout - and only if shutdown didn't happen within the timeout, shutdown is interrupted. Using `SigTerm`'s context will stop it immediately, without a graceful approach.

AlekSi: Then I guess we need to change "setup"/"shutdown" with "run until ctx is canceled" like other things there

rumyantseva: but who will decide when is ctx canceled? :)

AlekSi: The user will decide – by sending SIGTERM or another signal. Or I don't understand the question.

Note how context from `ctxutil.SigTerm` is passed to `debug.RunHandler`, `runTelemetryReporter`, etc. We should do the same.

cmd/ferretdb/main.go
```diff
@@ -395,6 +420,20 @@ func run() {
 		stop()
 	}()
 
+	otelShutdown := setupTracer(logger)
+
+	if otelShutdown != nil {
+		defer func() {
+			shutdownCtx, shutdownCancel := ctxutil.WithDelay(ctx.Done(), 3*time.Second) //nolint:mnd // simple shutdown timeout
```
rumyantseva, line 427: @AlekSi I added what you had suggested, but to me, it doesn't make sense. This function is already defered, it will be executed at the moment of exiting `main`. There are no `os.Exit` starting from here, it means, exit only happens after `ctx.Done` had been done, there is no way to do it diffrenetly. So, keeping `ctx.Done()` here is unnecessary, this information won't bring any value. A simple `WithTImeout` was enough.

rumyantseva:
Ready for review. Marking as "not ready" for merging because for defering shutdown I want to use the approach I had before. In the current version, I implemented the suggested approach, but I have concerns about it (see my comment).

integration/setup/startup.go
rumyantseva: Fixed (both cases have `-` now)

cmd/ferretdb/main.go
```diff
@@ -395,6 +420,20 @@ func run() {
...
+
+			if err := otelShutdown(shutdownCtx); err != nil {
+				logger.Sugar().Errorf("Failed to shutdown OpenTelemetry: %s.", err)
```
chilagrow, line 430:
```suggestion
				logger.Error("Failed to shutdown OpenTelemetry", zap.Error(err))
```
Easier to read the log using zap `Field`.

chilagrow (dismissed):

cmd/ferretdb/main.go
AlekSi: That's not what I suggested, however.

CleanShot 2024-06-27 at 08 18 37@2x

There is not Run method

rumyantseva: Could you please push what you suggested? I don't understand the idea.

AlekSi: `debug.RunHandler` runs the debug handler until the passed context is canceled.
`runTelemetryReporter` runs the telemetry reporter until the passed context is canceled.
`acceptLoop` runs the loop until the passed context is canceled (with a small delay).
OTel exporter uses start and stop functions instead and has problems with context. We should change it to use the Run function/method instead. The closest analogue is `acceptLoop`.

That's just a restatement of what I wrote [there](https://github.com/FerretDB/FerretDB/pull/4380#discussion_r1652985725).

rumyantseva: Sorry, I think I'm too confused. Do you mean, we should write a custom otel exporter instead of using the oltp one?

AlekSi: Does `debug.RunHandler` run a custom HTTP server instead of using `net/http`?

```go
// RunOtelExpoter runs OpenTelemetry exporter until the context is canceled.
func RunOtelExpoter(ctx context.Context, ...) {
```

rumyantseva: Well, `debug.RunHandler` has something that is run until it's stopped :) 
I don't understand what should be run in this case. Just a gorotine that waits to call `otelShutdown`? Something like this?

```
go func() {
    <-ctx.Done()
   otelShutdown(...)
}
```

AlekSi: We would no longer need `otelShutdown` (and `SetupOtel`) as separate functions.

The context passed to the TracerProvider.Shutdown method should be similar to https://github.com/FerretDB/FerretDB/blob/300fbc35ef965f282e19d0ecdfca8913093eeebb/internal/clientconn/listener.go#L274-L276

rumyantseva: Ah, I think I got it! You want to move this logic to the `observability` package, right?

AlekSi: Ether that, or it could be done the same way as `runTelemetryReporter` in the `main.go` and constructor and Run methods in their own package

internal/util/observability/observability.go
```diff
@@ -22,31 +22,61 @@ import (
...
-	exporter := must.NotFail(otlptracehttp.New(
+// OtelConfig is the configuration for OpenTelemetry.
+type OtelConfig struct {
+	Service  string
+	Version  string
+	Endpoint string
+}
+
+// RunOtel runs the OpenTelemetry system with the given configuration.
+func RunOtel(ctx context.Context, config OtelConfig, logger *zap.SugaredLogger) {
```
AlekSi, line 41: As discussed, we should either return an error from that function or make a separate construction function that could return an error. FerretDB should shutdown if there was an initialization error.

internal/util/observability/observability.go
```diff
@@ -22,31 +22,61 @@ import (
...
+	var exporter *otlptrace.Exporter
+	var err error
+
+	// Exporter and tracer are configured with the particular params on purpose.
+	// We don't want to let them being set through OTEL_* environment variables,
+	// but we set them explicitly.
+	exporter, err = otlptracehttp.New(
 		context.TODO(),
-		otlptracehttp.WithEndpoint("127.0.0.1:4318"),
+		otlptracehttp.WithEndpoint(config.Endpoint),
```
AlekSi, line 50: The empty value should return error (or panic)

integration/setup/startup.go
```diff
@@ -55,6 +51,18 @@ func Startup() {
...
+		Endpoint: "127.0.0.1:4318",
+	}
+
+	ot, err := observability.NewOtel(&otelConf, zap.L().Named("otel"))
+	if err != nil {
+		zap.S().Fatal(err)
+	}
+
+	go ot.Run(context.TODO())
```
rumyantseva, line 64: this one needs a better decision on how to pass `ctx.Done()`

internal/util/observability/observability.go
```diff
@@ -19,34 +19,82 @@ package observability
...
 
-	return tp.Shutdown
+	return &Otel{
+		l:  l,
+		tp: tp,
+	}, nil
+}
+
+// Run runs the OpenTelemetry system.
+func (o *Otel) Run(ctx context.Context) {
```
chilagrow, line 86: It seems `OpenTelemetry` is already started or at least this function doesn't seem to run it. It does has deferred call to shudown `TracerProvider` though. What does this function supposed to do?

AlekSi: Run "thing" until ctx is canceled, then shutdown gracefully. It follows the common approach in #4404. The comment could be improved, though.

internal/util/observability/observability.go
```diff
@@ -19,34 +19,82 @@ package observability
...
+}
+
+// Otel represents the OpenTelemetry system.
+type Otel struct {
+	l  *zap.Logger
+	tp *otelsdktrace.TracerProvider
+}
+
+// NewOtel sets up OTLP exporter and tracer provider.
+func NewOtel(config *OtelConfig, l *zap.Logger) (*Otel, error) {
```
AlekSi, line 48: We generally don't pass options structure and separate arguments at the same time. We could do `NewOtelOpts` as in other places

internal/util/observability/observability.go
```diff
@@ -19,34 +19,82 @@ package observability
...
-	exporter := must.NotFail(otlptracehttp.New(
+// OtelConfig is the configuration for OpenTelemetry.
+type OtelConfig struct {
+	Service  string
+	Version  string
+	Endpoint string
+}
+
+// Otel represents the OpenTelemetry system.
+type Otel struct {
```
AlekSi, line 42: I'm not sure what an "OpenTelemetry system" is. I guess the official name is "[OTLP exporter](https://opentelemetry.io/docs/languages/go/exporters/#otlp)".

We could also do the same protection as in #4404, because we modify the global state with `otel.SetTracerProvider`

integration/setup/startup.go
AlekSi: `shutdownOtel` becomes `context.CancelFunc` and called from `Shutdown`

AlekSi:
```go
ctx, shutdownOtel = context.WithCancel(context.Background())
```

integration/setup/startup.go
```diff
@@ -117,17 +125,10 @@ func Startup() {
 	} else {
 		zap.S().Infof("Compat system: none, compatibility tests will be skipped.")
 	}
-
-	shutdownOtel = observability.SetupOtel("integration")
 }
 
 // Shutdown cleans up after all tests.
 func Shutdown() {
```
AlekSi, line 131:
```go
shutdownOtel()
```

noisersup (approved):

chilagrow (approved):

AlekSi (approved):

PR #4398, opened 2024-06-26
Implement Kubernetes Readiness probe

### LINKED ISSUE #4306: Provide Kubernetes probes
# What should be done?

# DoD

noisersup:
# Description

**PR not cleared and finished, it's a PoC of the probe to discuss the approach.**

Closes #4306.

## Readiness checklist


internal/util/debug/debug.go
```diff
@@ -118,8 +128,22 @@ func Listen(opts *ListenOpts) (*Handler, error) {
...
+
+		ctx := conninfo.Ctx(r.Context(), connInfo)
+
+		_, err := opts.Backend.Status(ctx, nil)
+		if err != nil {
+			opts.L.Error("Backend returned error", zap.Error(err))
+			rw.WriteHeader(http.StatusServiceUnavailable)
+
+			return
+		}
```
noisersup, line 145: The backend is only provided if new authentication is enabled. If the old one is used, the check just returns straightforward `200`.

Providing backend, allows us to easily call `Status` which properly pings the database.

cmd/ferretdb/main.go
```diff
@@ -519,6 +498,33 @@ func run() {
...
+
+			h, err := debug.Listen(opts)
+			if err != nil {
+				l.Sugar().Fatalf("Failed to create debug handler: %s.", err)
+			}
+
+			h.Serve(ctx)
+		}()
+	}
+
```
noisersup, line 527: It's possible to not move the whole debug section from the beginning, by calling `h.Serve()`, `l.Run()`, and `r.Run()`, later in the code. If the approach is alright, I'll proceed to do that, for now less changes is better.

internal/util/debug/debug.go
chilagrow: Yeah for backend authentication, it won't be able to ping.

cmd/ferretdb/main.go
chilagrow: What is the issue here? Moving the code? :thinking:

cmd/ferretdb/main.go
```diff
@@ -519,6 +498,33 @@ func run() {
...
+			l := logger.Named("debug")
+
+			opts := &debug.ListenOpts{
+				TCPAddr: cli.DebugAddr,
+				L:       l,
+				R:       metricsRegisterer,
+			}
+
+			if cli.Setup.Database != "" {
```
chilagrow, line 515: `cli.Setup` may be nil

AlekSi: Nope, that's struct value, not pointer

cmd/ferretdb/main.go
noisersup: I believe we'd like to make sure that the debug handler is the first that is running

internal/util/debug/debug.go
```diff
@@ -125,8 +128,28 @@ func Listen(opts *ListenOpts) (*Handler, error) {
 		rw.WriteHeader(http.StatusOK)
 	})
 
-	readyHandler := http.HandlerFunc(func(rw http.ResponseWriter, _ *http.Request) {
-		// TODO https://github.com/FerretDB/FerretDB/issues/4306
+	// ready handler, which is used for readiness probe,
+	// returns StatusOK when Backend.Status() returned no error.
```
chilagrow, line 132: First two comment seems redundant of the implementation.

chilagrow (commented):
Just small comment

PR #4399, opened 2024-06-26
Implement Kubernetes startup probe

### LINKED ISSUE #4306: Provide Kubernetes probes

noisersup:
# Description

Closes #4306.

## Readiness checklist


internal/util/debug/debug.go
```diff
@@ -74,9 +74,21 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
 
 	must.NoError(r.Register(requestCount))
 
+	// started handler, which is used for startup probe, returns StatusOK when FerretDB listener were initialized.
+	// If it wasn't yet, the StatusInternalServerError is returned.
```
chilagrow, line 78: I think `503 Service Unavailable` is more suitable for this case https://www.rfc-editor.org/rfc/rfc7231#section-6.6.4.

cmd/ferretdb/main.go
```diff
@@ -397,12 +397,14 @@ func run() {
...
 
+	listenerStarted := make(chan struct{})
+
 	if cli.DebugAddr != "" && cli.DebugAddr != "-" {
 		wg.Add(1)
 
 		go func() {
 			defer wg.Done()
-			debug.RunHandler(ctx, cli.DebugAddr, metricsRegisterer, logger.Named("debug"))
+			debug.RunHandler(ctx, cli.DebugAddr, metricsRegisterer, logger.Named("debug"), listenerStarted)
```
chilagrow, line 407: How was this tested?

chilagrow (commented):
Could you add tests please :hugs:

internal/util/debug/debug.go
noisersup: I wasn't sure about that, as it is usually used for the server overload (which is more suitable for the readiness probe)

internal/util/debug/debug.go
```diff
@@ -44,7 +44,7 @@ const (
 )
 
 // RunHandler runs debug handler.
-func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *zap.Logger) {
+func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *zap.Logger, started <-chan struct{}) {
```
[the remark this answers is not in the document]
noisersup: It's mainly because of the usage of the `RunHandler` function inside of `envtool` and `integration` package, where we don't check the kubernetes probes at all.

cmd/ferretdb/main.go
noisersup: Manually, I'll add some unit tests.

internal/util/debug/debug.go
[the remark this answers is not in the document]
chilagrow: Can we pass a closed channel for  `envtool` and `integration` packages? Then it returns `http.StatusOK` without having special handling for nil.

```
started := make(chan struct{})
close(started)
go debug.RunHandler(ctx, "127.0.0.1:8089", prometheus.DefaultRegisterer, logger.Named("debug").Desugar(), started)
```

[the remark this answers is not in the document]
noisersup: sure

internal/util/debug/debug.go
chilagrow: Hmm, I see. You are right about usage for server overload. So the current internal error is probably most appropriate :+1:

internal/clientconn/listener.go
```diff
@@ -82,9 +82,45 @@ func NewListener(opts *NewListenerOpts) *Listener {
 	}
 }
 
+// Wait blocks until all of the expected listeners have started or returned the error.
+func (l *Listener) Wait() {
```
AlekSi, line 86: ooooooof…

noisersup: Do we have any better way of implementing that?

AlekSi: Hopefully. But I don't know what is that yet

internal/util/debug/debug_test.go
```diff
@@ -0,0 +1,87 @@
...
+	ctx, cancel := context.WithCancel(testutil.Ctx(t))
+
+	addr := l.Addr().(*net.TCPAddr)
+	started := make(chan struct{})
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go func() {
+		RunHandler(ctx, addr.String(), prometheus.NewRegistry(), testutil.Logger(t), started)
```
noisersup, line 52: We most likely need to test `RunHandler` function as a whole instead of extracting `startedHandler`, because the handler relies on the `started` channel which is passed from the `RunHandler`.

Technically, we could still extract the handler and test it with `httptest`. In such scenario we need to refer to the channel created in test directly, but that leaves more space for some untracked bugs related to the `RunHandler` implementation

internal/util/debug/debug_test.go
```diff
@@ -0,0 +1,87 @@
...
+		wg.Done()
+	}()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "http://"+addr.String()+"/debug/started", nil)
+	require.NoError(t, err)
+
+	// Wait for handler
```
chilagrow, line 59: Can we use `ctxutil.SleepWithJitter`?

internal/util/debug/debug_test.go
```diff
@@ -0,0 +1,87 @@
...
+	for range 5 {
+		_, err = http.DefaultClient.Do(req)
```
chilagrow, line 61: Why do you call it 5 times? What is the special condition about `break` when error is nil?

internal/clientconn/listener.go
AlekSi: #4404

internal/util/debug/debug.go
```diff
@@ -74,9 +76,15 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
 
 	must.NoError(r.Register(requestCount))
 
+	// started handler, which is used for startup probe, returns StatusOK when FerretDB listener were initialized.
```
AlekSi, line 79: That line describes some "listener" the debug handler knows nothing about. All it knows is that the "started handler" should return different values depending on whether the "started" channel is closed or not.

internal/util/debug/debug.go
```diff
@@ -74,9 +76,15 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
...
+	// If it wasn't yet, the StatusInternalServerError is returned.
```
AlekSi, line 80: That line just duplicates the code

AlekSi (changes requested):
Let's merge #4404 first. That listener change is a no-go: it adds yet another different mechanism for runnables, and we need fewer of them, not more.

chilagrow (approved):
Great work!

internal/util/debug/debug.go
```diff
@@ -63,9 +63,10 @@ type Handler struct {
 //
 //nolint:vet // for readability
 type ListenOpts struct {
-	TCPAddr string
-	L       *zap.Logger
-	R       prometheus.Registerer
+	TCPAddr         string
+	L               *zap.Logger
+	R               prometheus.Registerer
+	FerretdbStarted <-chan struct{}
```
AlekSi, line 69: Since we will need atomic for other probes, let's use atomic there too.

Also, the "FerretDB" prefix is excess. On the one hand, what else could it be? On the other hand – it does not even matter because the debug handler only knows about the probe that happened to be called "started". There will be no different logic for other probes.

internal/util/debug/debug_test.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+func TestDebugHandlerStartupProbe(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := context.WithCancel(testutil.Ctx(t))
+	started := make(chan struct{})
+
+	var wg sync.WaitGroup
+	wg.Add(1)
```
AlekSi, line 38: Move it closer to its usage below

AlekSi (approved):

chilagrow (approved):
Great!

cmd/ferretdb/main.go
```diff
@@ -521,7 +525,9 @@ func run() {
 
 	metricsRegisterer.MustRegister(l)
 
+	listenerStarted.Store(true)
```
rumyantseva, line 528: I see the PR is merged, but I still have a question.

At that moment, the listener hasn't started yet, right?
With the current approach, what is the difference between having this `listenerStarted` var and always returning `ok` in the probe? How to reproduce a moment when the probe doesn't return `ok`?

AlekSi: > At that moment, the listener hasn't started yet, right?

No, that's not right in the current code

PR #4404, opened 2024-07-01
Refactor runnables

### LINKED ISSUE #4403: Refactor "runnables"
### What should be done?

After #4306 and #4380, we should refactor things that run with context to be more consistent. For example:
* all should have constructors that could return initialization errors;
* all should have a `Run(ctx)` method that runs until `ctx` is canceled (with `ctxutil.WithDelay` inside is needed) that doesn't return errors (?).

### Where?

TBD

### Definition of Done

TBD

- handler updated;
- all backends updated;
- unit tests added/updated;
- spot refactorings done;
- something else?

AlekSi:
# Description

Return fatal errors from constructors. Do not return transient errors from main accept/serve loops.

Closes #4403.

## Readiness checklist

### DIFF
--- a/integration/setup/startup.go
+++ b/integration/setup/startup.go
@@ -37,8 +37,11 @@ import (
 // listenerMetrics are shared between tests.
 var listenerMetrics = connmetrics.NewListenerMetrics()
 
-// shutdownOtel is a function that stops OpenTelemetry provider.
+// shutdownOtel usage should be replaced by shutdown usage.
 var shutdownOtel func(context.Context) error
+
+// shutdown cancels context passed to startup components.
+var shutdown context.CancelFunc
 
 // Startup initializes things that should be initialized only once.
 func Startup() {
@@ -52,6 +55,9 @@ func Startup() {
 
 	prometheus.DefaultRegisterer.MustRegister(listenerMetrics)
 
+	var ctx context.Context
+	ctx, shutdown = context.WithCancel(context.Background())
+
 	// use any available port to allow running different configurations in parallel
 	h, err := debug.Listen(&debug.ListenOpts{
 		TCPAddr: "127.0.0.1:0",
@@ -62,10 +68,13 @@ func Startup() {
 		zap.S().Fatalf("Failed to create debug handler: %s.", err)
 	}
 
-	go h.Serve(context.Background())
+	go h.Serve(ctx)
 
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-	defer cancel()
+	// this should use ctx instead
+	shutdownOtel = observability.SetupOtel("integration")
+
+	startupCtx, startupCancel := context.WithTimeout(context.Background(), 5*time.Second) //nolint:mnd // for now
+	defer startupCancel()
 
 	// do basic flags validation earlier, before all tests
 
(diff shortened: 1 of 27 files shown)


chilagrow (approved):

integration/setup/startup.go
```diff
@@ -53,7 +53,16 @@ func Startup() {
...
+	h, err := debug.Listen(&debug.ListenOpts{
+		TCPAddr: "127.0.0.1:0",
+		L:       zap.L().Named("debug"),
+		R:       prometheus.DefaultRegisterer,
+	})
+	if err != nil {
+		zap.S().Fatalf("Failed to create debug handler: %s.", err)
+	}
+
+	go h.Serve(context.Background())
```
rumyantseva, line 65: So, we don't shut down the handler in tests? Is it because there is nothing to shutdown, or is it because there is no mechanism to do it?

internal/util/debug/debug.go
```diff
@@ -141,39 +174,55 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
...
 			panic(err)
 		}
 	}()
 
 	<-ctx.Done()
 
-	stopCtx, stopCancel := context.WithTimeout(context.Background(), time.Second)
-	defer stopCancel()
-	s.Shutdown(stopCtx) //nolint:contextcheck // use new context for cancellation
+	stopCtx, stopCancel := ctxutil.WithDelay(ctx)
```
rumyantseva, line 221: What's the purpose of using `ctxutil.WithDelay` here? Now it hides the delay timeout inside, and it makes it less readable to understand that the timeout is 3 secs. Also, an important logic of `WithDelay` is to wait for a parent context, but in this case `ctx.Done()` was clearly received (one line above). Why don't we use standard library functions in such cases? Is it because of the special cancel func related logic in `WithDelay`? But I don't see how it's used / why it's important to use it.

internal/util/debug/debug.go
```diff
@@ -141,39 +174,55 @@ func RunHandler(ctx context.Context, addr string, r prometheus.Registerer, l *za
...
+	defer stopCancel(nil)
+	_ = s.Shutdown(stopCtx)
```
rumyantseva, line 223: By the way, why don't we log potential errors? There are no potentially valuable errors for us? E.g. if context's deadline is exceeded and the server was force-stopped, it might be an interesting learning.

rumyantseva (commented):
I asked a couple of questions.

From my perspective, the new approach with `WithDelay` makes things less readable and harder to understand.

integration/setup/startup.go
AlekSi: It is mainly to make PR smaller. There is not much to properly shutdown, but since you are going to do it for the Otel exporter (which we must shutdown gracefully), it could be shutdown as well

internal/util/debug/debug.go
AlekSi: To pass values from the context and to make all graceful shutdown code look similar

internal/util/debug/debug.go
AlekSi: I don't think there is a way to make those functions return errors

rumyantseva (approved):
:shipit:

PR #4407, opened 2024-07-02
Test `authSource`

### LINKED ISSUE #4392: Check database name during authentication
### What should be done?

Should we check user's database and command's `$db` or that's a part of authorization? If yes, we should do it. If not, we should remove references.

### Where?

See references to this issue.

### Definition of Done

TBD

- handler updated;
- all backends updated;
- unit tests added/updated;
- spot refactorings done;
- something else?

chilagrow:
# Description

Closes #4392.

## Readiness checklist


integration/users/connection_test.go
```diff
@@ -307,6 +308,181 @@ func TestAuthenticationOnAuthenticatedConnection(t *testing.T) {
...
+		}).Err()
+		require.NoError(t, err)
+
+		t.Cleanup(func() {
+			err = db.Client().Database(dbName).RunCommand(ctx, bson.D{{"dropAllUsersFromDatabase", int32(1)}}).Err()
+			require.NoError(t, err)
+		})
+	}
+
+	for name, tc := range map[string]struct {
```
rumyantseva, line 438: There is one case that I don't see in this or the previous test.
What if `authSource` is set as something else than a `db1`? 
E.g. the user is correct, the password is correct, but `authSource` is set as a non-existent database, or `authSource` database exists, but no users were created with this database.

rumyantseva (commented):
Looks good overall, asked one question.

integration/users/connection_test.go
```diff
@@ -307,6 +308,181 @@ func TestAuthenticationOnAuthenticatedConnection(t *testing.T) {
...
+	t.Parallel()
+
+	s := setup.SetupWithOpts(t, nil)
+	ctx, db := s.Ctx, s.Collection.Database()
+
+	for name, tc := range map[string]struct {
+		username         string
+		password         string
+		authenticationDB string
+		path             string
```
AlekSi, line 321: The path component of MongoDB URI is officially called "defaultauthdb". Having those two separate fields makes it more complicated.

I wonder if having only two fields (a full URI with ":%d" for port and "current database") will make things easier.

AlekSi (commented):
In our dev setup, MongoDB authentication is disabled. Did we run those tests manually against MongoDB with auth enabled?

integration/users/connection_test.go
chilagrow: The test case is easier to read, body is slightly more complicated but I think it's better

integration/users/connection_test.go
```diff
@@ -307,6 +308,175 @@ func TestAuthenticationOnAuthenticatedConnection(t *testing.T) {
...
+			q := u.Query()
+			for k, v := range testURI.Query() {
+				q.Set(k, v[0])
+			}
+
+			u.RawQuery = q.Encode()
+
+			opts := options.Client().ApplyURI(u.String())
+
+			t.Log("connecting", u.String())
```
chilagrow, line 381:
```suggestion
			t.Log("Connecting", u.String())
```

rumyantseva (approved):
Looks good!

AlekSi:
@chilagrow You are right, I confused `mongodb.conf` and `mongodb_secured.conf`

noisersup (approved):

chilagrow:
Yeah, the dev environment setup is a bit too confusing and complicated...

AlekSi (approved):

PR #4418, opened 2024-07-03
Add an ability to exclude particular spans from being recorded in traces

### LINKED ISSUE #4345: Setup Otel exporter for FerretDB

rumyantseva:
# Description

Closes #4345.

The implementation is based on this discussion: https://github.com/open-telemetry/opentelemetry-go/issues/4718.

Also, if it were possible to propagate parent's attributes to all the children's spans automatically, the algorithm of span exclusion would be much easier. But there isn't an option for it, so we need to find all the children while exporting spans.

Please leave your feedback if you see that something can be simplified.

## Readiness checklist

### DIFF
--- a/internal/util/observability/observability.go
+++ b/internal/util/observability/observability.go
@@ -20,8 +20,11 @@ package observability
 import (
 	"context"
 	"errors"
+	"slices"
 	"sync/atomic"
 	"time"
+
+	"go.opentelemetry.io/otel/trace"
 
 	"go.opentelemetry.io/otel/attribute"
 
@@ -120,12 +123,12 @@ type ExporterWithFilter struct {
 	exporter otelsdktrace.SpanExporter
 }
 
-var ExclusionAttribute = attribute.KeyValue{Key: "excluded"}
+var ExclusionAttribute = attribute.KeyValue{Key: "excluded", Value: attribute.BoolValue(true)}
 
 func (e *ExporterWithFilter) ExportSpans(ctx context.Context, spans []otelsdktrace.ReadOnlySpan) error {
-	/*excluded := make(map[trace.SpanID]struct{})
+	excluded := make(map[trace.SpanID]struct{})
 
-		for i, span := range spans {
+	for i, span := range spans {
 		if slices.Contains(span.Attributes(), ExclusionAttribute) {
 			excluded[span.SpanContext().SpanID()] = struct{}{}
 		}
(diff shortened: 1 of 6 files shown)


internal/util/observability/observability.go
```diff
@@ -111,3 +121,61 @@ func (ot *OtelTracer) Run(ctx context.Context) {
...
+			}
+
+			delete(toProcess, parentID)
+		}
+	}
+
+	var filteredSpans []otelsdktrace.ReadOnlySpan
+	for _, sps := range toProcess {
+		filteredSpans = append(filteredSpans, sps...)
+	}
```
chilagrow, line 173:
```suggestion
filteredSpans := maps.Values(toProcess) 
```

internal/util/observability/observability.go
```diff
@@ -111,3 +121,61 @@ func (ot *OtelTracer) Run(ctx context.Context) {
...
+// ExporterWithFilter is a span exporter that filters out spans with ExclusionAttribute and all their subspans.
+type ExporterWithFilter struct {
+	exporter otelsdktrace.SpanExporter
+}
+
+// ExclusionAttribute is the attribute that marks spans to be excluded.
+var ExclusionAttribute = attribute.KeyValue{Key: "excluded", Value: attribute.BoolValue(true)}
+
+// ExportSpans implements otelsdktrace.SpanExporter interface.
+func (e *ExporterWithFilter) ExportSpans(ctx context.Context, spans []otelsdktrace.ReadOnlySpan) error {
```
chilagrow, line 134: Let me see if I understand what we are trying to do. Do we want to stop trace for envtool? I'm trying to understand the implementation of `ExclusionAttribute` to our usage, is it only for envtool? Currently, this is pretty tricky to understand.

Could a simpler solution be used? In https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#example-SpanProcessor-Filtered
there is `InstrumentationBlacklist` example which drops span with specified InstrumentationScope name.

chilagrow (commented):
I have a question about how this implementation is expected to be used. Why is it important to hide particular span? :thinking:

internal/util/observability/observability.go
```diff
@@ -111,3 +121,61 @@ func (ot *OtelTracer) Run(ctx context.Context) {
 
 	ot.l.Info("OTLP tracer stopped.")
 }
+
+// ExporterWithFilter is a span exporter that filters out spans with ExclusionAttribute and all their subspans.
+type ExporterWithFilter struct {
```
AlekSi, line 126: Please unexport identifiers that should not be exported

rumyantseva:
I close this PR as #4423 provides a simpler solution.

PR #4424, opened 2024-07-04
Document Kubernetes probes

AlekSi:
# Description

And simplify implementation.

## Readiness checklist

### DIFF
--- a/internal/clientconn/listener.go
+++ b/internal/clientconn/listener.go
@@ -51,6 +51,7 @@ type Listener struct {
 	tcpListenerReady  chan struct{}
 	unixListenerReady chan struct{}
 	tlsListenerReady  chan struct{}
+	listenersClosed   chan struct{}
 }
 
 // NewListenerOpts represents listener configuration.
@@ -84,13 +85,19 @@ func Listen(opts *NewListenerOpts) (*Listener, error) {
 		tcpListenerReady:  make(chan struct{}),
 		unixListenerReady: make(chan struct{}),
 		tlsListenerReady:  make(chan struct{}),
+		listenersClosed:   make(chan struct{}),
 	}
 
 	var err error
 
+	defer func() {
+		if err != nil {
+			l.Handler.Close()
+		}
+	}()
+
 	if l.TCP != "" {
 		if l.tcpListener, err = net.Listen("tcp", l.TCP); err != nil {
-			opts.Handler.Close()
 			return nil, lazyerrors.Error(err)
 		}
 
@@ -100,7 +107,6 @@ func Listen(opts *NewListenerOpts) (*Listener, error) {
 
 	if l.Unix != "" {
 		if l.unixListener, err = net.Listen("unix", l.Unix); err != nil {
-			opts.Handler.Close()
 			return nil, lazyerrors.Error(err)
 		}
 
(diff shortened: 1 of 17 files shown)


AlekSi:
https://github.com/FerretDB/FerretDB/labels/do%20not%20merge until everyone reviews

noisersup (dismissed):

rumyantseva (approved):

website/docs/configuration/observability.md
```diff
@@ -36,10 +36,38 @@ $ docker logs my-ferretdb
...
+:::note
+The set of metrics is not stable yet; metric and label names and value formatting might change in minor releases.
+:::
+
+### Probes
+
+FerretDB exposes the following probes that can be used for
+[Kubernetes health checks](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)
+and similar use cases.
+They return HTTP code 200 if a probe is successful and 500 otherwise.
```
noisersup, line 60:
```suggestion
They return HTTP code 200 if a probe is successful and 5xx otherwise.
```

For the readiness probe it makes more sense to return 503 (at least in some scenarios). The readiness probe is useful, when the service is temporarily unavailable, so it's not killed by kubelet, but it's not flooded with new queries.

internal/util/debug/debug.go
```diff
@@ -49,6 +49,11 @@ const (
 // setup ensures that debug handler is set up only once.
 var setup atomic.Bool
 
+// Probe should return true on success and false on failure.
+//
+// It must be thread-safe.
+type Probe func() bool
```
noisersup, line 55: Actually, when I think about it now. I think the function should include context as an argument.

The readiness probe has some operations behind it, so it'd be useful to have Cancellation option based on the request context.

website/docs/configuration/observability.md
AlekSi: I updated the docs, but I think we should still return 500, not 503, in practice. Why 503 and not 502, for example? Also, HTTP semantics go beyond status code names. For example, https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503 says, "a user-friendly page explaining the problem should be sent."

rumyantseva (approved):

noisersup (approved):

internal/clientconn/listener.go
```diff
@@ -128,33 +132,25 @@ func Listen(opts *NewListenerOpts) (*Listener, error) {
 	return l, nil
 }
 
+// Listening returns true if the listener is currently listening and accepting new connection.
+//
+// It returns false when listener is stopped
+// or when it is still running with established connections.
```
chilagrow, line 138: `// or when it is still running with established connections.` It means the listener hasn't started yet right?

chilagrow (approved):

internal/clientconn/listener.go
AlekSi: The opposite: the listener is closed, accept loop exited, but already accepted connection are still active

website/docs/configuration/observability.md
```diff
@@ -36,10 +36,39 @@ $ docker logs my-ferretdb
...
+:::note
+The set of metrics is not stable yet; metric and label names and value formatting might change in minor releases.
+:::
+
+### Probes
+
+FerretDB exposes the following probes that can be used for
+[Kubernetes health checks](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)
+and similar use cases.
+They return HTTP code 2xx if a probe is successful and 5xx otherwise.
```
Fashander, line 60: Will there be fixed HTTP codes for this? e.g. `200 or 500`

AlekSi: https://github.com/FerretDB/FerretDB/pull/4424#discussion_r1666007724

Fashander (approved):

PR #4426, opened 2024-07-05
Implement Kubernetes Readiness probe

### LINKED ISSUE #4306: Provide Kubernetes probes

noisersup:
# Description

Closes #4306.

## Readiness checklist


cmd/ferretdb/main.go
```diff
@@ -220,7 +220,18 @@ func main() {
...
+		logger := setupLogger(cli.Log.Format, "")
+		checkFlags(logger)
+
+		ready := ReadyZ{
+			l: logger,
+		}
+
+		ctx, _ := ctxutil.SigTerm(context.Background())
+		if !ready.Probe(ctx) {
+			os.Exit(1)
```
chilagrow, line 232: Elsewhere, we exit with non zero error code using `logger.Fatal`, why is this special?

cmd/ferretdb/ping.go
```diff
@@ -66,38 +72,8 @@ func ping() {
...
-			Scheme:   "mongodb",
-			Host:     net.JoinHostPort(host, port),
-			Path:     cli.Setup.Database,
-			User:     url.UserPassword(cli.Setup.Username, cli.Setup.Password),
-			RawQuery: values.Encode(),
-		}
-
-		urls = append(urls, u.String())
+		// TODO https://github.com/FerretDB/FerretDB/issues/4427
+		l.Warn("TLS ping is skipped, please check https://github.com/FerretDB/FerretDB/issues/4427 for more details.")
```
chilagrow, line 76: I don't understand why it doesn't work after reading the issue :thinking:

cmd/ferretdb/main.go
noisersup: In which case we use `logger.Fatal` instead of `os.Exit` when we don't return the error?

The error is already logged inside of the `Probe` function, hence nothing can be returned here. If `Probe` returns false, it means that the readiness check didn't perform correctly, so the caller can call os.Exit(1).

cmd/ferretdb/ping.go
noisersup: I left it here, with motivation to speed up the process of finishing the current issue. I wanted to work on this task right afterwards, and just quickly skip the code that makes TLS healthchecks unusable. I'll improve the issue description

cmd/ferretdb/ping.go
```diff
@@ -22,29 +22,35 @@ import (
...
-
-	"github.com/FerretDB/FerretDB/internal/util/ctxutil"
 )
 
-// ping creates connection to FerretDB instance specified by the flags, and runs `ping` command against it.
-// The check is only executed if --setup-database flag is set.
-func ping() {
-	logger := setupLogger(cli.Log.Format, "")
-	checkFlags(logger)
+// ReadyZ represents the Readiness probe, which is used to run `ping`
```
rumyantseva, line 27: I'm still not sure about the whole idea of ping, so please feel free to ignore my comment if it doesn't make sense.

Between having this called "ping" and "readyz", I'd prefer `ping` because it's more universal. Readiness probe is quite a Kubernetes-specific thing, and `ping` looks more general to me. So, I'd continue calling it `ping` :) 

Is there a separate PR that adds the necessary readiness probe logic to the debug handler?

noisersup: Sure, I'll rename it in a moment.

> Is there a separate PR that adds the necessary readiness probe logic to the debug handler?

It's already merged, and being used inside of `debug/debug.go`. The only thing that needs to be done is to provide the actual Probe function to the `debug.Handler` struct which is done in this PR

rumyantseva (approved):

cmd/ferretdb/ping.go
AlekSi: > Sure, I'll rename it in a moment.

#4306 is very clear that documentation from #4424 should be used, and that PR (approved by everyone and merged) is very clear what paths should be used. You updated the code, but did not update the docs. Prevert the path change.

> Between having this called "ping" and "readyz", I'd prefer ping because it's more universal

I'm sorry, but that's just too late.

cmd/ferretdb/main.go
```diff
@@ -220,7 +220,18 @@ func main() {
...
 	case "ping":
-		ping()
+		logger := setupLogger(cli.Log.Format, "")
+		checkFlags(logger)
+
+		ping := Ping{
+			l: logger,
+		}
+
+		ctx, _ := ctxutil.SigTerm(context.Background())
```
AlekSi, line 230: The stop function should be called, not ignored. It is useful for terminating process hung during shutdown by hitting Ctrl+C / sending SIGTERM twice

cmd/ferretdb/ping.go
```diff
@@ -22,29 +22,35 @@ import (
...
-
-	"github.com/FerretDB/FerretDB/internal/util/ctxutil"
 )
 
-// ping creates connection to FerretDB instance specified by the flags, and runs `ping` command against it.
-// The check is only executed if --setup-database flag is set.
-func ping() {
-	logger := setupLogger(cli.Log.Format, "")
-	checkFlags(logger)
+// Ping represents mongo driver client, which can be used to run `ping`
```
AlekSi, line 27:
```go
// Readyz represents the readiness probe […]
```

cmd/ferretdb/ping.go
AlekSi:
```
l.Warn("TLS ping is not implemented yet.")
```

TODO with an issue number is enough.

cmd/ferretdb/main.go
chilagrow: Let me see if I understand your question :thinking:. We don't call os.Exit in our code base directly. Something such as `Fatal("Ping failed")` gives user friendly message on what has caused exit 1.

cmd/ferretdb/ping.go
```diff
@@ -22,29 +22,35 @@ import (
...
+// command against the FerretDB instance specified by cli flags.
+type Ping struct {
+	l *zap.Logger
+}
+
+// Probe executes ping queries to open listeners, and returns true if they succeed.
+// Any errors that occure are passed through Ping.l listener.
+//
+// It is only executed if --setup-database flag is set.
```
chilagrow, line 36: Does the signature of this function `Probe(ctx) bool` have any significance related to kubernetes? Could you document that please? If it doesn't have any significance, I find it a bit strange to return boolean instead of an error indicating what failed.

cmd/ferretdb/ping.go
chilagrow: Yep good idea to scope out tricky things :+1:

cmd/ferretdb/ping.go
AlekSi: That was done in #4424. See how the signature is documented and how probes are used

chilagrow: Oops thank you :see_no_evil:

cmd/ferretdb/ping.go
```diff
@@ -22,29 +22,35 @@ import (
...
+}
+
+// Probe executes ping queries to open listeners, and returns true if they succeed.
+// Any errors that occure are passed through ReadyZ.l listener.
+//
+// It is only executed if --setup-database flag is set.
+func (ready *ReadyZ) Probe(ctx context.Context) bool {
+	logger := ready.l
 
 	l := logger.Sugar()
```
AlekSi, line 40: I'm not sure why we need both `logger` and `l`

AlekSi (approved):

chilagrow (approved):

rumyantseva (approved):

PR #4430, opened 2024-07-08
Extend `observability.FuncCall` for future needs

### LINKED ISSUE #3244: Extend `observability` package
### What should be done?

OTel traces, Prometheus metrics, Go execution tracing, /x/net/trace, pprof labels? Trace function entry/leave with a common root?

Since we don't know how exactly the solution looks like, make small PRs for each step.
The first PR probably should change `observability.FuncCall` signature to return (ctx, cancel).

### Where?

https://github.com/FerretDB/FerretDB/tree/main/internal/util/observability

### Definition of Done

- all handlers updated;
- unit tests added/updated;
- spot refactorings done;
- something else?

rumyantseva:
# Description

Closes #3244.

## Readiness checklist


internal/util/observability/funccall.go
```diff
@@ -53,14 +55,19 @@ func FuncCall(ctx context.Context) func() {
...
+	ctx, cancel = context.WithCancel(parent)
+	go fc.run(ctx)
+
+	return
 }
 
-// leave is called on function exit.
-func (fc *funcCall) leave() {
+// run waits until ctx is canceled.
+func (fc *funcCall) run(ctx context.Context) {
```
rumyantseva, line 68: I don't really like the "run" name because, in fact, this function doesn't run anything, it rather waits. I called it "run" to keep it consistent with other similar functions, but better naming proposals are welcome.

internal/util/observability/funccall.go
```diff
@@ -31,17 +31,19 @@ type funcCall struct {
...
-//	    // ...
+//		ctx, cancel := FuncCall(ctx)
+//		defer cancel()
+//		// ...
+//	}
 //
 // For the Go execution tracer, FuncCall creates a new region for the function call
 // and attaches it to the task in the context (or background task).
-func FuncCall(ctx context.Context) func() {
+func FuncCall(ctx context.Context) (context.Context, context.CancelFunc) {
```
rumyantseva, line 46: It would be great to add a test for this function - to call it from the test and then check that everything was stopped correctly, but I don't see a simple way to do it.

rumyantseva:
@AlekSi PTAL if I understood the idea correctly.

internal/util/observability/funccall.go
AlekSi, line 46: The documentation should be updated to mention:
* what context is returned;
* how it should be used;
* is it being canceled when returned function is called.

internal/util/observability/funccall.go
```diff
@@ -56,11 +58,16 @@ func FuncCall(ctx context.Context) func() {
 		fc.region = trace.StartRegion(ctx, funcName)
 	}
 
-	return fc.leave
+	stopCtx, stopCancel := context.WithCancel(context.Background())
+	go fc.run(stopCtx)
```
AlekSi, line 62: FuncCall can't create new goroutines because `Region.End` method must be called from the same goroutine where the region was started

AlekSi (changes requested):

AlekSi:
> PTAL if I understood the idea correctly

I don't know if the idea is understood correctly (documentation wasn't updated), but the implementation is just incorrect

internal/util/observability/funccall.go
rumyantseva: @AlekSi could you please describe these things? E.g. in the task description or right here. Based on your comments, I see I don't understand how this function is expected to work and what idea is described in the task. Also I don't see what should be used instead of `go fc.run(stopCtx)`.

AlekSi: > I see I don't understand how this function is expected to work

But you wrote that function?.. I don't follow. All I'm saying is that the function signature changed without a proper documentation update. Things that I wrote should described in the doc comment.

> Also I don't see what should be used instead of go fc.run(stopCtx).

I don't know. All I'm saying is that https://pkg.go.dev/runtime/trace#StartRegion description explicitly disallows the current implementation.

internal/util/observability/funccall.go
```diff
@@ -56,11 +58,16 @@ func FuncCall(ctx context.Context) func() {
 		fc.region = trace.StartRegion(ctx, funcName)
 	}
 
-	return fc.leave
+	stopCtx, stopCancel := context.WithCancel(context.Background())
```
chilagrow, line 61: We create new context instead of using ctx passed in this function on purpose right? Why is that?

rumyantseva: Yes, it was written on purpose. This context is not related to the initial context. Before this PR,`FuncCall` returned a function that should be called at the end of the caller function (so, it was deferred). One of the ideas here was to use context with cancelation instead of that deferred function, so this context is only related to cancelation. 

If we pass the original context `ctx` instead of `context.Background()`, and `ctx` is canceled, this `stopCtx` will be canceled before exiting from the caller function, my idea was to avoid it and only have `stopCtx` canceled when `stopCancel` is called.

AlekSi: That causes values inside original `ctx` to be lost

internal/backends/collection.go
```diff
@@ -103,7 +103,9 @@ type QueryResult struct {
 //
 // Limit, if non-zero, should be applied.
 func (cc *collectionContract) Query(ctx context.Context, params *QueryParams) (*QueryResult, error) {
-	defer observability.FuncCall(ctx)()
+	var cancel context.CancelFunc
+	ctx, cancel = observability.FuncCall(ctx)
```
AlekSi, line 107:
```suggestion
	ctx, cancel := observability.FuncCall(ctx)
```

Everywhere

internal/util/observability/funccall.go
```diff
@@ -18,52 +18,65 @@ import (
...
 
-	if trace.IsEnabled() {
-		pc := make([]uintptr, 1)
-		runtime.Callers(1, pc)
-		f, _ := runtime.CallersFrames(pc).Next()
-		funcName := f.Function
+	pc := make([]uintptr, 1)
+	runtime.Callers(1, pc)
+	f, _ := runtime.CallersFrames(pc).Next()
+	funcName := f.Function
```
chilagrow, line 64: Do we need this block outside `if trace.IsEnabed()`?

rumyantseva: @chilagrow we will need to know the name of the caller function for other observability things too, e.g. to use it for open telemetry traces, with something like `otel.Tracer("").Start(ctx, funcName)`.

chilagrow: Ah okay, I guess it's not used yet for those other things

internal/util/observability/funccall.go
```diff
@@ -18,52 +18,65 @@ import (
...
 	}
 
-	return fc.leave
+	return ctx, sync.OnceFunc(fc.cancel)
 }
 
-// leave is called on function exit.
-func (fc *funcCall) leave() {
+// cancel is called on function exit.
+func (fc *funcCall) cancel() {
```
rumyantseva, line 74: I'm not sure if `cancel` is a good name here. I'd say we rather want to end something here, not specifically cancel. So I'd prefer to call it `end`, but it's not a strong opinion.

internal/util/observability/funccall.go
rumyantseva: Yes, they will be added in the next PR. The change is made bt Alexey, personally, I don't mind having it inside the `if` and move it out later, but I also don't see a reason to move it back as the change is already made :)

chilagrow (dismissed):

internal/util/observability/funccall.go
```diff
@@ -18,52 +18,65 @@ import (
...
+//
+//		bar(ctx)
+//		// ...
+//	}
 //
 // For the Go execution tracer, FuncCall creates a new region for the function call
 // and attaches it to the task in the context (or background task).
-func FuncCall(ctx context.Context) func() {
+func FuncCall(ctx context.Context) (context.Context, context.CancelFunc) {
+	ctx, cancelCtx := context.WithCancel(ctx)
```
rumyantseva, line 53: To be honest, I don't understand the idea of having `context.WithCancel` here.

rumyantseva: Or, better say, I don't undertrstand why we want to return `ctx`. If a caller has a gorutine and passes this inherited context to that gorutine, we will have a problem.

internal/util/observability/funccall.go
```diff
@@ -18,52 +18,65 @@ import (
...
-func (fc *funcCall) leave() {
+// cancel is called on function exit.
+func (fc *funcCall) cancel() {
 	if fc.region != nil {
 		fc.region.End()
 	}
 
 	resource.Untrack(fc, fc.token)
+
+	fc.cancelCtx()
```
rumyantseva, line 81: The only idea I had - it would work if we don't cancel that context.

chilagrow: The failure is interesting. Because I didn't expect this. We create a copy of original context, when function exits and cancel that context all operation should be done. So I didn't expect to see `context canceled` error.

chilagrow: Actually reading the code, I understand what is happening. The `context canceled` came from integration test setup code to cleanup collection after test is completed.

See how observability is calling defer cancel?
https://github.com/FerretDB/FerretDB/blob/f79460f8b8a7973842c5dfe9005ca1a3cdb836dc/integration/setup/setup.go#L210-L211

See how `t.Cleanup()` is using canceled context because it was cancelled when that function exited. So it fails here.
https://github.com/FerretDB/FerretDB/blob/f79460f8b8a7973842c5dfe9005ca1a3cdb836dc/integration/setup/setup.go#L251-L258

What we can do is using a new context to call `t.Cleanup` or cancel the context in `t.Cleanup`. Which captures observability more accurately?

chilagrow: We would have same issue in other places where `tb.Cleanup` is used within integration test, such as in compatSetup.

rumyantseva: Yes, also in functions with goroutines, and, for example, `db.QueryContext`. I haven't found the whole list yet, `tb.Cleanup` was my last finding yesterday. My biggest concern is if we have so many exceptions, is it worth it at all?

chilagrow: I only spot two places, above and this https://github.com/FerretDB/FerretDB/blob/main/integration/setup/setup_compat.go#L158.

Where are other places?

rumyantseva: See the comments starting with `FuncCall's context can't be used`

chilagrow: I see one place where it is used. It's where we start FerretDB with `l.Run(ctx)` right? https://github.com/FerretDB/FerretDB/blob/main/integration/setup/listener.go#L252

Yes this one is quite naturally not gonna work.

I have a question, how should observability work for go routine? Do we start new observability for go routine? Do we ignore observability for go routine? In any case it cannot share the same observability/ctx from the function that created go routine.

chilagrow: I think it depends on how we want to capture observability on tools we use.

rumyantseva: > I have a question, how should observability work for go routine? Do we start new observability for go routine

To me, the question is what do we want to observe, how do we choose these places and why :)

And actually, I have an unpopular (at FerretDB) opinion here that with otel we should focus on request-based calls (however, in tests we do it differently). On the other hand, if we talk about pprof or metrics or execution tracing, these things don't have to be focused on requests, rather on interesting parts of the code.

chilagrow: I think you have a really good point. It's true that what we want to focus differs based on what we are using.
Also true, the test is different. For example for pprof and metrics, things such as setup in test might be interesting (in my opinion)

chilagrow: The current implementation detaches ctx and observability when go routine is around. Below are a few alternatives, code might not compile but just getting the idea across.

We could create observability for each go routine Can it even be useful for our tracking tool (maybe not?).
```
func setupListener(tb testtb.TB, ctx context.Context, logger *zap.Logger, opts *BackendOpts) string {
	tb.Helper()

	setupCtx, cancel := observability.FuncCall(ctx) // setupCtx is used for setting up a listener
	defer cancel()

	runDone := make(chan struct{})

	go func(ctx context.Context) {
		ctx, cancel := observability.FuncCall(ctx)
		defer cancel()

		defer close(runDone)

		l.Run(ctx)
	}(ctx)
```
We can ignore observability for go routine.
```
func setupListener(tb testtb.TB, ctx context.Context, logger *zap.Logger, opts *BackendOpts) string {
	tb.Helper()

	setupCtx, cancel := observability.FuncCall(ctx) // setupCtx is used for setting up a listener
	defer cancel()

	runDone := make(chan struct{})

	go func() {
		defer close(runDone)

		l.Run(ctx)
	}()
```

AlekSi: > To me, the question is what do we want to observe, how do we choose these places and why :)

You just answered your own questions

rumyantseva:
Ready for another round of review! In all the places where `FuncCall`'s context can't be used I added comments explaining why.

internal/backends/collection.go
```diff
@@ -103,7 +103,8 @@ type QueryResult struct {
 //
 // Limit, if non-zero, should be applied.
 func (cc *collectionContract) Query(ctx context.Context, params *QueryParams) (*QueryResult, error) {
-	defer observability.FuncCall(ctx)()
+	_, cancel := observability.FuncCall(ctx) // FuncCall's context can't be used for Query because of cancellation.
```
chilagrow, line 106: Is this due to getMore using the same context?

internal/clientconn/conn.go
```diff
@@ -575,7 +575,8 @@ func (c *conn) route(ctx context.Context, reqHeader *wire.MsgHeader, reqBody wir
 func (c *conn) handleOpMsg(ctx context.Context, msg *wire.OpMsg, command string) (*wire.OpMsg, error) {
 	if cmd, ok := c.h.Commands()[command]; ok {
 		if cmd.Handler != nil {
-			defer observability.FuncCall(ctx)()
+			_, cancel := observability.FuncCall(ctx) // FuncCall's context can't be used for cursors because of cancellation.
```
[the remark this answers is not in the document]
rumyantseva: Yeah, I like this idea! Let me try it to see what happens :)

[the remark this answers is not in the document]
AlekSi: That would cause `find` and `aggregate` commands to be unobservable, and that's kind of what we want. I think we should not forget what we are doing and why, not focus only on workarounds…

The simplest solution seem to be to add `observability.FuncCall` calls to each handler. After all, that the function we want to observe, and the reason why we abstract all things inside `FuncCall` – so we can add a single call to each handler, not a lot of calls for tracing/regions/etc

PR #4450, opened 2024-07-11
Port pgx logger

AlekSi:
# Description

Port a bridge between pgx and slog from v2.
It is not being used yet, that's for future PRs.

## Readiness checklist

### DIFF
--- a/internal/util/logging/pgx.go
+++ b/internal/util/logging/pgx.go
@@ -60,11 +60,8 @@ func (sl *pgxLogger) Log(ctx context.Context, level tracelog.LogLevel, msg strin
 
 	l, ok := pgxLogLevels[level]
 	if !ok {
-		l = slog.LevelError + 1
-		sl.l.LogAttrs(
-			ctx, slog.LevelWarn, "Invalid pgx log level",
-			slog.Int64("level", int64(level)), slog.String("original_msg", msg),
-		)
+		sl.l.LogAttrs(ctx, LevelDPanic, "Invalid pgx log level for: "+msg, slog.Int("level", int(level)))
+		l = slog.LevelError
 	}
 
 	if !sl.l.Enabled(ctx, l) {
--- a/internal/util/logging/pgx.go
+++ b/internal/util/logging/pgx.go
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !ferretdb_no_postgresql
+
 package logging
 
 import (
--- a/internal/util/logging/pgx.go
+++ b/internal/util/logging/pgx.go
@@ -1,0 +1,144 @@
+// Copyright 2021 FerretDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logging
+
+import (
+	"context"
+	"log/slog"
+	"runtime"
+	"slices"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/jackc/pgx/v5/tracelog"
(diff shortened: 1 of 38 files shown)


internal/util/logging/pgx.go
```diff
@@ -0,0 +1,146 @@
...
+	l, ok := pgxLogLevels[level]
+	if !ok {
+		l = slog.LevelError + 1
+		sl.l.LogAttrs(
+			ctx, slog.LevelWarn, "Invalid pgx log level",
+			slog.Int64("level", int64(level)), slog.String("original_msg", msg),
+		)
+	}
+
+	if !sl.l.Enabled(ctx, l) {
```
rumyantseva, line 70: Shouldn't this statement be first in the function?

internal/util/logging/pgx.go
```diff
@@ -0,0 +1,146 @@
...
+	// We can't call runtime.Caller(s) with some fixed skip value and then inspect a single frame
+	// because call Pgx calls Log method from different stack depths.
+
+	var pc uintptr
+	callers := make([]uintptr, 10)
+
+	if n := runtime.Callers(2, callers); n != 0 {
+		frames := runtime.CallersFrames(callers[:n])
+
+		for {
```
rumyantseva, line 83: One question I have here (and also in `FuncCall`) is how much overhead we bring by calling all these additional things, such as `frames.Next()`. As far as I understand, it doesn't bring a lot of overhead as we don't "calculate" any special things, they already exist, but still, it would be nice to run some benchmarks with and without these logs being turned on.

rumyantseva (commented):
Looks good overall, asked a couple of questions.

internal/util/logging/pgx.go
AlekSi: I like when you answer your own questions :-)

internal/util/logging/pgx.go
AlekSi: Why should it?

rumyantseva: Because if we are going to exit from this function, we don't need `now := time.Now()`.

AlekSi: But it does not answer why you think that line should be the first in the function — there is some other code in between

rumyantseva: The code in between is even more confusing to me, I expected it to be a panic.

chilagrow: Logger shouldn't panic, because often it's the last thing we use to debug. So panic in logging should be avoided at all cost.

The thing in the between converts `pgx` log level to `slog` log level. If pgx was configured to log level unknown to `slog`, we don't know how to map it. Best we can do is to log with a log level of our choice :hugs:

internal/util/logging/pgx.go
```diff
@@ -0,0 +1,146 @@
...
+	return &pgxLogger{l: l}
+}
+
+// Log implements [tracelog.Logger].
+func (sl *pgxLogger) Log(ctx context.Context, level tracelog.LogLevel, msg string, data map[string]any) {
+	now := time.Now()
+
+	l, ok := pgxLogLevels[level]
+	if !ok {
+		l = slog.LevelError + 1
```
chilagrow, line 63: why do we log `slog.LevelError + 1` instead of `slog.LevelError`? What does it do when this happens?

rumyantseva (dismissed):

internal/util/logging/pgx.go
chilagrow: Ah, it's equivalent of `LevelDPanic`. Is this allowed?
```suggestion
		l = LevelDPanic
```

internal/util/logging/pgx.go
AlekSi: Pgx documentation does not say what level values are used:
```go
// LogLevel represents the pgx logging level. See LogLevel* constants for
// possible values.
type LogLevel int

// The values for log levels are chosen such that the zero value means that no
// log level was specified.
const (
	LogLevelTrace = LogLevel(6)
	LogLevelDebug = LogLevel(5)
	LogLevelInfo  = LogLevel(4)
	LogLevelWarn  = LogLevel(3)
	LogLevelError = LogLevel(2)
	LogLevelNone  = LogLevel(1)
)
```

It is not clear if `LogLevelNone`, `LogLevel(0)`, and other values could be used.

On the other hand, that's what DPanic is for now. Updated.

`now := time.Now()` is on top to capture the time of the `Log` invocation as early as possible.

noisersup (approved):

chilagrow (approved):

Issue #4013, opened 2024-01-29
Use `slog` for logging


AlekSi:
### What should be done?

We should switch our logging to `log/slog` package:
* https://go.dev/blog/slog
* https://pkg.go.dev/log/slog

Among other things, that would make the usage of the embedded FerretDB package easier.

We probably should not remove all zap logging in this issue, but it should be hidden from the public interface of the `ferretdb` package.

That's a breaking change for v2.

### Where?

TBD

### Definition of Done

TBD

- handler updated;
- all backends updated;
- unit tests added/updated;
- integration/compatibility tests added/updated;
- spot refactorings done;
- user documentation updated or an issue to create documentation created;
- something else?

sbshah97:
Would love to start contributing to this whenever this is marked as ready. I believe it'll be many small PRs

PR #4462, opened 2024-07-17
Use `slog` in `main.go` and embedded `ferretdb` package

chilagrow:
# Description

Closes #4013.

## Readiness checklist


cmd/ferretdb/main.go
```diff
@@ -360,42 +367,44 @@ func run() {
...
-		zap.Bool("debugBuild", info.DebugBuild),
-		zap.Any("buildEnvironment", info.BuildEnvironment.Map()),
+	startupFields := []slog.Attr{
+		slog.String("version", info.Version),
+		slog.String("commit", info.Commit),
+		slog.String("branch", info.Branch),
+		slog.Bool("dirty", info.Dirty),
+		slog.String("package", info.Package),
+		slog.Bool("debugBuild", info.DebugBuild),
+		slog.Any("buildEnvironment", info.BuildEnvironment),
```
AlekSi, line 377: How that line is formatted?

ferretdb/ferretdb.go
```diff
@@ -229,20 +234,24 @@ func (f *FerretDB) MongoDBURI() string {
...
-		logger = zap.L()
+		opts := &logging.NewHandlerOpts{
+			Base:  "console",
+			Level: level,
+		}
+		logging.Setup(opts, "")
+		logger = slog.Default()
 	})
 
 	return logger
```
AlekSi, line 257: I think we should remove that code and just use `slog.Default()` if `SLogger` is nil

ferretdb/ferretdb.go
```diff
@@ -118,18 +121,20 @@ func New(config *Config) (*FerretDB, error) {
 
 	metrics := connmetrics.NewListenerMetrics()
 
-	log := config.Logger
+	log := config.SLogger
 	if log == nil {
 		log = getGlobalLogger()
 	} else {
-		log = logging.WithHooks(log)
+		log = logging.WrapLogger(log)
```
AlekSi, line 128: Ok, but why? We should just use the passed logger. It is the caller's responsibility to pass the logger (and handler) configured to their desires

chilagrow: If the logger isn't wrapped, `getLog` will be empty, and yes we can make that as the caller's responsibility. I'll make a comment about it on `Config.SLogger`.

ferretdb/ferretdb.go
chilagrow: Sure, I think `version.Get().DebugBuild` is left over from when there was no option to configure a logger.

cmd/ferretdb/main.go
chilagrow: It's `*types.Document`, so we would need `slog.LogValuer` implemented for this.

ferretdb/ferretdb.go
chilagrow: Just thinking, if `SLogger` is nil, we still want to call `logging.Setup` to populate `RecentEntries`. Do you recall why `sync.Once` is used to setup logger though?

cmd/ferretdb/main.go
chilagrow: Actually let's convert this to `*bson.Document`, that implements `slog.LogValuer` already https://github.com/FerretDB/FerretDB/blob/main/internal/bson/document.go#L220.

chilagrow: Ignore my last comment, handling it in https://github.com/FerretDB/FerretDB/pull/4479

noisersup (approved):
LGTM

rumyantseva (dismissed):
Looks good!

rumyantseva (approved):
🚀

ferretdb/ferretdb.go
AlekSi: `sync.Once` is used because setup functions change the global state: change the default logger, change what is being used by old `log` package in code like `log.Printf`

AlekSi (approved):

PR #4464, opened 2024-07-17
Use slog in `fsql`

chilagrow:
# Description

Closes #4013.

## Readiness checklist


internal/util/fsql/tx.go
```diff
@@ -116,8 +127,10 @@ func (tx *Tx) ExecContext(ctx context.Context, query string, args ...any) (sql.R
 		ra = &rav
 	}
 
-	fields = append(fields, zap.Int64p("rows", ra), zap.Duration("time", time.Since(start)), zap.Error(err))
-	tx.l.Sugar().With(fields...).Debugf("<<< %s", query)
+	fields = append(fields, slog.Any("rows", ra), slog.Duration("time", time.Since(start)), logging.Error(err))
```
AlekSi, line 130: `slog.Any("rows", ra)` is unlikely to be handled correctly by any handler

chilagrow: True, I've only tested in `console` handler, but you are right about other handlers unable to handle it.

chilagrow: I've changed the logic a bit, `rows` attr is added when exec is successful, only when there is no error. There is no need to set the attribute if its value is nil in my opinion.

internal/util/fsql/tx.go
```diff
@@ -86,13 +91,17 @@ func (tx *Tx) QueryRowContext(ctx context.Context, query string, args ...any) *s
 
 	start := time.Now()
 
-	fields := []any{zap.Any("args", args)}
-	tx.l.Sugar().With(fields...).Debugf(">>> %s", query)
+	fields := []any{slog.Any("args", args)}
+	if tx.l.Enabled(ctx, slog.LevelDebug) {
+		tx.l.With(fields...).DebugContext(ctx, fmt.Sprintf(">>> %s", query))
```
AlekSi, line 96: That check is not needed: `DebugContext` does nothing if debug level is not enabled

chilagrow: I placed the check because evaluating `fmt.Sprintf` on query may be undesirable since `fmt.Sprintf` is evaluated (and not used) if log level set larger than debug. I check code path and `query` can't be too big. The largest possible is `ExecContext` used by insert, but still not as big as I initially thought.

It might look like below for inserting 15 docs.
`>>> INSERT INTO whateverthetable (_ferretdb_sjson) VALUES (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?)...` 

Well let me simplify to remove `Enabled(ctx, slog.LevelDebug)` check everywhere

internal/util/fsql/db.go
```diff
@@ -80,13 +82,13 @@ func (db *DB) QueryContext(ctx context.Context, query string, args ...any) (*Row
 
 	start := time.Now()
 
-	fields := []any{zap.Any("args", args)}
-	db.l.Sugar().With(fields...).Debugf(">>> %s", query)
+	fields := []any{slog.Any("args", args)}
```
chilagrow, line 85: Args can be `[]any` including such as byte, string, int. As far as I'm aware, they are all primitive types.

chilagrow: @AlekSi any though how I would want to handle this case?

AlekSi: That's one place where `slog.Any` is fine

noisersup (approved):

rumyantseva (approved):
Looks good! I like your attention to details!

AlekSi (approved):

PR #4470, opened 2024-07-17
Use `slog` in handler

chilagrow:
# Description

Closes #4013.

## Readiness checklist


internal/handler/handler.go
```diff
@@ -290,11 +297,11 @@ func (h *Handler) Collect(ch chan<- prometheus.Metric) {
 
 // cleanupAllCappedCollections drops the given percent of documents from all capped collections.
 func (h *Handler) cleanupAllCappedCollections(ctx context.Context) error {
-	h.L.Debug("cleanupAllCappedCollections: started", zap.Uint8("percentage", h.CappedCleanupPercentage))
+	h.L.DebugContext(ctx, "cleanupAllCappedCollections: started", slog.Any("percentage", h.CappedCleanupPercentage))
```
AlekSi, line 300: In general, we should try to avoid using `slog.Any` – we don't know what types are really supported by handlers. In that case, we should convert to an integer type

chilagrow: Sure, I think only time we use `slog.Any` is reserved for logging `struct`, but even then you are correct about not knowing how handlers handle types.

chilagrow: I've left the usage of `slog.Any` for `struct` type and `any` type. For `struct` type it's possible to implement `slog.LogValuer` interface, if we would like. That should be respected by handlers.

internal/handler/common/unimplemented.go
```diff
@@ -60,12 +59,12 @@ func UnimplementedNonDefault(doc *types.Document, field string, isDefault func(v
...
 
 // Ignored logs a message if doc has any of the given fields.
-func Ignored(doc *types.Document, l *zap.Logger, fields ...string) {
+func Ignored(doc *types.Document, l *slog.Logger, fields ...string) {
 	for _, field := range fields {
 		if v, err := doc.Get(field); err == nil {
 			l.Debug(
 				"ignoring field",
-				zap.String("command", doc.Command()), zap.String("field", field), zap.Any("value", v),
+				slog.String("command", doc.Command()), slog.String("field", field), slog.Any("value", v),
```
chilagrow, line 67: `slog.Any("value", v)` can be any value from `types` which implies we don't know how some handler logs it.

internal/handler/handler.go
chilagrow: In fact `struct` type should always implement `slog.LogValuer`

internal/handler/common/unimplemented.go
chilagrow: That will be handled by https://github.com/FerretDB/FerretDB/pull/4479

rumyantseva (approved):
Looks good!

rumyantseva (approved):
:shipit:

AlekSi (approved):

PR #4483, opened 2024-07-23
Prototype OTel context propagation

### LINKED ISSUE #4456: Prototype OTel context propagation
### What should be done?

Let's prototype support for sqlcommenter-like context propagation of `trace_id` and `span_id` via `$comment` and/or `comment` parameters. Integration tests should send them, FerretDB should create child spans if those values are present.

### Where?

TBD

### Definition of Done

TBD

- handler updated;
- all backends updated;
- unit tests added/updated;
- spot refactorings done;
- something else?

rumyantseva:
# Description

Closes #4456

## Readiness checklist

## Demo

```
task test-integration-postgresql TEST_RUN='TestOtelComment'
```

Screenshot 2024-08-01 at 14 12 02

Example of trace in Jaeger UI.


AlekSi (commented):
Please push the actual code

internal/clientconn/otel_propagation.go
```diff
@@ -0,0 +1,51 @@
...
+	type TraceData struct {
+		TraceID [16]byte `json:"ferretTraceID"`
+		SpanID  [8]byte  `json:"ferretSpanID"`
+	}
+
+	var data TraceData
+
+	err := json.Unmarshal([]byte(comment), &data)
+	if err != nil {
+		return trace.SpanContext{}, lazyerrors.Error(err)
```
chilagrow, line 42: I suppose we don't stop users from setting other comment of their choice. Is this error message sufficient to tell if that happens such as any string.

internal/clientconn/conn.go
```diff
@@ -441,6 +442,24 @@ func (c *conn) route(connCtx context.Context, reqHeader *wire.MsgHeader, reqBody
 
 		command = document.Command()
 
+		if comment, cerr := document.Get("comment"); cerr == nil {
+			if cmt, ok := comment.(string); ok {
+				var spanCtx trace.SpanContext
+
+				spanCtx, cerr = spanContextFromComment(cmt)
+				if cerr != nil {
+					c.l.ErrorContext(connCtx, "Failed to extract span context from comment", logging.Error(err))
```
chilagrow, line 451: Do you think error level is too high for this log? Users can set string comment and that's expected behaviour.

internal/clientconn/otel_propagation.go
rumyantseva: I think it's enough as a prototype :)

Ideally, we should also validate if `TraceID` and `SpanID` are correct. I suggest, we decide first if we do anything with this approach, and then we can have better errors etc.

internal/clientconn/conn.go
```diff
@@ -441,6 +442,24 @@ func (c *conn) route(connCtx context.Context, reqHeader *wire.MsgHeader, reqBody
...
+
+				spanCtx, cerr = spanContextFromComment(cmt)
+				if cerr != nil {
+					c.l.DebugContext(connCtx, "Failed to extract span context from comment", logging.Error(err))
+				}
+
+				connCtx = trace.ContextWithRemoteSpanContext(connCtx, spanCtx)
+
+				var remoteCtxSpan trace.Span
+				connCtx, remoteCtxSpan = otel.Tracer("").Start(connCtx, command)
```
AlekSi, line 457: Let's create an extra span only if span info was given, not for any string comment

internal/clientconn/conn.go
```diff
@@ -441,6 +442,24 @@ func (c *conn) route(connCtx context.Context, reqHeader *wire.MsgHeader, reqBody
 
 		command = document.Command()
 
+		if comment, cerr := document.Get("comment"); cerr == nil {
```
AlekSi, line 445:
```go
var comment any
if comment, err = …
```

Otherwise, it is easy to use `err` instead of `cerr`, just like below

internal/clientconn/conn.go
```diff
@@ -441,6 +442,24 @@ func (c *conn) route(connCtx context.Context, reqHeader *wire.MsgHeader, reqBody
...
+			if cmt, ok := comment.(string); ok {
+				var spanCtx trace.SpanContext
+
+				spanCtx, cerr = spanContextFromComment(cmt)
+				if cerr != nil {
+					c.l.DebugContext(connCtx, "Failed to extract span context from comment", logging.Error(err))
```
AlekSi, line 451: `cerr` (but see above)

internal/clientconn/otel_propagation.go
```diff
@@ -0,0 +1,51 @@
...
+// If the comment is empty, it returns an empty span context and no error.
+func spanContextFromComment(comment string) (trace.SpanContext, error) {
+	if comment == "" {
+		return trace.SpanContext{}, nil
+	}
+
+	type TraceData struct {
+		TraceID [16]byte `json:"ferretTraceID"`
+		SpanID  [8]byte  `json:"ferretSpanID"`
+	}
```
AlekSi, line 36: Let's add injecting/extracting code into `internal/util/observability` package so the format is not duplicated.
Let's also improve the format (right now, it is a JSON array, which is weird: https://go.dev/play/p/lE11AQSHbsa) and add a top-level field instead of adding prefixes to fields. After all, it is not some special FerretDB Trace ID – it is a normal OTel Trace ID, just wrapped with another object to avoid clashes.

internal/clientconn/conn.go
rumyantseva: Is there any particular reason why we should do it?
I think, it would be nice to have a parent span somewhere, and this place is a good candidate.

internal/clientconn/conn.go
rumyantseva: It's not really possible because a few lines above, there is `err` returned from `msg.Dcoument`, and there are two checks of this `err` below.

internal/util/observability/otel_propagation.go
```diff
@@ -0,0 +1,124 @@
...
+
+	"go.opentelemetry.io/otel/trace"
+
+	"github.com/FerretDB/FerretDB/internal/util/lazyerrors"
+)
+
+// TraceData represents OpenTelemetry tracing information that can be used to restore the info about parent span.
+type TraceData struct {
+	TraceID [16]byte `json:"traceID"`
+	SpanID  [8]byte  `json:"spanID"`
```
chilagrow, line 29: Is there a need for `json:"traceID"` tags? I'm seeing anonymous struct is actually used for marshall/unmarshall.

internal/util/observability/otel_propagation.go
```diff
@@ -0,0 +1,124 @@
...
+		TraceID: hex.EncodeToString(td.TraceID[:]),
+		SpanID:  hex.EncodeToString(td.SpanID[:]),
+	})
+}
+
+// UnmarshalJSON implements json.Unmarshaler interface.
+func (td *TraceData) UnmarshalJSON(data []byte) error {
+	var stringData struct {
+		TraceID string `json:"traceID"`
+		SpanID  string `json:"spanID"`
```
chilagrow, line 47: I think there was an earlier comment about not duplicated format such as `json:"spanID"`, I think we can only define format once right?

noisersup (approved):

AlekSi (approved):

PR #4503, opened 2024-07-30
Fix log message

chilagrow:
# Description

## Readiness checklist


internal/clientconn/listener.go
```diff
@@ -111,22 +113,22 @@ func Listen(opts *NewListenerOpts) (*Listener, error) {
...
-		ll.Info(fmt.Sprintf("Listening on Unix %s...", l.UnixAddr()))
+		ll.InfoContext(ctx, fmt.Sprintf("Listening on Unix %s...", l.UnixAddr()))
 	}
 
 	if l.TLS != "" {
 		var config *tls.Config
 
 		if config, err = tlsutil.Config(l.TLSCertFile, l.TLSKeyFile, l.TLSCAFile); err != nil {
-			return nil, err
+			return nil, lazyerrors.Error(err)
```
AlekSi, line 123: That's one of the few places where the error string is user-visible. Try to start FerretDB with a non-existing TLS file and see the output. It should be nice, and lazyerrors will make it worse.

Let's add a comment so we don't forget.

rumyantseva (commented):
Looks good apart from what Alexey noted.

I see we use `...` almost everywhere, but not on the line `211`. Personally, I like having `...` because they give an idea that the process hasn't finished yet (but it's not a strong opinion).

internal/clientconn/listener.go
chilagrow: Indeed, lazyerror is not pretty. Added a comment

chilagrow:
> Looks good apart from what Alexey noted.
> 
> I see we use `...` almost everywhere, but not on the line `211`. Personally, I like having `...` because they give an idea that the process hasn't finished yet (but it's not a strong opinion).

I've left `...` in L211

internal/clientconn/conn.go
```diff
@@ -167,7 +167,7 @@ func (c *conn) run(ctx context.Context) (err error) {
 
 	defer func() {
 		if p := recover(); p != nil {
-			c.l.LogAttrs(ctx, logging.LevelDPanic, fmt.Sprint("Failed to run", p), logging.Error(err))
+			c.l.LogAttrs(ctx, logging.LevelDPanic, fmt.Sprint("Failed to run: ", p), logging.Error(err))
```
AlekSi, line 170: If that PR intends to sync changes with v2 (as the branch name indicates), then this line looks different there

AlekSi (commented):
Let's that be the last PR that fixes logging for a while because I'm concerned that we are in the endless loop of "port those changes with slight improvements that we need to port back (with slight improvements…)"

internal/clientconn/conn.go
chilagrow: Yeah let's make the same.

rumyantseva (approved):
Looks good!

noisersup (approved):

PR #4695, opened 2025-01-24
Skip private issue links in `checkcomments` by a flag

chilagrow:
# Description

Also added todos and made simple fixes to make CI pass.

Closes #4693.

## Readiness checklist

- [x] I added/updated unit tests (and they pass).
- [ ] I added/updated integration/compatibility tests (and they pass).
- [ ] I added/updated comments and checked rendering.
- [x] I made spot refactorings.
- [ ] I updated user documentation.
- [ ] I ran `task all`, and it passed.
- [x] I ensured that PR title is good enough for the changelog.
- [x] (for maintainers only) I set Reviewers ([`@FerretDB/core`](https://github.com/orgs/FerretDB/teams/core)), Milestone (`Next`), Labels, Project and project's Sprint fields.
- [x] I marked all done items in this checklist.

PR #4703, opened 2025-01-27
Refactor extract and convert in `genwrap` tool

### LINKED ISSUE #4700: Refactor and fix genwrap tool
### What should be done?

Currently, it can't generate code that would compile for `documentdb_api` schema (and probably others) due to duplicate function names, unhandled types, etc. See reference to this issue where we skip some functions.

In general, we should refactor the whole tool to better separate the Extract, Convert, and Generate stages. For example:
1. The current Extract stage converts data to another format, not actually extracts data from the database. The database connection, query, etc should be in `extract.go`, not `main.go`. It should not contain code for conversion like `extractedRoutineParam.toDataType()`. The test should also connect to a real database and compare a couple of results, not be a hermetic unit test.
2. The Convert stage should return data in a format suitable for the Generate stage. Currently, the Generate stage has to do its own conversions, effectively making all three stages do conversions.
3. All Go code should be generated in `generate.go`, not having a part of it in `genwrap.go`.

### Where?

`documentdb/genwrap` package.

### Definition of Done

- `genwrap` works for all DocumentDB schemas, even internal ones;
- unit tests added/updated;
- spot refactorings done.

chilagrow:
# Description

Failing linter would be solved by #4695.

Closes #4700.

## Readiness checklist

### DIFF
--- a/internal/documentdb/genwrap/convert.go
+++ b/internal/documentdb/genwrap/convert.go
@@ -22,6 +22,9 @@ import (
 	"unicode"
 )
 
+// Converter converts SQL data types to Go data types.
+type Converter struct{}
+
 // convertedRoutineParam represents name and type of Go parameter.
 type convertedRoutineParam struct {
 	Name string
@@ -47,7 +50,9 @@ type convertedRoutine struct {
 // For an anonymous SQL parameter, it assigns a unique name.
 // It also produces SQL query placeholders and return parameters in strings.
 func Convert(rows []map[string]any) map[string]convertedRoutine {
-	routineParams := groupBySpecificName(rows)
+	c := new(Converter)
+
+	routineParams := c.groupBySpecificName(rows)
 
 	routines := map[string]convertedRoutine{}
 
@@ -67,7 +72,7 @@ func Convert(rows []map[string]any) map[string]convertedRoutine {
 			}
 
 			if row["parameter_mode"] == "IN" {
-				name = uniqueName(paramNames, name)
+				name = c.uniqueName(paramNames, name)
 			}
 
 			paramNames = append(paramNames, name)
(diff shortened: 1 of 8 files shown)


internal/documentdb/genwrap/convert.go
```diff
@@ -37,117 +38,111 @@ type convertedRoutine struct {
...
 
-	var comment []string
-	var sqlParams []string
-	var sqlReturns []string
+	routines := map[string]convertedRoutine{}
 
-	uniqueNameChecker := map[string]struct{}{}
-	var placeholderCounter int
+	specificNames := slices.Collect(maps.Keys(routineParams))
+	slices.Sort(specificNames)
```
AlekSi, line 55: https://pkg.go.dev/slices#Sorted

internal/documentdb/genwrap/convert.go
```diff
@@ -289,3 +284,73 @@ func handleFunctionOverloading(f *convertedRoutine) {
 		f.Name = funcName
 	}
 }
+
+// groupBySpecificName groups rows by specific_name.
+func groupBySpecificName(rows []map[string]any) map[string][]map[string]any {
```
AlekSi, line 289: I wonder if it makes sense to make all those free-standing functions methods on a Converter type just to group them together

chilagrow: It might make it cleaner visually.

chilagrow: Actually this also allows (unnecessary) conversion done in generate.go. So it added value.

internal/documentdb/genwrap/convert.go
chilagrow: Much better :smile:

noisersup (approved):

AlekSi (approved):

PR #4714, opened 2025-01-29
Check DocumentDB issues in `checkcomments` and `checkdocs`

### LINKED ISSUE #4702: Make `checkcomments` check DocumentDB issues
### What should be done?

We should extend our `// TODO <URL>` check to cover https://github.com/microsoft/documentdb/issues/XXX URLs, too.
There is no need to extend it to different organizations, though.

See #4701 for a motivation.

### Where?

https://github.com/FerretDB/FerretDB/tree/main/tools/checkcomments

### Definition of Done

- https://github.com/microsoft/documentdb/issues/XXX issues are checked;
- unit tests added/updated;
- spot refactorings done.

### LINKED ISSUE #4715: Make `checkdocs` check issue URLs in documentation
### What should be done?

We should check that linked issues in our documentation are still open. Currently, `checkIssueURLs` is not being called from non-testing code, and the `main` function seems to check only blog posts.

### Where?

https://github.com/FerretDB/FerretDB/tree/main/tools/checkdocs

### Definition of Done

- issues are checked in all `*.md` and `*.mdx` files in the `website/`;
- documentation updated.

chilagrow:
# Description

Closes #4702.
Closes #4715.

## Readiness checklist


tools/checkdocs/checkdocs.go
```diff
@@ -43,10 +43,10 @@ func main() {
 	}
 }
 
-// issueRE represents FerretDB issue url with label in Markdown format,
+// issueRE represents FerretDB issue or microsoft/document issue url with label in Markdown format,
 // such as: `[Label](https://github.com/FerretDB/FerretDB/issues/1)`.
-// It returns url as a submatch.
-var issueRE = regexp.MustCompile(`\[\w+]\((\Qhttps://github.com/FerretDB/\E[-\w]+/issues/\d+)\)`)
+// It returns url and owner/repo name as submatches.
+var issueRE = regexp.MustCompile(`\[\w+]\((\Qhttps://github.com/\E(FerretDB/[-\w]+|microsoft/documentdb)/issues/\d+)\)`)
```
chilagrow, line 49: Do we want to check just the simple url? At the moment, it checks the following format only
```
[Label](https://github.com/FerretDB/FerretDB/issues/1)
```

AlekSi: Yes, because we want to check URLs in comments, for example

chilagrow: Updated to check simple URLs.

tools/checkcomments/checkcomments.go
```diff
@@ -31,7 +31,8 @@ import (
 )
 
 // todoRE represents correct "// TODO" comment format.
-var todoRE = regexp.MustCompile(`^// TODO (\Qhttps://github.com/FerretDB/\E[-\w]+/issues/\d+)$`)
+// It returns url and owner/repo name as submatches.
+var todoRE = regexp.MustCompile(`^// TODO (\Qhttps://github.com/\E(FerretDB/[-\w]+|microsoft/documentdb)/issues/\d+)$`)
```
AlekSi, line 35: I think that's a bit too complicated 😬 I think we could just check the org name and ignore repository name. So this code would also check issues in other MS repos – but it is highly unlikely we will have any references to them

tools/github/client.go
```diff
@@ -110,9 +111,9 @@ func CacheFilePath() (string, error) {
 }
 
 var (
-	// Correctly formatted FerretDB issue.
-	// It returns repository name and the issue number as it's submatches.
-	urlRE = regexp.MustCompile(`^\Qhttps://github.com/FerretDB/\E([-\w]+)/issues/(\d+)$`)
+	// Checks correctly formatted FerretDB issue and microsoft/documentdb issue.
+	// It returns owner/repository name and the issue number as it's submatches.
+	urlRE = regexp.MustCompile(`^\Qhttps://github.com/\E(FerretDB/[-\w]+|microsoft/documentdb)/issues/(\d+)$`)
```
AlekSi, line 116: Hm, I think it is strange that we have that regex in 3 places. Like, we check if something matches regexp, then pass it there, where we check regex again? Given that regexes are the same now, we might simplify that

chilagrow: Thanks that's true, it makes sense to pass them so I removed regex check in client.go.

tools/checkcomments/checkcomments.go
chilagrow: Now checking any MS issues :+1:

tools/github/client.go
```diff
@@ -110,57 +108,22 @@ func CacheFilePath() (string, error) {
...
-		return
-	default:
-		return
-	}
-}
-
 // IssueStatus returns issue status.
 // It uses cache.
 //
 // If URL formatting is incorrect it returns `ErrIncorrectURL`, or `ErrIncorrectIssueNumber` error.
```
AlekSi, line 121: Is it still?

That's something I noticed in the genwrap PR, too – only small parts are updated, making other things incorrect. If something is being changed, both callers and comments should be analyzed and updated.

chilagrow: Thanks, the comment was no longer correct. Updated relevant parts :pray:

noisersup (approved):

Issue #4431, opened 2024-07-08
Add support for named `slog` loggers


AlekSi:
### What should be done?

Zap logger had support for named loggers:

```
2024-07-10T17:10:31.882+0400  INFO  telemetry  telemetry/telemetry.go:120  Telemetry enabled.
```

Our current console slog output looks like that
```
2024-07-10T17:13:33.817+0400  INFO  telemetry/telemetry.go:120  Telemetry enabled.  {"name":"telemetry"}
```

We probably should add a special handling for named loggers.

TBD

PR #4716, opened 2025-01-29
Add `mongo` slog format

### LINKED ISSUE #4347: Add "mongo" slog format
### What should be done?

We should add another log format / handler that would include JSON fields like https://www.mongodb.com/docs/manual/reference/log-messages/#structured-logging That would improve compatibility with existing tools.

The order of attributes should be stable (but they should not necessarily be sorted alphabetically).

We should keep our other formats, including JSON, that should be more readable.

### Where?

* https://github.com/FerretDB/FerretDB/tree/main/internal/util/logging
* see references to this issue

### Definition of Done

- log handler created;
- CLI updated;
- unit tests added/updated;
- spot refactorings done;
- user documentation updated.

noisersup:
# Description

Closes #4347.

## Readiness checklist


AlekSi (commented):
Make sure to read https://golang.org/s/slog-handler-guide.

internal/util/logging/mongo_handler.go
```diff
@@ -0,0 +1,240 @@
...
+	Components string             `bson:"c"`
+	ID         int                `bson:"id"`
+	Ctx        string             `bson:"ctx"`
+	Svc        string             `bson:"svc,omitempty"`
+	Msg        string             `bson:"msg"`
+	Attr       map[string]any     `bson:"attr,omitempty"`
+	Tags       []string           `bson:"tags,omitempty"`
+	Truncated  map[string]any     `bson:"truncated,omitempty"`
+	Size       map[string]any     `bson:"size,omitempty"`
+}
```
noisersup, line 65: @AlekSi Some of fields here (such as id,or components), are not used by us, but specified and required by the format. Not sure if we want to stay 100% consistent with present fields or not, that's something to discuss. For now I left them empty here, but in further PRs we can remove/modify some of them if needed.

internal/util/logging/mongo_handler.go
```diff
@@ -0,0 +1,240 @@
...
+
+// mongoLog represents a single log message in mongo structured JSON format.
+// For now some fields are ignored and may be empty.
+//
+//nolint:vet // to preserve field ordering
+type mongoLog struct {
+	Timestamp  primitive.DateTime `bson:"t"`
+	Severity   string             `bson:"s"`
+	Components string             `bson:"c"`
```
chilagrow, line 56: I can see it's how `c`, `svc` and `id` are useful for parsing logs but we don't have useful value to populate. Our loggers are often initialized with `logging.WithName(l, "handler")` but that value is already available in `attr` :+1:

chilagrow (commented):
Great work, just a minor refactoring comment :pray:

chilagrow (commented):
There are still few TODO comments left in the code base for the issue.

website/docs/configuration/observability.md
```diff
@@ -21,9 +21,10 @@ FerretDB provides the following log formats:
 <!-- https://github.com/FerretDB/FerretDB/issues/4438 -->
 
 - `console` is a human-readable format with optional colors;
+- `mongo` is machine-readable structured JSON format;
```
AlekSi, line 24: Well, that needs more details

internal/util/logging/handler_test.go
```diff
@@ -43,6 +43,9 @@ func TestHandler(t *testing.T) {
 		"console": `2024-05-31T09:26:42.000Z	INFO	logging/handler_test.go:34	` +
 			"multi\nline\nmessage\t" +
 			`{"g2":{"g1":{"k1":42,"k2":7000000000},"g3":{"s":"a"},"i":1,"k3":"dup","name":"test.logger"}}` + "\n",
+		"mongo": `{"t":{"$date":"2024-05-31T09:26:42Z"},"s":"I","c":"","id":0,"ctx":"logging/handler_test.go:34",` +
```
chilagrow, line 46: Shouldn't it have `"c":"test.logger"` instead of `"c":""`? And `"name":"test.logger"` removed from attrs?
https://github.com/FerretDB/FerretDB/pull/4716/files#diff-220e72c093d41a2d3fa06d5c2123a2ed3ffaa8c4eec66bfeac8df5887ce62d87R128

internal/handler/msg_getlog.go
```diff
@@ -162,18 +162,19 @@ func (h *Handler) MsgGetLog(connCtx context.Context, msg *wire.OpMsg) (*wire.OpM
...
-				},
-			})
+			ml := logging.MongoLog{
+				Msg:       line,
+				Tags:      []string{"startupWarnings"},
+				Severity:  "I",
+				Component: "STORAGE",
+				ID:        42000,
+				Ctx:       "initandListen",
+				Timestamp: primitive.NewDateTimeFromTime(time.Now()),
```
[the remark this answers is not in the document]
noisersup: It's the same, `NewDateTimeFromTime` constructs the `DateTime` based on Unix time.

internal/util/logging/handler_test.go
noisersup: Nice catch, I forgot the name attribute shouldn't be checked only at the top level.
While trying to fix that I realized that doing it outside of `attrs()` function is impossible, as we convert attributes to map, which doesn't store information about the latest group that is used by handler.

[For now](https://github.com/FerretDB/FerretDB/pull/4716/commits/accb7414ee9215c78f313ae7d54c7f092dc17115) I've added the name checking as a part of the function, it simply adds name for both `mongo` and `console` handlers. It's not best, as for example it doesn't support chained names, so #4431 is still relevant and I left the proper TODO comment for it.

I think that group/attribute handling deserves its own struct with proper methods, so it can be reused easily and cleanly in logging handlers, but that's a lot to refactor for a single PR, I think it could be a part of #4431 to handle that.

internal/util/logging/attrs.go
```diff
@@ -0,0 +1,97 @@
...
+
+		for _, attr := range goa.attrs {
+			if attr.Key != nameKey || name != "" {
+				m[attr.Key] = resolve(attr.Value)
+				continue
+			}
+
+			v, ok := attr.Value.Any().(string)
+			if !ok {
+				return nil, "", lazyerrors.Errorf("attribute %q should be a string but was %T", nameKey, v)
```
chilagrow, line 71: I think logger should not return an error lightly. Often logging is the last result for debugging something not working, in that case, any written log would be useful :hugs:.

For this, how about using `Printf()` or something to make the value into string? Then we can keep the function signature without an error.

noisersup: I think we want to omit situation where values that're not strings are just `Sprintf`d into string. That'd result in really weird logger names if for example name is a struct, or pointer.

To eliminate error from the function, I'd suggest to treat `name` as any other attribute in cases where it's not a string. PTAL at the latest PR

chilagrow: Much better! Thank you!

chilagrow (dismissed):

cmd/ferretdb/main.go
```diff
@@ -125,7 +125,7 @@ var (
 		slog.LevelError.String(),
 	}
 
-	logFormats = []string{"console", "text", "json"}
+	logFormats = []string{"console", "mongo", "text", "json"}
```
AlekSi, line 128: Is that the best order?

website/docs/configuration/observability.md
```diff
@@ -21,9 +21,12 @@ FerretDB provides the following log formats:
...
 
 - `console` is a human-readable format with optional colors;
+- `mongo` is machine-readable structured JSON format, similar to the one used in MongoDB.
+  It follows the Relaxed Extended JSON specification.
+  Fields required in the output format but not yet implemented, will be included empty, to maintain compatibility;
 - `text` is machine-readable [logfmt](https://brandur.org/logfmt)-like format
   (powered by [Go's `slog.TextHandler`](https://pkg.go.dev/log/slog#TextHandler));
-- `json` if machine-readable JSON format
+- `json` is machine-readable JSON format
   (powered by [Go's `slog.JSONHandler`](https://pkg.go.dev/log/slog#JSONHandler)).
```
AlekSi, line 30: Is that the best order?
When should semicolons be used in a list?

internal/util/logging/console_handler.go
```diff
@@ -54,32 +54,27 @@ type consoleHandler struct {
...
+	h := consoleHandler{
 		opts:      opts,
 		testAttrs: testAttrs,
 		m:         new(sync.Mutex),
 		out:       out,
 	}
+
+	if h.opts.Level == nil {
+		h.opts.Level = slog.LevelInfo
+	}
```
AlekSi, line 70: Now, we modify the input parameter that is not documented to be modified by this function.

I think we should revert to the previous implementation.

internal/util/logging/handler_test.go
```diff
@@ -72,6 +75,12 @@ func TestHandler(t *testing.T) {
 			l := WithName(slog.New(h), "test.logger")
 			require.NoError(t, l.Handler().Handle(ctx, r))
 
+			// mongo handler doesn't preserve the order of attributes
```
AlekSi, line 78: Well, but does it keep it stable? I find it strange that we text exact values for the JSON format but not for the mongo format that uses the same JSON primitives under the hood.

internal/util/logging/mongo_handler.go
```diff
@@ -0,0 +1,227 @@
...
+	Timestamp primitive.DateTime `bson:"t"`
+	Severity  string             `bson:"s"`
+	Component string             `bson:"c"`
+	ID        int                `bson:"id"`
+	Ctx       string             `bson:"ctx"`
+	Svc       string             `bson:"svc,omitempty"`
+	Msg       string             `bson:"msg"`
+	Attr      map[string]any     `bson:"attr,omitempty"`
+	Tags      []string           `bson:"tags,omitempty"`
+	Truncated map[string]any     `bson:"truncated,omitempty"`
```
AlekSi, line 63: Let's remove fields we don't use, like this one

internal/util/logging/mongo_handler.go
```diff
@@ -0,0 +1,227 @@
...
+	m   *sync.Mutex
+	out io.Writer
+}
+
+// MongoLog represents a single log message in mongo structured JSON format.
+// For now some fields are ignored and may be empty.
+//
+//nolint:vet // to preserve field ordering
+type MongoLog struct {
+	Timestamp primitive.DateTime `bson:"t"`
```
AlekSi, line 54: Do we need to expose `primitive.DateTime` to the caller? Isn't the encoder support `time.Time`?

Also, "t" likely means time, not timestamp (which is a separate BSON/ExtJSON type).

internal/util/logging/handler_test.go
noisersup: Mongo handler doesn't keep the order of attributes stable. 

As the attributes are stored in map, the ordering is random.

Both `json`, and `console` handlers use `encoding/json`, which sorts attribute map keys lexicographically.
The `mongo` handler uses `bson.MarshalExtJSON` function, which doesn't perform any sorting.

internal/util/logging/mongo_handler.go
noisersup: Both `primitive.DateTime`, and "Timestamp" are used, based on the structured logging documentation:

```
{
"t": <Datetime>, // timestamp
"s": <String>, // severity
"c": <String>, // component
"id": <Integer>, // unique identifier
"ctx": <String>, // context
"svc": <String>, // service
"msg": <String>, // message body
"attr": <Object>, // additional attributes (optional)
"tags": <Array of strings>, // tags (optional)
"truncated": <Object>, // truncation info (if truncated)
"size": <Object> // original size of entry (if truncated)
}
```

As `time.Time` marshals to `primitive.DateTime`, I've changed the type in the struct.

internal/util/logging/handler_test.go
AlekSi: > As the attributes are stored in map, the ordering is random.

Could we use `bson.D`?

cmd/ferretdb/main.go
noisersup: Fixed the ordering, so `mongo` is next to the `json` handler.

internal/util/logging/handler_test.go
noisersup: Our logging handlers use [shared function](https://github.com/FerretDB/FerretDB/pull/4716/files#diff-ed4f1a531b6586a30017cb944e778ad59d2e61bfc4119017ae95334da77a7bc1R35) that returns attributes in map.

Changing `mongoLog`'s field from map to `bson.D` wouldn't change anything.

We could add additional logic that would iterate through map, create new `bson.D` object with sorted fields, and then pass it to `bson.MarshalExtJSON`, but I don't see why would we want that. As far as I'm aware the attribute ordering is not expected anywhere, besides the test case.

chilagrow: Usage of `map` in `attrs(..)` function fully internal, wouldn't replacing with `bson.D` allow preserve the order for all our custom handlers?

I don't think we need to iterate the `map` and then sort it, we can let it have the order defined by the logger.

internal/util/logging/mongo_handler.go
```diff
@@ -0,0 +1,223 @@
...
+	testAttrs map[string]any
+
+	m   *sync.Mutex
+	out io.Writer
+}
+
+// MongoLog represents a single log message in mongo structured JSON format.
+// For now it lacks the fields that are not used by any caller.
+//
+//nolint:vet // to preserve field ordering
```
chilagrow, line 52: This struct would be used a lot. Wouldn't be better to optimize the size?

noisersup: It could be, although the order of fields will be affected.

internal/util/logging/handler_test.go
noisersup: I'm not sure if using `bson.D` internally is something that we want to do. The `attrs` function is meant to be used by our logging handlers. In that case for `consoleHandler`, we'd need to take this bson.D output and convert it to value that'd be able to be consumed by `encoding/json` encoder.

AlekSi: > Our logging handlers use shared function

Then we probably should un-share it. Or maybe generalize it by using generics and/or iterators (if that would result in simpler code).
Having attributes in a stable order seems to be useful properly for logs – it makes them more readable, which is kind of important.
Let's to that in a separate PR, but in this issue.

internal/util/logging/console_handler.go
```diff
@@ -102,10 +96,18 @@ func (ch *consoleHandler) Handle(ctx context.Context, r slog.Record) error {
...
 		}
 	}
 
+	m, name := attrs(r, ch.ga)
+
+	// TODO https://github.com/FerretDB/FerretDB/issues/4431
+	if name != "" {
+		buf.WriteString(name)
+		buf.WriteRune('\t')
+	}
```
AlekSi, line 109: Let's not try to change name handling in this PR – it is already quite complicated (and in progress for a long time), and that's a separate issue anyway

internal/util/logging/mongo_handler.go
```diff
@@ -0,0 +1,223 @@
...
+type MongoLog struct {
```
AlekSi, line 53: Could we think of a better name? Even `MongoLogRecord` (similar to `slog.Record`) is probably better.

website/docs/configuration/observability.md
AlekSi: > When should semicolons be used in a list?

When items could be concatenated to form a single sentence:

- John, the baker;
- Simon, the policeman;
- Toby, the architect;

=>

John, the baker; Simon, the policeman; and Toby, the architect.

Using semicolons makes it weird if a list item contains multiple complete sentences. Just use periods.

chilagrow (approved):

website/docs/configuration/observability.md
```diff
@@ -23,8 +23,11 @@ FerretDB provides the following log formats:
 - `console` is a human-readable format with optional colors;
 - `text` is machine-readable [logfmt](https://brandur.org/logfmt)-like format
   (powered by [Go's `slog.TextHandler`](https://pkg.go.dev/log/slog#TextHandler));
-- `json` if machine-readable JSON format
-  (powered by [Go's `slog.JSONHandler`](https://pkg.go.dev/log/slog#JSONHandler)).
+- `json` is machine-readable JSON format
+  (powered by [Go's `slog.JSONHandler`](https://pkg.go.dev/log/slog#JSONHandler));
+- `mongo` is machine-readable structured JSON format, similar to the one used in MongoDB;
+  It follows the Relaxed Extended JSON specification;
+  Fields required in the output format but not yet implemented, will not be included.
```
AlekSi, line 30: https://github.com/FerretDB/FerretDB/pull/4716#discussion_r1962030153. Use dots.

AlekSi (approved):

website/docs/configuration/observability.md
AlekSi: LGTM, but please resolve that in the next PR, maybe after the `dbStats` issue.

internal/handler/msg_getlog.go
```diff
@@ -163,18 +162,19 @@ func (h *Handler) MsgGetLog(connCtx context.Context, msg *wire.OpMsg) (*wire.OpM
...
-					"$date": time.Now().UTC().Format("2006-01-02T15:04:05.999Z07:00"),
-				},
-			})
+			ml := logging.MongoLogRecord{
+				Msg:       line,
+				Tags:      []string{"startupWarnings"},
+				Severity:  "I",
+				Component: "STORAGE",
+				ID:        42000,
+				Ctx:       "initandListen",
```
AlekSi, line 171: Copilot noted that it changed from `initandlisten` to `initandListen`. Was it intentional?

internal/util/logging/mongo_handler.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// MongoLogRecord represents a single log message in mongo structured JSON format.
+// For now it lacks the fields that are not used by any caller.
+//
+//nolint:vet // to preserve field ordering
+type MongoLogRecord struct {
+	Timestamp time.Time      `bson:"t"`
+	Severity  string         `bson:"s"`
+	Component string         `bson:"c"`
```
AlekSi, line 56:
```suggestion
	Component string         `bson:"c"` // TODO https://github.com/FerretDB/FerretDB/issues/4431
```

internal/handler/msg_getlog.go
noisersup: That was mistake. Fixed

chilagrow (approved):

AlekSi (approved):

PR #4787, opened 2025-02-19
Reformat request/response `.js` files

### LINKED ISSUE #4774: Use consistent code formatting in documentation
### What should be done?

We should use `mongosh` output as is, not reformat it to change quotes, add or remove colons and semicolons, etc.
All other JS code snippets should use the same style as `mongosh`.
We should adjust linter settings in all contexts: inside Markdown files, for separate JS files, etc.
We should completely remove `json5` and use `js` (not `json`!) for `mongosh` requests and responses, and `json` where needed.
We should also update our writing guides.

AlekSi:
# Description

Ensure that `mongosh`'s requests and responses are not reformatted.

The problem is that the response is a valid JS snipped (not JSON, not JSON5) in Markdown but not a valid JS code in a separate file. Adding `response = ` hack resolves that. Maybe that's good enough.

Closes #4774.
But we still need to do the rest of the files, update the writing guide, and do the rest in that issue.

## Readiness checklist


build/legacy-mongo-shell/test.js
```diff
@@ -1,28 +1,28 @@
 // Please do not merge changes in this file.
 
-(function () {
-  "use strict";
+;(function () {
```
AlekSi, line 3: That semicolon is very unfortunate

website/docs/guides/responses/3-hnsw-search.json
AlekSi, line 1: That wasn't even the full response

Fashander (approved):
LGTM

PR #4863, opened 2025-03-04
Update TODO URLs for `listDatabase` tests

### LINKED ISSUE #4722: Add filtering and sizes to `listDatabases` command
### What should be done?

DocumentDB part for filtering was implemented https://github.com/microsoft/documentdb/issues/45. There is no issue for sizes in the DocumentDB repo yet (?).

We should use `documentdb_api.list_databases` in the implementation of the `listDatabases` command. It would look similar to `listCollections`.

- [x] Check test (it does not seem to check size correctly).
- [x] Create new issue.
- [x] Update this issue.

### Where?

* https://github.com/FerretDB/FerretDB/blob/main/internal/handler/msg_listdatabases.go
* See references to this issue in integration tests, etc.

### Definition of Done

- handler updated;
- spot refactorings done.

noisersup:
# Description

Closes #4722.

## Readiness checklist


integration/commands_administration_test.go
```diff
@@ -270,7 +269,7 @@ func TestListDatabases(t *testing.T) {
 			}
 			actual.TotalSize = 0
 
-			assert.EqualValues(t, tc.expected, actual)
+			assert.ElementsMatch(t, tc.expected.Databases, actual.Databases)
```
AlekSi, line 272: I believe they should be sorted, just like in `listCollections`

chilagrow: I saw CI sometimes failing, yep let's sort it as the order is not deterministic :see_no_evil:.

chilagrow: I saw CI sometimes failing, yep let's sort it as the order is not deterministic :see_no_evil:.

chilagrow (commented):
Let's apply @AlekSi's comment

integration/commands_administration_test.go
AlekSi: MongoDB's responses to `listDatabases` and `listCollections` are always sorted. We have a workaround for that DocumentDB problem in `listCollection` implementation. We should apply the same for `listDatabases`, not in test.

AlekSi: * https://www.mongodb.com/docs/manual/reference/command/listDatabases/
* https://www.mongodb.com/docs/manual/reference/command/listCollections/

Hm, both are now explicitly documented as unsorted

AlekSi: 5d72de9d74a2e67070fc3fab996d6e34dff5a9fa

@noisersup Please check if we could simplify `listCollections` tests using `assert.ElementsMatch`, too

noisersup: I couldn't find any test that we could apply the same logic there.

AlekSi (commented):
https://github.com/FerretDB/FerretDB/pull/4863#discussion_r1981124413

chilagrow (dismissed):

AlekSi (commented):
We should also update `TestListCollectionsCompat`

integration/commands_administration_test.go
noisersup: We use `AssertEqualDocumentsSlice` in this test, which ignores the order of fields of nested documents. The tests fail if `assert.ElementsMatch` is used, because of some fields being in different order.

integration/commands_administration_compat_test.go
```diff
@@ -57,6 +58,13 @@ func TestListCollectionsCompat(t *testing.T) {
 	err = compat.All(ctx, &compatRes)
 	require.NoError(t, err)
 
+	compatNames := make([]string, len(compatRes))
+	for i, doc := range compatRes {
+		compatNames[i] = doc.Map()["name"].(string)
+	}
+
+	require.True(t, slices.IsSorted(compatNames), "compat collections are not sorted")
```
AlekSi: We should not check that MongoDB collections are sorted because they are not:

> The `listCollections` command returns a document that contains an **unsorted** list of all collections and views in the database

From https://www.mongodb.com/docs/manual/reference/command/listCollections/

AlekSi (changes requested):

chilagrow (dismissed):
The sorting seems complicated, but ok for now

integration/commands_administration_compat_test.go
```diff
@@ -112,6 +106,39 @@ func TestListCollectionsCompat(t *testing.T) {
...
+			for _, field := range a {
+				if aName != "" {
+					break
+				}
+
+				switch field.Key {
+				case "name":
+					if assert.IsType(t, "", field.Value) {
+						aName = field.Value.(string)
+					}
```
AlekSi, line 121: I agree with Copilot that using `assert` (that would continue the test) inside comparison function is quite weird. Also, I think that's quite unlikely that the name will not be a string, so using a regular assertion would be fine:

```go
if field.Key == "name" {
  aName := field.Value.(string)
  break
}
```

(using `if` there also removes the need to add it above as `break` would work as we want it to)

AlekSi (approved):

PR #4904, opened 2025-03-10
Add colorized levels to `console` logger

### LINKED ISSUE #4438: Add colors to `console` logger
### What should be done?

Let's add very basic colors to the console logs handler.

Let's colorize only the level for now.

Let's use https://pkg.go.dev/golang.org/x/term for both detecting if output is TTY and for coloring itself.

### Where?

TBD

### Definition of Done

TBD

noisersup:
# Description

Closes #4438.

## Readiness checklist


internal/util/logging/console_handler.go
```diff
@@ -75,6 +92,23 @@ func (ch *consoleHandler) Enabled(_ context.Context, l slog.Level) bool {
 	return l >= minLevel
 }
 
+func coloredLevel(l slog.Level) string {
+	t := term.NewTerminal(nil, "")
+
+	switch l {
+	case slog.LevelInfo:
+		return string(t.Escape.Green) + l.String() + string(t.Escape.Reset)
```
noisersup, line 100: @AlekSi So, it's technically possible to use color escape codes from `term` package, but as they're [unexported](https://cs.opensource.google/go/x/term/+/master:terminal.go;l=26), the only way to get them is by creating Terminal, and then using its Escape field.

As we don't use term.Terminal (which takes io.ReadWriter, as an argument, and does things as taking an input, or autocomplete handling), initializing it on side just to takes those values doesn't make much sense. We can store necessary escape codes in variables on the logging package, and just use them as well.

The `term.IsTerminal` is still useful, as it will be used to decide whether colored output should be enabled.

AlekSi: > which takes io.ReadWriter

So we use type assertion.

>  and does things as taking an input, or autocomplete handling

I believe it does that only in raw mode, no?

> We can store necessary escape codes in variables on the logging package

The whole point was to not have terminal magic outside of a dependency that deals with terminals.

internal/util/logging/console_handler.go
```diff
@@ -57,6 +70,10 @@ type consoleHandler struct {
 func newConsoleHandler(out io.Writer, opts *NewHandlerOpts, testAttrs map[string]any) *consoleHandler {
 	must.NotBeZero(opts)
 
+	if !term.IsTerminal(int(os.Stdout.Fd())) {
```
AlekSi, line 73: We should use `out` with a type assertion, not `os.Stdout`

internal/util/logging/console_handler.go
```diff
@@ -180,10 +191,39 @@ func (ch *consoleHandler) WithGroup(name string) slog.Handler {
...
+func (ch *consoleHandler) colorizedLevel(l slog.Level) string {
+	if !ch.isColorized() {
+		return l.String()
+	}
+
+	format := "%s%s%s"
+
+	switch l {
+	case slog.LevelDebug:
+		return fmt.Sprintf(format, ch.esc.Magenta, l, ch.esc.Reset)
```
chilagrow, line 215: Could we use more gray or blue scale color for debug log?

internal/util/logging/console_handler.go
```diff
@@ -180,10 +197,39 @@ func (ch *consoleHandler) WithGroup(name string) slog.Handler {
...
+		return fmt.Sprintf(format, ch.term.Escape.Magenta, l.String(), ch.term.Escape.Reset)
+	case slog.LevelInfo:
+		return fmt.Sprintf(format, ch.term.Escape.Green, l.String(), ch.term.Escape.Reset)
+	case slog.LevelWarn:
+		return fmt.Sprintf(format, ch.term.Escape.Yellow, l.String(), ch.term.Escape.Reset)
+	case slog.LevelError:
+		return fmt.Sprintf(format, ch.term.Escape.Red, l.String(), ch.term.Escape.Reset)
+	default:
+		return l.String()
+	}
```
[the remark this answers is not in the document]
chilagrow: Also there are very commonly used `slog.LevelError + n` constants.

https://github.com/FerretDB/FerretDB/blob/fd460f08fe7aa9da273b987e4a907258e6287cbe/internal/util/logging/logging.go#L27-L34

internal/util/logging/console_handler.go
```diff
@@ -49,20 +50,31 @@ type consoleHandler struct {
...
+	ch := &consoleHandler{
 		opts:      opts,
 		testAttrs: testAttrs,
 		m:         new(sync.Mutex),
 		out:       out,
 	}
+
+	if f, ok := out.(*os.File); ok && term.IsTerminal(int(f.Fd())) {
+		t := term.NewTerminal(f, "")
+		ch.out = t
```
AlekSi, line 73: https://github.com/FerretDB/FerretDB/pull/4904#discussion_r1989355307

internal/util/logging/console_handler.go
```diff
@@ -180,10 +191,41 @@ func (ch *consoleHandler) WithGroup(name string) slog.Handler {
...
 		out:       ch.out,
+		esc:       ch.esc,
 		testAttrs: ch.testAttrs,
 	}
 }
 
+// isColorized returns true if the handler is able to print colorized messages.
+func (ch *consoleHandler) isColorized() bool {
+	return ch.esc != nil
+}
```
AlekSi, line 202: Why do we need that function now?

internal/util/logging/console_handler.go
```diff
@@ -180,10 +191,41 @@ func (ch *consoleHandler) WithGroup(name string) slog.Handler {
...
+		return fmt.Sprintf(format, ch.esc.Green, l, ch.esc.Reset)
+	case l == slog.LevelWarn:
+		return fmt.Sprintf(format, ch.esc.Yellow, l, ch.esc.Reset)
+	case l >= slog.LevelError:
+		return fmt.Sprintf(format, ch.esc.Red, l, ch.esc.Reset)
+
+	default:
+		return l.String()
+	}
+}
```
AlekSi, line 227: https://github.com/FerretDB/FerretDB/pull/4904#discussion_r1989383413

internal/util/logging/console_handler_test.go
```diff
@@ -49,3 +54,96 @@ func TestConsoleHandler(t *testing.T) {
...
+			expected: "\033[32mINFO\033[0m\tfoobar\r\n",
+		},
+		"Warn": {
+			level:    slog.LevelWarn,
+			expected: "\033[33mWARN\033[0m\tfoobar\r\n",
+		},
+		"Error": {
+			level:    slog.LevelError,
+			expected: "\033[31mERROR\033[0m\tfoobar\r\n",
+		},
```
AlekSi, line 106: You even wrote tests that assume that there are no log levels between, say, warn and error – but there are

AlekSi (changes requested):
Existing comments were not addressed

internal/util/logging/console_handler.go
noisersup: To output non-colorized output if the `io.Writer` is not a terminal, which is a case in some tests, as well as possible in embedded FerretDB.

AlekSi: I'm not asking why we need the `!=` part. But that function is used precisely once, below. That's just more code for no good reason.

internal/util/logging/console_handler.go
```diff
@@ -180,10 +191,39 @@ func (ch *consoleHandler) WithGroup(name string) slog.Handler {
...
+	if !ch.isColorized() {
+		return l.String()
+	}
+
+	format := "%s%s%s"
+
+	switch {
+	case l < slog.LevelInfo:
+		return fmt.Sprintf(format, ch.esc.Blue, l, ch.esc.Reset)
+	case l >= slog.LevelInfo && l < slog.LevelWarn:
```
AlekSi, line 216: Just `l < XXX`. Cases are evaluated sequentially

internal/util/logging/console_handler.go
AlekSi: Well, readability is definitely better – there are more things to read…

Yes, we don't need a function for a single comparison in a single place, where it is clear from the context what is going on.

AlekSi (dismissed):

internal/util/logging/console_handler.go
```diff
@@ -180,10 +191,41 @@ func (ch *consoleHandler) WithGroup(name string) slog.Handler {
...
+	switch {
+	case l < slog.LevelInfo:
+		// debug levels
+		return fmt.Sprintf(format, ch.esc.Blue, l, ch.esc.Reset)
+	case l == slog.LevelInfo:
+		return fmt.Sprintf(format, ch.esc.Green, l, ch.esc.Reset)
+	case l == slog.LevelWarn:
+		return fmt.Sprintf(format, ch.esc.Yellow, l, ch.esc.Reset)
+	case l >= slog.LevelError:
+		return fmt.Sprintf(format, ch.esc.Red, l, ch.esc.Reset)
```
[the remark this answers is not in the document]
chilagrow: Let's just use color :hugs:

chilagrow (dismissed):

AlekSi (approved):

chilagrow (approved):

Issue #1997, opened 2023-02-15
Proxy should provide metrics and traces


AlekSi:
### What should be done?

We should provide metrics and traces for proxy modes, too. That should help us investigate compatibility and performance problems.

We should set the same labels and attributes. That would require changing how incoming messages are handled – the message should be partially decoded in the `clientconn` to extract command, etc.

Metrics handling could be another "middleware", so when `dataapi` calls that handler directly, those requests are visible in metrics.

We could also send the message to the proxy in parallel with calling a handling function. That would allow us to compare performance directly and would nicely align traces.

⚠️ But before doing any of that, we should refactor `clientconn` to make it simpler to change (without actually changing behaviour in this first PR).

### Where?

* https://github.com/FerretDB/FerretDB/tree/main/internal/handlers/proxy
* https://github.com/FerretDB/FerretDB/tree/main/internal/clientconn

See references to this issue.

### Definition of Done

- metrics are available;
- `clientconn` package is refactored.

PR #4951, opened 2025-03-21
Implement error middleware

chilagrow:
# Description

Closes #1997.

## Readiness checklist

### DIFF
--- a/internal/clientconn/conn.go
+++ b/internal/clientconn/conn.go
@@ -396,19 +396,28 @@ func (c *conn) route(connCtx context.Context, reqHeader *wire.MsgHeader, reqBody
 		query := reqBody.(*wire.OpQuery)
 		resHeader.OpCode = wire.OpCodeReply
 
-		// do not store typed nil in interface, it makes it non-nil
-
-		o := &middleware.Observability{
+		observabilityHandler := &middleware.Observability{
 			L: logging.WithName(c.l, "observability"),
 		}
 
-		replyHandler := o.HandleOpReply(c.h.CmdQuery)
+		errHandler := middleware.NewError("", logging.WithName(c.l, "error"))
+
+		replyHandler := errHandler.HandleOpReply(c.h.CmdQuery)
+		replyHandler = observabilityHandler.HandleOpReply(replyHandler)
 
 		var reply *middleware.ReplyResponse
 		reply, err = replyHandler(connCtx, &middleware.QueryRequest{OpQuery: query})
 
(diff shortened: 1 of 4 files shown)


internal/clientconn/conn.go
```diff
@@ -396,19 +396,28 @@ func (c *conn) route(connCtx context.Context, reqHeader *wire.MsgHeader, reqBody
...
 
 		var reply *middleware.ReplyResponse
-		reply, err = replyHandler(connCtx, &middleware.QueryRequest{OpQuery: query})
 
-		if reply != nil && reply.OpReply != nil {
-			resBody = reply.OpReply
+		if reply, err = replyHandler(connCtx, &middleware.QueryRequest{OpQuery: query}); err != nil {
+			// must not be reachable, error is handled by middleware.Error
+			panic(err)
+		}
```
AlekSi, line 414: `must.NotFail` could be used there

internal/handler/middleware/error.go
```diff
@@ -0,0 +1,87 @@
...
+				resp = &MsgResponse{
+					OpMsg: mErr.Msg(),
+					Error: mErr,
+				}
+
+				err = nil
+			}
+		}()
+
+		return next(ctx, req)
```
AlekSi, line 58:
```suggestion
		resp, err = next(ctx, req)
		return
```

It is generally not advised to combine named results and non-bare returns – it looks a bit magical: what is assigning `err`?

internal/handler/middleware/middleware.go
```diff
@@ -33,6 +34,7 @@ type MsgRequest struct {
 // MsgResponse represent outgoing response to the client.
 type MsgResponse struct {
 	*wire.OpMsg
+	*mongoerrors.Error
```
AlekSi, line 37:
```suggestion
	err *mongoerrors.Error
```

Embedding two fields could lead to clashes of method sets. And we don't actually want to embed any of them; we just embedded OpMsg to reduce the number of changes in handlers for now.

internal/handler/middleware/error.go
chilagrow: Yeah let's use naked return statement to make it more readable.

internal/handler/middleware/middleware.go
chilagrow: That's true. Newly added fields doesn't need to be embedded.

noisersup (approved):

AlekSi (approved):

PR #4961, opened 2025-03-22
Remove MongoDB driver v1

### LINKED ISSUE #4798: Remove MongoDB driver v1
### What should be done?

We should switch to using [`go.mongodb.org/mongo-driver/v2`](https://pkg.go.dev/go.mongodb.org/mongo-driver/v2) everywhere. See https://www.mongodb.com/docs/drivers/go/current/whats-new/#what-s-new-in-2.0

Let's start from the FerretDB itself, then update integration tests.

### Where?

Everywhere. See references to this issue.

### Definition of Done

- `go.mongodb.org/mongo-driver/bson` and `go.mongodb.org/mongo-driver/mongo` are no longer imported;
- `depguard` rules updated;
- spot refactorings done.

KrishnaSindhur:
# Description
Removed MongoDB driver v1 only for FerretDB

Closes #4798.

## Readiness checklist


internal/dataapi/server/json.go
AlekSi, line 1: @noisersup PTAL

internal/dataapi/server/json.go
```diff
@@ -15,32 +15,25 @@
...
 	if err != nil {
 		return lazyerrors.Error(err)
 	}
 
-	encoder, err := bson.NewEncoder(vw)
-	if err != nil {
-		return lazyerrors.Error(err)
-	}
-
-	encoder.IntMinSize()
```
chilagrow, line 41: @noisersup could you check  the necessity of this option please :pray:

noisersup: It's not required as far as I can see.

Overall, it seems that the usage `bson.MarshalExtJSON` seems to work here. From what I remember I had some issues with that while implementing it, but maybe some things changed between the driver versions.

noisersup (dismissed):
LGTM

internal/dataapi/server/json.go
```diff
@@ -15,32 +15,25 @@
...
-
-	err = encoder.Encode(bson.Raw(raw))
+	_, err = io.Copy(jsonDst, bytes.NewReader(b))
```
chilagrow, line 36: Let's keep the implementation more closely to the example.
https://pkg.go.dev/go.mongodb.org/mongo-driver/v2/bson#example-Encoder-ExtendedJSON

I think we should be able to do something like below.
```
encoder := bson.NewEncoder(bson.NewExtJSONValueWriter(jsonDst, false, false))

if err := encoder.Encode(bson.Raw(raw)); err != nil {
	return lazyerrors.Error(err)
}
```
If encoder needs special option, it's simple to make the change :hugs:

AlekSi: @KrishnaSindhur PTAL

chilagrow (approved):
Thanks for your contribution!

AlekSi (approved):

AlekSi:
Thank you!

PR #5038, opened 2025-04-09
Supervise services in evaluation docker image

### LINKED ISSUE #4976: Provide evaluation Docker image with production builds
### What should be done?

No matter how much we try to discourage them, some people will use evaluation Docker images in production. Many do that for convenience—deploying one container is more straightforward than two.

Let's update our existing evaluation Docke with development built to use ~[runit](https://smarden.org/runit/) as we [did in 1.x](https://github.com/FerretDB/FerretDB/tree/main-v1/build/ferretdb/all-in-one)~ supervisord. Make sure that termination signals are handled correctly.

After that, let's rename that image to `ferretdb-eval-dev`.

After that, let's make a `ferretdb-eval` image using production builds of FerretDB and DocumentDB.

Let's build both like we build `ferretdb-eval` now – for every commit into `main` branch and for every PR with `packages` and `trust` labels (and unlike our main production image).

### Where?

Build system, CI.

### Definition of Done

- images are built;
- spot refactorings done;
- user documentation updated.

chilagrow:
# Description

Closes #4976.

## Readiness checklist


build/ferretdb/evaluation.Dockerfile
```diff
@@ -116,6 +116,14 @@ EOF
 COPY --from=evaluation-build /src/bin/ferretdb /usr/local/bin/ferretdb
 COPY build/ferretdb/99-start-ferretdb.sh /docker-entrypoint-initdb.d/
 
+COPY --from=evaluation-build /src/build/ferretdb/evaluation/ferretdb.sh /etc/service/ferretdb/run
+COPY --from=evaluation-build /src/build/ferretdb/evaluation/postgresql.sh /etc/service/postgresql/run
+COPY --from=evaluation-build /src/build/ferretdb/evaluation/entrypoint.sh /entrypoint.sh
+
+# send termination signal to each process, see https://smarden.org/runit/runsvdir.8
+STOPSIGNAL SIGHUP
```
chilagrow, line 124: I am not 100% on this solution, as it's said in https://smarden.org/runit/runsvdir.8 below, after sending TERM signals to processes, runsvdir exits with 111. Perhaps it's possible to exit container while pg is still shutting down.

> If runsvdir receives a HUP signal, it sends a TERM signal to each runsv(8) process it is monitoring and then exits with 111.

AlekSi: I think it should be possible to achieve what we want using control files:
*https://smarden.org/runit/runsv.8
* https://wiki.gentoo.org/wiki/Runit#Custom_process_control
* don't forget to ask Copilot Chat

AlekSi: And _maybe_ we should use `runit` binary, not `runsv`

AlekSi: Another option is https://skarnet.org/software/s6/, but let's save it for later. Let's use `supervisord` in another PR if `runit` does not work.

noisersup (approved):

build/ferretdb/evaluation.Dockerfile
chilagrow: So it seems non that I tried worked. Basically, `runsvdir` doesn't wait for processes to finish, and by returning before finishing processes, docker shutdowns the container. 

- control directory files doesn't get executed. I think it's the same issue that HUP signal sent is captured by runsvdir, it sends TERM signal to control file but doesn't wait. https://github.com/vulk-archive/runit/blob/master/src/runsvdir.c#L280 (link to copy of official https://smarden.org/runit/)

- Attempted to use `runit` instead of `runsvdir`. In theory, it should start services but failed to start postgres and other errors (probably due to expected way of executing)
```
$ docker run --rm --name ferretdb -p 27017:27017 --platform linux/amd64 ferretdb-local                                                                       
- runit: $Id: 25da3b86f7bed4038b8a039d2f8e8c9bbcf0822b $: booting.                                                                                                                                          
- runit: warning: unable to open /dev/console: file does not exist                                                                                                                                          
- runit: enter stage: /etc/runit/1                                                                                                                                                                          
- runit: leave stage: /etc/runit/1                                                                                                                                                                          
- runit: enter stage: /etc/runit/2                                                                                                                                                                          
runsvchdir: default: current.                                                                                                                                                                               
Enabling additional executable binary formats: binfmt-supportmount: /proc/sys/fs/binfmt_misc: permission denied.                                                                                            
       dmesg(1) may have more information after failed mount system call.                                                                                                                                   
update-binfmts: warning: Couldn't mount the binfmt_misc filesystem on /proc/sys/fs/binfmt_misc.                                                                                                             
.                                                                                                                                                                                                           
No PostgreSQL clusters exist; see "man pg_createcluster" ... (warning).
```

- `finish` script also doesn't get executed :thinking: 

- I think most promising is `/etc/runit/ctrlaltdel`, which should be executed on INT signal before its sent to processes.

PR #5055, opened 2025-04-16
Implement handler `Handle` function

### LINKED ISSUE #5046: Make Data API server use `Handle` method
### What should be done?

Currently, the Data API server looks up the handler method to use via the `Commands` method. That makes it completely bypass our future middlewares.

We should do four steps in four PRs, one after another:

1. We should implement Handler's `Handle` method. There is a similar code for that in the `clientconn/conn.go` file. At this point, it will not be used yet.
2. We should update the Data API code to use the `Handle` method.
3. We should apply the smallest change to `clientconn/conn.go` to use that method, too, and remove `Commands` method.
4. We should unexport Handler's methods for commands.

No changes in tests are expected.

### Where?

- https://github.com/FerretDB/FerretDB/tree/main/internal/clientconn
- https://github.com/FerretDB/FerretDB/tree/main/internal/dataapi
- https://github.com/FerretDB/FerretDB/tree/main/internal/handler

See references to this issue.

### Definition of Done

Code updated and tested manually, existing tests pass.
Do not attempt refactoring of existing code in this issue.

chilagrow:
# Description

Closes #5046.

## Readiness checklist


internal/handler/handler.go
```diff
@@ -135,8 +137,32 @@ func (h *Handler) Run(ctx context.Context) {
 
 // Handle processes a request.
 func (h *Handler) Handle(ctx context.Context, req *middleware.Request) (*middleware.Response, error) {
-	// TODO https://github.com/FerretDB/FerretDB/issues/5046
-	panic("not implemented")
+	switch {
+	case req.OpMsg != nil:
+		doc, err := req.OpMsg.Document()
```
chilagrow, line 142: Ideally decoded `OpMsg.Document()` could be in `*middleware.Request`, but that would require some refactoring so not in this issue.

internal/handler/handler.go
```diff
@@ -135,8 +137,32 @@ func (h *Handler) Run(ctx context.Context) {
...
+		}
+
+		msgCmd := doc.Command()
+
+		var cmdHandler middleware.HandleFunc
+
+		cmd, ok := h.commands[msgCmd]
+		if !ok {
+			cmdHandler = notFound(msgCmd)
+		} else if cmd.Handler == nil {
```
AlekSi, line 154: We should be able to rewrite it with `else if` by using `switch`, early `return`, or something. `} else {` is generally not good, but `else if` is the worst.

But also, I think that branch is not possible: `initCommands` already set all handlers and uses `notImplemented` if needed.

internal/handler/handler.go
AlekSi: No, `OpMsg.Document()` itself should cache the result. See https://github.com/FerretDB/wire/issues/94

internal/handler/handler.go
```diff
@@ -151,6 +177,16 @@ func (h *Handler) Collect(ch chan<- prometheus.Metric) {
...
+// notFound returns a handler that returns not found error.
+func notFound(command string) middleware.HandleFunc {
+	return func(context.Context, *middleware.Request) (*middleware.Response, error) {
+		return nil, mongoerrors.New(
+			mongoerrors.ErrCommandNotFound,
+			fmt.Sprintf("no such command: '%s'", command),
+		)
+	}
+}
+
```
AlekSi, line 189: Let's move it to `commands.go` next to `notImplemented`

internal/handler/handler.go
chilagrow: Thanks you are right, it's not possible to reach `notImplemented` here.

internal/handler/handler.go
chilagrow: Excellent thanks!

internal/handler/handler.go
```diff
@@ -135,8 +135,26 @@ func (h *Handler) Run(ctx context.Context) {
...
+		if ok && cmd.Handler != nil {
+			return cmd.Handler(ctx, req)
+		}
+
+		return notFound(msgCmd)(ctx, req)
+	case req.OpQuery != nil:
+		return h.CmdQuery(ctx, req)
+	default:
+		panic("unsupported request")
+	}
```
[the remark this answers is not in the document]
chilagrow: panic is what we want, because it should be unreachable.

PR #5077, opened 2025-04-19
Use `runtime.Cleanup` for resource tracking

### LINKED ISSUE #5047: Use `runtime.Cleanup` for resource tracking
### What should be done?

Currently, we use [finalizers](https://pkg.go.dev/runtime#SetFinalizer), but there can be only one.

<details>
<summary>joke</summary>

Image

</details>

We should update our code to use [`runtime.Cleanup`](https://pkg.go.dev/runtime#Cleanup). See also [a blog post about it](https://go.dev/blog/cleanups-and-weak).

We probably could also add a unit test.

### Where?

- https://github.com/FerretDB/FerretDB/blob/main/internal/util/resource/resource.go

### Definition of Done

- unit tests added;
- code updated;
- unit test added before code change still passes.

sahinakyol:
# Description
Replaced SetFinalizer with runtime.AddCleanup, which supports multiple callbacks and doesn’t pause GC cycles.
Store each cleanup handle so Untrack can call Stop() and cancel it.
Added self‑cleanup in the callback to remove the token from the map and profile, preventing memory / profile leaks when Untrack is forgotten.

Closes #5047.

## Readiness checklist


internal/util/resource/resource_test.go
AlekSi, line 1: Please use testify, as in our other tests

internal/util/resource/resource.go
```diff
@@ -76,28 +80,30 @@ func Track(obj any, token *Token) {
...
+	// Delete token from map and profile to avoid leaks.
+	c := runtime.AddCleanup(obj, func(msg string) {
+		cleanupByToken.Delete(token)
 
-		msg := fmt.Sprintf("%T has not been finalized", obj)
-		if stack != "" {
-			msg += "\nObject created by " + stack
+		if p := pprof.Lookup(name); p != nil {
+			p.Remove(token)
 		}
```
AlekSi, line 94: I'm not sure why we do that. We use this package precisely to track leaks. We don't want to avoid them implicitly there.

internal/util/resource/resource.go
```diff
@@ -76,28 +80,30 @@ func Track(obj any, token *Token) {
 	// because otherwise profile will hold a reference to obj and finalizer will never run
```
AlekSi, line 80: We should update comments like this one that refer to finalizers

internal/util/resource/resource.go
```diff
@@ -27,6 +27,10 @@ import (
 	"github.com/FerretDB/FerretDB/v2/internal/util/devbuild"
 )
 
+// cleanupByToken maps each *Token to its runtime.Cleanup handle.
+// That lets Untrack cancel the scheduled cleanup and avoids leaks.
+var cleanupByToken sync.Map
```
AlekSi, line 32: Do we need both a global map and a token? Can't the token store cleanup function?

internal/util/resource/resource.go
sahinakyol: Understood, I haven't to looked with this eye. Addressed.

internal/util/resource/resource.go
sahinakyol: I was considering violating this comment, 
`
// Token should be a field of a tracked object.
//
// The underlying type is not struct{} because (from the Go spec)
// "Two distinct zero-size variables may have the same address in memory",
// and they do.
` 
but if we create token object as in my last change, we can handle the zero size problem. I've implemented this based on your suggestion.

AlekSi:
@sahinakyol, please sign the CLA in [this comment](https://github.com/FerretDB/FerretDB/pull/5077#issuecomment-2816697506), fix linter warnings, and [re-request review](https://github.com/FerretDB/FerretDB/blob/main/CONTRIBUTING.md#submitting-pr).

internal/util/resource/resource.go
```diff
@@ -22,21 +22,25 @@ import (
...
-// The underlying type is not struct{} because (from the Go spec)
-// "Two distinct zero-size variables may have the same address in memory",
-// and they do.
-type Token byte
+// Contains a byte field to ensure unique memory addresses, as Go spec states
+// "Two distinct zero-size variables may have the same address in memory".
+// Also stores cleanup handle for proper resource management.
+type Token struct {
+	cleanup atomic.Pointer[runtime.Cleanup]
+	token   byte
```
AlekSi, line 38: We no longer need that field. The comment should be updated, too.

internal/util/resource/resource.go
```diff
@@ -22,21 +22,25 @@ import (
...
 //
-// The underlying type is not struct{} because (from the Go spec)
-// "Two distinct zero-size variables may have the same address in memory",
-// and they do.
-type Token byte
+// Contains a byte field to ensure unique memory addresses, as Go spec states
+// "Two distinct zero-size variables may have the same address in memory".
+// Also stores cleanup handle for proper resource management.
+type Token struct {
+	cleanup atomic.Pointer[runtime.Cleanup]
```
AlekSi, line 37: Let's store data like msg or a stack trace there, not a function pointer. There is also no need to use `atomic.Pointer` because there is no need to swap the pointer.

internal/util/resource/resource.go
```diff
@@ -49,8 +53,9 @@ func profileName(obj any) string {
 
 // Track tracks the lifetime of an object until Untrack is called on it.
 //
-// Obj should a pointer to a struct with a field "token" of type *Token.
-func Track(obj any, token *Token) {
+// Obj should be a pointer to a struct with a field "token" of type *Token.
+// Uses generics to provide type safety.
```
AlekSi, line 57: It is already clear from the function signature :)

internal/util/resource/resource_test.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// runGC forces several GC cycles to give the runtime a chance to run cleanups.
+func runGC(t *testing.T) {
+	t.Helper()
+
+	for i := 0; i < 8; i++ {
+		runtime.GC()
+		time.Sleep(10 * time.Millisecond)
+	}
```
AlekSi, line 37: Why eight iterations? Why 10 milliseconds?

internal/util/resource/resource_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	p := pprof.Lookup(profileName(obj))
+	if p != nil {
+		return p.Count()
+	}
+
+	return 0
+}
+
+func TestTrackProfileEntryAdded(t *testing.T) {
```
AlekSi, line 52: It looks like many of those test cases could be merged into one

internal/util/resource/resource_test.go
```diff
@@ -0,0 +1,100 @@
...
+	// GC should not run cleanup because obj is still reachable.
+	runGC(t)
+
+	runtime.KeepAlive(obj)
+
+	assert.Equal(t, 1, entryCount(t, obj), "cleanup shouldn't run while object is reachable")
+}
+
+func TestTrackCleanupRunsWhenAbandoned(t *testing.T) {
+	t.Skip("cleanup panic occurs in another goroutine; needs sub‑process test")
```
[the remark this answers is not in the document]
sahinakyol: No, how can I recover panic scenario which is creating a new routine ? You are right it seems like but I just wanted to show which case causes to run cleanup in rest of test :D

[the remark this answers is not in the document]
AlekSi: I don't know if we could test it automatically. But we could, for example, check if the `FERRETDB_TEST_MANUAL` environment variable is set and run this test, so we could run it manually and check that it panics.

internal/util/resource/resource_test.go
sahinakyol: Loop cycle is just an random number, sleep time is big enough. There is no deterministic reason.

internal/util/resource/resource_test.go
sahinakyol: Sure will do, generally I prefer to seperate tests to small parts that's why I did.

internal/util/resource/resource.go
sahinakyol: Let me check your suggestions, thanks a lot 👍

internal/util/resource/resource_test.go
AlekSi: I wonder if just one or maybe two `runtime.GC()` would be reliable, too. If not, we probably could use `runtime.Gosched()`

internal/util/resource/resource.go
```diff
@@ -73,31 +76,26 @@ func Track(obj any, token *Token) {
 	}
 
 	// use token instead of obj itself,
-	// because otherwise profile will hold a reference to obj and finalizer will never run
+	// because otherwise profile will hold a reference to obj and cleanup will never run
 	p.Add(token, 1)
 
-	var stack string
+	token.msg = fmt.Sprintf("%T has not been finalized", obj)
```
chilagrow, line 82: Why does `token.msg` need to be part of token struct :thinking:? It does not seem to be used outside this function.

internal/util/resource/resource.go
```diff
@@ -73,31 +73,26 @@ func Track(obj any, token *Token) {
 	}
 
 	// use token instead of obj itself,
-	// because otherwise profile will hold a reference to obj and finalizer will never run
+	// because otherwise profile will hold a reference to obj and cleanup will never run
 	p.Add(token, 1)
 
-	var stack string
+	var errMsg = fmt.Sprintf("%T has not been finalized", obj)
```
chilagrow, line 79: Let's use the short form to be consistent with the rest of the code :hugs: 

```suggestion
	errMsg := fmt.Sprintf("%T has not been finalized", obj)
```

chilagrow (commented):
Just a tiny final comment, thanks for your contribution :hugs:

internal/util/resource/resource.go
sahinakyol: Done :)

chilagrow (approved):
Thanks for your contribution!

noisersup (approved):

PR #5161, opened 2025-05-20
Add documentation for logical operators

Fashander:
# Description

Closes #<issue_number>.

## Readiness checklist

### DIFF
--- a/website/docs/reference/query-operators/logical-operators.json
+++ b/website/docs/reference/query-operators/logical-operators.json
@@ -1,0 +1,291 @@
+{
+  "1-drop-collection": {
+    "request": {
+      "drop": "books",
+      "$db": "{{.Database}}"
+    },
+    "response": {
+      "nIndexesWas": {
+        "$numberInt": "1"
+      },
+      "ns": "{{.Database}}.books",
+      "ok": {
+        "$numberDouble": "1.0"
+      }
+    }
+  },
+  "2-insert-dataset": {
+    "request": {
+      "insert": "books",
+      "$db": "{{.Database}}",
+      "documents": [
+        {
+          "_id": "pride_prejudice_1813",
+          "title": "Pride and Prejudice",
+          "authors": [
+            {
+              "name": "Jane Austen",
+              "nationality": "British"
+            }
+          ],
+          "genres": ["Romance", "Classic", "Historical Fiction"],
+          "rating": {
+            "$numberDouble": "4.5"
+          }
+        },
+        {
+          "_id": "moby_dick_1851",
+          "title": "Moby Dick",
+          "authors": [
+            {
+              "name": "Herman Melville",
+              "nationality": "American"
+            }
+          ],
+          "genres": ["Adventure", "Classic", "Sea Story"],
+          "rating": {
+            "$numberDouble": "4.3"
+          }
+        },
+        {
+          "_id": "clean_code_2008",
+          "title": "Clean Code: A Handbook of Agile Software Craftsmanship",
+          "authors": [
+            {
+              "name": "Robert C. Martin",
+              "nationality": "American"
+            }
+          ],
+          "genres": ["Computer Science", "Software Engineering", "Programming"],
+          "rating": {
+            "$numberDouble": "4.7"
+          }
+        }
+      ]
+    },
+    "response": {
+      "n": {
+        "$numberInt": "3"
+      },
+      "ok": {
+        "$numberDouble": "1.0"
+      },
+      "createdCollectionAutomatically": false
+    }
+  },
+  "3-and-operator": {
+    "request": {
+      "find": "books",
+      "$db": "{{.Database}}",
+      "filter": {
+        "$and": [
+          {
+            "authors.nationality": "British"
+          },
+          {
+            "rating": {
+              "$gte": {
+                "$numberDouble": "4.5"
+              }
+            }
+          }
+        ]
+      }
+    },
+    "response": {
+      "cursor": {
+        "id": {
+          "$numberLong": "0"
+        },
+        "ns": "{{.Database}}.books",
+        "firstBatch": [
+          {
+            "_id": "pride_prejudice_1813",
+            "title": "Pride and Prejudice",
+            "authors": [
+              {
+                "name": "Jane Austen",
+                "nationality": "British"
+              }
+            ],
+            "genres": ["Romance", "Classic", "Historical Fiction"],
+            "rating": {
+              "$numberDouble": "4.5"
+            }
+          }
+        ]
+      },
+      "ok": {
+        "$numberDouble": "1.0"
+      },
+      "createdCollectionAutomatically": false
+    }
+  },
+  "4-or-operator": {
+    "request": {
+      "find": "books",
+      "$db": "{{.Database}}",
+      "filter": {
+        "$or": [
+          {
+            "authors.nationality": "British"
+          },
+          {
+            "rating": {
+              "$gte": {
+                "$numberDouble": "4.5"
+              }
+            }
+          }
+        ]
+      }
+    },
+    "response": {
+      "cursor": {
+        "id": {
+          "$numberLong": "0"
+        },
+        "ns": "{{.Database}}.books",
+        "firstBatch": [
+          {
+            "_id": "pride_prejudice_1813",
+            "title": "Pride and Prejudice",
+            "authors": [
+              {
+                "name": "Jane Austen",
+                "nationality": "British"
+              }
+            ],
+            "genres": ["Romance", "Classic", "Historical Fiction"],
+            "rating": {
+              "$numberDouble": "4.5"
+            }
+          },
+          {
+            "_id": "clean_code_2008",
+            "title": "Clean Code: A Handbook of Agile Software Craftsmanship",
+            "authors": [
+              {
+                "name": "Robert C. Martin",
+                "nationality": "American"
+              }
+            ],
+            "genres": ["Computer Science", "Software Engineering", "Programming"],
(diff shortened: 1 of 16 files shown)


website/docs/reference/query-operators/logical-operators.json
```diff
@@ -0,0 +1,291 @@
+{
```
Fashander, line 1: @noisersup Please could you take a look at this setup? I don't understand why CTS is failing since the expected response is just as described here.

chilagrow:
See failed CTS run https://github.com/FerretDB/FerretDB/actions/runs/15146587081/job/42583647021?pr=5161

website/docs/reference/operators/query/logical-operators/and.mdx
```diff
@@ -0,0 +1,64 @@
...
+The `$and` operator combines multiple query expressions and returns documents that match all the expressions.
+
+The operator accepts an array of query expressions as shown below, and all expressions must be satisfied for a document to be included in the result set.
+
+```js
+db.runCommand({
+  find: '<collection_name>',
+  filter: {
+    $and: [
+      { <query-expression1> },
```
chilagrow, line 22: Let's use either `_` or `-` but not mix them like `<collection_name>` and `<query-expression1>`

website/docs/reference/operators/query/logical-operators/or.mdx
```diff
@@ -0,0 +1,62 @@
...
+---
+
+# `$or`
+
+import CodeBlock from '@theme/CodeBlock';
+import Dataset from '!!raw-loader!./logical-operators/2-insert-dataset.request.js';
+import OrOperatorRequest from '!!raw-loader!./logical-operators/4-or-operator.request.js';
+import OrOperatorResponse from '!!raw-loader!./logical-operators/4-or-operator.response.js';
+
+The `$or` operator Combines multiple query expressions and returns documents that match at least one expression.
```
chilagrow, line 13: Is Combine in capital on purpose? A bit weird :thinking:

website/docs/reference/operators/query/logical-operators/and.mdx
```diff
@@ -0,0 +1,64 @@
...
+      { <query-expression2> }
+    ]
+  }
+})
+```
+
+A query expression can be any valid query such as `{ <field>: <value> }`, `{ <field>: { <operator>: <value> } }`, `{ <field>: /<pattern>/ }`, etc.
```
chilagrow, line 29: Could you think of a better way to display this? The new line is making it harder to read

image

website/docs/reference/operators/query/logical-operators/not.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+The following examples demonstrate how to use the `$not` operator in queries using documents in the following `books` collection:
+
+<!-- markdownlint-disable MD033 -->
+
+<CodeBlock language="js">{Dataset}</CodeBlock>
+
+### Filter documents that do not match a query expression
+
+Using the books collection, return all documents where the rating is not less than `4.5`.
```
chilagrow, line 38: Could we use a bit simpler example? _not less than_ requires a bit of thinking, as we would prefer to use _greater than or equal_. Maybe something like not British?

website/docs/reference/operators/query/logical-operators/nor.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+The following examples demonstrate how to use the `$nor` operator in queries using documents in the following `books` collection:
+
+<!-- markdownlint-disable MD033 -->
+
+<CodeBlock language="js">{Dataset}</CodeBlock>
+
+### Filter documents with multiple query expressions
+
+Using the books collection, return all documents where the author is not `British` AND the rating is not greater than or equal to `4.5`.
```
chilagrow, line 41: Same comment as $or, let's use a simpler operator. Possibly something doesn't involve `$gt` or `$lt` :see_no_evil:

website/docs/reference/operators/query/logical-operators/logical-operators/2-insert-dataset.request.js
```diff
@@ -0,0 +1,41 @@
+db.runCommand({
+  insert: 'books',
+  documents: [
```
chilagrow, line 3: Are we using the same set of data for createIndexes? :thinking: Maybe we should create a fixture if that's the case

Fashander: The createIndexes doc doesn't have an insert, commands are only for creating indexes

website/docs/reference/operators/query/logical-operators/not.mdx
Fashander: Thanks for asking this. I was able to go deeper and `$not` op only works for operator expressions and regex

website/docs/reference/operators/query/logical-operators/nor.mdx
Fashander: For this, I think an example that showcases 2 of the listed query expressions is better. So one uses the not British and the other filters for documents with ratings that are less than 4.5

website/docs/reference/operators/query/logical-operators/and.mdx
Fashander: PTAL

website/docs/reference/operators/query/logical-operators/logical-operators/2-insert-dataset.request.js
chilagrow: Let me rephrase, I've see similar data inserted like this. It probably wasn't for createIndexes command but for other documentation. Is it worth creating a fixture that contains subset of books collection. You've mentioned that you don't want to use existing books fixture because it contains field like `vector`. I think we should create a fixture that can be used for majority of data.

website/docs/reference/operators/query/logical-operators/and.mdx
```diff
@@ -0,0 +1,68 @@
...
+      { <query_expression1> },
+      { <query_expression2> }
+    ]
+  }
+})
+```
+
+A query expression can be any valid query such as:
+
+- `{ <field>: <value> }`
```
chilagrow, line 31: Could we use ```json format where syntax highlighting is used? Or possibly `js`? I think we should use syntax highlighting as it's come across easier read.

website/docs/reference/operators/query/logical-operators/and.mdx
```diff
@@ -0,0 +1,68 @@
...
+The AND operation returns `pride_prejudice_1813` as the document that matches both expressions.
+
+<CodeBlock language="js">{AndOperatorResponse}</CodeBlock>
+
+Even though `clean_code_2008` matches the second query expression in the earlier example, it is not returned since it does not match the first one.
+
+## Implementation notes and constraints
+
+- The `$and` operator cannot be used directly as a top-level aggregation stage; it must be used within a query context like `$match`.
+- For array fields, ensure that the query expressions are structured to match the elements within the array, or use `$elemMatch` for more complex queries.
```
chilagrow, line 58: > the query expressions are structured to match the elements within the array

Could you simplify this please?

website/docs/reference/operators/query/logical-operators/logical-operators/2-insert-dataset.request.js
Fashander: While I understand the importance of a standard fixture, there will always be conmands or operation examples that favor some document fields more than others. And if we have to have an overload of fields to cover all use cases, then across our docs, it won't be beginner friendly. I believe we should have a balance between standard and simplicity.

And example like this is simple and easy to understand for a `findOne` command:

```
db.books.insertMany([
  { title: "1984", author: "George Orwell" }
])
```

But something with 8 fields and nested documents to cover multiple scenarios introduces unnecessary complexity.

Fashander: What we agreed on was to work on each command independently so as to avoid this kind of complexity, as long as its tested with CTS.

website/docs/reference/operators/query/logical-operators/not.mdx
chilagrow: Can we use not equal then such as `$not: {$eq: xxxx}`? the usage of `not less than 4.5` is just weird as we would use `greater than or equal to 4.5`.

website/docs/reference/operators/query/logical-operators/and.mdx
Fashander: We should not use JSON syntax highlighting here since it's not really talking about an operation but a query expression which can be written in different ways. Its main use is to explain what's already written within the syntax.

website/docs/reference/operators/query/logical-operators/logical-operators/2-insert-dataset.request.js
Fashander: Yes. With each set of commands having their own variation of fields, we have flexibility and can always cover a lot of basis and scenarios, as long as they remain tested and validated with CTS.

chilagrow: Thanks for the explanation :hugs:

website/docs/reference/operators/query/logical-operators/nor.mdx
chilagrow: Yes 2 conditions are great for this example. How about `not British` and `not Classic`? Using `$gt` or `$lt` operator with `not` condition is not straightforward to process :upside_down_face:

PR #5203, opened 2025-05-30
Add `*slog.logger` for embedded FerretDB config

### LINKED ISSUE #4750: Add configuration options for embeddable package
### What should be done?

- [x] Update https://github.com/FerretDB/embedded-example
- [ ] Refactor listener and CLI to extract more common code?
- [x] Make telemetry configurable
- [ ] Make logging configurable: allow to set `slog.Handler` or `slog.Logger`
- [ ] Make metrics configurable (and register them)
- [ ] Make tracing configurable
- [ ] Make debug handler configurable (?)
- [ ] Make authentication configurable
- [ ] Make TLS configurable
- [ ] Make proxy mode configurable

### Where?

- https://github.com/FerretDB/FerretDB/tree/main/ferretdb
- https://github.com/FerretDB/FerretDB/tree/main/cmd/ferretdb

### Definition of Done

- the `ferretdb` package updated and documented;
- the `main` package updated;
- unit tests added;
- spot refactorings done;
- user documentation updated (quite a lot could be copied from v1 docs).

jyz0309:
# Description
Add `*slog.logger` config for embedded FerretDB config

Closes #4750.

## Readiness checklist


ferretdb/ferretdb_test.go
```diff
@@ -83,6 +84,54 @@ func Example() {
...
+	require.NoError(t, err)
+
+	res := client.Database("admin").RunCommand(ctx, bson.D{{Key: "getLog", Value: "global"}})
+
+	var actual bson.D
+	err = res.Decode(&actual)
+	require.NoError(t, err)
+
+	require.Len(t, actual, 3)
+	require.Equal(t, "log", actual[0].Key)
```
chilagrow, line 123: What does actual[0] look like? Current assertions do not test `getLog` retrieves logs from handler :thinking:. Let's make sure to test that logs from handler are fetched.

internal/util/logging/handler.go
```diff
@@ -140,6 +141,8 @@ func NewHandler(out io.Writer, opts *NewHandlerOpts) *Handler {
 		h = slog.NewTextHandler(out, stdOpts)
 	case "json":
 		h = slog.NewJSONHandler(out, stdOpts)
+	case "custom":
```
chilagrow, line 144: Perhaps calling it `embeddable` might be more precise name?

ferretdb/ferretdb.go
```diff
@@ -102,11 +104,21 @@ func New(config *Config) (*FerretDB, error) {
...
+		Level:      logLevel,
+	}
+
+	var logger *slog.Logger
+
+	if config.Logger == nil {
+		lOpts.Base = "console"
+		logger = logging.WithName(logging.Logger(logOutput, lOpts, ""), "ferretdb")
+	} else {
+		handler := config.Logger.Handler()
```
chilagrow, line 117: If I set config.Logger with a logger that configured with log level such as `logger.Enabled(ctx, slog.LevelDebug)`, how is it work?

I think it should set `NewHandlerOpts.Level` with the configured level of the `config.Logger`. Could it be tested?

chilagrow (commented):
Thanks for your contribution! Great start, I have a few comments :pray:

ferretdb/ferretdb_test.go
jyz0309: The `actual["log"]` value is the log in FerretDB. Maybe we can use regexp to make sure that the value is the log fetched by handler like `.*Listening on TCP.*` (the first log output by FerretDB)? If you have a better way, plz tell me :)

chilagrow: I think we can execute a command which would produce a log, for example like below command produces `WARN` level log with `no such command: 'nonExistentCommand'`.

```
RunCommand(ctx, bson.D{{Key: "nonExistentCommand", Value: 1}})
```

Then we can check that specific log in the test assertion. I think startup logs such as `.*Listening on TCP.*` works slightly differently and we want to make sure the logs produced by commands work as expected :hugs:.

Doing test this way would also test `WARN` log level can be set :hugs:

ferretdb/ferretdb.go
chilagrow: I see your point, the simplest way is using `LogLevel slog.Leveler` configuration option. Otherwise, you would need to figure out the level by calling `Enabled()` to determine the level set by the logger.

jyz0309: @chilagrow I think we can use `LogLevel slog.Level` configuration option, how about create a new handler named `embeddableHandler` in `logging` package, and make the `custom logger.Handler` as a field in `embeddableHandler`? Then we can use LogLevel for the `custom logger`

internal/util/logging/embeddable_handler.go
```diff
@@ -0,0 +1,74 @@
...
+// WithAttrs implements [slog.Handler].
+func (h *embeddableHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
+	if len(attrs) == 0 {
+		return h
+	}
+
+	h.m.Lock()
+	defer h.m.Unlock()
+
+	h.opts.Handler = h.opts.Handler.WithAttrs(attrs)
```
chilagrow, line 61: How was this tested? See how other handlers are tested in https://github.com/FerretDB/FerretDB/blob/main/internal/util/logging/handler_test.go. We should do the same :pray:

internal/util/logging/embeddable_handler.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// WithAttrs implements [slog.Handler].
+func (h *embeddableHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
+	if len(attrs) == 0 {
+		return h
+	}
+
+	h.m.Lock()
+	defer h.m.Unlock()
```
chilagrow, line 59: Could you add a comment about why mutex lock is needed here please :pray:

internal/util/logging/embeddable_handler.go
chilagrow, line 1: We should have test for handler using `slogtest.Run`, see how it's done in other handler test such as
https://github.com/FerretDB/FerretDB/blob/main/internal/util/logging/mongo_handler_test.go

ferretdb/ferretdb_test.go
```diff
@@ -83,6 +84,57 @@ func Example() {
...
+	require.Regexp(t, ".*no such command: 'nonExistentCommand'.*", actual[0].Value)
```
chilagrow, line 127: `require.Contains` is probably simpler and sufficient here :hugs:

chilagrow (commented):
Thanks :pray: Great work, I have a few suggestions!

internal/util/logging/embeddable_handler.go
```diff
@@ -0,0 +1,122 @@
...
+	if h.opts.Level != nil {
+		minLevel = h.opts.Level.Level()
+	}
+
+	return l >= minLevel
+}
+
+// Handle implements [slog.Handler].
+func (h *embeddableHandler) Handle(ctx context.Context, record slog.Record) error {
+	if h.testAttrs != nil {
```
chilagrow, line 58: `h.testAttrs` is only used for tests. This means that this condition applies only for tests. Instead, we should make sure that options apply even if `h.testAttrs == nil` for normal cases. See implementation for other handlers :hugs:

internal/util/logging/handler.go
```diff
@@ -47,8 +47,9 @@ type Handler struct {
 //
 //nolint:vet // for readability
 type NewHandlerOpts struct {
-	Base         string // base handler to create: "console", "text", "json" or "mongo"
+	Base         string // base handler to create: "console", "text", "json", "mongo" or "embeddable"
 	Level        slog.Leveler
+	Handler      slog.Handler
```
chilagrow, line 52: Let's be specific that `Handler` is only for embeddable base. We could name it `EmbeddedHandler`, and add a comment.

internal/util/logging/handler.go
```diff
@@ -140,6 +141,8 @@ func NewHandler(out io.Writer, opts *NewHandlerOpts) *Handler {
 		h = slog.NewTextHandler(out, stdOpts)
 	case "json":
 		h = slog.NewJSONHandler(out, stdOpts)
+	case "embeddable":
+		h = newEmbeddableHandler(opts, nil)
```
chilagrow, line 145: This one is odd one out, it implicitly doesn't use `io.Writer` instead it uses `opts.Handler`. Could you add a comment for special handling of embedded handler please. Also checks like below helps set expected values.
```
must.BeZero(out)
must.NotBeZero(opts.Handler)
```

ferretdb/ferretdb.go
```diff
@@ -100,11 +102,22 @@ func New(config *Config) (*FerretDB, error) {
 	}
 
 	lOpts := &logging.NewHandlerOpts{
-		Base:       "console",
-		Level:      logLevel,
 		SkipChecks: true,
+		Level:      logLevel,
+	}
```
chilagrow, line 107: We often avoid unnecessary `else` statement. Also I don't think we need to instantiate the logger separately for example. Could it be simplified something like below?
```
	var embeddedHandler slog.Handler
	logBase := "console"

	if config.Logger != nil {
		logBase = "embeddable"
		embeddedHandler = config.Logger.Handler()
		logOutput = nil
	}

	lOpts := &logging.NewHandlerOpts{
		Base:       logBase,
		Level:      logLevel,
		Handler:    embeddedHandler,
		SkipChecks: true,
	}

	logger := logging.WithName(logging.Logger(logOutput, lOpts, ""), "ferretdb")

```

internal/util/logging/handler_test.go
```diff
@@ -53,16 +53,29 @@ func TestHandler(t *testing.T) {
...
 
-			var buf bytes.Buffer
-			var h slog.Handler = NewHandler(&buf, &NewHandlerOpts{
-				Base:  base,
-				Level: slog.LevelInfo,
-			})
+			var (
+				buf bytes.Buffer
+				h   slog.Handler
+			)
```
chilagrow, line 65: consistency with rest of our code :pray: 
```suggestion
			var buf bytes.Buffer
			var h   slog.Handler
```

chilagrow (commented):
Embeddable handler doesn't handle options for non test scenario currently.

internal/util/logging/embeddable_handler.go
jyz0309: I think we also need to ignore the RemoveLevel/Source fields, cause we can not control the logic in  `embeddableHandler.Handle()` function by the user :)

chilagrow (commented):
Let's clarify what this mutex is used for :hugs:

chilagrow (approved):
Thank you for your contribution!

internal/util/logging/handler.go
```diff
@@ -140,6 +141,12 @@ func NewHandler(out io.Writer, opts *NewHandlerOpts) *Handler {
 		h = slog.NewTextHandler(out, stdOpts)
 	case "json":
 		h = slog.NewJSONHandler(out, stdOpts)
+	case "embeddable":
+		// embeddable handler implicitly doesn't use io.Writer
+		// instead it uses opts.EmbeddedHandler
+		must.BeZero(out)
+		must.NotBeZero(opts.EmbeddedHandler)
+		h = newEmbeddableHandler(opts, nil)
```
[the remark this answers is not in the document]
jyz0309: because the `opts.embeddedHandler` has its logLevel, if we need to use the `LogLevel` in opts we need to use `embeddable_handler` and set the logLevel for it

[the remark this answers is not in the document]
jyz0309: Okay, I will delete the embeddable_handler

AlekSi:
@jyz0309, sorry for the delay. We will return to this PR this week and resolve the merge conflict ourselves. Thank you for your contribution!

PR #5244, opened 2025-06-10
Test Script: Try to Repro Lookup + Unwind

Aweptimum:
I have a suspicion that this will work, but I cannot get the query to yield results in my local ferret instance


Aweptimum:
Ok clearly I have no idea what modules or functions are accessible in the test.js file, so I guess I'll just try ditching the Binary type and use a string instead.

AlekSi:
It might be fixed by https://github.com/microsoft/documentdb/issues/198

Aweptimum:
Not to be too bothersome, but when could we expect the next tagged release to ship with the update? I'm using the docker image in my project and so I am at the mercy of the tagged builds 😅

Aweptimum:
Oh, awesome! Thank you! I guess I missed those somehow.

PR #5361, opened 2025-07-17
Use Codecov for flaky integration tests detection

### LINKED ISSUE #5339: Use Codecov for flaky tests detection
### What should be done?

After #5185, let's try to use Codecov for flaky tests detection. See https://docs.codecov.com/docs/test-analytics.

Let's use https://github.com/AlekSi/gocov-xml and/or https://github.com/jstemmer/go-junit-report for converting test/coverage results to XML.

And let's actually test it by adding an integration test that fails approximately once in 5 runs.

### Where?

- CI configuration;
- integration tests.

### Definition of Done

- CI configuration updated;
- results are visible somewhere at https://app.codecov.io/gh/FerretDB/FerretDB/;
- detection actually works.

noisersup:
# Description

Closes #5339.

https://app.codecov.io/gh/FerretDB/FerretDB/tests/noisersup%2FFerretDB%3Acodecov-flaky-detection

## Readiness checklist


Taskfile.yml
```diff
@@ -232,6 +232,10 @@ tasks:
         -postgresql-url='postgres://username:password@127.0.0.1:5432/postgres'
         -compat-url='mongodb://username:password@127.0.0.1:47017/?replicaSet=rs0'
         -no-xfail={{.NO_XFAIL}}
+      - >
+        ../bin/go-junit-report -parser=gojson -iocopy
+        -in postgresql-{{.SHARD_INDEX}}-{{.SHARD_TOTAL}}-stdout.json.tmp
+        -out postgresql-{{.SHARD_INDEX}}-{{.SHARD_TOTAL}}.xml
```
chilagrow, line 238: I don't think this works if we use `task test-integration-postgresql RUN_TEST=xxx`, which is a common usage for local development.

Taskfile.yml
```diff
@@ -258,6 +262,17 @@ tasks:
...
+  convert-test-results:
+    desc: "Convert test results to JUnit XML format"
+    dir: integration
+    cmds:
+      - >
+        ../bin/go-junit-report -parser=gojson -iocopy
+        -in {{.IN_FILE}}
+        -out {{.OUT_FILE}}
+    requires:
+      vars: [IN_FILE, OUT_FILE]
```
chilagrow, line 274: Could we have default values? Let's say I want to run it locally, it's not trivial to know what values to set.

Taskfile.yml
noisersup: It works, the default `SHARD_INDEX` and `SHARD_TOTAL` is zero, so the file `postgresql-0-0-stdout.json.tmp` is created.

Taskfile.yml
noisersup: What would be a default value for those? This helper function is used for both `postgresql` and `mongodb` files. I don't think it has any deafault usage (similary to the `docker-build` task for example).

Taskfile.yml
noisersup: > Any reason why we do it only for postgresql?

My mistake :sweat_smile:

chilagrow (dismissed):

AlekSi (commented):
Why don't we do it for unit tests (Go workflow, `short-test` and `test` jobs)?

chilagrow (approved):

.gitignore
```diff
@@ -12,6 +12,7 @@ telemetry.json
 *.out
 *.tmp
 *.txt
+*.xml
 !website/static/*.txt
```
AlekSi, line 16: Lines covering `*.txt` should be together

Taskfile.yml
```diff
@@ -197,7 +197,7 @@ tasks:
     desc: "Run all unit tests"
     cmds:
       - go test -c -o /dev/null ./...
-      - go test -count=1 -timeout={{.TEST_TIMEOUT}} {{.RACE_FLAG}} -tags={{.BUILD_TAGS}} -shuffle=on -coverprofile=cover.txt ./...
+      - go test -v -count=1 -timeout={{.TEST_TIMEOUT}} {{.RACE_FLAG}} -tags={{.BUILD_TAGS}} -shuffle=on -coverprofile=cover.txt ./... 2>&1 | ./bin/go-junit-report -out=unit.xml -iocopy
```
AlekSi, line 200: Let's use `tee`, as we do in `bench-unit`

.github/workflows/_integration.yml
```diff
@@ -125,6 +125,20 @@ jobs:
...
+          bin/task convert-test-results
+          IN_FILE=${{ inputs.task }}-${{ inputs.shard_index }}-${{ inputs.shard_total }}-stdout.json.tmp
+          OUT_FILE=${{ inputs.task }}-${{ inputs.shard_index }}-${{ inputs.shard_total }}.xml
+
+      - name: Upload test results to Codecov
+        if: always()
+        uses: codecov/test-results-action@v1
+        with:
+          flags: integration,${{ inputs.task }}-${{ inputs.shard_index }}
+          file: ./integration/${{ inputs.task }}-${{ inputs.shard_index }}-${{ inputs.shard_total }}.xml
```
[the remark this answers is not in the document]
AlekSi: I semi-agree with copilot there – we don't have to copy the same naming scheme everywhere. For example, the input could be just `*.json.tmp`, and the output `integration.xml` or something

Taskfile.yml
```diff
@@ -261,6 +261,17 @@ tasks:
         -target-backend=mongodb
         -no-xfail={{.NO_XFAIL}}
 
+  convert-test-results:
```
AlekSi, line 264: Do we even need this Taskfile target, given that we don't use it for unit tests?

AlekSi (commented):
See my comments

AlekSi (commented):
See comment

PR #5372, opened 2025-07-23
Skip YugabyteDB setup for `arm64`

### LINKED ISSUE #85: Test with Yugabyte
### What should be done?

https://www.yugabyte.com/blog/mongodb-api-in-yugabytedb/

Let's always start next to PostgreSQL in Docker Compose (i.e., don't use Compose profiles, etc).
Let's add all needed configurations next to PostgreSQL (i.e., `task test-integration-yugabyte`, etc).
Let's add CI configuration.

Updating integration tests (fixing, skipping, or making them use expected failures) is not in the scope. That CI job could fail.

### Where?

- https://github.com/FerretDB/FerretDB/blob/main/Taskfile.yml
- https://github.com/FerretDB/FerretDB/blob/main/integration/Taskfile.yml
- https://github.com/FerretDB/FerretDB/blob/main/docker-compose.yml
- CI configuration
- envtool?

### Definition of Done

- Yugabyte has been added to CI and the local setup.
- Yugabyte's setup should be skipped on arm64 with a TODO comment pointing to #5369.

chilagrow:
# Description

Closes #85.

## Readiness checklist

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -16,7 +16,8 @@ vars:
   RACE_FLAG: -race={{and (ne OS "windows") (ne ARCH "arm") (ne ARCH "riscv64")}}
   BUILD_TAGS: ferretdb_dev
   MODE: diff-normal
-  SERVICES: jaeger mongodb mongodb-secure postgres yugabytedb
+  # TODO https://github.com/FerretDB/FerretDB/issues/5369
+  SERVICES: jaeger mongodb mongodb-secure postgres {{if (ne ARCH "arm64")}}yugabytedb{{end}}
 
 tasks:
   # invoked when `task` is run without arguments
@@ -105,7 +106,7 @@ tasks:
   env-setup:
     deps: [gen-version]
     cmds:
-      - bin/envtool{{exeExt}} setup
+      - bin/envtool{{exeExt}} setup {{if (eq ARCH "arm64")}}--no-yugabyte-db{{end}}
 
   env-logs:
     cmds:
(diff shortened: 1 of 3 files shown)


cmd/envtool/setup.go
```diff
@@ -172,6 +172,12 @@ func setup(ctx context.Context, logger *slog.Logger) error {
 		return lazyerrors.Error(err)
 	}
 
+	// TODO https://github.com/FerretDB/FerretDB/issues/5369
+	if !yugabyteDB {
+		logger.InfoContext(ctx, "Skipping YugabyteDB setup")
+		return nil
+	}
```
AlekSi, line 179: That means we will not print the final 'Done' below. I think that's the case where early return is not warranted.

Taskfile.yml
```diff
@@ -16,7 +16,8 @@ vars:
   RACE_FLAG: -race={{and (ne OS "windows") (ne ARCH "arm") (ne ARCH "riscv64")}}
   BUILD_TAGS: ferretdb_dev
   MODE: diff-normal
-  SERVICES: jaeger mongodb mongodb-secure postgres yugabytedb
+  # TODO https://github.com/FerretDB/FerretDB/issues/5369
+  SERVICES: jaeger mongodb mongodb-secure postgres {{if (ne ARCH "arm64")}}yugabytedb{{end}}
```
AlekSi, line 20: I don't think we need that complication. It will try to start and fail, but that by itself does not break anything

AlekSi (commented):
Overall, that's fine, but `if runtime.GOARCH == "arm64"` in the code without any CLI flags, etc would be enough.

Taskfile.yml
chilagrow: Sure, let's remove this complication

noisersup (approved):

AlekSi (approved):

PR #5385, opened 2025-07-29
Collapse and hide sidebar

### LINKED ISSUE #5384: Collapse sidebar and ToC
### What should be done?

We want to provide options for readers to make the screen wider, either via collapsing/hiding left sidebar and right ToCs.

Image

Fashander:
# Description

Closes #5384.

## Readiness checklist


website/docusaurus.config.js
```diff
@@ -104,6 +104,11 @@ const config = {
   themeConfig:
     /** @type {import('@docusaurus/preset-classic').ThemeConfig} */
     ({
+      docs: {
+        sidebar: {
+          hideable: true,
+        },
+      },
```
AlekSi, line 111: Do we need it if it does not work?

AlekSi (changes requested):
CleanShot 2025-07-30 at 18 58 02@2x

AlekSi (changes requested):
https://github.com/FerretDB/FerretDB/blob/b1189f2594c9c448f6d011648fe4b7a44712025f/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L2

AlekSi (commented):
> Seems there's no automatic way to disable ToC across the entire doc; we can only do it per page with hide_table_of_contents: true

So let's add it do all pages. I added a linter in #5391.

website/docs/configuration/flags.md
chilagrow, line 1: There is this weird transition from section page to detail page. For example, when you are on `Configuration` page, it's width is different from `Configuration flags` page. You will notice it when you click through `next`. May be this is okay? 

image

image

Fashander: That's because of the table. It's actually the reason for the PR so we can expand the tables rather than have them squeezed in. That way, tables just expands to its full width. Other content remains within the previous width.

AlekSi: Because that [autogenerated category page](https://docusaurus.io/docs/sidebar/autogenerated) does not seem to support that option.

@Fashander, please search Docusaurus issues/discussions about that, maybe create a new one. Maybe there is a way to solve it.

